Issue #365, opened 2014-10-07
Faking virtual members before constructor execution


matkoch:
In our codebase we need to mock objects, whose constructors contain virtual/abstract member calls. (For us, it is unfortunately not an option to remove this code smell.) Using FIE, there is no way to setup calls to these members, which results either in an exception (MissingMethod) for the abstract member, or an modifiable behavior of the virtual member. We like FIE a lot for it's syntax, but so far this is a problem that we could only solve with converting to Moq. Moq, postpones the creation of the proxy type until `mock.Object` is called. Thus, I can make all the setups in front of the fake type generation/creation.

So far as explanation for my problem. Hope to see some solutions :)

blairconrad:
The changes requested in this issue are covered by the following:

* #368 (fixed in release 1.25.1), which provides better default behaviour for virtual (and abstract) methods called by constructors, and 
* #371 (in progress, as I type), which will allow configuration of virtual methods _before_ they are called by constructors

Accordingly, I'm closing this as a duplicate.

Issue #367, opened 2014-10-08
Bad exception thrown when fake's base's constructor fails


matkoch:
Related to #365, FIE throws a `FakeCreationException` when calling a fakable method (with struct return type) in the constructor. However, the current exception message isn't really helpful, because it doesn't outline that possibility.

```
FakeItEasy.Core.FakeCreationException : 
  Failed to create fake of type "ClassLibrary7.DomainType".

  Below is a list of reasons for failure per attempted constructor:
    No constructor arguments failed:
      No usable default constructor was found on the type ClassLibrary7.DomainType.
      An exception was caught during this call. Its message was:
      Exception has been thrown by the target of an invocation.
```

As long as #365 is not solved somehow, FIE should either state this situation in the exception, or, what i prefer, introduce a new exception type for virtual/abstract calls in the constructor. Also, the thrown `FakeCreationException` doesn't expose the inner exception - it is null. I think it could be wrapped.

blairconrad:
@matkoch, thanks very much for your work on this issue. Look for your name in the release notes. :trophy:

This issue has been fixed in [release 2.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/2.0.0).

PR #372, opened 2014-10-17
Allow fake configuration before faked type's constructor is called.

### LINKED ISSUE #371: Virtual calls should behave the same way during the constructor as after the constructor call
Given a class with the anti-pattern of virtual calls during the constructor.

```c#
public class Foo
{
    public Foo()
    {
        Console.WriteLine(this.VirtualMethod("call in constructor"));
    }

    public virtual string VirtualMethod(string parameter)
    {
        return "implementation value";
    }
}
```

When creating a fake of such a class, FIE does not process the calls *during* the constructor call. 

As a consequence ...
* The [default behavior](https://github.com/FakeItEasy/FakeItEasy/wiki/Default-fake-behavior) of virtual members isn't available during the constructor, i.e. `VirtualMethod()` returns null instead of the empty string in the sample above. Note that this is the cause for #368. (@blairconrad: and the deficiency has since been resolved in that issue)
* Calls during the constructor cannot be asserted, i.e. `A.CallTo (() => foo.VirtualMethod ("call in constructor")).MustHaveHappened();` will fail.
* The fake options Strict(), CallsBaseMethods(), and Wrapping() do not apply to virtual methods during the constructor call.
* Also scope configurations using Fake.CreateScope(IFakeObjectContainer) don't apply to virtual methods during the constructor call.

So at the moment FIE simply doesn't support virtual calls in the constructor(s) of faked classes.

I already prepared a PR for this issue and will propose it today.

ulrichb:
Fixes #371.

Before this PR the `FakeManager` is created / attached and configured (`IFakeObjectConfigurator`) by the `FakeObjectCreator` *after* the proxy object was created and returned by `IProxyGenerator`. That's too late for the c'tor call. The `Strict()` and `CallsBaseMethods()` fake options were applied *even later* in the `DefaultFakeAndDummyManager` when evaluating the `OnFakeCreatedActions`. The `Wrapping()` fake option was also applied in the `DefaultFakeAndDummyManager`.

The idea of this pull request is to encapsulate the creation and initial configuration into the new `LazyFakeManagerProvider`. The first `LazyFakeManagerProvider.Fetch(): IInterceptionSink` call initializes the fake manager and returns an interface, which is used by the proxy interceptor to notify the `FakeManager` about an intercepted call. The `IInterceptionSink` interface replaces the old mechanism using the `CallWasIntercepted` event, which was exposed by the interceptor.

The advantage of this solution is that during the execution of an constructor, the very first call to a virtual member invokes `LazyFakeManagerProvider.Fetch()` and therefore configures the fake manager just in time. If there was no virtual call during the c'tor, then a call to `LazyFakeManagerProvider. EnsureInitialized()` initializes the fake manager *after the c'tor* as it happened before this PR.

For the application of the `Strict()` and `CallsBaseMethods()` options, there are now new `OnFakeConfiguration` callbacks, which are evaluated inside the `LazyFakeManagerProvider`. The existing `OnFakeCreated` callbacks stay in the `DefaultFakeAndDummyManager` to avoid breaking changes. The handling of `Wrapping()` is now also inside `LazyFakeManagerProvider`. So in the end `LazyFakeManagerProvider` is a cohesive service, which does all configuration steps before giving the fake instance to the user.

An interesting side effect of the new `OnFakeConfiguration` callbacks is that this also solves #365, because we can do something like this ...

```c#
A.Fake<ClassWithVirtualCallsInConstructor>(o => 
    o.OnFakeConfiguration(f => 
        A.CallTo(() => f.VirtualMethod(A<string>._)).Returns("configured value")));
```

... which also applies to calls within the constructor.

Note:
* There is a new `ConstructorWithVirtualCallsSpecs` with the formalized specifications of #371.
* Because there is a lot going on in this PR I tried to avoid any unnecessary changes. After this PR is accepted, I will propose a few small refactorings.

### DIFF
--- a/Source/FakeItEasy.Specs/ConstructorWithVirtualCallsSpecs.cs
+++ b/Source/FakeItEasy.Specs/ConstructorWithVirtualCallsSpecs.cs
@@ -146,6 +146,21 @@
         };
     }
 
+    public class when_faking_a_class_which_calls_virtual_member_in_constructor_with_Wrapping_fake_option
+    {
+        static SimpleVirtualCallInConstructor fake;
+
+        Because of = () =>
+            fake = A.Fake<SimpleVirtualCallInConstructor>(
+                    o => o.Wrapping(new SimpleVirtualCallInConstructor()));
+
+        It should_delegate_to_the_wrapped_instance_inside_and_outside_of_the_constructor_call = () =>
+        {
+            fake.VirtualMethodValueDuringConstructorCall.Should().Be("implementation value");
+            fake.VirtualMethod(null).Should().Be("implementation value");
+        };
+    }
+
     public class when_faking_a_class_which_calls_virtual_property_members_in_constructor
     {
         static FakedClass fake;
--- a/Source/FakeItEasy.Specs/ConstructorWithVirtualCallsSpecs.cs
+++ b/Source/FakeItEasy.Specs/ConstructorWithVirtualCallsSpecs.cs
@@ -39,6 +39,22 @@
             A.CallTo(() => fake.VirtualMethod("call in constructor")).MustHaveHappened();
     }
 
+    public class when_faking_a_class_which_calls_virtual_member_in_constructor_with_changed_configuration_before_the_constructor
+    {
+        static SimpleVirtualCallInConstructor fake;
+
+        Because of = () => 
+            fake = A.Fake<SimpleVirtualCallInConstructor>(
+                    o => o.OnFakeConfiguration(
+                        f => A.CallTo(() => f.VirtualMethod(A<string>._)).Returns("configured value in fake options")));
+
+        It should_return_the_configured_value_inside_and_outside_of_the_constructor_call = () =>
+        {
+            fake.VirtualMethodValueDuringConstructorCall.Should().Be("configured value in fake options");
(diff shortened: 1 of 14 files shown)


blairconrad:
Happy belated New Year, @ulrichb.

There's been quite a lot of work on FakeItEasy since this PR was opened, so we're due for a rebase as well as some other adjustments:

1. you'll find that `when faking a class which calls virtual member in constructor with Strict option
» should throw an exception during the constructor call because of the unconfigured call` fails, as a result of the changes for #367 
1. as discussed above, we think that we should not add `OnFakeConfiguration`, and instead should just call `OnFakeCreation` earlier, but that this method should be renamed `ConfigureFake` to match what `IFakeConfigurator` uses
1. I can also reread the code to supply those "line notes" I promised on [November 14](https://github.com/FakeItEasy/FakeItEasy/pull/372#issuecomment-63157353).

Are you still interested in working on this PR? If so, then great, and we look forward to it. With its breaking changes, it's a perfect fit for the 2.0 release.

If you're not interested in continuing the work for any reason, then I'd be willing to take what you have now and perform the appropriate rebase and test-fixing and rename the methods and whatever to complete the fix. Your contributions would survive, and you'd get credit for them. 

What do you think?

adamralph:
Superseded by #447

PR #405, opened 2014-11-30
refactor: event raising implementation

adamralph:
The broken commit. For reference only (even just for me).


adamralph:
I still have the branch for comparison, should I need it.

PR #419, opened 2014-12-18
402 increase power of extensions

### LINKED ISSUE #402: increase power of IDummyDefinition and IFakeConfigurator
About a year ago, in https://github.com/FakeItEasy/FakeItEasy/issues/156#issuecomment-28899986 @patrik-hagne said

> … it would be simpler if the Task-dummies where implemented using IDummyDefinitions, I realize that there needs to be a change in the way dummy definitions work for that though

This stuck in my mind, and it's come back to me with the recent activity in #358. There's currently no good way to create dummy definitions for generic types—we only allow definitions for specific types.  `IFakeConfigurator` has the same restriction. The latter touches on the answer that @philippdolder and I gave for Stack Overflow question [FakeItEasy sometimes fails to create a fake when tests are run in parallel](http://stackoverflow.com/questions/26501952/fakeiteasy-sometimes-fails-to-create-a-fake-when-tests-are-run-in-parallel). Were we able to define a `IFakeConfigurator` that acted on any type, it would simplify the process of synchronizing _all_ fakes. There are likely other benefits.

I propose changing the interface definitions to allow user-specified type matching. I'll use `IDummyDefinition` as an example. Instead of

```csharp
public interface IDummyDefinition
{
    Type ForType { get; }
    object CreateDummy();
}
```

we could have 

```csharp
public interface IDummyDefinition
{
    int Priority { get; }
    bool CanCreateDummyOfType(Type type);
    object CreateDummyOfType(Type type);   
}
```

`Priority` would be used to choose between matching definitions, if there's more than one, just as `IArgumentValueFormatter.Priority` is. The methods should be self-explanatory.

This would be a breaking change, but the effects could be greatly mitigated by updating `DummyDefinition<T>`. I'll send a PR that shows how it can be changed so dummy definitions extending `DummyDefinition<T>` won't have to be touched.

blairconrad:
Fixes #402.


Source/FakeItEasy/Core/Cache.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// to create one, which will then be cached.
+        /// </summary>
+        /// <param name="key">The key of the value to get.</param>
+        /// <returns>The cached value for the key, or a newly-created and cached value.</returns>
+        public TValue this[TKey key]
+        {
+            get { return this.cachedValues.GetOrAdd(key, this.valueFactory); }
+        }
+    }
+}
```
blairconrad: Correction! We'd need no extension method on `ConcurrentDictionary`. `GetOrAdd` has the same semantics (and signature) of the `Intern` you suggest, and I see little value in adding an extension method just to change the name.

So I think the choices are: keep (something like) `Cache` or use `GetOrAdd`, no?

Source/FakeItEasy.IntegrationTests/DummyTests.cs
```diff
@@ -38,6 +39,33 @@ public void Correct_exception_should_be_thrown_when_dummy_is_requested_for_non_f
...
+        [SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "CanCreateDummyOfType",
+            Justification = "That's really the name of the method.")]
+        [Test]
+        public void Definition_for_type_should_be_cached_after_first_use()
+        {
+            var firstDummy = A.Dummy<Dummy>();
+
+            var secondDummy = A.Dummy<Dummy>();
+
+            Assert.That(secondDummy.Count, Is.EqualTo(firstDummy.Count), "CanCreateDummyOfType was called the wrong number of times.");
```
adamralph: How about something like:

```C#
[Test]
public void Dummy_Definitions_are_polled_only_once()
{
    A.Dummy<Dummy>();
    A.Dummy<Dummy>();

    DummyTestsDummyDefinition.TimesCanCreateDummyOfTypeWasCalled.Should().Be(1);
}
```

Does that test enough?

Source/FakeItEasy/Core/Cache.cs
adamralph: Aha! Well I think that's a no brainer then: we should use `ConcurrentDictionary<TKey, TValue>.GetOrAdd`.

Source/FakeItEasy.IntegrationTests/DummyTests.cs
blairconrad: Regrettably, no. I'd've preferred it as well. 

`CanCreateDummyOfType` will be called on a dummy definition when we're looking for any (not yet cached) dummy definition to use. So if other tests have already tried to make dummies, there's a chance that the framework will have already called `CanCreateDummyOfType`.

Right now if I make the change, it fails, as the method has been called 5 times, not 1.

Of course, as I type this, I realize that I can reset the count in the test. Two secs.

```csharp
[Test]
public void CanCreateDummyOfType_should_not_be_called_again_after_definition_for_a_dummy_type_is_found()
{
    // Clear the count, since CanCreateDummyOfType may have been called for other types
    // in other tests.
    DummyTestsDummyDefinition.TimesCanCreateDummyOfTypeWasCalled = 0;

    A.Dummy<Dummy>();
    DummyTestsDummyDefinition.TimesCanCreateDummyOfTypeWasCalled.Should()
        .Be(1, "we need to call CanCreateDummyForType to create the Dummy");

    A.Dummy<Dummy>();

    DummyTestsDummyDefinition.TimesCanCreateDummyOfTypeWasCalled.Should()
        .Be(1, "the definition should've been cached, so there's no need to call CanCreateDummyForType");
}
```

The first  `Should` is practically a precondition test, which I'm not always so keen on. Say the word and I'll strike it.

adamralph: Ah, I see. I'm not keen in the resetting either so I think your original suggestion was best.

Perhaps we can trim the names down a bit? E.g.:

```csharp
[Test]
public void Definition_for_type_should_be_cached_after_first_use()
{
    A.Dummy<Dummy>();
    var firstCallCount = DummyTestsDummyDefinition.CanCreateDummyOfTypeCallCount;

    A.Dummy<Dummy>();
    var secondCallCount = DummyTestsDummyDefinition.CanCreateDummyOfTypeCallCount;

    firstCallCount.Should().Be(
        secondCallCount, "once a dummy has been created for a type, its dummy definition should be cached");
}
```

blairconrad:
Thanks for the great feedback, @adamralph. I think I've addressed everything in the latest push.
Note that if/when this is merged, I'll need to rebase #421, since I grabbed `BeAFake()` from that PR.

adamralph:
Great work @blairconrad! As usual.

PR #427, opened 2015-01-08
#425 remove obsolete type members

### LINKED ISSUE #425: Remove obsolete type members from API
Spun off from https://github.com/FakeItEasy/FakeItEasy/issues/410

blairconrad:
Fixes #425.

These are all the _obsolete_ type members I could find. There are other types that could be made internal or moved to different namespaces, though. Unless there's an objection, I'll make another issue for those.

### DIFF
--- a/Source/FakeItEasy.Tests/ATests.cs
+++ b/Source/FakeItEasy.Tests/ATests.cs
@@ -3,6 +3,7 @@
     using System;
     using System.Collections.Generic;
     using FakeItEasy.Creation;
+    using FluentAssertions;
     using NUnit.Framework;
 
     [TestFixture]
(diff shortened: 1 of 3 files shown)


Source/FakeItEasy.Tests/ATests.cs
```diff
@@ -79,17 +66,17 @@ public void CollectionOfFakes_should_delegate_to_fake_creator()
             A.CallTo(() => creator.CollectionOfFake<IFoo>(10)).Returns(returnedFromCreator);
 
             // Act
-            var result = A.CollectionOfFake<IFoo>(10);
+            object result = A.CollectionOfFake<IFoo>(10);
```
blairconrad: FluentAssertions wouldn't give me .BeSameAs() so long as it knew that result was a collection. It was this or an ugly cast on (new) line 72.

Issue #280, opened 2014-03-27
Overhaul project naming


adamralph:
My proposal:

![image](https://cloud.githubusercontent.com/assets/677704/2536072/bfbefddc-b598-11e3-83ee-65b3ecc41c5c.png)

Points to note:

* What we are currently calling 'specs' I tend to refer to as 'acceptance tests'. I.e. black box tests written in the language of the public API which ignore all implementation details
* What we are currently calling 'integration tests' are what I tend to refer to as 'component tests'. I.e. semi-white box testing of subsets of the implementation details behind the public API
* Unit test remain as the name for white box testing of individual classes which may or may not form part of the public API
* On disk, all test projects will be moved to a 'tests' subfolder of 'Source' (or 'src' if we agree on #279)
* The framework suffix is mandatory on all projects to avoid ambiguity and the suffixes used match the nuget framework identifiers but using .NET casing rules
* 'FakeItEasy.IntegrationTests.External' is removed and the formatter it contains is moved to 'FakeItEasy.Examples.Net40'. If we choose to implement #281 then we will need to do something else. Perhaps we can keep the project and just rename it to 'FakeItEasy.Tests.Component.External.Net40'.
* The rake tasks `spec` and `integ` would be renamed to `accept` and `component`

Looking forward to hearing any opinions on this.

blairconrad:
Instead of separate projects for the various framework projects, would we ever consider one project (of each type - the production code, unit tests, component tests, acceptance tests) with separate configurations for the frameworks? I know the magic ["include lots of files" trick](https://github.com/xbehave/xbehave.net/blob/011a9b11290a8f21a39dfe4e3b2ae02efed4c7c5/src/Xbehave.Net45/Xbehave.Net45.csproj#L67) can make it easy to include all the code (and all the tests) for the framework-specific projects, but so does multiple configurations, and I thought this might be another option.
It seems to work well enough for Castle.Core, and makes for a smaller/more manageable Solution Tree in the IDE.

adamralph:
IIRC I tried that initially with xbehave but soon went off the idea and switched to separate projects instead. The exact reasoning escapes me however.

I think it might work quite well when you only have progressive versions of the desktop CLR but once you started mixing in Silverlight and Win8 etc. the configs start to diverge a lot and all of a sudden you have to try and move stuff which normally sits outside the 'debug' and 'release' config properties to inside them and things can start to get a bit hairy. 

[...]

PR #430, opened 2015-01-10
fixing aberrant test class namespaces

blairconrad:
While working on the "internalizing" issue, I broke the Win8/Win8.1 tests and couldn't tell which tests were failing at first, due to the less-than-ideal namespace. So I figured, "let's fix those up while we're deconfusing ourself".

### DIFF
--- a/Source/FakeItEasy-SL.Tests/FakeItEasy-SL.Tests.csproj
+++ b/Source/FakeItEasy-SL.Tests/FakeItEasy-SL.Tests.csproj
@@ -9,7 +9,7 @@
     <ProjectTypeGuids>{A1591282-1198-4647-A2B1-27E5FF5F6F3B};{fae04ec0-301f-11d3-bf4b-00c04f79efbc}</ProjectTypeGuids>
     <OutputType>Library</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
-    <RootNamespace>FakeItEasy.Tests</RootNamespace>
+    <RootNamespace>FakeItEasy_SL.Tests</RootNamespace>
     <AssemblyName>FakeItEasy-SL.Tests</AssemblyName>
     <TargetFrameworkIdentifier>Silverlight</TargetFrameworkIdentifier>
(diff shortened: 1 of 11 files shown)


adamralph:
Given the current state of the test projects, this makes sense. However, I don't think the current state of the test projects makes sense :wink:. IMO the content of these projects should be identical. Otherwise, what are they really testing?

Linked files are usually the best way to achieve this and is the reason why, in most cases, these kinds of projects share identical namespaces, e.g. https://github.com/xbehave/xbehave.net/tree/dev/src/test.

Ultimately I think we should move to that model. We can do this right now for the Net35 project but as long as we are on NUnit, we can't do it for the Win8, Win81 or SL projects. Later, we can consider switching to xunit 2.0 (when it goes stable) so we can do this for all the projects.

Thoughts?

Incidentally, we should probably rename `FakeItEasy-SL.Tests` to `FakeItEasy.SL5.Tests`. Or should it be `FakeItEasy.Sl5.Tests`?. We should also rename `FakeItEasy.Tests` to `FakeItEasy.Net40.Tests`. More on that over at https://github.com/FakeItEasy/FakeItEasy/issues/280.

blairconrad:
First, from your xbehave.net link:

```xml
<ItemGroup>
    <Compile Include="..\Xbehave.Test.Acceptance.Net40\**\*.cs">
    <Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
    </Compile>
</ItemGroup>
```
I did not know that this was a thing. Assuming it does what it looks like it does, my mind is blown.

Your point about the more comprehensive #280 and renaming the Silverlight tests is a good one. I can muddle by with what we have until we get there. :smile:

Sorry for wasting your time.

Oh, and regarding SL/Sl: two-letter acronyms should have both letter capitalized, no? So based on that, I'd say "SL". Of course, "Silverlight" looks kind of like one word, so maybe "S". But that's ridiculous. Me, I'd go with SL. But we've time for that.

adamralph:
I got that trick from the xunit 2.0 codebase :wink:

Yeah, I think you're right. It should be SL5.

PR #431, opened 2015-01-12
428 internalize types and members

### LINKED ISSUE #428: Internalize types and members that don't need to be exposed

blairconrad:
Fixes #428.

One member—`ArgumentCollection.Empty` was removed, not internalized.

`ITaggable`, which I'd thought could be internalized, couldn't be so much. Doing so broke fake-making on SL5, something for which Castle.Core has no solution (there may not be one). I thought to remove it, and attach FakeManagers to DynamicProxies via the same mechanism as we use for Delegates, but that breaks serialization of fakes (fake Delegates can't be faked). So I'm stuck. I think we should at least finish off #428, and then either leave ITaggable or retackle later.


blairconrad:
Crafty. Yes, I think so.

PR #438, opened 2015-01-14
432 improve namespaces

### LINKED ISSUE #432: move types out of FakeItEasy namespace where appropriate
Scope to be determined from #374.

blairconrad:
Fixes #432.

### DIFF
--- a/Source/FakeItEasy.Tests/Configuration/ArgumentCollectionTests.cs
+++ b/Source/FakeItEasy.Tests/Configuration/ArgumentCollectionTests.cs
@@ -1,10 +1,11 @@
-﻿namespace FakeItEasy.Tests
+﻿namespace FakeItEasy.Tests.Configuration
 {
     using System;
     using System.Collections;
     using System.Collections.Generic;
     using System.Globalization;
     using System.Linq;
+    using FakeItEasy.Configuration;
     using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
     using NUnit.Framework;
--- a/Source/FakeItEasy.Tests/Configuration/ArgumentCollectionTests.cs
+++ b/Source/FakeItEasy.Tests/Configuration/ArgumentCollectionTests.cs
@@ -33,7 +33,7 @@
 
             var arguments = new ArgumentCollection(new object[] { "foo", "bar" }, method);
 
-            Assert.That(arguments.ArgumentNames.SequenceEqual(new string[] { "argument", "argument2" }));
+            arguments.ArgumentNames.ShouldBeEquivalentTo(new[] { "argument", "argument2" });
         }
 
         [Test]
@@ -51,7 +51,7 @@
         {
             var arguments = this.CreateFakeArgumentList("foo", "bar");
 
-            Assert.That(arguments.Get<string>(1), Is.EqualTo("bar"));
+            arguments.Get<string>(1).Should().Be("bar");
         }
 
         [Test]
@@ -59,7 +59,7 @@
         {
             var arguments = this.CreateFakeArgumentList(new[] { "foo", "bar" }, 1, 2);
 
-            Assert.That(arguments.Get<int>("bar"), Is.EqualTo(2));
+            arguments.Get<int>("bar").Should().Be(2);
         }
 
         [Test]
@@ -67,8 +67,9 @@
         {
             var arguments = this.CreateFakeArgumentList(new[] { "foo", "bar" }, 1, 2);
 
-            Assert.Throws<ArgumentException>(() =>
-                arguments.Get<int>("unknown"));
+            var exception = Record.Exception(() => arguments.Get<int>("unknown"));
+
+            exception.Should().BeAnExceptionOfType<ArgumentException>();
         }
 
         [Test]
@@ -76,7 +77,7 @@
         {
             var arguments = this.CreateFakeArgumentList(new[] { "foo", "bar" }, 1, 2);
 
-            Assert.That(arguments.Count, Is.EqualTo(2));
+            arguments.Count.Should().Be(2);
         }
 
         [Test]
@@ -91,7 +92,7 @@
                 found.Add(enumerator.Current);
             }
 
-            Assert.That(found, Is.EquivalentTo(new object[] { 1, 2 }));
+            found.ShouldBeEquivalentTo(new object[] { 1, 2 });
         }
 
         [Test]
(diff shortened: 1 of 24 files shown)


Source/FakeItEasy.Tests/Configuration/ArgumentCollectionTests.cs
```diff
@@ -98,22 +100,24 @@ public void GetEnumerator_returns_enumerable_that_enumerates_the_arguments_of_th
...
 
         [Test]
-        public void Should_be_serializable()
+        public void Should_be_serializable_when_arguments_are_equatable()
         {
             // Arrange
-            var collection = new ArgumentCollection(new object[] { new object() }, typeof(object).GetMethod("Equals", new[] { typeof(object) }));
+            object collection = new ArgumentCollection(
+                new object[] { "first argument" },
+                typeof(string).GetMethod("Equals", new[] { typeof(string) }));
```
blairconrad: * var → object to so FluentAssertions will let me `BeBinarySerializable`.
* changed the type of the argument because FluentAssertions is stricter than NUnit when checking binary serializability—it compares the objects after deserialization, and a `new object()` would not compare properly

Source/FakeItEasy.Tests/Configuration/ArgumentCollectionTests.cs
```diff
@@ -32,7 +33,7 @@ public void Constructor_that_takes_method_should_set_argument_names()
 
             var arguments = new ArgumentCollection(new object[] { "foo", "bar" }, method);
 
-            Assert.That(arguments.ArgumentNames.SequenceEqual(new string[] { "argument", "argument2" }));
+            arguments.ArgumentNames.ShouldBeEquivalentTo(new[] { "argument", "argument2" });
```
adamralph: By using `ShouldBeEquivalentTo()` we're losing the assertion of order. We should be using `Should().Equal()` instead.

adamralph:
The third refactoring commit should really only either touch tests or non-tests, not both, but in this case the changes are low risk so I'm OK to merge as is.

Source/FakeItEasy.Tests/Configuration/ArgumentCollectionTests.cs
adamralph: Looks fine to me.

adamralph:
I've finished reviewing the PR. Looks good apart from one thing - https://github.com/FakeItEasy/FakeItEasy/pull/438#discussion-diff-23125075

Source/FakeItEasy.Tests/Configuration/ArgumentCollectionTests.cs
blairconrad: *gasp!* You're right. How embarrassing. Will amend.
There are two other `ShouldBeEquivalentTo`s in here (that were originally `Is.EquivalenTo`) that I think should also care about order as well. I will change them all.

blairconrad:
> (refactoring) should really only either touch tests or non-tests, not both

Oh? In all my time here, I've never made any attempt to restrict the cleanup to either tests or non-tests. I don't mind, and would even have been happy to separate this one, but I'm curious—have we just gotten lucky up 'til now and it's never come up?

adamralph:
It's about risk. By leaving one of the tests or non-tests constant when refactoring you lessen the chance that you inadvertently change behaviour. I think we've been lucky so far ;-). Mind you, if you don't verify each commit on the build server then it's still up to the PR sender to verify that each refactoring commit is still green. On that note, it might be worth switching our CI config to build each commit separately. We tend not to have PR's with a huge number of commits and our build time is relatively small.

blairconrad:
I understood the motivation, once you mentioned it. It's just that it hadn't seemed to be important to you up 'til now, so I was a little surprised. I've no objection.
To tell the truth, I think the risk would be similar when we're making the commits that are actually intended to fix issues, but we routinely change the production code and the tests in a single commit.

If we had the CI server build each commit separately, and someone had the habit of committing red-green-refactor, we'd get 1/3 of the commits failing. I assume the CI server would ultimately give a pass and we'd just be able to look at the history for our own information?

adamralph:
Regarding the CI server, yes, the red commits would indeed be red, but I don't think that's necessarily bad, (perhaps it's good?).

So here's my take on red-green-refactor, after having spent most of the last year practising it almost exclusively:

For a red commit, in the extreme, you could say that any properly formed test code is valid, whether or not it compiles. Another approach is to require that the tests compile, and when run they give meaningful failure messages. The latter approach is what I favour. Now, with something like scriptcs, an executable, it's possible to have fully compiling and 'properly failing' acceptance (black box) tests with a red commit which only touches the test code. This is because the acceptance test project doesn't even reference the executable. Each acceptance test prepares an appropriate working directory, invokes the exe with relevant arguments and asserts the outcomes based on the resulting state of the file system and what it captures from stdout and stderr. With an API, e.g. FakeItEasy, you need to at least alter the surface area of the API in the same commit in order to satisfy the compilation and failure constraints. For me, this is an acceptable approach and low risk enough. The changes to the non-acceptance test code should be as simple and crude as possible just to get the code to compile. This includes any white box tests (unit and component).

When making a green commit, it's important that acceptance tests are not touched since the acceptance criteria must not change during the writing of the green commit. If that happens, then the red commit was flawed and should be amended before the green commit is made. Unit and component tests however, if present, must change by necessity as part of the green commit, since they are concerned with implementation, which is what the green commit is all about.

With refactoring commits, changes to black box tests must be mutually exclusive with all other changes. This is the only way to ensure that you really are refactoring and not 'refucktoring', which is what Greg Young calls a refactoring commit which changes acceptance tests and production code. I.e. the unchanged state of one validates the changes to the other. If refactoring, you are not changing the acceptance criteria, so this makes sense. Again, white box tests are separate from black box tests, so changes to them would be done in the same commit as the changes to the production code.

Aaaaaaand I'm done. That's really a blog post :stuck_out_tongue_closed_eyes:. I'm sure you're already familiar with most if not all of the above but I thought I may as well lay my cards on the table. It may even be overkill for FakeItEasy but I find it interesting nonetheless :smile:.

blairconrad:
> that's really a blog post"

Go for it!

Everything you say makes sense to me, and I'm mostly familiar with the reasoning. Still, it's interesting, and I understand the usefulness of separating the changes to the production and test code.
(And I even think that if you _wanted_ to, you could extend the discipline so that commits touch either production code or white box tests, but not both. I'm not advocating for that, just saying that it's an option.)

I'm not against  adopting the method you describe, but also wonder if it's not too much for FakeItEasy, or really any open source project. I am happy to comply with any process that we all think is reasonable, but worry about introducing additional barriers to contributors.

I hesitate to bring this up, as it sounds like arguing about your comment on 7dc539b92cdeac5021ea136b2b8adcb6fc2c9197, which I'm not trying to do, but the terms are a little funny here. If we consider only the Specs to be black box tests, then the commit did only touch production code and white box tests. My guess is that you commented because many of our "unit" tests are operating at a white box level. Unfortunately, this does make it much harder to determine when we're mixing change that we shouldn't. Again, not fighting, and not looking for an action item here. Just food for our collective thought, I think.

adamralph:
Yeah, I also thought a little about applying the same thing to prod/white box tests and I think that it many cases that's possible. E.g. when you refactor the internals of a test or the internals of a type. I think there are other cases where you couldn't necessarily make the separation e.g. you change some internal types around completely, and some or all may have unit tests.

Regarding project policy, I agree that something like this, in its entirety is probably too much to insist on from contributors. I think elements of it may be applicable. E.g. the separation of refactoring commits into back box/other. But insisting on separate red/green for instance is probably too much. Even then, I'd probably only *insist* on refactoring separation if it was quite a huge refactoring commits. The size of the most of the PR's we receive fall well under this threshold.

You're absolutely right about my original comment that brought about this conversation, it's in fact applying the richer rules around white box testing that we've only just started to talk about :smile:.

PR #444, opened 2015-01-25
allow faking of types implementing IDisposable

### LINKED ISSUE #440: Cannot fake types implementing IDisposable
**Update by @adamralph**

See https://github.com/FakeItEasy/FakeItEasy/issues/440#issuecomment-71312010. When a type implements `IDisposable` correctly, and that type is faked, the call to `virtual Dispose(false)` from the finalizer of the type causes an `ArgumentException` to be thrown.

----

Hi there,
yesterday we updated from FakeItEasy 1.14.0 to FakeItEasy 1.25.1.
Now we have several errors during our unit-test teardown.
We pointed out that the error occurs when calling any faked method in an object finalizer.

```csharp
internal class Program {
		private static void Main(string[] args) {
			new Foobar(A.Fake<IDisposable>());
			GC.Collect();
			GC.WaitForPendingFinalizers();
		}
	}

	internal class Foobar : IDisposable {
		private readonly IDisposable _something;
		private bool _disposed;

		public void Dispose() {
			Dispose(true);
			GC.SuppressFinalize(this);
		}

		~Foobar() {
			Dispose(false);
		}

		public Foobar(IDisposable something) {
			_something = something;
		}

		protected virtual void Dispose(bool free_managed) {
			if (_disposed) {
				return;
			}

			if (free_managed) {
			}
			_something.Dispose();
			_disposed = true;
		}
	}
```

error

blairconrad:
Fixes #440.

### DIFF
--- a/Source/FakeItEasy.Specs/DisposableSpecs.cs
+++ b/Source/FakeItEasy.Specs/DisposableSpecs.cs
@@ -1,0 +1,55 @@
+﻿namespace FakeItEasy.Specs
+{
+    using System;
+    using FluentAssertions;
+    using Machine.Specifications;
+
+    public class when_faking_a_disposable_class
+        : EventRaisingSpecs
+    {
+        static IDisposable fake;
+        private static Exception exception;
+
+        Establish context = () =>
+        {
+            AppDomain.CurrentDomain.UnhandledException += UnhandledExceptionHandler;
+
+            fake = A.Fake<SomeDisposable>();
+        };
+
+        Because of = () =>
+        {
+            fake = null;
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+        };
+
+        It should_not_throw_when_finalized = () => exception.Should().BeNull();
+
+        Cleanup after = () =>
+        {
+            AppDomain.CurrentDomain.UnhandledException -= UnhandledExceptionHandler;
+        };
+
+        private static void UnhandledExceptionHandler(object sender, UnhandledExceptionEventArgs e)
+        {
+            exception = (Exception)e.ExceptionObject;
+        }
+
+        public abstract class SomeDisposable : IDisposable
+        {
+            ~SomeDisposable()
+            {
+                this.Dispose(false);
+            }
+
+            public void Dispose()
+            {
+                this.Dispose(true);
+                GC.SuppressFinalize(this);
+            }
+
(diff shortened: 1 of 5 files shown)


Source/FakeItEasy.Specs/DisposableSpecs.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            fake = A.Fake<SomeDisposable>();
+        };
+
+        Because of = () =>
+        {
+            fake = null;
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+        };
```
adamralph: Does this fail without the fix? I guess it probably does, but the reason I put in the extra collects in https://github.com/FakeItEasy/FakeItEasy/issues/440#issuecomment-71342374 was to ensure that the deallocation of everything else happens *before* the finalization of the fake. Of course, `GC.Collect()` is, contrary to popular belief, only a *hint* to the GC that it can perform collection so I guess this doesn't make a lot of difference anyway.

adamralph: I took it for a whirl locally and it does indeed fail (not that I doubted you, I just wanted to see it with my own eyes :wink:). I suppose, in this case, it's impossible to write a test which is *guaranteed* to fail, only one which *happens* to fail :confused:.

adamralph: But at least, in our current environments, it has proved that the fix works, and that's the most important thing.

blairconrad: And I thought I was the champion of "conversations with myself in fora and PRs".

Yup, it failed. Your way looked good too, but since it was more complicated and, as you say, these sorts of tests can never _prove_ anything, I figured, "go for the low line count".
I am in _no way_ offended that you tried it yourself. Prudent.

To tell the truth, I didn't understand why you had so many collects. What was the benefit of checking that "the deallocation of everything else happens before the finalization of the fake"? The fake's interceptor holds the `FakeManager`, so everything that _mattered_ would all still be live until we break reachability of the fake.

And thanks for the speedy review and merge!

adamralph: I think me having so many collects was an effect of my ignorance of what's happening internally :stuck_out_tongue:

Issue #442, opened 2015-01-24
1.25.2 Release


adamralph:
**Ready** when all other issues forming part of the release are **Done**.

- [x] run code analysis in VS in *Release* mode and address violations (send a regular PR which must be merged before continuing)
- [x] raise PR from hotfix branch to master - https://github.com/FakeItEasy/FakeItEasy/pull/445
- [x] merge hotfix to master - https://github.com/FakeItEasy/FakeItEasy/pull/445
- [x] if necessary, change `VERSION_SUFFIX` on [CI Server](http://teamcity.codebetter.com/admin/editBuildParams.html?id=buildType:bt929)
      to appropriate "-beta123" or "" (for non-betas) value and initiate a build
- [x] check build
-  edit draft release in [GitHub UI](https://github.com/FakeItEasy/FakeItEasy/releases):
    - [x] complete release notes, mentioning non-owner contributors, if any
    - [x] attach nupkg
    - [x] publish the release
- [x] push NuGet package
- [x] copy release notes from GitHub to NuGet
- [x] de-list pre-release or superseded buggy NuGet packages if present (copy any release notes forward to the new version)
- [x] update website with contributors list (if in place)
- [x] tweet, mentioning contributors and post link as comment here for easy retweeting ;-) - https://github.com/FakeItEasy/FakeItEasy/issues/440#issuecomment-71358769
- [x] post tweet in JabbR ([fakeiteasy][1] and [general-chat][2]) and Gitter ([FakeItEasy/FakeItEasy][3])
- [x] post links to the NuGet and GitHub release in each issue in this milestone, with thanks to contributors
- [x] close all issues on this milestone
- [ ] close this milestone

[1]: https://jabbr.net/#/rooms/fakeiteasy
[2]: https://jabbr.net/#/rooms/general-chat
[3]: https://gitter.im/FakeItEasy/FakeItEasy

adamralph:
@blairconrad I've gone as far as I dare without getting your seal of approval to perform the release. I have dash now so if you're happy to release then please go ahead, otherwise I can pick it up again later.

blairconrad:
This issue has been fixed in release 1.25.2.

https://www.nuget.org/packages/FakeItEasy/1.25.2
https://github.com/FakeItEasy/FakeItEasy/releases/tag/1.25.2

blairconrad:
https://twitter.com/fakeiteasyfx/status/559320796418805760

blairconrad:
[...]

PR #445, opened 2015-01-25
merge Hotfix/1.25 to master

adamralph:
re #442

### DIFF
--- a/Source/CommonAssemblyInfo.cs
+++ b/Source/CommonAssemblyInfo.cs
@@ -5,9 +5,9 @@ using System.Resources;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyVersion("1.25.1")]
-[assembly: AssemblyFileVersion("1.25.1")]
-[assembly: AssemblyInformationalVersion("1.25.1")]
+[assembly: AssemblyVersion("1.25.2")]
+[assembly: AssemblyFileVersion("1.25.2")]
+[assembly: AssemblyInformationalVersion("1.25.2")]
 
 [assembly: AssemblyCompany("Patrik Hägne")]
 [assembly: AssemblyProduct("FakeItEasy")]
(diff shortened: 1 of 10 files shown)


adamralph:
It wil have to be manual because of the conflicting version change.

blairconrad:
@adamralph, I have merged in my local repo by checking out the branch into master and creating a new commit.
Then I reverted the version number changes and resolved the (dead simple) conflict in FakeItEasy.Specs.csproj.

As much as it pains me, it looks like this:

image

I can push. 

On the other hand, could we not have just merged 440-disposable? It makes a cleaner graph and we don't have to undo the version change…

image

(pretend "t" is master)

Happy to conform. Just curious.

adamralph:
I would prefer the history to show a merge from the patch branch to master as it's the simplest way for the history to convey that patches have made back into the mainline branch. Otherwise you have to manually match up what went into the patch branch and what went into the master branch each time you want to browse that part of the history. The way we do this at work is that manual merges (if any are required) are done by a pair to satisfy the four eyes rule, but I guess that doesn't really carry over so well to the OSS world.

An alternative would be to create a new branch on master, merge the hotfix into it, correcting any conflicts, and then send a PR from that new branch into master. I guess that would satisfy all requirements of our workflow but I wonder if it's over complication? I guess it could be worth a try to see how it looks. Not that I envisage us having to do this kind of thing too much in the future. Or at least hopefully not :wink:.

blairconrad:
So at your day job, you do the "change the version number/change it back" dance?
If it works for you, I defer to your experience. (In Subversion, we've got a whole other headache around merging fixes.)
If you've no objection to how my first screenshot looks, I'm content to push.

For your suggested alternative, would we merge just my two commits (005807b806674c2f1992d324f83955b7be75e822 and fe39e93fd5768d572bef52bba5798655c5cbdab7) to the branch off master, or did you mean merge the whole patch branch? I think the latter would look like my first screenshot, only with more lines.

adamralph:
I would merge the whole match branch. Sec, I'll prepare an alternate PR to see how that looks.

adamralph:
Alternate PR: https://github.com/FakeItEasy/FakeItEasy/pull/446

PR #447, opened 2015-01-31
Allow configuration of methods before the fake is created

### LINKED ISSUE #371: Virtual calls should behave the same way during the constructor as after the constructor call

### LINKED ISSUE #454: rename `IFakeOptionsBuilder.OnFakeCreated` to `ConfigureFake`
Spun off from #371, as a result of https://github.com/FakeItEasy/FakeItEasy/pull/447#issuecomment-72695211. Reflects the fact that the call be made before the fake is created.

blairconrad:
Fixes #371.
Fixes #454.
Supersedes #372.

I've included the commits from the previous pull request and:

* fixed test errors introduced by changes that had been merged to master while the PR was open (ordinarily I'd've fixed those in existing commits when rebasing, but I added a877d022c5f0c1429f7a76e4045dc74571b8b3a1 because I was afraid that editing an earlier commit would somehow taint the earlier commits and @ulrichb wouldn't get credit anymore)
* consolidated `OnFakeConfiguration` and `OnFakeCreated`, merging into `ConfigureFake`
* mucked about, tidying


adamralph:
I'm on this one. Do you think we should specifically make this part of 2.0? Or equally happy to have it in 2.1?

blairconrad:
We're renaming `OnFakeCreated` to `ConfigureFake`. I think we have to have it in a package with a  major version bump.

adamralph:
Ah OK Should we mark #371 as breaking? Or should we spin off another issue and mark this PR as closing it? Is the rename actually a requirement of the feature? (I forget)

blairconrad:
The rename isn't a requirement of the feature. It would've featured along using `OnFakeCreated`, but it would've been misleading.

One could consider #371 to be breaking even without the rename since the configuration actions may be applied earlier than before, and multiple times, although the likelihood that anyone will be affected is very low.

The extra issue is not a bad idea. It would make the rename more visible.

Source/FakeItEasy.Specs/ConstructorWithVirtualCallsSpecs.cs
```diff
@@ -106,16 +260,9 @@ public FakedClass()
             {
                 this.StringPropertyValueDuringConstructorCall = this.StringProperty;
                 this.ValueTypePropertyValueDuringConstructorCall = this.ValueTypeProperty;
-                this.InterfacePropertyMethodResultDuringConstructorCall = this.InterfaceProperty.InnerProperty;
 
                 this.StringProperty = "value set in constructor";
                 this.ValueTypeProperty = 123456;
-                this.InterfaceProperty.InnerProperty = "value set in constructor";
```
ulrichb: The idea of the `InterfaceProperty`-tests was to prove even the fake *creation*+configuration works within virtual c'tor calls (using an virtual interface property as an example). Is this not a valid corner case?

Source/FakeItEasy.Specs/ConstructorWithVirtualCallsSpecs.cs
```diff
@@ -65,38 +92,165 @@ public class when_faking_a_class_which_calls_virtual_member_in_constructor_withi
...
+
+        It should_return_the_configured_value_during_the_constructor =
+            () => fake.VirtualMethodValueDuringConstructorCall.Should().Be("configured value of strict fake");
+
+        It should_return_the_configured_value_after_the_constructor =
+            () => fake.VirtualMethod(null).Should().Be("configured value of strict fake");
     }
 
-    public class when_faking_a_class_which_calls_virtual_property_members_in_constructor
+    public class when_wrapping_an_object_in_a_fake_that_calls_virtual_member_in_constructor
```
ulrichb: Are these new "feature intersection" tests really directly related to the "virtual calls c'tor" stuff (also the previous `when_configuring_a_method_called_by_a_constructor_that_is_also_configured_by_a_fake_configurator
`)? If not, move them into a new spec, maybe "ConfigurationOptionsSpec"?

Source/FakeItEasy.Specs/DomainEvents.cs
```diff
@@ -1,8 +1,26 @@
 ﻿namespace FakeItEasy.Specs
 {
+    using System;
```
ulrichb: Until now this (test data) file (`DomainEvents.cs`) has a far more specific purpose (in the configuration/factory specs). Do we really want to reuse it in tests regarding the virtual calls in a c'tor?

Also: Shouldn't we give it a name where the intended purpose becomes clearer, anyways?

Source/FakeItEasy.Specs/ConstructorWithVirtualCallsSpecs.cs
blairconrad: The benefit the test brought appeared to be that `InterfaceProperty`'s behaviour is properly configured by the time it's used during the constructor—it returns a fake `IInterface`. It seemed to me that this is no different than returning the proper values from the other properties, one of which was a value type and one a reference type.

Source/FakeItEasy.Specs/ConstructorWithVirtualCallsSpecs.cs
blairconrad: I think so. Were we we not making the virtual calls in the constructor, the tests would be moot, no?

Source/FakeItEasy.Specs/DomainEvents.cs
blairconrad: I can go either way regarding reusing sample classes between test groups.
I suppose it opens the door for clashing tests (although all one has to do is add more tests that use the class until there's a problem and _then_ make a new one), but it also helps with what would otherwise be a proliferation of test classes.

As for the name, if the class _is_ to be shared, then naming becomes more difficult, as one has to keep renaming the class every time it's used for a new purpose. Me, I think the class names provide info about the intent and capabilities of the classes used therein, but the authors always think their makes sense!

That being said, other than a perhaps too-strong attachment to the phrase "RunsAmok" in the one class name, I'm not bothered at the prospect of adding extra classes, or renaming, if this is the popular opinion.

Source/FakeItEasy.Specs/ConstructorWithVirtualCallsSpecs.cs
ulrichb: What I meant with "feature intersection tests" are the additional test cases you added. E.g. "Wrapping+ConfigureFake";  "ConfigureFake+Wrapping"; "Wrapping+FakeConfigurator". Are these *additional* cases directly related to "virtual calls in c'tor"? If not moving them into a new spec reduces the complexity of `ConstructorWithVirtualCallsSpecs`.

blairconrad: No worries, "feature intersection tests" was clear to me.

As you point out, they do test the interaction between two features or aspects of FakeItEasy, which causes a problem. Where should tests of interacting features live? With tests for one feature? For the other?

They could move somewhere else, but I don't think "ConfigurationOptionsSpec" is necessarily the answer for all, although some would be a good fit. But I'd be more inclined to name the spec closer to the interface involved, maybe FakeOptionsBuilderSpecs. Perhaps that's too close to the implementation and not how the user would think of it, though. Still, these specs aren't user docs.

Truth be told, "ConstructorWithVirtualCallsSpecs" never really sat _that_ well with me as a concept, as in and of itself, a constructor with virtual calls isn't something that's part of FakeItEasy. Sure, it's important to know how we deal with those things, but I would've approached it from a "feature" direction, rather than a "property of the thing we're faking" direction. However, I didn't make a fuss because
1. I'm new to the whole Acceptance Test (or Spec) milieu, and 
2. grouping these classes is sort of like naming them, and that is one of the [Two Hard Things](http://martinfowler.com/bliki/TwoHardThings.html).

However, now that we've opened the discussion, I might as well propose a full-blown alternative, for comment. Were I to start moving things around on my own, I might do something like this (some classes would need to be renamed, and some _existing_ test classes might move, for example `when_asserting_must_have_happened_when_did_not_happen`):

- FakeConfiguratorSpecs
    - when_configuring_a_method_called_by_a_constructor_that_is_also_configured_by_a_fake_configurator
- FakeOptionsBuilderSpecs
    - when_configuring_a_method_called_by_a_constructor_to_be_strict
    - when_configuring_a_method_called_by_a_constructor_to_be_strict_and_adding_configuration
    - when_configuring_a_method_called_by_a_constructor_to_call_base_method
    - when_configuring_a_method_called_by_constructor_and_then_wrapping_an_object
    - when_faking_a_class_which_calls_configured_virtual_member_in_constructor
    - when_faking_a_class_whose_first_constructor_fails
    - when_wrapping_an_object_and_then_configuring_a_method_called_by_constructor
    - when_wrapping_an_object_in_a_fake_that_calls_virtual_member_in_constructor
    - when_wrapping_an_object_with_a_fake_that_has_a_fake_configurator (a tough one, as there's definite intersection here, but I picked the option builder specs because that's what's _visible_ to the user)
- FakeScopeSpecs?
    - when_configuring_a_method_called_by_a_constructor_from_within_a_scope
- UnconfiguredFakeSpecs?  I thought this was also tough.The virtual calls in constructor do tie these methods together, but I think the feature we're testing here is how an otherwise unconfigured fake behaves. 
    - when_faking_a_class_which_calls_virtual_member_in_constructor
    - when_faking_a_class_which_calls_virtual_properties_in_constructor

I think this make sense, but that could just be because everyone's ideas seem best to _them_.

adamralph: Our 'specs' are the closest thing we have to white box/acceptance tests, which should always be expressed in the implementation agnostic language of the user. A heuristic I use for other projects, e.g. https://github.com/xbehave/xbehave.net is that the language of the acceptance tests should look natural alongside the language used in the wiki, which forms the documentation. If this proposal is a step in that direction then I'm all for it.

blairconrad: I think the categorization I laid out would more closely mirror the organization of the wiki. The containing file names are very close (and could be tweaked) to the concepts one sees in the wiki's table of concepts.
The class names could also be massaged to express concepts the user is familiar with, although I think they are most of the way there already.
I'll work toward the change I outlined.

adamralph: :+1: sounds good!

ulrichb: The idea of the "ConstructorWithVirtualCallsSpecs" was to be a specification for *the new feature "support for mocked virtual calls during the c'tor"* (which could also get a new wiki page). All the scenarios for the different configuration options are just there to prove that all them *also work during the c'tor*.

I'm with you that that there should be separate "FakeScope/FakeConfigurator/FakeOptions" specs (which would duplicate some of the scenario(names) of `ConstructorWithVirtualCallsSpecs`). But personally I think that the scenarios in `ConstructorWithVirtualCallsSpecs` have a high cohesion and should not be separated (except the mentioned "feature intersection" scenarios).

adamralph:
I decided to merge as is since I'm perfectly happy with all the library changes, and the tests, whilst there is merit in the proposals made in the renaming/reorganisation discussion, do cover all the cases which constitute the new features. Great work all round @ulrichb and @blairconrad!

More than happy for the spec refactorings to be sent in another PR or PRs.

PR #451, opened 2015-02-03
441 rename dummy definition

### LINKED ISSUE #441: Rename IDummyDefinition, DummyDefinition
As suggested by @patrik-hagne  in https://github.com/FakeItEasy/FakeItEasy/issues/402#issuecomment-70158778 and subsequently refined by @adamralph, `IDummyDefinition` could have a better name. 

Current best suggestion (as I see it): `IDummyFactory`.

Make the change **and** the corresponding one for `DummyDefinition<T>`.

adamralph:
fixes #441


adamralph:
I went further with the third commit, and shortened the method names. Given the objects are now called factories, 'create' is sufficient language to convey the semantics of the methods, whereas before, it was probably necessary to convey explicitly what was being created.

blairconrad:
Very thorough. Nice job.

PR #459, opened 2015-02-17
#312 - give unfakeable auto property stable return values

### LINKED ISSUE #312: magic property get followed by another get returns different object when property type is not fakeable
```csharp
public class SealedProperties
{
    public sealed class Duple
    {
        public int First { get; set; }
        public int Second { get; set; }

        public Duple(int first, int second)
        {
            First = first;
            Second = second;
        }

        public override string ToString()
        {
            return string.Concat('(', First, ", ", Second, ")");
        }
    }

    public interface InterfaceWithADuple
    {
        Duple Duple { get; set; }
    }

    [Test]
    public void Should_be_able_to_get_the_same_sealed_type_twice()
    {
        var fake = A.Fake<InterfaceWithADuple>();

        var initialDuple = fake.Duple;

        initialDuple.First = 1;
        initialDuple.Second = 2;

        var fetchedDuple = fake.Duple;

        Assert.That(fetchedDuple, Is.SameAs(initialDuple));
    }
}
```

Fails:

```
  Expected: same as <(1, 2)>
  But was:  <(0, 0)>

   at NUnit.Framework.Assert.That(Object actual, IResolveConstraint expression, String message, Object[] args)
   at NUnit.Framework.Assert.That(Object actual, IResolveConstraint expression)
   at FakeItEasyQuestions.SealedProperties.Should_be_able_to_get_the_same_sealed_type_twice() in SealedProperties.cs: line 59
```

(Really, setting the First and Second wasn't necessary. It just gives a more dramatic error message.)

I found this while working on #293 and trying to use an array as a return type. 
The cause is that `FakeManager.AutoFakePropetyRule` checks to see if the `TypeIsFakable` before saying that the rule is applicable to an incoming call. 

I'd be inclined to say that we should allow the rule if we can make a Dummy for the type, but I think that stirs up #156, which I don't mind, but it could be another month of @patrik-hagne and I going in circles. :wink:

blairconrad:
Fixes #312

### DIFF
--- a/Source/FakeItEasy/Core/DefaultReturnValueRule.cs
+++ b/Source/FakeItEasy/Core/DefaultReturnValueRule.cs
@@ -17,6 +17,17 @@
             get { return ServiceLocator.Current.Resolve<IFakeAndDummyManager>(); }
         }
 
+        public static object ResolveReturnValue(IInterceptedFakeObjectCall fakeObjectCall)
+        {
+            object result;
+            if (!FakeManager.TryCreateDummy(fakeObjectCall.Method.ReturnType, out result))
+            {
+                return fakeObjectCall.Method.ReturnType.GetDefaultValue();
+            }
+
+            return result;
+        }
+
         public bool IsApplicableTo(IFakeObjectCall fakeObjectCall)
         {
             return true;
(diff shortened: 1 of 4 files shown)


Source/FakeItEasy/Core/DefaultReturnValueRule.cs
```diff
@@ -17,6 +17,17 @@ private static IFakeAndDummyManager FakeManager
...
+        {
+            object result;
+            if (!FakeManager.TryCreateDummy(fakeObjectCall.Method.ReturnType, out result))
+            {
+                return fakeObjectCall.Method.ReturnType.GetDefaultValue();
+            }
+
+            return result;
+        }
+
```
blairconrad: I considered whether to move `ResolveReturnValue` to another location, but it felt like it _wanted_ to belong to `DefaultReturnValueRule`.
Happy to put somewhere else if the "one rule calling another's static method" doesn't sit well.

adamralph:
Sorry it took me so long to get onto this one!

blairconrad:
No trouble, big bubble. (As my father says. When I was a child, this was hilarious.)

Thanks for your efforts.

Issue #461, opened 2015-02-24
Replace IFakeConfigurator with IFakeOptionsBuilder


blairconrad:
Currently `IFakeConfigurator` provides only one mechanism for configuring the fakes:

```
void ConfigureFake(object fake);
```

which allows users to set properties and/or set expectations via `A.CallTo`, but only after the Fake has been constructed.

**It doesn't easily allow making a fake Strict, having it call base methods, having it wrap an existing instance, having it implement additional interfaces, or giving it additional attributes.**

Replacing `IFakeConfigurator` with 

```c#
public interface IFakeOptionsBuilder
{
    /// <summary>
    /// Gets the priority of the options builder. When multiple builders that apply to
    /// the same type are registered, the one with the highest priority is used.
    /// </summary>
    /// <remarks>Negative values are reserved for use by FakeItEasy.</remarks>
    int Priority { get; }

    /// <summary>
    /// Whether or not this object can build options for a Fake of type <paramref name="type"/>.
    /// </summary>
    /// <param name="type">The type of fake to build options for.</param>
    /// <returns>
    /// <c>true</c> if this object can build options for a
    /// Fake of type <paramref name="type"/>. Otherwise <c>false</c>.
    /// </returns>
    bool CanBuildOptionsForFakeOfType(Type type);

    /// <summary>
    /// Manipulates <paramref name="options"/>, which will later be used to 
    /// create a Fake.
    /// </summary>
    /// <param name="typeOfFake">The type the fake object represents.</param>
    /// <param name="options">The fake options to manipulate.</param>
    void BuildOptions(Type typeOfFake, IFakeOptions options);
}
```

Will allow implementations to effect all the changes listed above, via methods we're used to from `A.Fake<T>(Action<IFakeOptionsBuilder<T>>)`. (The [Explicit Creation Options](https://github.com/FakeItEasy/FakeItEasy/wiki/Creating-Fakes#explicit-creation-options).)

We'd provide the `IFakeOptionsBuilder` interface as above, as well as a `FakeOptionsBuilder<TFake>` abstract base class that would provide its own `CanBuildOptionsForFakeOfType` method, which would only return true if the new Fake were to fake out a `TFake`.

[...]

PR #465, opened 2015-03-04
Adding fake creation options characterization tests

blairconrad:
As a precursor to the real work of #461, I

- moved some of the existing specs around according to the suggestion I had in https://github.com/FakeItEasy/FakeItEasy/pull/447#discussion_r24468982 and 
- added extra _characterization_ specs for the behaviour of the explicit fake creation options interactions, as laid out in https://github.com/FakeItEasy/FakeItEasy/issues/461#issuecomment-76596031. There are currently some deviations from our described ideal, which may cause us to re-evaluate that ideal:
    - WithAdditionalAttributes called twice: last one in wins, rather than stacking
    - ConfigureFake followed by Wrapping - uses configured, not wrapped, behaviour
    - CallsBaseMethods followed by Wrapping - calls base method
    - Strict followed by Wrapping - acts strict

Note that I have not changed any behaviour in the production code, only added some tests and moved a few between files.


Source/FakeItEasy.sln
```diff
@@ -64,6 +64,9 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Integration", "Integration", "{5D41D463-D53B-4ADC-90AD-44AA03D6340D}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Specs", "Specs", "{575A3C3B-17BA-4A8A-9F53-05320913D0A6}"
+	ProjectSection(SolutionItems) = preProject
+		CreationOptionsSpecs.cs = CreationOptionsSpecs.cs
+	EndProjectSection
```
adamralph: What's this doing here? :stuck_out_tongue_closed_eyes:

adamralph: In fact, this looks like a no-op change. Can this whole be reverted?

adamralph:
In the second commit, some of the test names seemed to have become less descriptive, in that they had the 'which calls a virtual xxx in constructor' removed. Is that intentional?

In the third commit, some new tests have been added which also omit that from their descriptions.

adamralph:
Finally got round to reviewing this :wink: Sorry for taking so long and thanks for being patient.

I've finished my first pass.

Source/FakeItEasy.sln
blairconrad: Whoa. Yes. I do not know what was going on there.
Worse, I can't explain why I didn't catch it in pre-PR review. :blush:

blairconrad:
> Finally got round to reviewing this :wink: Sorry for taking so long and thanks for being patient.

Please, no need to apologize. Everyone's busy, and your comments are worth waiting for! :wink:

Yes, removing "which calls a virtual xxx in constructor" was intentional. I don't think it adds value. We're either talking about configuring xxxes, which is nonsensical for non-virtual xxxes, or the test class indicates that the xxxes are virtual, and the `It`s indicate when the xxxes have been called from within a constructor. (In which case I suppose the "which calls a virtual xxx in constructor" was in some sense moved, not removed.) 

If you don't care for this approach, I can reword.

If you're okay with it, I think my only action is to push a commit reverting the sln, which I can do at almost any time.

blairconrad:
I semi-accidentally pushed the sln revert anyhow, so I figure @adamralph might as well own the PR for now. I'm still happy to rename the tests if the current names are not satisfactory.

adamralph:
You're reasoning about the test names does indeed make sense :smile:

PR #478, opened 2015-04-14
making ArgumentConstraintTrap threadsafe

### LINKED ISSUE #476: ArgumentConstraintTrap is not threadsafe
This is one of the problems preventing us from supporting parallel test execution out of the box (#60), but may not be the only factor, so I preferred to start a new issue. It is the only concrete 

The internal `ArgumentConstraintTrap` class, used to capture argument constraints using the `That` syntax (`A<SomeClass>.That.Matches(…)`), is not threadsafe. It uses a single static variable to keep trap of the captured arguments. This can cause parallel test execution to fail, whether the `That` is used to configure a fake or to assert that certain calls were made to the fake.

I believe that replacing the static `List<IArgumentConstraint>` with a `ThreadLocal<List<IArgumentConstraint>>` will fix the problem, and preliminary testing has supported this.
Of course, testing never _proves_ anything, but I think making the field ThreadLocal is a reasonable step.

blairconrad:
Fixes #476.

### DIFF
--- a/Source/FakeItEasy.Tests/Core/ArgumentConstraintTrapTests.cs
+++ b/Source/FakeItEasy.Tests/Core/ArgumentConstraintTrapTests.cs
@@ -2,13 +2,14 @@
 {
     using System.Threading.Tasks;
     using FakeItEasy.Core;
+    using FluentAssertions;
     using NUnit.Framework;
     using TestHelpers;
 
     [TestFixture]
     public class ArgumentConstraintTrapTests
     {
-        private ArgumentConstraintTrap trap = new ArgumentConstraintTrap();
+        private readonly ArgumentConstraintTrap trap = new ArgumentConstraintTrap();
 
         [Test]
         public void Should_return_constraints_that_has_been_trapped()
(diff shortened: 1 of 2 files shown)


blairconrad:
I agonized over the test. I don't like having multithreading in the tests, but that's the entire point of the change, and I didn't want to have nothing. Then I agonized over where to put them. In the end, I let the convenience of augmenting the existing ArgumentConstraintTrapTests push me to put the test there.
I'm happy to make any changes, of course.

adamralph:
@blairconrad finished my review. Over to you for comment

blairconrad:
Thanks for the comments. I really enjoyed the "test" one. (Not that I didn't the other.)

blairconrad:
@adamralph, as you'll see, I ended up using two locks to completely lock down the order in which each thread starts and finishes recording the argument constraints.
Thanks so much for the comment.

adamralph:
:heart: the new test, specially the comment :stuck_out_tongue:

blairconrad:
> :heart: the new test, specially the comment :stuck_out_tongue: 

ASCII art FTW.

PR #471, opened 2015-03-27
Add generic overload of Implements

thomaslevesque:
Fixes #470

PR #481, opened 2015-04-27
Issue480

### LINKED ISSUE #480: NullReferenceException when trying to fake a non-virtual generic method method
If I try to fake a non-virtual, non-generic method, I get a `FakeConfigurationException` because the method is non-virtual (as expected). But if the method is generic, FIE throws a `NullReferenceException`.

Example:

```c#
public class Foo
{
    public int GetValue()
    {
        return 0;
    }

    public T GetValue<T>()
    {
        return default(T);
    }
}
```

This throws the expected `FakeConfigurationException`:

```c#
A.CallTo(() => foo.GetValue()).Returns(42);
```

This throws a `NullReferenceException`:

```c#
A.CallTo(() => foo.GetValue<int>()).Returns(42);
```

thomaslevesque:
Changed `HasSameBaseMethod` to ensure we're always working with the generic method definition

fixes #480

### DIFF
--- a/Source/FakeItEasy.IntegrationTests/GeneralTests.cs
+++ b/Source/FakeItEasy.IntegrationTests/GeneralTests.cs
@@ -161,6 +161,21 @@ namespace FakeItEasy.IntegrationTests
         }
 
         [Test]
+        public void ErrorMessage_when_configuring_generic_function_call_that_cannot_be_configured_should_be_correct()
+        {
+            // Arrange
+            var fake = A.Fake<TypeWithNonConfigurableMethod>();
+
+            // Act
+            var exception = Record.Exception(() =>
+                A.CallTo(() => fake.GenericNonVirtualFunction<int>(string.Empty, 1)).Returns(10));
+
+            // Asssert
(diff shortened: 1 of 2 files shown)


blairconrad:
Thanks, @thomaslevesque. Generally looks good, I think, but I think we may be a while before we get other owners' eyes on it.

One thing that I would suggest though, and maybe I'm being overly picky, is that the commit messages could be more descriptive. It's true that they are a test and fix for #480, but it'd be nice to have more context, such as what you provided for your commits in #471.

We appreciate your input.

thomaslevesque:
> One thing that I would suggest though, and maybe I'm being overly picky, is that the commit messages could be more descriptive. It's true that they are a test and fix for #480, but it'd be nice to have more context, such as what you provided for your commits in #471.

Sure, I can reword the commits

thomaslevesque:
I pushed the reworded commits

adamralph:
Thanks @thomaslevesque! Look out for your name in the release notes :trophy:

PR #483, opened 2015-04-28
Updated ILMerge NuGet package

thomaslevesque:
ILMerge 2.13.0307 was generating an invalid assembly (methods with missing
implementation) when merging assemblies built with the C# 6 compiler.
Version 2.14.1208 works as expected.

### DIFF
--- a/Source/FakeItEasy-SL/FakeItEasy-SL.csproj
+++ b/Source/FakeItEasy-SL/FakeItEasy-SL.csproj
@@ -721,7 +721,7 @@
     </VisualStudio>
   </ProjectExtensions>
   <Target Name="AfterBuild">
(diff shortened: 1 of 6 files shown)


blairconrad:
@thomaslevesque, thanks for the contribution.

Perhaps I'm missing something, but we don't build with the C#6 compiler. Is this something you encountered when building on your own machine?

thomaslevesque:
@blairconrad, yes, I tried to build with VS2015 yesterday and the generated assembly had types with missing methods. This caused TypeLoadExceptions when trying to load those types. I saw that the original unmerged assembly in the obj directory was correct, which meant that ILMerge was causing the problem. Updating it fixed the problem.

adamralph:
Thanks @thomaslevesque!

I've marked this as a refactoring since, given the current platform it doesn't have any functional effect. It's nice to know that future proofing for C# 6 is a side effect though :wink:

PR #487, opened 2015-05-06
Aligning creation options interaction

### LINKED ISSUE #467: improve fake creation options interaction
spun off from/a precursor to #461, specifically https://github.com/FakeItEasy/FakeItEasy/issues/461#issuecomment-76596031, where I advocate this set of interactions:

- `CallsBaseMethods` - should trump prior invocations of `ConfigureFake`, `Strict`, and `Wrapping`
- `ConfigureFake` - should "stack" with prior invocations, just as
  multiple `A.CallsTo` do, which would include overriding earlier
  `CallsBaseMethods`, `Strict`, and `Wrapping`
- `Implements` - should "stack" with prior invocations, ever accumulating more interfaces
- `Strict` - should trump prior invocations of `CallsBaseMethods`,
  `ConfigureFake`, and `Wrapping`
- `WithAdditionalAttributes` - should "stack" with prior invocations, ever accumulating more attributes
- `WithArgumentsForConstructor` - should trump prior invocations
- `Wrapping` - should trump prior invocations of `CallsBaseMethods`,
  `ConfigureFake`, `Strict`, and `Wrapping`

While writing characterization tests in #465, I noticed deviations, even when just setting competing options in the `A.Fake()` calls:

- WithAdditionalAttributes called twice: last one in wins, rather than stacking
- ConfigureFake followed by Wrapping - uses configured, not wrapped, behaviour
- CallsBaseMethods followed by Wrapping - calls base method
- Strict followed by Wrapping - acts strict

I'd like to smooth out these deviations before extending `IFakeConfigurator`.

I am very close to a PR (in the "cleanup" phase) that would build upon the work in #465.

blairconrad:
Fixes #467.


adamralph:
@blairconrad I'm going through this and it looks good so far. I only have the big commit, 'making Wrapping override earlier configuration', left to go.

The only thing I'm not too sure about so far is the exact purpose of the 'adding specs for SelfInitializedFakes' commit. Can you shed some light?

Source/FakeItEasy/Core/DefaultFakeWrapperConfigurer.cs
```diff
@@ -6,21 +6,27 @@ namespace FakeItEasy.Core
     /// Handles configuring of fake objects to delegate all their calls to a wrapped instance.
     /// </summary>
     internal class DefaultFakeWrapperConfigurer
-        : IFakeWrapperConfigurer
```
adamralph: Given that the removal of the interface means that an alternate implementation can no longer be used anywhere, I suggest this class is renamed to `FakeWrapperConfigurer`.

Source/FakeItEasy/Core/DefaultFakeWrapperConfigurer.cs
```diff
@@ -6,21 +6,27 @@ namespace FakeItEasy.Core
...
     {
+        public DefaultFakeWrapperConfigurer(object wrappedObject)
+        {
+            this.WrappedObject = wrappedObject;
+        }
+
+        public object WrappedObject { get; private set; }
+
+        public ISelfInitializingFakeRecorder Recorder { get; set; }
```
adamralph: The introduction of a mutable field always sets off an alarm bell in my head but I'm tying myself in knots trying to explore a way to avoid it. I don't really understand the collaboration of the various types well enough.

If you believe this is the best solution then I'm happy to roll with it.

Source/FakeItEasy/Creation/FakeOptions.cs
```diff
@@ -20,9 +22,19 @@ public static FakeOptions Empty
...
+            get
+            {
+                return this.wrapper;
+            }
 
-        public ISelfInitializingFakeRecorder SelfInitializedFakeRecorder { get; set; }
+            set
+            {
+                this.wrapper = value;
+                this.FakeConfigurationActions.Add(fake => this.wrapper.ConfigureFakeToWrap(fake));
```
adamralph: I wonder if we can come up with a better shape for this class. In it's current form, setting this property obviously has a side effect on `FakeConfigurationActions` which itself has a property setter.

adamralph: Ah, it's `internal`. But still...

adamralph:
@blairconrad I've finished my first pass.

adamralph:
Sorry it took so long for me to get onto it! :cold_sweat:

Source/FakeItEasy/Core/DefaultFakeWrapperConfigurer.cs
blairconrad: Good idea.
I noticed the tests `DefaultFakeWrapperConfigurator`, and "Configurator" is more prevalent in the solution than Configurer (this is the only type name containing "configurer"). So `FakeWrapperConfigurator`?
I could give `DynamicContainerTests.availableConfigurers` the same treatment while I'm at it.

adamralph: Good point, yeah let's change all occurrences of 'configurer' to 'configurator'.

blairconrad:
> The only thing I'm not too sure about so far is the exact purpose of the 'adding specs for
> SelfInitializedFakes' commit. Can you shed some light?

Of course. The SelfInitializedFakes are bound up in the `Wrapping` functionality, as you noted when pointing out the less savoury of the changes I've made. The next step I was going to make in the change was to rework how Wrapping was done, so I wanted to have tests in place to make sure I didn't break the SelfInitiializedFakes. The only existing tests (`RecordingManagerTests`) were explicit, narrow in scope, performed weak checks, hit live web sites, and wrote files to a path that most people wouldn't have on their machines. Commit c7753edb5636ddedcbfc0b5b845a2ee2bcdd5a2d was intended to fix those problems (okay, nearly all; the remaining tests that access the filesystem are still explicit).

adamralph:
OK, thanks. That makes sense.

Do you have any more thoughts about my comments on `DefaultFakeWrapperConfigurer.Recorder` and `FakeOptions.SelfInitializedFakeRecorder`?

Given these are both internal, unless some obvious improvement springs to mind, I'm happy to go with them as they are. I wouldn't want to hold things up with any significant rework.

blairconrad:
Sorry, I've haven't had the time to muster enough concentration to address the `DefaultFakeWrapperConfigurer.Recorder` and `FakeOptions.SelfInitializedFakeRecorder` concerns yet. Will try later today.

Source/FakeItEasy/Core/DefaultFakeWrapperConfigurer.cs
blairconrad: I'll take a look in a bit. I did wrestle a while with how to implement the wrapping/recording while introducing the "last configuration in wins" semantics. It was tough to come up with something, which may mean there isn't a better option on the horizon, or it could mean that I was just so happy to find something that worked that I stopped looking.

If it's any consolation, the mutable field isn't so much introduced as moved from `FakeOptions.SelfInitializedFakeRecorder`. :wink:

I will take a look again, and at [your concerns about the FakeOptions.Wrapper setter](https://github.com/FakeItEasy/FakeItEasy/pull/487/files#r32370991).

blairconrad: I think the mutable field or some variant of it is the way to go. We can't know that there's a recorder when we create the wrapper, it seems we need to add one later (as we are now), replace the wrapper with one that has a recorder, which means we're mutating the Wrapper field in FakeOptions, or (as was done before) have a parallel mutable field in FakeOptions. 
I think the current approach is better than the other two, or at least no worse.

blairconrad:
I've come up with what "improvements" (unless they're not) that I can.
Much of the awkwardness comes from the fact that at the time we call `Wrapping` on the options builder action, we need to _immediately_ add the `FakeConfigurationOption`, or the `Wrapping`'s order will not be preserved relative the other options, such as `CallsBaseMethods` and `ConfigureFake`, but then we need to retain access to the wrapper so it's available to add any late-breaking  "add a RecordedBy" action.

Oh! How embarrassing. I suppose we _could_ change the API so there's `Wrapping` by itself, and then another call on the fake options builder for "wrap and also record actions in this thing". That would make things a little simpler on the inside, but at the cost of changing the API. Not that I'm _in love_ with

```csharp
var fakeService = A.Fake<ILibraryService>(options => options
    .Wrapping(realServiceWhileRecording).RecordedBy(recorder));
```

but I don't think a two-parameter replacement for `Wrapping` is likely to improve things… I'm inclined to leave the API and push the internal changes that I have, for your consideration.

Source/FakeItEasy/Creation/FakeOptions.cs
```diff
@@ -2,34 +2,72 @@ namespace FakeItEasy.Creation
...
+    using FakeItEasy.Core;
 
     internal class FakeOptions
     {
+        private readonly List<Type> additionalInterfacesToImplement;
+        private readonly List<Action<object>> fakeConfigurationActions;
+        private readonly List<CustomAttributeBuilder> additionalAttributes;
+        private FakeWrapperConfigurator wrapper;
+
         public FakeOptions()
```
adamralph: Minor point, but this constructor can be removed and replaced with field initializers. Won't hold up the PR.

PR #497, opened 2015-06-24
removing FakeOptions's parameterless constructor

blairconrad:
Per @adamralph's [comment in #487](https://github.com/FakeItEasy/FakeItEasy/pull/487/files#r33120984):

> Minor point, but this constructor can be removed and replaced with field initializers.


blairconrad:
Thanks, @adamralph!

PR #499, opened 2015-06-26
Use ConcurrentDictionary for MethodCache…

TimLovellSmith:
… in MethodInfoManager. For just slightly less threads blocking other threads.

### DIFF
--- a/Source/FakeItEasy/Core/MethodInfoManager.cs
+++ b/Source/FakeItEasy/Core/MethodInfoManager.cs
@@ -1,6 +1,7 @@
 namespace FakeItEasy.Core
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
@@ -11,7 +12,7 @@ namespace FakeItEasy.Core
     /// </summary>
     internal class MethodInfoManager
     {
-        private static readonly Dictionary<TypeMethodInfoPair, MethodInfo> MethodCache = new Dictionary<TypeMethodInfoPair, MethodInfo>();
+        private static readonly ConcurrentDictionary<TypeMethodInfoPair, MethodInfo> MethodCache = new ConcurrentDictionary<TypeMethodInfoPair, MethodInfo>();
 
         /// <summary>
         /// Gets a value indicating whether the two instances of <see cref="MethodInfo"/> would invoke the same method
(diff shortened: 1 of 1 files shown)


blairconrad:
Hi, @TimLovellSmith. Thanks for thinking of us.

Generally seems good, with one note

>  For just slightly less threads blocking other threads.

… with the small possibility that we will execute `FindMethodOnTypeThatWillBeInvokedByMethodInfo` multiple times for the same key.
So there's a small tradeoff. Not one that I think is necessarily a big deal.

adamralph:
This isn't necessarily an improvement. The situation is actually quite complex and depends on the usage profile, see http://www.codeproject.com/Articles/548406/Dictionary-plus-Locking-versus-ConcurrentDictionar

blairconrad:
Indeed, @adamralph. Very interesting read. I hadn't realized that 

> … the way the ConcurrentDictionary allocates the nodes and puts them into its buckets is different. It is optimized to allow lock-free reading. But, when items are added, allocating such node is expensive.

focusing instead on the cost of computing the value multiple times if we're trying to access the same key concurrently, which I was prepared to discount as not worth worrying over.

All this is supposition, but I would expect that

1. reads are much more common than writes during a typical test run (and when they're not, it's because there are very few tests and so overall speed doesn't matter much), and
2. `FindMethodOnTypeThatWillBeInvokedByMethodInfo`, while not the cheapest thing in the world (if it were, we wouldn't cache it), is probably not prohibitively expensive

That being said, we really don't have numbers, and if we did, they'd be for some of our own projects (likely) and we wouldn't know whether they were typical or not.
Something that stuck with me from the end of @adamralph's linked-to article:

> All dictionaries are fast. I was creating millions of items and it is still fast. We will usually only create a small amount of items and read them with intervals, so we will not perceive the time spent reading them;

So it probably will not make a noticeable difference. In this case, the "enterprise software developer" part of my brain leans toward "if it's not (demonstrably) broken, don't fix it".
Were there some great improvement in readability, I'd still probably opt for the `ConcurrentDictionary`, but to me, they're about the same.

adamralph:
If it's a matter of micro-optimisation then I agree, let's go with the simpler option. In that case, I'd prefer `ConcurrentDictionary` since there is less code and it removes the need to remember about the locking each time the dictionary is touched.

blairconrad:
As we've determined that this is a micro-optimisation, unlikely to bring observable benefit, I've rebranded this a **refactoring**.

blairconrad:
If it's all the same to the owners, don't merge. I'll send a superseding PR soonish.

TimLovellSmith:
Withdrawing this PR in favor of the more complete #503. :)

blairconrad:
Thanks, @TimLovellSmith. I felt bad about potentially asking you to clean up the stylistic warnings in what was an otherwise complete PR.

PR #501, opened 2015-06-26
Fix non-thread-safeness of cachedFormatters dictionary in ArgumentVal…

### LINKED ISSUE #500: FakeItEasy - NullRef when using ArgumentValueFormatter concurrently
I had a one off failure with xunit 2 which I think is a concurrency bug.

```
Result Message:	System.NullReferenceException : Object reference not set to an instance of an object.
Result StackTrace:	
at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at FakeItEasy.Core.ArgumentValueFormatter.GetArgumentValueAsString(Object argumentValue) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\ArgumentValueFormatter.cs:line 37
   at FakeItEasy.Core.DefaultFakeObjectCallFormatter.GetArgumentValueAsString(Object argumentValue) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\DefaultFakeObjectCallFormatter.cs:line 170
   at FakeItEasy.Core.DefaultFakeObjectCallFormatter.AppendArgumentValue(StringBuilder builder, ArgumentValueInfo argument) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\DefaultFakeObjectCallFormatter.cs:line 162
   at FakeItEasy.Core.DefaultFakeObjectCallFormatter.AppendArguments(StringBuilder builder, IEnumerable`1 arguments) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\DefaultFakeObjectCallFormatter.cs:line 180
   at FakeItEasy.Core.DefaultFakeObjectCallFormatter.AppendArgumentsList(StringBuilder builder, IFakeObjectCall call) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\DefaultFakeObjectCallFormatter.cs:line 148
   at FakeItEasy.Core.DefaultFakeObjectCallFormatter.GetDescription(IFakeObjectCall call) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\DefaultFakeObjectCallFormatter.cs:line 40
   at FakeItEasy.Core.CallWriter.WriteCalls(IEnumerable`1 calls, IOutputWriter writer) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\CallWriter.cs:line 40
   at FakeItEasy.Core.FakeAsserter.AppendCallList(IEnumerable`1 calls, CallWriter callWriter, IOutputWriter writer) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\FakeAsserter.cs:line 86
   at FakeItEasy.Core.FakeAsserter.CreateExceptionMessage(IEnumerable`1 calls, CallWriter callWriter, String callDescription, String repeatDescription, Int32 matchedCallCount) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\FakeAsserter.cs:line 46
   at FakeItEasy.Core.FakeAsserter.AssertWasCalled(Func`2 callPredicate, String callDescription, Func`2 repeatPredicate, String repeatDescription) in c:\repo\FakeItEasy\Source\FakeItEasy\Core\FakeAsserter.cs:line 31
   at FakeItEasy.Configuration.RuleBuilder.MustHaveHappened(Repeated repeatConstraint) in c:\repo\FakeItEasy\Source\FakeItEasy\Configuration\RuleBuilder.cs:line 114
   at FakeItEasy.Configuration.RuleBuilder.ReturnValueConfiguration`1.MustHaveHappened(Repeated repeatConstraint) in c:\repo\FakeItEasy\Source\FakeItEasy\Configuration\RuleBuilder.cs:line 168
   at FakeItEasy.AssertConfigurationExtensions.MustHaveHappened(IAssertConfiguration configuration) in c:\repo\FakeItEasy\Source\FakeItEasy\AssertConfigurationExtensions.cs:line 18
   at [testcode]
```

The only reason I can think of this can nullref, given the FakeItEasy source I compiled from is that it's a plain Dictionary<> being shared across threads without locking, since it appears `ArgumentFormatter` is a singleton.

Sending a PR for this soon!

TimLovellSmith:
…ueFormatter singleton. I think this should fix #500.

### DIFF
--- a/Source/FakeItEasy/Core/ArgumentValueFormatter.cs
+++ b/Source/FakeItEasy/Core/ArgumentValueFormatter.cs
@@ -1,17 +1,18 @@
 namespace FakeItEasy.Core
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Linq;
 
     internal class ArgumentValueFormatter
     {
-        private readonly Dictionary<Type, IArgumentValueFormatter> cachedFormatters;
+        private readonly ConcurrentDictionary<Type, IArgumentValueFormatter> cachedFormatters;
         private readonly IEnumerable<IArgumentValueFormatter> typeFormatters;
 
         public ArgumentValueFormatter(IEnumerable<IArgumentValueFormatter> typeFormatters)
         {
-            this.cachedFormatters = new Dictionary<Type, IArgumentValueFormatter>();
+            this.cachedFormatters = new ConcurrentDictionary<Type, IArgumentValueFormatter>();
 
             this.typeFormatters = typeFormatters.Concat(
                 new IArgumentValueFormatter[]
@@ -31,11 +32,7 @@ namespace FakeItEasy.Core
             var argumentType = argumentValue.GetType();
 
             IArgumentValueFormatter formatter;
-            if (!this.cachedFormatters.TryGetValue(argumentType, out formatter))
-            {
-                formatter = this.ResolveTypeFormatter(argumentType);
-                this.cachedFormatters.Add(argumentType, formatter);
-            }
+            formatter = this.cachedFormatters.GetOrAdd(argumentType, this.ResolveTypeFormatter);
 
             return formatter.GetArgumentValueAsString(argumentValue);
         }


blairconrad:
Thanks, @TimLovellSmith.
As this is a change that's visible from the outside, I'd be inclined to add a test to verify your change (not that I doubt it) and ensure that we don't backslide.
I'll take a closer look at the change as time allows. I hope over the weekend.

adamralph:
As per https://github.com/FakeItEasy/FakeItEasy/pull/499#issuecomment-116252064 we need to make sure we are choosing the right option for concurrency (`lock` + `Dictionary` vs `ConcurrentDictionary`).

blairconrad:
Agreed, @adamralph.
Since this is likely to be an ongoing concern, and in many cases we're won't have a clear-cut answer to the question of which option is better, I think we should establish a guideline for the "we really have no idea" case.

Note that we already use `ConcurrentDictionary` in `DynamicContainer`, used to hold `IDummyFactory`s and `IFakeConfigurator`s. This was added in 93f5bf701300ba13e82bb52c14324b01ff367a29 for #402.
And in `EventHandlerArgumentProviderMap`, we maintain a `Dictionary` and manage the locks ourselves. In that case, we have the value up front, so don't need to have the dictionary call a factory for us, so there's no concern about multiple factories being run concurrently. It's just the performance of the dictionary implementation that we'd have to worry about.

I think I came down a little on the "manage the locks ourselves" side in #499, at least for that particular case. However, there we only had one use of the dictionary: `GetOrAdd`, so I didn't find a big difference in the readability. In general, and especially where we have to manage concurrent access to a dictionary from multiple call sites, I think `ConcurrentDictionary` has the edge in readability.
Keeping in mind the fact that we're unlikely to be accessing our dictionaries so heavily that the time will make a difference, I think I'd lean toward using `ConcurrentDictionary` unless there was a reason to believe it wouldn't perform adequately in a particular situation.

adamralph:
Agreed.

blairconrad:
Okay, `ConcurrentDictionary` it is. Huzzah.

@TimLovellSmith, the implementation looks generally good to me, and I'd be happy to merge it with an accompanying test case.
I'll also make some stylistic notes in the file, since we're getting StyleCop warnings with this build. I wouldn't have bothered if they were the only comment, but I figure since I'm asking for a change _anyhow_…

Source/FakeItEasy/Core/ArgumentValueFormatter.cs
```diff
@@ -31,11 +32,8 @@ public virtual string GetArgumentValueAsString(object argumentValue)
             var argumentType = argumentValue.GetType();
 
             IArgumentValueFormatter formatter;
-            if (!this.cachedFormatters.TryGetValue(argumentType, out formatter))
-            {
-                formatter = this.ResolveTypeFormatter(argumentType);
-                this.cachedFormatters.Add(argumentType, formatter);
-            }
+            formatter = this.cachedFormatters.GetOrAdd(argumentType, (argType) =>
+                this.ResolveTypeFormatter(argType));
```
blairconrad: The change looks technically sound, but has introduced some StyleCop warnings which could be cleared up by moving some arguments around. However, for my money, a superior formulation would be 

`var formatter = this.cachedFormatters.GetOrAdd(argumentType, this.ResolveTypeFormatter);`

adamralph: :+1:

TimLovellSmith:
Clarification: would you like the test to verify the concurrent scenario, or just sanity check that the code works? Which test project is the best place? Unit? Integration?

blairconrad:
@TimLovellSmith, thanks for asking!
The concurrent scenario, ideally. We have coverage to ensure that ArgumentFormatters work in general, so I think a test to ensure that the "feature" of concurrent argument formatting would be a good idea. 

Yes, Unit or Integration. :wink:. I'm not in a position to play with the code right now, but a quick scan suggests that adding something in or new `ArgumentValueFormatterTests` might be viable.

I'm not sure it'll be helpful, but @adamralph recently helped me write a non-deterministic test of concurrent processes: https://github.com/FakeItEasy/FakeItEasy/commit/02a9f2c953ba6b45a7f1ebe80c7d876f0b164f47

TimLovellSmith:
non-deterministic test = :cool: :+1: unfortunately I don't think that style of test verification really works well to 'expose a bug' on the concurrent dictionary code, or even the non-concurrent dictionary code for that matter, in this caching scenario where there's no data race, so I think I'm just going to have to do a test with some parallelism that will sometimes fail if there's a bug.

blairconrad:
@TimLovellSmith, sorry for the delay. I somehow missed that you pushed a new commit after your last comment. I will take a look.
Although it's probably worth noting that I had an error in my [comment of 11 days ago](https://github.com/FakeItEasy/FakeItEasy/pull/501#issuecomment-116857294). I'd mistyped, suggesting a non-deterministic test where I really meant to encourage a deterministic one. From your comment, I'm not sure that tripped you up, though.
It sounds like you prefer a non-deterministic (no typo that time) test over a deterministic one in this case.
Can you explain why that is? You say you went with (and I've taken a look at the code, and I believe it to be true) 

>  a test with some parallelism that will sometimes fail if there's a bug

Why not a test with some parallelism that will _always_ fail if there's a bug?

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
```diff
@@ -142,6 +143,40 @@ public void Should_prefer_exact_type_formatter_to_interface_formatter()
...
+                new { l = 0 },
+                new { m = 0 },
+                new { n = 0 },
+                new { o = 0 },
+                new { p = 0 },
+                new { q = 0 },
+                new { r = 0 },
+                new { s = 0 },
+                new { t = 0 },
+            };
```
blairconrad: The array of anonymous types is clever.

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
```diff
@@ -142,6 +143,40 @@ public void Should_prefer_exact_type_formatter_to_interface_formatter()
...
+
+            Parallel.For(0, 20, _i => 
+                {
+                    localFormatter.GetArgumentValueAsString(a[_i]);
+                });
+            Assert.That(localFormatter.cachedFormatters.Count, Is.EqualTo(20));
+        }
+
```
blairconrad: As I mentioned in the conversation, I'm not sold on this approach. In my admittedly unscientific testing, using `Dictionary` for the `cachedFormatters`, I see a failure quite infrequently. Certainly no more than once in every 4 runs. I'll not comment more, waiting for your response about why you think a non-deterministic test is superior.

blairconrad:
Ah

>  I don't think that style of test verification really works well to 'expose a bug' on the concurrent dictionary code, or even the non-concurrent dictionary code for that matter

I'd interpreted that as an expression of preference, but do you mean that you don't think it's technically feasible in this case?
After rereading your comment and the code, I am inclined to agree.

TimLovellSmith:
Yup, that's what I meant. We can't make a deterministic race, because every action we are doing on the dictionary is already atomic.

blairconrad:
Yeah. in my head, we'd been constructing the formatters as we tried to add them to the dictionary, which is silly, as we wouldn't be able to call `ForType` on them.

We _could_ make a deterministic test that ensured that we only call `ForType` on a formatter once per type being formatted, but I don't think that's a big deal.

@FakeItEasy/owners, I doubt we'll get a better test than @TimLovellSmith has already added. Given that, it's non-deterministic, would only infrequently detect a problem, and forces us to make `localFormatter.cachedFormatters` internal instead of private. (Although that last step could be omitted by instead having a formatter that tracks the number of times `ForType` is called and, after we've registered all the formatters by formatting everything, taking another pass over the types to format, ensuring that the number of times `ForType` was called does not change.)
Alternatively, we could say that I was foolish to have asked for a test given the circumstances, and have it removed altogether.

I'm on the fence about the options, so if anyone cares at all, pipe up, and push me over to one side. Alternatively, if we don't want to decide yet, the fix is sound, and anything else we do later would be a test refactoring, not at all visible to users. In order to get @TimLovellSmith off his tenterhooks, we could take the PR as is, and then debate what to do about the test later. I'd be happy to clean it up in whatever way the other owners like. We'll need to clean up some style warnings anyhow, so it's no hardship for me.

adamralph:
Personally, I think a non-deterministic test is worse than no test at all, since it gives a false sense of security. The internal field is a nasty code smell too, so I'd rather we remove the test and revert the field back to private.

Source/FakeItEasy/Core/ArgumentValueFormatter.cs
```diff
@@ -31,11 +32,7 @@ public virtual string GetArgumentValueAsString(object argumentValue)
             var argumentType = argumentValue.GetType();
 
             IArgumentValueFormatter formatter;
-            if (!this.cachedFormatters.TryGetValue(argumentType, out formatter))
-            {
-                formatter = this.ResolveTypeFormatter(argumentType);
-                this.cachedFormatters.Add(argumentType, formatter);
-            }
+            formatter = this.cachedFormatters.GetOrAdd(argumentType, ResolveTypeFormatter);
```
blairconrad: I'm away from a compiler now, but I think `ResultTypeFormatter` is a style warning (no `this.`). 
If you come back in to remove the new test, how do you feel about cleaning it up while you're here?
Alternatively, I can do later.

blairconrad:
I second @adamralph's position that "Blair was foolish to even have asked for a test"<sup>1</sup>.
Sorry for the thrash, @TimLovellSmith.
Would you like to remove the test?
----
<sup>1</sup> I'm kidding. He's way too nice to say that.

TimLovellSmith:
@blairconrad no problem - cleaned it up!

adamralph:
Thanks @TimLovellSmith! Look out for your name in the release notes! :trophy:

PR #503, opened 2015-06-28
Use ConcurrentDictionary instead of Dictionary and locks

blairconrad:
As discussed in #499 and #501.
Resolves #499.

### DIFF
--- a/Source/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
+++ b/Source/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
@@ -6,24 +6,17 @@
 
     internal class EventHandlerArgumentProviderMap
     {
-        private readonly ConcurrentDictionary<Delegate, IEventRaiserArguments> map
-            = new ConcurrentDictionary<Delegate, IEventRaiserArguments>(new EventRaiserDelegateComparer());
+        private readonly ConcurrentDictionary<Delegate, IEventRaiserArgumentProvider> map
+            = new ConcurrentDictionary<Delegate, IEventRaiserArgumentProvider>(new EventRaiserDelegateComparer());
 
-        public void AddArgumentProvider(Delegate eventHandler, IEventRaiserArguments argumentProvider)
+        public void AddArgumentProvider(Delegate eventHandler, IEventRaiserArgumentProvider argumentProvider)
         {
             this.map[eventHandler] = argumentProvider;
         }
 
-        public IEventRaiserArguments TakeArgumentProviderFor(Delegate eventHandler)
+        public bool TryTakeArgumentProviderFor(Delegate eventHandler, out IEventRaiserArgumentProvider argumentProvider)
         {
-            IEventRaiserArguments provider;
-            this.map.TryRemove(eventHandler, out provider);
-            return provider;
-        }
-
-        public bool Contains(Delegate eventHandler)
-        {
-            return this.map.ContainsKey(eventHandler);
+            return this.map.TryRemove(eventHandler, out argumentProvider);
         }
 
         /// <summary>
--- a/Source/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
+++ b/Source/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
@@ -1,29 +1,24 @@
 ﻿namespace FakeItEasy.Core
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
 
     internal class EventHandlerArgumentProviderMap
     {
-        private readonly Dictionary<Delegate, IEventRaiserArguments> map
-            = new Dictionary<Delegate, IEventRaiserArguments>(new EventRaiserDelegateComparer());
+        private readonly ConcurrentDictionary<Delegate, IEventRaiserArguments> map
+            = new ConcurrentDictionary<Delegate, IEventRaiserArguments>(new EventRaiserDelegateComparer());
 
         public void AddArgumentProvider(Delegate eventHandler, IEventRaiserArguments argumentProvider)
         {
-            lock (this.map)
-            {
-                this.map[eventHandler] = argumentProvider;
-            }
+            this.map[eventHandler] = argumentProvider;
         }
 
         public IEventRaiserArguments TakeArgumentProviderFor(Delegate eventHandler)
         {
-            lock (this.map)
-            {
-                var provider = this.map[eventHandler];
-                this.map.Remove(eventHandler);
-                return provider;
-            }
+            IEventRaiserArguments provider;
+            this.map.TryRemove(eventHandler, out provider);
+            return provider;
         }
 
         public bool Contains(Delegate eventHandler)
--- a/Source/FakeItEasy/Core/MethodInfoManager.cs
+++ b/Source/FakeItEasy/Core/MethodInfoManager.cs
@@ -1,6 +1,7 @@
 namespace FakeItEasy.Core
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
@@ -11,7 +12,7 @@ namespace FakeItEasy.Core
     /// </summary>
     internal class MethodInfoManager
     {
-        private static readonly Dictionary<TypeMethodInfoPair, MethodInfo> MethodCache = new Dictionary<TypeMethodInfoPair, MethodInfo>();
+        private static readonly ConcurrentDictionary<TypeMethodInfoPair, MethodInfo> MethodCache = new ConcurrentDictionary<TypeMethodInfoPair, MethodInfo>();
 
         /// <summary>
         /// Gets a value indicating whether the two instances of <see cref="MethodInfo"/> would invoke the same method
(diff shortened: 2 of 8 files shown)


Source/FakeItEasy/Core/MethodInfoManager.cs
```diff
@@ -170,12 +160,23 @@ private static bool TypeImplementsInterface(Type type, Type interfaceType)
...
+            public Type Type { get; private set; }
 
             public override int GetHashCode()
             {
-                return this.Type.GetHashCode() ^ this.MethodInfo.GetHashCode();
+                unchecked
+                {
+                    return (Type.GetHashCode() * 23) + MethodInfo.GetHashCode();
+                }
             }
```
blairconrad: Technically the changes to TypeMethodInfoPair weren't needed, but the original GetHashCode 

1. exhibited the suboptimal "just xor things together" pattern, and
2. built the hashcode from non-readonly members, so it was technically possible to mutate the object and mess up the hashing

Source/FakeItEasy/Core/MethodInfoManager.cs
```diff
@@ -183,7 +184,7 @@ public override bool Equals(object obj)
             {
                 var other = (TypeMethodInfoPair)obj;
 
-                return this.Type.Equals(other.Type) && this.MethodInfo == other.MethodInfo;
+                return this.Type == other.Type && this.MethodInfo == other.MethodInfo;
```
blairconrad: This was just so things looked prettier. `Type` has an `operator==`, so why not use it?

Source/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
```diff
@@ -1,29 +1,24 @@
...
         {
-            lock (this.map)
-            {
-                var provider = this.map[eventHandler];
-                this.map.Remove(eventHandler);
-                return provider;
-            }
+            IEventRaiserArguments provider;
+            this.map.TryRemove(eventHandler, out provider);
+            return provider;
```
adamralph: Does this not change the behaviour of the method if the dictionary does not contain the specified key? I.e. the old version would throw `KeyNotFoundException`, the new would return null?

Source/FakeItEasy/Core/MethodInfoManager.cs
adamralph: :+1:

adamralph:
@blairconrad initial review done

Source/FakeItEasy/Core/MethodInfoManager.cs
adamralph: Indeed!

Source/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
blairconrad: Good point. It _does_ change the behaviour of the method, but not the overall behaviour. See `FakeManager.EventRule.HandleEventCall`. It will not call `RaiseEvent` (which is the only caller of this method) unless the eventCall is an event raiser, which means that the `EventHandlerArgumentProviderMap` `Contains` the `eventHandler`.

Because of this and the coupling of the argument provider registration and the raising of the event (it all happens in `robot.FellInLove += Raise.With(someEventArgs)`), there's no chance for the argument provider to be removed from the map between the `Contains` and `TakeArgumentProviderFor` being called.

I believe the code is sound as is, but if you'd like, I can make an adjustment.
Would you prefer an exception be thrown if `TryRemove` finds nothing?

adamralph: I suspected that the public behaviour of package hasn't changed, thanks for confirming that.

So we know the way in which the type is currently used means the package works. The remaining question is, does this type make sense in isolation? I.e. does it represent a sensible encapsulation and do it's method signatures convey the right semantics? I appreciate it's an internal type, but these questions are still important, since *someone* still has to work with it (contributors).

blairconrad: I'm not entirely sure how to answer this. The questions sound slightly leading, which makes me think you're trying to guide me to a realization about a deficiency in the type "in isolation". I fear I'm not getting it, but let's see if I can work myself up to it. Let's see where we end up.

I agree that the type should represent a sensible encapsulation and convey the right semantics. As such, I'd been satisfied with the type as it is. Setting aside the point that the signatures have not changed, so they convey the same semantics that they always did, we have a unit whose job is to provide a lookup table from event handlers to objects that can provide arguments that should be passed to an event handler. To support that, it offers the ability to add a `<hander, argument provider>` pair, to fetch and remove one, and also to check if there is such a record for a given handler.
This seemed to me a cohesive set of operations, working around a single responsibility.
However, there's nothing inherent in the signature of `TakeArgumentProviderFor` that indicates one way or another the behaviour when there's no matching argument provider for a supplied event handler, which could be considered a problem. We could change the signature to something like
`public bool TryTakeArgumentProviderFor(Delegate eventHandler, out IEventRaiserArguments argumentProvider)` in order to more clearly signal to the clients what's going on.
I don't hate it. I kind of like it, but will need to gauge how this affects the rest of the class and its clients

1. Given such a change, I think the need for `Contains` is greatly diminished, since production code only cares about whether an item is the map when it wants to use the arguments. As such, `EventRule.HandleEventCall` could be rewritten to attempt to take the arguments, and  on failure, add an event listener instead of firing one. A positive change, eliminating what I'd initially thought might've been your confusion about whether we'd ever try to take something from the map without it being there. Let's accept that.
    1. If `Contains` goes, `RaiseTests.Should_not_leak_handlers_when_raising` would have to change. I think it could use `TryTakeArgumentProviderFor` instead.
2. One may then notice that the signatures of the remaining methods on `EventHandlerArgumentProviderMap` match methods on `ConcurrentDictionary<Delegate, IEventRaiserArguments>` (One could even argue that `TryTakeArgumentProviderFor` could've been named `TryRemoveArgumentProviderFor`, as it more closely resembles the `TryRemove` precedent set by ConcurrentDictionary. Me, I prefer the `Take` in the name. I think it more clearly expresses "remove and return".),  and wonder about collapsing the class entirely. I'd be inclined against it myself, for reasons both aesthetic and practical:
    1. We'd have to decide where to hang the `EventRaiserDelegateComparer`. I think it fits very nicely nested inside its parent.
    2. I prefer to have a small, single-purpose class exposing its API to the clients, rather than the sprawling `ConcurrentDictionary` where we're supposed to limit our actions
    3. I think if we collapsed the type, we'd either have to start finding the instance using something like `ServiceLocator.Current.Resolve<ConcurrentDictionary<Delegate, IEventRaiserArguments>>()` or agument `ServiceLocator` to look items up by name or some such. Neither particularly appeals to me.

So, I've convinced myself that we may be better off with something like:

```csharp
internal class EventHandlerArgumentProviderMap
{
    public void AddArgumentProvider(Delegate eventHandler, IEventRaiserArgumentProvider argumentProvider)
    public bool TryTakeArgumentProviderFor(Delegate eventHandler, out IEventRaiserArgumentProvider argumentProvider)
}
```

(I sneaked a type rename in there:  `IEventRaiserArguments`  → `IEventRaiserArgumentProvider`, since the former made it sound like we had arguments, when really what we have is a thing that can make arguments for us.)

Has that at all ended up where you'd thought we might, or at least in pleasing location? Or have I misinterpreted your intent and further muddied the waters?

blairconrad: Although either way, I wonder if whatever change we want to the outside of `EventHandlerArgumentProviderMap` shouldn't be another PR…

blairconrad: I've pushed to https://github.com/blairconrad/FakeItEasy/tree/rejig-EventHandlerArgumentProviderMap, in case you want to see what the changes from 2 comments would look like.

Source/FakeItEasy/Core/MethodInfoManager.cs
TimLovellSmith: Technically this is breaking, is it really worth fixing the xor pattern?

Source/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
TimLovellSmith: >does this type make sense in isolation?
I vote it makes no more nor less sense now than it did before, and this PR is a net positive as is, waiting on a further refactoring of the class shouldn't block this?

blairconrad:
@adamralph, I think you're just busy at the meetup or whatever, but if I'm wrong, is there something more that you're waiting for from me at this point?
I'm happy to discuss and make changes to the PR, but amn't sure where to go right now.

If you're just busy, never mind and sorry for prodding.

adamralph:
@blairconrad I was just busy at our company meeting all of last week. I'll respond to your comments ASAP.

blairconrad:
@adamralph I figured, but sometimes issues do fall through the cracks. Sorry to be a bother.
Sounds like you had a good time.

Source/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
adamralph: @blairconrad I like the proposed change and, yes, the conversation has definitely ended up in a place I hope it would :smile: 

@TimLovellSmith a fair comment, and I feel bad for not responding until now. However, this is going to be a 2.0 feature so it's not going to be released as stable very soon anyway. Given the change is simple, I'd prefer to get it in here now before merging.

blairconrad: Yes, @TimLovellSmith, I should've also thanked you for your enthusiasm for getting things moving along. Fortunately, as @adamralph says, we're in no particular rush for this change (although maybe we'll release another beta soonish), and I have pretty high patience for insightful (and nitpicky :wink:) code review comments that have even a chance of producing better code, which this one did.

Thanks, @adamralph. I may be slow, but I get there in the end. Sometimes "there" is even the right place!
I'll add the new commit to the original PR branch soonish.

blairconrad:
Ah. I just came in to check to see if I'd assigned to you, @adamralph. Sigh. Thanks.

blairconrad:
This issue has been fixed in [release 2.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/2.0.0).

PR #509, opened 2015-07-18
Handle OutAttribute (not out) parameters properly

### LINKED ISSUE #508: Configuring Stream.Read(byte[], int, int) does not work - complains of mismatched out/ref args
From [Richard Lang's Stackoverflow Question "A.Fake<Stream>().Read(…) throwing InvalidOperationException"](http://stackoverflow.com/questions/31419162/a-fakestream-read-throwing-invalidoperationexception):

When running 

```csharp
[Test]
public void test_whatIsUpWithStreamRead()
{
    Stream fakeStream = A.Fake<Stream>();

    A.CallTo(() => fakeStream.Read(A<byte[]>.Ignored, A<int>.Ignored, A<int>.Ignored))
        .Returns(1);

    byte[] buffer = new byte[16];

    int numBytesRead = fakeStream.Read(buffer, 0, 16);

    Assert.AreEqual(1, numBytesRead);
}
```

The `fakeStream.Read` 
throws a System.InvalidOperationException with the message:

> The number of values for out and ref parameters specified does not match the number of out and ref parameters in the call.

I've reproduced, and started debugging, but do not yet know where we're going wrong.

blairconrad:
Fixes #508.

### DIFF
--- a/Source/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/Source/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -1,9 +1,7 @@
 ﻿namespace FakeItEasy.Specs
 {
     using System;
-    using System.Collections;
     using System.Collections.Generic;
-    using System.Diagnostics.CodeAnalysis;
     using System.Runtime.InteropServices;
     using FluentAssertions;
     using Machine.Specifications;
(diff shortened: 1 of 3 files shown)


blairconrad:
I'd considered changing `ExpressionArgumentConstraintFactory.IsOutArgument` to check the `IsByRef` as well, which would've worked, and left `GetArgumentConstraint` unchanged, but I felt like the `IsOut` and `IsRef` names would then be lies, since `IsRef` wouldn't exclude `IsOut`.
An third alternative would've been to make the change I just mentioned and add a
`!argument.ArgumentInformation.IsOut`
to `IsRefArgument`.

I'm happy to change the implementation to either other approach. Or something different, if anyone would like.

blairconrad:
Hmm. I created a PR to fix on the 2.0.0 release stream, as I thought that was what we did last time we needed to hotfix something (#440). Now that I look back, it seems we had actually made a hotfix branch for 1.25.2 and after merging it, ported forward to 2.0.0.

I'll repush in a bit.

blairconrad:
Replaced with #511.

PR #513, opened 2015-07-19
make hotfix/1.25 build in new TeamCity environment

blairconrad:
Began by cherry-picking @adamralph's #473.
Will need to adjust build to put artifacts in "artifacts" as well, as was done during part of #442.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 [Oo]bj
 [Bb]in
-Build*/*
+artifacts
 
 *.csproj.user
 *.vbproj.user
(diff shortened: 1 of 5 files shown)


blairconrad:
Okay. I think that's it. b0e29fecf821cfbffecd15ba3d20c3709c1bbd54 had to be cobbled together from @adamralph's 8acc95ba672a6ac447adb02f9c45995658089b98 and bits of changes (mostly directory creation) that had been added to `master` in the future. There may have been a more idiomatically gitish way to do it, but we ended up with good content, I believe.

Issue #520, opened 2015-08-01
Rename `IFakeOptionsBuilder` to `IFakeOptions`


blairconrad:
Extracted from https://github.com/FakeItEasy/FakeItEasy/issues/461#issuecomment-118212168, since

1. I think it's a good move anyhow, and 
2. If we do this, it'll affect #461, and may simplify our discussion of that issue

----

The more I look at things, the more I think that what users see as "fake options builders" could really just be "fake options" to them. That would alleviate _some_ of the naming pressure (for the API) in #461, not just in the discoverable interface we're talking about there, but when using `A.Fake<T>(Action<IFakeOptionsBuilder<T>>)` too.

If we made the change, the internal `FakeOptions` would collide with the class we'd use to implement the public `IFakeOptions`.
We could:

1. rename `FakeOptions` to something like `FakeCreationOptions`, or
2. collapse `FakeOptionsBuilder` and `FakeOptions`, as I'm not sure the latter is pulling its weight). If we did this we _could_ still implement two interfaces, to separate the "build up the options" actions from the "just get me the data so I can make a proxy" operations.

blairconrad:
I know when I first started using the `A.Fake<T>(Action<IFakeOptionsBuilder<T>>)` method, I found it a little confusing:

>I want to set the fake's options. Oh. There's an options builder. But I don't supply that, I supply an action that _acts_ on a builder.

That kind of thing. At least the interface is fluent, so it's fairly easy to get where you're going once you get started.

adamralph:
Sounds fine to me. We'd definitely have to get rid of `FakeOptions`, not least because ∃ `FakeOptionsBuilder<T>.Options` and we certainly wouldn't want that to become `FakeOptions<T>.Options`.

We'd also need to rename `IFakeOptionsBuilderForWrappers<T>` to `IFakeOptionsForWrappers<T>`. 

I'm also wondering we have this as a separate interface. There is only one implementation of `IFakeOptionsBuilder<T>` and that is `FakeOptionsBuilder<T>`, which also implements `IFakeOptionsBuilderForWrappers<T>`.

Could `IFakeOptionsBuilderForWrappers<T>` be folded into `IFakeOptionsBuilder<T>`? Or does it still need to exist because `IFakeOptionsBuilderForWrappers<T>.RecordedBy()` needs to return a `IFakeOptionsBuilder<T>` and not a `IFakeOptionsBuilderForWrappers<T>`?

[...]

PR #522, opened 2015-08-02
moving wrapper from FakeOptions to FakeOptionsBuilder

blairconrad:
I think an improvement over the arrangement introduced in #487.
@adamralph had [expressed doubts](https://github.com/FakeItEasy/FakeItEasy/pull/487#discussion_r32370948) about the mutable `FakeOptionsBuilder<T>.Recorder` field.
At the time, I couldn't think of a better alternative.
While preparing for #520, it seemed we'd be better off if `Wrapper` weren't on `FakeOptions`.
The change makes `FakeWrapperConfigurator` a stronger participant in the fluent fake configuration API.

I tried to preserve existing unit tests, but where it was super awkward, and where the conditions being tested were covered by all the new specs I'd added recently, I just dropped the tests.

### DIFF
--- a/Source/FakeItEasy.Tests/Core/FakeWrapperConfiguratorTests.cs
+++ b/Source/FakeItEasy.Tests/Core/FakeWrapperConfiguratorTests.cs
@@ -2,6 +2,7 @@
 {
     using System.Linq;
     using FakeItEasy.Core;
+    using FakeItEasy.Creation;
     using FakeItEasy.SelfInitializedFakes;
     using FluentAssertions;
     using NUnit.Framework;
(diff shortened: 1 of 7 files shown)


adamralph:
:+1: for dropping the unit tests which are covered by specs.

adamralph:
This seems much cleaner than before. Do I understand correctly that, if one is using just the fluent API, i.e. the regular 'faking' API then there are no changes required and it is only when someone is doing something funky with extending/customising FIE behaviours that their code will need to change?

blairconrad:
@adamralph, my understanding is a little different. Everything I touched was internal. Nobody outside should see _any_ change. Maybe I missed something?

adamralph:
@blairconrad sorry, you're absolutely correct.

PR #524, opened 2015-08-08
Remove sl5

### LINKED ISSUE #507: Drop support for Silverlight / Win8 / Win8.1
Castle.Core may be dropping support for Silverlight soon (see castleproject/Core#100).
If this happens (and nobody's arguing against it as is), we have no way to support Silverlight.
Win8 and Win8.1 support relies on our SL5 DLL, so that would go too. (And 8.1 may not be functional anyhow.)

Shall we keep an eye on castleproject/Core#100 and, if it's accepted, drop Silverlight support in 2.0?
Or just drop support in 2.0 anyhow?

blairconrad:
Fixes #507.


blairconrad:
Woo hoo! Thanks. Now I don't have to edit the SL solution when I move stuff around for #520.

PR #526, opened 2015-08-10
For discussion: Rename fake options builder

### LINKED ISSUE #520: Rename `IFakeOptionsBuilder` to `IFakeOptions`

blairconrad:
This is the general approach I'd like to take to fix #520, but wasn't quite sure about the ending and wanted some feedback.

1. After #522 was merged, the old `FakeOptions.fakeConfigurationActions` member was only used to configure our proxies, so I renamed to `proxyConfigurationActions`.
2. Then it was clear that `FakeOptions` was only used to set proxies' options, so I renamed that class to `ProxyOptions` (in order to make way for `FakeOptionsBuilder` to become `FakeOptions`)
3. At this point, I was able to rename `IFakeOptionsBuilder`, `IFakeOptionsBuilderForWrappers`, and `DefaultFakeCreatorFacade.FakeOptionsBuilder`, dropping the `Builder`. (At this point, I renamed parameters of type `<Action<FakeOptions>>` to `optionsBuilder`, rather than just `options`.)
4. Now `ProxyOptions` is used quite a bit internally, but only the `FakeOptions` class really needs to use the mutators on the class, so I extracted `IProxyOptions`, a read-only interface, to be used elsewhere.
5. Finally, I had `FakeOptions` absorb `ProxyOptions`, as @adamralph and I had discussed.

I'm generally happy with the arrangement, but before proceeding with a PR, I wanted feedback on a few points.

1. Do you think step 4 brings value? I like separating the read-only interface, just so it's clear that once the `ProxyOptions` are constructed, they aren't changed
2. `ProxyOptions` and `IProxyOptions` are not the best pair of names. I could rename to emphasize readonlyness of one or mutability of the other. `MutableProxyOptions` and `IProxyOptions`, or `ProxyOptions` and `IReadonlyProxyOptions`?
3. I'm not convinced that absorbing `ProxyOptions` is the best step. I was happy after step 4, and found that, due to `ProxyOptions`'s non-genericness, I had to introduce a new `EmptyProxyOptions` class in order to create a default proxy options object in the cases where this was used. I'd probably stop at step 4, if it were up to me.

Thoughts?

Thanks!
  B

### DIFF
--- a/Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
+++ b/Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
@@ -438,11 +438,11 @@
     <Compile Include="..\FakeItEasy\Creation\IFakeObjectCreator.cs">
       <Link>Creation\IFakeObjectCreator.cs</Link>
     </Compile>
-    <Compile Include="..\FakeItEasy\Creation\IFakeOptionsBuilder.cs">
-      <Link>Creation\IFakeOptionsBuilder.cs</Link>
+    <Compile Include="..\FakeItEasy\Creation\IFakeOptions.cs">
+      <Link>Creation\IFakeOptions.cs</Link>
     </Compile>
-    <Compile Include="..\FakeItEasy\Creation\IFakeOptionsBuilderForWrappers.cs">
-      <Link>Creation\IFakeOptionsBuilderForWrappers.cs</Link>
+    <Compile Include="..\FakeItEasy\Creation\IFakeOptionsForWrappers.cs">
+      <Link>Creation\IFakeOptionsForWrappers.cs</Link>
     </Compile>
     <Compile Include="..\FakeItEasy\Creation\IProxyGenerator.cs">
       <Link>Creation\IProxyGenerator.cs</Link>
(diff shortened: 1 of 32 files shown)


adamralph:
1. I think that step 4 brings value. This is the type used in method signatures and is a nice way for those methods to declare what they require from a proxy options object, whereas `ProxyOptions` is only ever new'd up and mutability is an implementation detail.
1. I don't mind `ProxyOptions` and `IProxyOptions`. As in previous point, the mutability of `ProxyOptions` is an implementation detail.
1. I think step 5 is a step too far. I also vote for stopping at step 4.

blairconrad:
@adamralph, music to my ears. I love it when we agree, although I learn less than when we don't. :wink: 

I'll give everything a reread and submit a proper 4-step PR as soon as I'm able.

PR #527, opened 2015-08-10
refactor: remove redundant package entries in Source/packages/repositories.config

adamralph:
re #507

When I reviewed #524 locally in VS, this change was left over in my working dir.


blairconrad:
Embarrassing. I'm not sure how I missed this.
I was using fancy tools to find SL and Silverlight and everything, I thought.

Thanks for catching this!

Issue #484, opened 2015-05-04
Make FakeScope threadsafe


TimLovellSmith:
![whattt](https://cloud.githubusercontent.com/assets/2711795/7458929/43b123e4-f24d-11e4-89ea-54eeb42efc71.png)

This is really confusing me. How can this be a failure - doesn't that mean the 1 call it got matches?
The code in question looks like: 

```C#
var fakeDns = A.Fake<IAzureDnsResourceManager>();

/// do stuff

var theCall = A.CallTo(() => fakeDns.GetCNameAync("busy123"));
theCall.Returns(Task.FromResult("busy123456"));

/// do stuff

theCall.MustHaveHappened();
```

FakeItEasy version 1.25.1. What I'm doing *different* from before is running tests in parallel... if I rerun this test on its own, it passes.

TimLovellSmith:
Ah, looks like this is issue #60. Sorry :)

blairconrad:
@TimLovellSmith, I completely believe that you're running afoul of #60, but are you sure that the code you're having problems with is

`var theCall = A.CallTo(() => fakeDns.GetCNameAync("busy123"));`?

I have only ever seen problems when we use `That.Matches` or `Ignored`, and I'd be very excited/sad to find out that the multiple threads are causing problems with other kinds of verifications.

(Also, your screenshoot appears to include a `Ignored`, so that makes me feel hopeful, but I thought I'd check just in case.)

Thanks.

blairconrad:
Oh, and if you're interested, you may want to take the nupkg from
http://teamcity.codebetter.com/viewLog.html?buildId=189740&tab=artifacts&buildTypeId=bt929#!132is1d

for a spin. It contains my proposed fix for #478, which is a facet of #60.

TimLovellSmith:
Yes, I'm completely sure that's the code I'm having problem with. I'll give your nupkg a spin and see if it helps, thanks!

TimLovellSmith:
So. I'm seeing less problems with the new nupkg, I think. But I'm still seeing some problems, and the problem basically seems to be that the fake call that *should* be configured did not actually get invoked. Here is some code to explain. I have this line of code:

StorageAndAffinityGroup sa = await AzureResourceManager.CreateStorageAccountInAnySubscription(scope, Pool, dep.Name, dep.Region, suggestedName, token);

That call is supposed to get faked out by the following:

[...]

Issue #458, opened 2015-02-16
Switch from MSpec to xbehave


blairconrad:
I've lately been preferring to write comprehensive end-to-end tests (our "specs" or acceptance tests) to unit or integration (component) tests whenever it's feasible.
To this day, every time I do something more complicated than add a single test case, I have to look up the MSpec syntax and/or copy an existing spec.
I also found it difficult and cumbersome to reuse anything other than the test setups.

Perhaps we could adopt another framework? One option would be [xbehave.net](http://xbehave.github.io/). The first time I used it, I found the test pattern easy to understand and apply.
In addition, being based on XUnit, one can leverage the lifecycle and reuse patterns in that framework, and use any testrunners that work for XUnit.

Thoughts? Objections?

adamralph:
Obviously I'm all for [xbehave.net](http://xbehave.github.io/) :wink:

adamralph:
https://github.com/ursenzler/MSpec2xBehaveConverter /via @ursenzler

Just saying' :wink:

blairconrad:
Interesting. I'd rather save this kind of activity for post-2.0 release, but a tool sounds fun, and looks like it would give us a leg up.

philippdolder:
appccelerate is migrating/ed as well :-)

ursenzler:
If you need help with the migrator, let me know. It probably can be tweaked to better match your coding style.

adamralph:
@philippdolder the testing toolkit of champions :wink: 

@ursenzler to help get us started, it would be really great if you could run your migrator against the FakeItEasy source and open a PR, even with just the raw results of the migration.

ursenzler:
I can do that for you.

matkoch:
I authored a new test framework called [TestFx](https://github.com/matkoch/TestFx). I just [ported some tests](https://gist.github.com/matkoch/79a8d5d5c4ea477673e9) from the `CreationSpecs` *and* from the `NullGuardedConstraintTests` (although it is a NUnit test - just for showcasing).

ursenzler:
See pull request https://github.com/FakeItEasy/FakeItEasy/pull/528 for how it would like with my converter.

Let me know, whether I should convert the rest of the specs.

Issue #437, opened 2015-01-13
Adding additional attributes to a fake of an already faked type fails silently


adamralph:
Spun off from #436.

``` C#
var constructor = typeof(BarAttribute).GetConstructor(new[] { typeof(string) });

var fake1 = A.Fake<Foo>(x => x.WithAdditionalAttributes(new[] { new CustomAttributeBuilder(constructor, new object[] { "1" }) }));
var fake2 = A.Fake<Foo>(x => x.WithAdditionalAttributes(new[] { new CustomAttributeBuilder(constructor, new object[] { "2" }) }));
var fake3 = A.Fake<Foo>(x => x.WithAdditionalAttributes(new[] { new CustomAttributeBuilder(constructor, new object[] { "3" }) }));

Assert.AreEqual("3", fake3.GetType().GetCustomAttributes(typeof(BarAttribute), true).Cast<BarAttribute>().First().Value);
```

This test will fail because the attribute with the value of "1" is returned although I'd expect it to return the attribute with value "3" (since I'm checking fake3). This is because all fakes are of the same type (Castle.Proxies.FooProxy, DynamicProxyGenAssembly2, DynamicProxyGenAssembly2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=a621a9e7e5c32e69) and that type is only created once.

Until a solution to #436 can be found (e.g. each fake has it's own type) we should at least throw an exception when the second fake is created, stating that the type has already been created and therefore additional attributes cannot be added. Nasty, but at least it's owning up and telling the truth, rather than silently failing and leading the test author into forensics to discover why their tests aren't working.

adamralph:
Hold off for now - only required if we can't find a (timely) solution to #436.

blairconrad:
> Hold off for now - only required if we can't find a (timely) solution to #436.

Based on https://github.com/FakeItEasy/FakeItEasy/issues/436#issuecomment-222655976, I don't think we're getting a fix for #436 anytime soon. Do we still see a reason to hold off?
I wouldn't wait just to avoid extra work, and I don't think that introducing the exception and then rolling it out when #436 is breaking, so I think we could keep the minor release train going.

adamralph:
:+1: to go ahead with this now.

adamralph:
Would we consider this a breaking change?

blairconrad:
I wouldn't consider it to be breaking. It already doesn't work. We'd just be telling people now.

[...]

PR #528, opened 2015-08-12
conversion of MSpec to xBehave specifications

ursenzler:
1. I added nuget references to xBehave.Core on the specs assembly.
1. I converted the Lazy Spec for first feedback from you guys
 - manually added a Subject()
 - run converter
 - minimal manual layouting
 - renamed File

Questions:
- does it look like you want it to look?
 - file name without Specs so that class and filename match
 - layout of the file content
- should I continue with the other specs?

Of course, the texts of the scenarios and especially the establishs are not very meaningful after the raw conversion and need manual "optimization".


blairconrad:
Thanks for this, @ursenzler. It's very interesting. And thanks for doing one file first, so as not to overwhelm us!

I'm an xBehave novice, so am not familiar with all the idioms, but what you have reads well, I think, and looks good.

One downside is that the change has introduced a StyleCop warning, since we now have an interface at the bottom of the class, instead of before any of the test methods. We'd have to figure out whether we want to relax the rule for the Specs project, or whether we'd want to move `IFoo` up to the top of `Lazy`. I do hate to relax rules, or have inconsistent ones ("interfaces are at the top for most projects, but bottom for the specs" (if that's even possible)), but the interface would be annoying clutter at the top…

I think, too, we'd need to establish (no pun intended) conventions for the step names in the specs as well. 
For example, I'm a fan of Given/When/Then(/And) over Establish/When/It, but I imagine this is something that requires human intervention to get right anyhow.
Or is the tool be able to make Givens out of Establishes, for example?

Please don't take all my comments as complaints or lack of interest. I like where this is headed!

ursenzler:
I did one file to get exactly this kind of feedback. It's your (the team) project and it has to be the way you like it.

I think that renaming _establish_, _when_, _it should_ to something else could possibly be done with a simple Search/Replace or a R# Structural Search.

Were the interfaces should go and whether to relax StyleCop rules for specs, is completely up to you.

For better readability, I'd replace the `"establish"` with something more expressive, like `"establish a LazyFactory"`. And I'd replace the scenario (method) name with a short name for the scenario, like `ReturningLazies`


As a side note: when all specs are written with xBehave, maybe you could take use of my xBehaveMarkdownReport to create a markdown report out of the specs. Great when you want to add them to the documentation.

Second side note: I disabled StyleCop on my machine because in my current project, we use StyleCop.CodeAnalyzers (VS 2015), so it didn't realize this because it did not show an issue.

adamralph:
This is awesome!!! :dizzy: 

I think we can ignore the StyleCop violations for now, and just blitz through the warnings as a final manual commit to put them right.

If we have some control over the output, I'd prefer this more idiomatic format:

```C#
"Given ..."
    .x(() =>
    {
        fake = A.Fake<ILazyFactory>();
    });

"When calling a method that returns a lazy"
    .x(() =>
    {
        lazy = fake.Create();
    });

"Then it should return a lazy"
    .x(() =>
    {
        lazy.Should().NotBeNull();
    });

"Then it should return a lazy whose value is a dummy"
    .x(() =>
    {
        lazy.Value.Should().Be(FooFactory.Instance);
    });
```

Even better would be if the tool could detect that a lambda is a single statement and collapse if so:

```C#
"Given ..."
    .x(() => fake = A.Fake<ILazyFactory>());

"When calling a method that returns a lazy"
    .x(() => lazy = fake.Create());

"Then it should return a lazy"
    .x(() => lazy.Should().NotBeNull());

"Then it should return a lazy whose value is a dummy"
    .x(() => lazy.Value.Should().Be(FooFactory.Instance));
```

ursenzler:
The tool cannot detect a single statement because it is purely text based and does not know about any semantics. This would be a manual step (Alt-Enter Spree with R#).

I can add an option to the tool to specify, whether it should use `_` or `x` as the function. But I prefer `_` because it's less noise to my eyes. But if you want `x` I'll add it.

I can also add an option for the additional newline after the string.

adamralph:
@ursenzler that sounds great. No problem with the single statement, we can fix that up afterwards. Can it also do Given/When/Then?

ursenzler:
I could replace establish with Given, but the text should be updated manually anyway. The `when_this_and_that` and `it should` are taken from the already existing texts (except the leading `it`) so it has to be improved manually, too.

blairconrad:
@ursenzler, good point about structural search and replace. That could definitely be done at the end of the process.

As you say, all the text has to be updated manually, but an option for (for example) turning `It` into something that starts with "Then it" instead of "it" might relieve some of the tedium.
Then again, perhaps a simple search and replace could help with this as well, but I'm forgetful, so always like to get tools to do as much as possible. :wink:

That's not to discount the value of the conversion that the tool brings already. It is quite nice.

@adamralph, you'd suggested the following as a preferred format:

```csharp
"Then it should return a lazy"
    .x(() => lazy.Should().NotBeNull());

"Then it should return a lazy whose value is a dummy"
    .x(() => lazy.Value.Should().Be(FooFactory.Instance));
```

Educate me, please, as on my own, I might've gone for something like

```csharp
"Then it should return a lazy"
    .x(() => lazy.Should().NotBeNull());

"And the lazy's value is a dummy"
    .x(() => lazy.Value.Should().Be(FooFactory.Instance));
```

Is this not established BDD syntax?

adamralph:
@blairconrad oops, yes, it should have been `And` on the second assertion

blairconrad:
@adamralph Look at me, making wild guesses that turn out to be right!
Having used up my luck for the day, I will refrain from playing backgammon until tomorrow.

adamralph:
@blairconrad and if I send you a :four_leaf_clover: for your :game_die:?

ursenzler:
Just to add some thoughts about _then_, _should_ and so on.

Although writing the specs like _given this_, _and that_, _when doing something_, _then should that, _and that_ results in a description that looks more like prose, I prefer something different after some years of writing specs in different forms:

I like to write the _given_ statements as plain facts like: _the system is running_, _the user is logged in_ etc.

I write the actions starting with _when_ (no debate here :-) )

And finally, the assertions start all with _should_ to emphasize that that is a wish and maybe not the reality (the spec could fail).

The advantage is that ordering among all assertions is not important. It is really not hard to add the _and_ just in your head when reading a report. The _given_ and _and_ is noise to my eyes.

And of course, I could do that without the need for manual editing ;-)


For the conversion itself: it is easier to convert one spec at a time and get it running before going to the next. Therefore, we should know what the end result should look like and not just convert everything into something that cannot be executed.

adamralph:
> results in a description that looks more like prose

I think that's the whole idea :wink:

Anyway, I'm not sure it's worth getting into a debate about the string literals. The biggest value is to make the switch, so I think just going with the idiomatic pattern is the simplest thing to do for now. Note that this isn't an invention of xbehave, it comes from Gherkin, which is idiomatic (or even strict) in several other frameworks, e.g. Cucumber and SpecFlow.

ursenzler:
@adamralph one of the reasons I like xBehave more than the other frameworks is that I'm not bound to strict rules :-)

adamralph:
That's good and I'm glad to hear it :smile: 

Let's make this democratic then. I vote for Given/When/Then garnished with And :stuck_out_tongue_closed_eyes: 

Reasons:
* Idiomatic. All samples and documentation use this convention.
* Follows Gherkin. Familiar to anyone coming from Cucumber or SpecFlow backgrounds or any other Gherkin based tool.
* Reads well. I like the fact that you can read the test output as natural English sentences.

blairconrad:
I'm a BDD neophyte, but do like the idea of following along with Gherkin. +1 to @adamralph's suggestion.

ursenzler:
Okay, to sum it up.

Step 1 (running with xBehave)
1. I'll do a raw convertion
1. I'll make manual adjustments where needed to make the specs runnable

Step 2 (make tests nice)
1. change to Given/When/Then/And (partly done with search/replace)
2. clean-up (manually)

ursenzler:
I ran into a problem with a spec that runs in isolation, but fails when run with others. We need to check the spec after the conversion. It's `when_WithAdditionalAttributes_is_used_to_configure_a_fake`

ursenzler:
Things to clean up:

[ ] check test `when_WithAdditionalAttributes_is_used_to_configure_a_fake`
[ ] replace class Catch with Record.Exception (just did this for faster conversion)
[ ] use Given/When/Then/And everywhere
[ ] use lambda statements where for one line steps

blairconrad:
Wow. @ursenzler. An impressive amount of work. Thank you. I really like your checklist of things to clean up as well.

I see we have an "All checks have passed". Do you think this indicates that whatever was wrong with `when_WithAdditionalAttributes_is_used_to_configure_a_fake` has improved, or do you suspect non-determinism?

ursenzler:
The test still fails on my machine when not run isolated. It fails always. I think, one of you should check this test (isolated and with all others) on a dev machine when run with R# xunit test runner.

ursenzler:
Okay, the raw conversion is done and the reference to Machine.Specifications is removed. see checklist above for TODOs.

blairconrad:
@ursenzler, I think that `when_WithAdditionalAttributes_is_used_to_configure_a_fake` is failing when run in conjunction with `when_WithAdditionalAttributes_is_used_to_configure_a_fake_twice` due to 
#437. Before conversion, the tests were applying attributes to two different types (`when_WithAdditionalAttributes_is_used_to_configure_a_fake.IInterfaceThatWeWillAddAttributesTo ` and `when_WithAdditionalAttributes_is_used_to_configure_a_fake_twice.IInterfaceThatWeWillAddAttributesTo `). During your conversion, you collapsed the two types.

I notice a few other failures when running all the tests at once. Some of which pop in and out.
I haven't had too much time to dig into them, but  think at least some of them are probably arising because Fake Scopes aren't threadsafe.

ursenzler:
@blairconrad good catch.

ursenzler:
All specs run now (run from console).

I added  `[assembly:Xunit.CollectionBehavior(MaxParallelThreads = 1)]` to an assembly info so that tests are not run in parallel and introduced multiple interfaces for the specs adding attributes to an interface.

blairconrad:
Wow, thanks @ursenzler!
I was swamped last night, and (obviously) got nothing done.
Running the specs in one thread for now is an acceptable pragmatic step.
We have the longer-term problem in the library itself of the non-threadsafe Scopes, but that's known - see #484.
Until we get it resolved, I think "not antagonizing the FakeScope" is probably the best approach.

ursenzler:
Quick question:

Which style do you prefer for the placement of the `.x(() => `:

on same line as the step implementation:
```
        [Scenario]
        public void VirtualMethod(
            MakesVirtualCallInConstructor fake)
        {
            "when faking a class with a virtual method"
                .x(() => fake = A.Fake<MakesVirtualCallInConstructor>());

            "then should return a default value when the method is called during the constructor"
                .x(() => fake.VirtualMethodValueDuringConstructorCall.Should().Be(string.Empty));

            "and should return a default value when the method is called after the constructor"
                .x(() => fake.VirtualMethod("call after constructor").Should().Be(string.Empty));

            "and should record the method call during the constructor"
                .x(() => A.CallTo(() => fake.VirtualMethod("call in constructor")).MustHaveHappened());

            "and should record the method call after the constructor"
                .x(() =>
                    {
                        A.CallTo(() => fake.VirtualMethod("call after constructor")).MustHaveHappened();
                    });        }
```

on the same line as the text:
```
[Scenario]
        public void VirtualProperties(
            FakedClass fake)
        {
            "when faking a class with a virtual method".x(() => 
                fake = A.Fake<FakedClass>());

            "then should return a default value when a reference type property is called during the constructor".x(() => 
                fake.StringPropertyValueDuringConstructorCall.Should().Be(string.Empty));

            // The test for the default value of a value type is a regression test for https://github.com/FakeItEasy/FakeItEasy/issues/368:
            "and should return a default value when a value type property is called during the constructor".x(() => 
                fake.ValueTypePropertyValueDuringConstructorCall.Should().Be(0));

            "and should return the value assigned during constructor when a reference type property is gotten after the constructor".x(() => 
                fake.StringProperty.Should().Be("value set in constructor"));

            "and should return the value assigned during constructor when a value type property is gotten after the constructor".x(() =>
                {
                    fake.ValueTypeProperty.Should().Be(123456);
                });
        }
```

Please note that the last step of both examples would be an expression lambda, too. I just wrote it as a statement lambda just for showing how it would like for multi-line steps.

The second example goes better with code layouting of Visual Studio and R# (but that could just be my installation - not sure)

Is there anything in these examples that do not match the style you'd like to have?

Once I know what you prefer, I can layout the other specs.

blairconrad:
Thanks for asking, @ursenzler.

I've no real preference myself. @adamralph has more experience with xBehave.net, and is opinionated, but is somewhat unavailable for a little bit. We may get lucky and hear from him.
If you can't defer the decision around `.x()=>` placement (And why not? Take a break! You've been working hard.), perhaps best to conform to the examples in the [xBehave.net docs](https://github.com/xbehave/xbehave.net/wiki/Quickstart):

```c#
"Given the number 1"
                .x(() => x = 1);
```

ursenzler:
I'm finished on my end now. 

I think it makes more sense if a member of the team rewrites the step names to Given/When/Then/And and verifies whether the texts are meaningful.

Killing the reference to the unit test assembly is out of the scope of the conversion.

Running the spec marked with a Trait has to be done on the build server, I guess.

Let me know if I missed anything or I can help in any other way.

blairconrad:
I agree that it's probably up to us to finalize the language of the G/W/T/A, or at least unfair to ask you to do! :wink:

And yeah, the other things you mention are for later or another issue.

Thanks very much for all your hard work @ursenzler. :pray:

I'm currently focused (when I get FakeItEasy time) on a few of the issues we still need for [2.0](https://github.com/FakeItEasy/FakeItEasy/issues?q=is%3Aopen+is%3Aissue+milestone%3A2.0.0), and would be inclined to defer to @adamralph for most of the niggly xBehave syntax phrasing anyhow, so won't jump on this right away.

One comment is that in order to merge this, we'd want it to be coming in from a non-master branch, but I also wonder if we shouldn't merge it into a non-master branch as well, to get it into the project. Then we'd have the option of picking away at it while we work on more urgent issues.
I think this would make more sense, but would like buy-in from another owner before taking that step.

ursenzler:
The current state of the specs rendered as markdown: https://gist.github.com/ursenzler/5ef86b4ca1cd776ad3cf

adamralph:
:heart_eyes: that is so awesome.

Thank you so much @ursenzler, this really is superb work.

Regarding the branch, it doesn't matter that the PR is from @ursenzler's master branch. It's a straight line of commits from the upstream master so that's fine. It will only be problematic for @ursenzler when he send his next PR :wink: since he'll have to reset and force push his master to get it back in line.

Since there are some more commits to be added (by either myself or @blairconrad) I think the best way to handle this is to send a superset PR. When that is merged, this PR will also be marked as merged.

adamralph:
Linking to #458

adamralph:
FYI - I've started looking at this. So far all looks good. I may just squash some commits.

blairconrad:
Cool.

Source/FakeItEasy.Specs/AssertingOnSetOnlyProperties.cs
```diff
@@ -0,0 +1,36 @@
...
+    public  class AssertingOnSetOnlyProperties
+    {
+        [Scenario]
+        public void SetOnlyProperties(
+            ISetOnly setOnly)
+        {
+            "establish"
+                .x(() => setOnly = A.Fake<ISetOnly>());
+
+            "when assertion on set only properties"
```
blairconrad: super-nitpick: consider "set-only"

blairconrad: Oh, and "asserting". Maybe this will all be covered in the FakeItEasy-team-lead step name rewriting?

blairconrad:
> I may just squash some commits.

@adamralph does this mean we're dispensing with (or delaying) the "Given/When/Then/And Then" rewrite?

Also, how can I help? More annoying nitpicks like in AssertingOnSetOnlyProperties?
Or should I stand back for now?

adamralph:
I think, to unblock other work, we can delay the "GWT" re-write for later. It can happen as an ongoing process.

I'm just tidying up a couple of things as discussed. I should be able to send a superset PR fairly soon.

blairconrad:
Okay. GWT rewrites to be completed in the background and certainly as we touch a file, not unlike the NUnit Assert → FluentAssertions coversions? I can live with that.
I've suggesetd jimmyheaddon add his specs to `CreationOptionsSpecs.cs`, so wouldn't mind tidying that one up early, just so he doesn't feel he has to!

adamralph:
Yep, very much like the gradual FluentAssertions conversion.

adamralph:
Thanks  @ursenzler !

PR #529, opened 2015-08-12
Rename IFakeOptionsBuilder to IFakeOptions

### LINKED ISSUE #520: Rename `IFakeOptionsBuilder` to `IFakeOptions`

blairconrad:
Fixes #520.

Contains steps 1–4 from #526, plus the obligatory "cleanup" commit.

connects to #520

### DIFF
--- a/Source/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
+++ b/Source/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
@@ -11,7 +11,7 @@
         private class AutoFakePropertyRule
             : IFakeObjectCallRule
         {
-            public FakeManager FakeManager { get; set; }
+            public FakeManager FakeManager { private get; set; }
 
             public int? NumberOfTimesToCall
             {
--- a/Source/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
+++ b/Source/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
@@ -56,7 +56,7 @@
 
             private static bool TryCreateFake(Type type, out object fake)
             {
-                return FakeAndDummyManager.TryCreateFake(type, new FakeOptions(), out fake);
+                return FakeAndDummyManager.TryCreateFake(type, new ProxyOptions(), out fake);
             }
 
             private static object DefaultReturnValue(IInterceptedFakeObjectCall fakeObjectCall)
--- a/Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
+++ b/Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
@@ -438,11 +438,11 @@
     <Compile Include="..\FakeItEasy\Creation\IFakeObjectCreator.cs">
       <Link>Creation\IFakeObjectCreator.cs</Link>
     </Compile>
-    <Compile Include="..\FakeItEasy\Creation\IFakeOptionsBuilder.cs">
-      <Link>Creation\IFakeOptionsBuilder.cs</Link>
+    <Compile Include="..\FakeItEasy\Creation\IFakeOptions.cs">
+      <Link>Creation\IFakeOptions.cs</Link>
     </Compile>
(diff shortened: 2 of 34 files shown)


jimmyheaddon:
For reference, pull request [#536](https://github.com/FakeItEasy/FakeItEasy/pull/536) will need merging with this one at some point.

blairconrad:
Yup, @jimmyheaddon. My plan was to, assuming the other owners have no problem with #536, merge that one first and rebase my PR.

blairconrad:
Rebased! I think we're ready for review.

blairconrad:
Unassigning. Since we now have a veritable army of owners, in general there's no benefit in assigning a PR to anyone in particular.

Source/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
```diff
@@ -11,7 +11,7 @@ public partial class FakeManager
         private class AutoFakePropertyRule
             : IFakeObjectCallRule
         {
-            public FakeManager FakeManager { get; set; }
+            public FakeManager FakeManager { private get; set; }
```
thomaslevesque: The write-only property is a bit weird... Wouldn't it be better to make it private and initialize it from the constructor?

thomaslevesque:
I just reviewed the code. There are many changes, but most of them are trivial, so I don't have much to say (just one minor remark).

(I did learn something along the way: I didn't know about the `[Fake]` and `[UnderTest]` attributes. That's great stuff! I was using an auto-mocking container to achieve the same result, but the attribute-based approached is much better.)

blairconrad:
Thanks. I enjoyed your remark and will look at tonight.

The attributes aren't bad. I've never used them on my own. Sometimes I think it's a "cool feature that maybe belongs outside the core package". But I think things like that about lots of stuff.
I'm anti-feature, in general.

adamralph:
@blairconrad for my understanding, can you please explain the rationale behind the first commit:

> renaming fakeConfigurationActions to proxyConfigurationActions

adamralph:
Oh, and also the second commit:

> renaming FakeOptions to ProxyOptions

thomaslevesque:
@adamralph, it's explained in #526

adamralph:
@thomaslevesque thanks! Memory like a sieve!

BTW - @blairconrad, I love the separation of commits here, it makes this so much easier to review (and only heightens my embarrassment for not doing so sooner).

Source/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
adamralph: Agreed, but I think we can deal with this in a separate PR. This change only highlights the design problem by minimising the accessibility to what is needed, and thus highlights the design problem.

blairconrad:
Thanks, gentlemen. I was a little occupied last night and today or would've gotten into the conversation.
@thomaslevesque, nice explaining about #526. I probably would've forgotten as well.
@adamralph, if you like those commits, wait until you see how I broke the next change up into 22 simple commits!
And don't be embarrassed.

Issue #462, opened 2015-02-24
complain when IFakeOptionsBuilder.Implements is passed a non-interface


blairconrad:
It's supposed to allow us to add interfaces to a fake, but it accepts types.

```c#
A.Fake<IInterface>(builder => builder.Implements(typeof (SomeClass)));
```

passes silently, but with no obvious benefit.

Casting to `SomeClass` fails.

Found while investigating an answer to [Having an interface fake inherit from abstract while both share same interface inheritance](http://stackoverflow.com/questions/28697918/fakeiteasy-having-an-interface-fake-inherit-from-abstract-while-both-share-sam).

blairconrad:
Changing to **enhancement** since FakeItEasy is doing nothing wrong when used as intended. This is just a bonus warning for users to help them use it correctly.

jimmyheaddon:
Anyone mind if I take this?  Seems like a relatively simple modification for my first contribution

blairconrad:
Hi, @jimmyheaddon.
Not at all. Thanks for the interest. I'll take a gander at the pull request as soon as I can.

blairconrad:
Fixed by #536.

blairconrad:
@jimmyheaddon, we often thank people on Twitter and add them to a "contributors" list.
Based on your GitHub handle and avatar you're using here (which I'm taking as a hint to your geographical location), I think I know who you are there.
Am I right?
Do you mind?

In related news, thanks for your hard work.
Look for your name in the release notes! :trophy:

jimmyheaddon:
@blairconrad feel free, I think I'm the only person of my namesake on Twitter and LinkedIn :smile:

Hopefully the next PR won't be such hard work for you guys, now I have a better understanding how you work!

adamralph:
@jimmyheaddon on the contrary, it's always a pleasure to help get a new contributor on board.

blairconrad:
Ditto. It's been a delight, @jimmyheaddon.

blairconrad:
@jimmyheaddon, thanks very much for your work on this issue. Look for your name in the release notes. :trophy:

This issue has been fixed in [release 2.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/2.0.0).

PR #536, opened 2015-09-16
#462 - Throw when IFakeOptionsBuilder.Implements is passed a non-interface

jimmyheaddon:
Modifications to throw when IFakeOptionsBuilder.Implements() is passed a non-interface
#462

### DIFF
--- a/Source/FakeItEasy.Tests/Creation/DefaultFakeCreatorTests.cs
+++ b/Source/FakeItEasy.Tests/Creation/DefaultFakeCreatorTests.cs
@@ -16,8 +16,8 @@ namespace FakeItEasy.Tests.Creation
 
         [SuppressMessage("Microsoft.Performance", "CA1823:AvoidUnusedPrivateFields", Justification = "Used reflectively.")]
         private object[] optionBuilderCalls = TestCases.Create<Func<IFakeOptionsBuilder<Foo>, IFakeOptionsBuilder<Foo>>>(
-                x => x.Wrapping(A.Fake<Foo>()).Implements(typeof(Foo)),
-                x => x.Implements(typeof(Foo)),
+                x => x.Wrapping(A.Fake<Foo>()).Implements(typeof(IFoo)),
+                x => x.Implements(typeof(IFoo)),
                 x => x.WithArgumentsForConstructor(() => new Foo()),
                 x => x.WithArgumentsForConstructor(new object[] { A.Fake<IServiceProvider>() }),
                 x => x.Wrapping(A.Fake<Foo>()).RecordedBy(A.Fake<ISelfInitializingFakeRecorder>()))
--- a/Source/FakeItEasy/Creation/FakeOptions.cs
+++ b/Source/FakeItEasy/Creation/FakeOptions.cs
@@ -2,6 +2,7 @@ namespace FakeItEasy.Creation
 {
     using System;
     using System.Collections.Generic;
+    using System.Globalization;
     using System.Reflection.Emit;
 
     internal class FakeOptions
@@ -29,6 +30,17 @@ namespace FakeItEasy.Creation
 
         public void AddInterfaceToImplement(Type interfaceType)
         {
+            Guard.AgainstNull(interfaceType, "interfaceType");
+
+            if (!interfaceType.IsInterface)
+            {
+                throw new ArgumentException(
+                    string.Format(
+                        CultureInfo.CurrentCulture,
+                        "The specified type '{0}' is not an interface",
+                        interfaceType.FullNameCSharp()));
+            }
+
             this.additionalInterfacesToImplement.Add(interfaceType);
         }
 
--- a/Source/FakeItEasy.Specs/CreationOptionsSpecs.cs
+++ b/Source/FakeItEasy.Specs/CreationOptionsSpecs.cs
@@ -341,6 +341,19 @@
         }
 
         [Scenario]
+        public static void ImplementsNonInterfaceType(
+            Exception exception)
+        {
+            "When a fake is built to implement a non-interface type"
+                .x(() => exception = Record.Exception(() => A.Fake<MakesVirtualCallInConstructor>(options => options
+                                                                .Implements(typeof(MakesVirtualCallInConstructor)))));
+
+            "Then it should throw an argument exception"
+                .x(() => exception.Should().BeAnExceptionOfType<ArgumentException>()
+                             .WithMessage("*The specified type 'FakeItEasy.Specs.MakesVirtualCallInConstructor' is not an interface*"));
+        }
+
+        [Scenario]
         public static void MultipleImplementsConfigurations(
             MakesVirtualCallInConstructor fake)
         {


jimmyheaddon:
For reference I've updated both implementations, and have added a Guard check along with a Type check. 

The XML exception comments already existed, therefore I've matched the exception string and exception type.

Source/FakeItEasy.Tests/Creation/DefaultFakeCreatorTests.cs
```diff
@@ -114,6 +121,21 @@ public void CreateFake_should_pass_options_builder_that_returns_itself_for_any_c
             Assert.That(call.Invoke(builderPassedToAction), Is.SameAs(builderPassedToAction));
         }
 
+        [TestCaseSource("invalidOptionBuilderCalls")]
+        public void CreateFake_should_throw_when_type_passed_to_options_builder_is_not_an_interface(Func<IFakeOptionsBuilder<Foo>, IFakeOptionsBuilder<Foo>> call)
```
blairconrad: I don't understand the benefit of parameterizing this test. What do we gain over a single test that just calls `Implements` with a non-interface?

Source/FakeItEasy/Core/FakeWrapperConfigurator.cs
```diff
@@ -47,6 +47,13 @@ public IFakeOptionsBuilder<T> WithAdditionalAttributes(IEnumerable<CustomAttribu
 
         public IFakeOptionsBuilder<T> Implements(Type interfaceType)
         {
+            Guard.AgainstNull(interfaceType, "interfaceType");
+
+            if (!interfaceType.IsInterface)
+            {
+                throw new ArgumentException("The specified type is not an interface");
+            }
+
```
blairconrad: The change looks good to me, but I'd've thought to just make the adjustment to `FakeOptions.AddInterfaceToImplement`, as that's where all implementations end up.
I'm not necessarily against the way you've done it, but wondered if you had a reason to prefer putting the change in two spots.

Source/FakeItEasy/Core/FakeWrapperConfigurator.cs
```diff
@@ -47,6 +47,13 @@ public IFakeOptionsBuilder<T> WithAdditionalAttributes(IEnumerable<CustomAttribu
 
         public IFakeOptionsBuilder<T> Implements(Type interfaceType)
         {
+            Guard.AgainstNull(interfaceType, "interfaceType");
+
+            if (!interfaceType.IsInterface)
+            {
+                throw new ArgumentException("The specified type is not an interface");
```
blairconrad: It's unlikely that a user would be tripped up for long, but if the message indicated what type was supplied, it may be helpful.

blairconrad:
Thanks, @jimmyheaddon. I've taken a quick look, and of course everything looks technically correct. Also, you've done a nice job conforming to our coding standards.
I had a few questions (this is not uncommon for me) which I've put in as line notes.

Source/FakeItEasy.Tests/Creation/DefaultFakeCreatorTests.cs
jimmyheaddon: I went with the convention that existed for the valid instantiations in the test above (CreateFake_should_pass_options_builder_that_returns_itself_for_any_call).  I'm happy to split this out, but it seemed like a decent candidate for parameterising given that there are 3 inputs which should all caused failures now.

Source/FakeItEasy/Core/FakeWrapperConfigurator.cs
jimmyheaddon: Nope I'm just not that familiar with the code base, I'm happy to move that if you think it's better suited there

Source/FakeItEasy/Core/FakeWrapperConfigurator.cs
jimmyheaddon: Yep I can reword that, I literally took the comments from the existing XML exceptions on IFakeOptionsBuilder.Implements(), would you like me to update that also?

According to IFakeOptionsBuilder.Implements<TInterface>(), that implementation should also throw for non-interfaces.  I don't have the code to hand, but do you want the same fix applied to that (if required)?

jimmyheaddon:
Thanks @blairconrad, that's very much appreciated.  I'm not familiar with the code base and thus wanted to follow the existing standards to avoid deviation.  I'll try address your comments tonight.

Source/FakeItEasy.Tests/Creation/DefaultFakeCreatorTests.cs
blairconrad: For `CreateFake_should_pass_options_builder_that_returns_itself_for_any_call`, the `Wrapping.Implements` case was just added because `Wrapping` _doesn't actually return itself_—it's an oddball, but putting another action on the end of `Wrapping` means we do get the original object back again. I picked `Implements` out of a hat for the additional action.
So I think the first test case in `invalidOptionBuilderCalls` could be removed. It's not wrong or harmful, but may be superfluous.

More confusing to me is how the third case is related to `CreateFake_should_throw_when_type_passed_to_options_builder_is_not_an_interface`.
`WithArgumentsForConstructor` seems completely unrelated to passing interfaces. Sure, `() => CreateFoo()` is not a good value, but we wouldn't require an interface there: the problem is that we've been passed something that's not a constructor call.
It's a good test case, but I think it's been lumped into another test with a different purpose. I think the list of test cases with a more generic name or (my preference) completely separate tests with specific names would make more sense.
While I'm on the subject of names, I notice now that the existing test name is probably a little too generic. If we go with specific test names, how do you feel about `CreateFake_should_throw_when_type_passed_to_implements_is_not_an_interface`? It more closely matches the existing `CreateFake_should_pass_all_interfaces_from_implements_to_fake_and_dummy_manager`. Also, `Implements` is the only thing that cares about interfaces.

Source/FakeItEasy/Core/FakeWrapperConfigurator.cs
blairconrad: Oh, I see. I had always took the bit inside the `<exception>` element as a general description of the condition that would cause a condition, not a copy of the message that the exception would have.
I think the XML docs are fine as is, but unless you think there's no value in including the passed-in type in the exception message, perhaps update the message?

> According to `IFakeOptionsBuilder.Implements<TInterface>()`, that implementation should also throw for non-interfaces. I don't have the code to hand, but do you want the same fix applied to that (if required)?

(Note that I edited your comment above to put back ticks around  `IFakeOptionsBuilder.Implements<TInterface>()`. Without them, GitHub/HTML eat the `<TInterface>` and it looks like just IFakeOptionsBuilder.Implements().)
I think you propose to update those docs as well, if the other docs had required an update. That's generous, but I think both can stay as-is.

Source/FakeItEasy/Core/FakeWrapperConfigurator.cs
blairconrad: Yeah, it's not obvious at first glance, but since both classes are internal (or better) and built the way they are to support the fluent syntax, I think it's safe to collapse the checks onto `FakeOptions.AddInterfaceToImplement`, if you don't mind. All else being equal, I always vote for fewer lines of code. Unfortunately at this point, that's more work for you. Then again, if you update the exception message with the supplied type, you'll only have to do it once. :grinning:

blairconrad:
Thank you again, @jimmyheaddon. I've responded to all your comments, I think. Please don't be discouraged. It may seem like I hate your contribution, but it's not true. I just obsess (sometimes wrongly) over little thing. We do appreciate your efforts!

jimmyheaddon:
Thanks @blairconrad, I'll take a look tonight.  I'm new to Git, so I'm curious how do you prefer to work with pull requests, would you accept additional commits and squash them on your end, or would you rather I tore down this request and created a new one?

blairconrad:
Good of you to ask, @jimmyheaddon. Actually, we like "add a new commit to the pull request". Then when everyone's happy, we merge, so you'd see a few of your commits come into the branch.
That way, you get all the credit, and the PR comments stay relevant. If anyone ever comes back to this very conversation, they can see the original commits that we were wrangling over.
Here's an example of another PR with this process in action: https://github.com/FakeItEasy/FakeItEasy/pull/487#commits-pushed-18431d2.

Had you just squashed or something, we wouldn't have complained, though. We're not fanatics. :wink:

Source/FakeItEasy.Tests/Creation/DefaultFakeCreatorTests.cs
jimmyheaddon: Good spot on the third parameterised test, I'm not sure why I pulled that in but I've removed it.  I've renamed the test as suggested, and to reflect this I've changed the name of the test case source from ```invalidOptionBuilderCalls``` to ```invalidImplementsCalls```.

blairconrad: Oh! Yes. Crucial test case source renaming! Nice.

Source/FakeItEasy/Core/FakeWrapperConfigurator.cs
jimmyheaddon: Done - I've consolidated both checks to just ```FakeOptions.AddInterfaceToImplement``` and I've added the ```Type.Name``` to the ArgumentException.

Do you also want me to add a test for the new ```Guard.AgainstNull```?  Essentially it would be:
```
// Arrange, Act, Assert
NullGuardedConstraint.Assert(() => this.creator.CreateFake<IFoo>(x => x.Implements(null)));
NullGuardedConstraint.Assert(() => this.creator.CreateFake<IFoo>(x => x.Wrapping(A.Fake<IFoo>()).Implements(null)));
```

Source/FakeItEasy/Core/FakeWrapperConfigurator.cs
jimmyheaddon: Thanks for that, my Markdown leaves much to be desired!  As mentioned above, I've included the ```Type.Name``` in the exception message now, so for the failing tests it reads *"The specified type 'Foo' is not an interface"*.

jimmyheaddon:
OK I took "add a new commit to the pull request" a little further in that I broke down each change in to separate commits, hopefully these amendments are more in line with your expectations :smile: 

A quick question, rakefile.rb looks for MSBUILD v12.0, unfortunately I have only v14.0 with Visual Studio 2015.  I've been swapping them when building locally, but is your preference that we always build against v12.0?

blairconrad:
The commit granularity looks good Thanks. I'll review the content sometime tonight (I'm in UTC-4 right now).

Yeah, we haven't really gotten the build process on the v14.0 bandwagon yet. It's possible we could adjust it to look for v12.0 and fall back to v14.0, I guess.

In the end, I don't think it much matters what you build with since the TeamCity server will build with v12.0, so it'll catch any incompatibilities and the releases should be all good.

At the Day Job, I'm stuck in VS2010 land (when I'm not supporting an even older version of our product... yeesh), so I don't pay as much attention to the bleeding edge of tooling as I should. Truth be told, the only reason I have VS2013 at home is for FakeItEasy.

Source/FakeItEasy/Creation/FakeOptions.cs
```diff
@@ -29,6 +30,17 @@ public IEnumerable<CustomAttributeBuilder> AdditionalAttributes
...
         {
+            Guard.AgainstNull(interfaceType, "interfaceType");
+
+            if (!interfaceType.IsInterface)
+            {
+                throw new ArgumentException(
+                    string.Format(
+                        CultureInfo.CurrentCulture,
+                        "The specified type '{0}' is not an interface",
+                        interfaceType.Name));
```
blairconrad: I feel terrible for not having thought of this before, but we can probably do better than just `Name`.
Here's a sample failure message when we just use `Name` and try to use `List<Foo>` as an interface (which it's not):

```
System.ArgumentException : The specified type 'List`1' is not an interface
```

However, there's an extension method, `FullNameCSharp`, which will give this output:

```
System.ArgumentException : The specified type 'System.Collections.Generic.List<FakeItEasy.Tests.Foo>' is not an interface
```

How do you feel about that?

Source/FakeItEasy/Creation/FakeOptions.cs
```diff
@@ -29,6 +30,17 @@ public IEnumerable<CustomAttributeBuilder> AdditionalAttributes
...
+            }
```
blairconrad: I feel terrible for not having thought of this before, but we can probably do better than just `Name`.
Here's a sample failure message when we just use `Name` and try to use `List<Foo>` as an interface (which it's not):

```
System.ArgumentException : The specified type 'List`1' is not an interface
```

However, there's an extension method, `FullNameCSharp`, which will give this output:

```
System.ArgumentException : The specified type 'System.Collections.Generic.List<FakeItEasy.Tests.Foo>' is not an interface
```

How do you feel about that?

Source/FakeItEasy.Tests/Creation/DefaultFakeCreatorTests.cs
```diff
@@ -114,6 +120,29 @@ public void CreateFake_should_pass_options_builder_that_returns_itself_for_any_c
...
+            Assert.Throws<ArgumentException>(() => call.Invoke(builderPassedToAction));
+        }
+
+        [Test]
+        public void CreateFake_implements_should_be_null_guarded()
+        {
+            // Arrange, Act, Assert
+            NullGuardedConstraint.Assert(() => this.creator.CreateFake<IFoo>(x => x.Implements(null)));
+            NullGuardedConstraint.Assert(() => this.creator.CreateFake<IFoo>(x => x.Wrapping(A.Fake<IFoo>()).Implements(null)));
+        }
```
blairconrad: @jimmyheaddon, I like the new test, but while I'm complaining about small things (:wink:), this is sort of testing two things (ultimately the same implementation, but we don't know that from where we stand).
Would you consider splitting it in two? Maybe the new test could be `CreateFake_implements_after_wrapping_should_be_null_guarded` or similar?
Otherwise, if the first Assert fails, we'll never know what happens with the second one…

blairconrad:
Thanks, @jimmyheaddon. I've made two minor notes (yes, again).
After that, I think we'll be good, unless one of the other @FakeItEasy/owners chimes in with something.

Source/FakeItEasy/Creation/FakeOptions.cs
jimmyheaddon: I've never come across that before, thanks! :smile:

Source/FakeItEasy.Tests/Creation/DefaultFakeCreatorTests.cs
jimmyheaddon: My bad, fixed!

jimmyheaddon:
@blairconrad, not a problem, all good suggestions!  I've pushed the changes.

Source/FakeItEasy/Creation/FakeOptions.cs
blairconrad: Oh, that's because its ours.

blairconrad:
@jimmyheaddon, it looks good to me. Thanks very much.

@FakeItEasy/owners, I'm content to merge the PR as is. Holler if you have concerns.

jimmyheaddon:
Thanks for that :) I've submitted another PR for reference, not that dissimilar

adamralph:
The change itself looks good to me, but the placement of the tests looks a little odd (our existing tests may be the chief culprit).

The change is to `FakeOptions` but the tests are added to `DefaultFakeCreatorTests` under `FakeItEasy.Tests` which purports to be our 'unit test' project. The two seem loosely related. I'd prefer to omit the tests from `FakeItEasy.Tests` and instead see a black box test in `FakeItEasy.Specs`.

blairconrad:
@adamralph, excellent point, and my fault. The test placement made more sense before I asked for the change in the production code to move. I should've been on top of the ramifications of the new location.
I also prefer specs, but didn't want to completely @jimmyheaddon on his first day in FakeItEasyland.

jimmyheaddon:
I'm happy to refactor them as Spec's using MSpec, or is the intention longer term to use xBehave Given/When/Then style tests, as per #528

blairconrad:
@jimmyheaddon, we're moving toward xBehave eventually, but it's a bigger undertaking (one I'd prefer not to engage in until we release FakeItEasy 2.0, to tell the truth). MSpec is great. Thanks.

jimmyheaddon:
OK so just to double-check, several tests in ```DefaultFakeCreatorTests``` assert on ```FakeOptions/Builder``` behaviours, rather than ```CreateFake``` behaviour.  These tests which assert only on ```FakeOptions/Build``` should be pulled out and placed in a new ```FakeOptions/Builders.Specs``` file/s?

adamralph:
I'm wondering if it might make things easier to hold off on this work until we merge the mspec to xbehave conversion. After that, this branch can be rebased and the black box tests can be written using xbehave. Otherwise, we'll have a nasty merge on our hands.

jimmyheaddon:
@adamralph, if that's an easier path then that's fine with me, I only want to contribute something valuable  and help out.  For reference I've just committed the change to migrate the new tests to ```FakeOptionsSpecs```.

adamralph:
@jimmyheaddon understood, and your efforts are hugely appreciated. All I meant was, that it might not have been worth doing that last commit until after we'd converted from mspec to xbehave since I think it will largely have to be re-written during a rebase. The first two commits will probably just rebase with any conflicts.

jimmyheaddon:
Ah I see, apologies.  I'll leave this as-is for now then, unless there are any other comments?

blairconrad:
As @adamralph said, we do very much appreciate your work, @jimmyheaddon. It's bad luck that you've shown up in the middle of the work to rewrite the specs. I hope we're not turning you off the work.
If you've the patience, we'll get that straightened out and then your last commit can be redone.

While I'm here, though:

1. I like how you picked up `Record.Exception`
2. I wouldn't have created new file for the specs. To me, they would be better in `CreationOptionsSpecs.cs`. I guess that would be `FakeItEasy.Specs/CreationOptions.cs` once the specs have been moved to xbehave tests…

Thanks!

adamralph:
@jimmyheaddon now that we've switched from mspec to xbehave.net, you can go ahead and rebase this  branch onto the latest commit on `master`, see https://github.com/FakeItEasy/FakeItEasy/blob/master/CONTRIBUTING.md#handling-updates-from-upstreammaster.

I believe the only conflict you will have is on the very last commit, which changes the specs. Indeed, the easiest way may be to discard this commit and redo it, but I'll leave that up to you.

blairconrad:
@jimmyheaddon, I don't mean to rush you, and if you are busy and need more time, that's perfectly fine, but I was wondering if you're still pursuing this pull request. We'd love to have your contribution when you are able to replace (or update) the spec commit. On the other hand, if you prefer, I can fold your earlier commits into a superseding PR, provide the new-style specs, and we'll proceed from there. You'd still get credit for your work.

Again, no pressure, and if you'd just like a little more time before returning to this PR, that's great.

jimmyheaddon:
@blairconrad I still intend to wrap this up yes, unfortunately one my girl has been in hospital this week so I've had no time. I'll try again this week now things have calmed down. Cheers!

blairconrad:
My goodness! I hope she is recovered from whatever it was.
Good luck, and please do not let this PR divert you from your family.

jimmyheaddon:
OK so I rebased as Adam suggested, however the TeamCity build fails when compiling FakeOptionsSpecs.cs.  This file has been deleted in my [latest commit](https://github.com/FakeItEasy/FakeItEasy/commit/9eaca9d72efa1f6cba291514b47a7e609bb44de0), any ideas?

blairconrad:
@jimmyheaddon weird. Ah well, I'm okay chalking it up to "one of those things".
I like the new specs. Thanks!

@adamralph, if you're content, I would be happy to see this PR merged.

jimmyheaddon:
@blairconrad - great, I based them on the existing ones to keep them consistent.  I'm around this evening if any changes are required :)

adamralph:
@blairconrad :+1: from me to merge.

If I were to nitpick, I'd ask whether the scenarios `WrappingImplements`, `WrappingImplementsNullType` and `ImplementsNullType` are redundant since `ImplementsNonInterfaceType` already covers what `WrappingImplements` covers and I don't think it's necessary to test guard clauses. I'd probably also ask for the commits to be reformed to a red commit (specs) and a green commit (unit test and impl.). This may be regarded as nothing more than cosmetic commentary. It's fine to merge as-is.

jimmyheaddon:
@adamralph - yep agreed, I simply rolled with the tests that existed before my changes, but in retrospect they're duplicating effort.  I'm happy to remove the tests if you don't believe that code path will deviate to the point where separate tests are required?

I'm a bit of a Git newbie I'm afraid, am I right in saying that you'd like me to back out my current commits, and then add a "red" commit (presumably a commit with specs which fail) and a subsequent "green" commit (the implementation and thus passing specs)?

adamralph:
@jimmyheaddon that's the general idea. I think the extra tests can be safely removed.

None of this *needs* doing, but if you have the will and the energy, I'm more than happy to help.

There are various ways to do it but what you outlined is a nice simple way and probably the way I'd do it. I.e.:

* Reset your branch to the latest `master` commit using **mixed** mode to preserve your changes, e.g.
  * `git checkout myBranch`
  * `git reset upstream/master --mixed`
* Remove the redundant tests
* Commit the specs change
* Commit the implementation and unit test changes
* Force push your branch to your fork

adamralph:
:ok_hand: :sparkles: :lipstick: :heart_eyes: 

It's now a text book PR :smiley:

jimmyheaddon:
@adamralph excellent, the build is running through now should be finished shortly :smiley:

adamralph:
Thanks @jimmyheaddon! Look out for your name in the release notes! :trophy:

Issue #530, opened 2015-08-17
say what the object is when A.CallTo complains "The specified object is not recognized as a fake object.”


blairconrad:
Inspired by @AlexAtNet's question 

> [A.CallTo(…).ReturnsLazily(…) throws “The specified object is not recognized as a fake object.”](http://stackoverflow.com/questions/32052187/a-callto-returnslazily-throws-the-specified-object-is-not-recognized)

over at StackOverflow.

Why don't we say what the object is? It might help people out a bit.

taken by @jimmyheaddon

jimmyheaddon:
@blairconrad are you thinking that we update DefaultFakeManagerAccessor.GetFakeManager() to improve the clarity of exception messages from multiple sources, or did you want this explicitly in RuleBuilder.ReturnsLazily()?

blairconrad:
@jimmyheaddon, not just for `ReturnsLazily`. It was a general comment, just inspired by @AlexAtNet's specific example. Thanks for asking!

blairconrad:
@jimmyheaddon, since you've gone quiet on this issue, I'm going to close it and return it back to the generally available pool. Thanks for your work so far. If you decide you'd like to come back to work on this or anything else, we'd welcome it.

blairconrad:
@rocifier has [indicated interest](https://gitter.im/FakeItEasy/FakeItEasy?at=58c09ee97ceae5376a8ea0ab) in this issue.

adamralph:
Thanks @rocifier!

blairconrad:
This issue was fixed in [Release 3.2.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/3.2.0).

@rocifier, look for your name in the release notes! :trophy:

PR #538, opened 2015-09-17
#530 - Say what the object is when A.CallTo complains "The specified object is not recognized as a fake object.”

jimmyheaddon:
#530 - Modifications to say what the object is when A.CallTo complains "The specified object is not recognized as a fake object.”

Previously both tests expected the same exception message, except that a ```Type.Name``` could not be resolved if the ```proxy.Tag == null```.  Therefore, I modified the logic of ```GetFakeManager``` to throw a separate exception which the first test now catches.  The second test continues to check for the original exception, only with the addition of the ```proxy.Tag.GetType().Name``` in the exception message.

connects to #530

### DIFF
--- a/Source/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
+++ b/Source/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
@@ -35,7 +35,7 @@ namespace FakeItEasy.Core
                     string.Format(
                         CultureInfo.CurrentCulture,
                         "The specified object of type '{0}' is not recognized as a fake object.",
-                        taggable.Tag.GetType().Name));
+                        taggable.Tag.GetType().FullNameCSharp()));
             }
 
             return result;
--- a/Source/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
+++ b/Source/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
@@ -1,8 +1,7 @@
 namespace FakeItEasy.Core
 {
     using System;
-    using System.Collections.Generic;
-    using System.Linq;
+    using System.Globalization;
     using System.Runtime.CompilerServices;
     using FakeItEasy.Creation;
 
@@ -23,11 +22,20 @@ namespace FakeItEasy.Core
 
             var taggable = AsTaggable(proxy);
 
+            if (taggable.Tag == null)
+            {
+                throw new ArgumentException("The specified object does not have a FakeManager attached.");
+            }
+
             var result = taggable.Tag as FakeManager;
 
             if (result == null)
             {
-                throw new ArgumentException("The specified object is not recognized as a fake object.");
+                throw new ArgumentException(
+                    string.Format(
+                        CultureInfo.CurrentCulture,
+                        "The specified object of type '{0}' is not recognized as a fake object.",
+                        taggable.Tag.GetType().Name));
             }
 
             return result;
--- a/Source/FakeItEasy.Tests/Core/DefaultFakeManagerAccessorTests.cs
+++ b/Source/FakeItEasy.Tests/Core/DefaultFakeManagerAccessorTests.cs
@@ -110,7 +110,7 @@ namespace FakeItEasy.Tests.Core
             // Assert
             Assert.That(
                 () => this.accessor.GetFakeManager(proxy),
-                Throws.ArgumentException.With.Message.EqualTo("The specified object of type 'Object' is not recognized as a fake object."));
+                Throws.ArgumentException.With.Message.EqualTo("The specified object of type 'System.Object' is not recognized as a fake object."));
         }
     }
 }
--- a/Source/FakeItEasy.Tests/Core/DefaultFakeManagerAccessorTests.cs
+++ b/Source/FakeItEasy.Tests/Core/DefaultFakeManagerAccessorTests.cs
@@ -95,7 +95,7 @@ namespace FakeItEasy.Tests.Core
             // Assert
             Assert.That(
                 () => this.accessor.GetFakeManager(proxy),
-                Throws.ArgumentException.With.Message.EqualTo("The specified object is not recognized as a fake object."));
+                Throws.ArgumentException.With.Message.EqualTo("The specified object does not have a FakeManager attached."));
         }
 
         [Test]
@@ -110,7 +110,7 @@ namespace FakeItEasy.Tests.Core
             // Assert
             Assert.That(
                 () => this.accessor.GetFakeManager(proxy),
-                Throws.ArgumentException.With.Message.EqualTo("The specified object is not recognized as a fake object."));
+                Throws.ArgumentException.With.Message.EqualTo("The specified object of type 'Object' is not recognized as a fake object."));
         }
     }
 }


blairconrad:
Thanks, @jimmyheaddon.
Just to confuse you, I'll note that, since there was no discussion about your commits, when you added the second one to update the type name that's output (nice touch, by the way), there would've been no confusion for people coming by and reading the PR later, so had you wanted to squash, that would've been completely appropriate.
What you've done is also fine.

blairconrad:
Thanks for submitting this, @jimmyheaddon. I like the direction of the change, but think it has missed the mark just a little.
I think this partly because of the slightly confusing tests that already existed in `DefaultFakeManagerAccessorTests`.
First, let's talk about the user experience, then I'll get into the details about your change and the original tests.
(I worry that I often come off as a jerk in written (and live) communication, so I'm obliged to state that this is not my intent.
I'm just trying to explain how I think your change could be improved, so we can have a discussion and come to a mutually
satisfactory solution. No offence is meant.)

I'm not an F# person, so won't replicate the original StackOverflow question that prompted the issue, but the gist is that some object was passed to `A.CallTo`, and the object was not recognized as either an `Expression<Func<>>` (or action) or a fake object, and the error message we gave was not very specific. The goal is to help the user figure out how they're misusing the API.
Let's make our own example misuse, one that I've seen numerous times on StackOverflow and in chat:

```c#
var notAFake = new Foo();
A.CallTo(() => notAFake.VirtualProperty).Returns("hippo");
```

In released FakeItEasy, here's the outcome:

```
System.ArgumentException : The specified object is not recognized as a fake object.
   at FakeItEasy.Core.DefaultFakeManagerAccessor.GetFakeManager(Object proxy)
   at FakeItEasy.Expressions.ExpressionParser.GetFakeManagerCallIsMadeOn(LambdaExpression fakeObjectCall)
   at FakeItEasy.Configuration.FakeConfigurationManager.CallTo(Expression`1 callSpecification)
   at FakeItEasyQuestions.NotAFake.Should_give_good_error_message_from_outside() in NotAFake.cs: line 13
```

Technically accurate, but not so helpful.

With your PR included, the exception changes to

```
System.ArgumentException : The specified object does not have a FakeManager attached.
```

Again, technically accurate, but not as helpful as it could be. The goal would be to tell the user that they passed in a Foo. Something like

```
System.ArgumentException : The specified object of type 'FakeItEasy.Tests.Foo' is not recognized as a fake object.
```

As a side note, the message about the FakeManager could also be confusing to people as FakeManager isn't really a first-class member of the public API. It's public, but really only because it traditionally has been and there's some quite advanced usage that it supports. I have never used it in any of my tests at the Day Job. I worry that the message would lead the novice user to start looking for a way to attach FakeManagers, and that's not going to be fruitful.

> Previously both tests expected the same exception message, except that a Type.Name could not be resolved if the proxy.Tag == null.

I applaud the effort to provide a full-fidelity solution with a graceful fallback, but in this case it's actually the type of the `proxy` that matters. That's what the user has passed into the `A.CallTo`, so I think (I haven't tried it) it's best to throw the same exception regardless of whether we can't find a tag or if the tag is "just" not a `FakeManager`.

Oh, yeah. Back to the tests that I though got you down the wrong path.

The original `Should_fail_when_getting_manager_from_object_where_a_fake_manager_has_not_been_attached` and `Should_fail_when_getting_manager_from_object_where_a_wrong_typed_fake_manager_has_been_attached` tests fall into the same trap (I think) as the original exception messages: technically correct, but not that helpful. Well, that's perhaps unfair. They test what they say they will. And those are the two conditions for failure (although how "wrong_typed_fake_manager" would ever come up is beyond me). It's just that they don't do a great job of what the effect is for the user. Which is fine, since they're unit tests, concerned with the messy internals of the framework. But it does obscure a little what's going on.

I think the best path forward would be to collapse the cases you added to the production code and issue the same exception, based on the type of the `proxy`.
Of course the unit tests would need to be adjusted accordingly.

There's another thing we could do as well. (I know, more work for you. Yay!)
We could add a couple of acceptance tests ("specs"). These are black box tests written "from the outside", that test the system end-to-end. They mimic features as the user would see them, not unlike the code snippet I included above. We've been moving more and more toward writing and updating specs when adding or changing features, and I've been quite happy with the result. In some cases, I've not bothered with unit tests at all, or I've even added specs to cover the unit tests, then a spec for the new/changed functionality, and then implemented the change.

In addition the specs are "refactor-resistant", tend to represent the user experience a little better, and even can work as examples for users, if we want. (`var proxy = A.Fake<ITaggable>(); proxy.Tag = new object();` is an acceptable test, but a terrible use case :wink:)

If  you're up for it, I'd consider adding two specs, probably to `ConfigurationSpecs.cs`, showing how a `A.CallTo` can fail when passed

1. a lambda that makes a call on a non-fake as in `A.CallTo<T>(Expression<Func<T>> callSpecification)` (or `A.CallTo(Expression<Action> callSpecification)`), and
2. a non-fake object as in `A.CallTo(object fake)`

So. A lot to take in. Sorry if I'm overwhelming you. I'm happy to chat about any and all of this at your convenience (and mine—I'm leaving soon for the day). As we discussed elsewhere, if you want live chat, Jabbr or Gitter are fine (although I'm more likely to see the notification from Gitter). Or we can muddle through using "highly asynchronous chat" in the comments here.
Of course, maybe you disagree with any or all of my points. If so, let's chat about that!

Oh. And if you prefer, once we settle on a path forward, we could consider this whole PR an "exploratory session" and close it before proceeding with a fresh one. But perhaps that's a discussion for later.

Thanks. We do appreciate the effort you're putting into this!

Source/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
```diff
@@ -23,11 +22,20 @@ public FakeManager GetFakeManager(object proxy)
 
             var taggable = AsTaggable(proxy);
 
+            if (taggable.Tag == null)
+            {
+                throw new ArgumentException("The specified object does not have a FakeManager attached.");
```
adamralph: I'm wondering how much this adds. Is this the message that will be surfaced to the end user?

My concern here is that `taggable.Tag == null` and "The specified object does not have a FakeManager attached." do not have any intuitive connection, and I'm wondering whether the message will mean anything at all to the end user.

blairconrad: Agreed, @adamralph. Check out my [screed in the conversation tab](https://github.com/FakeItEasy/FakeItEasy/pull/538#issuecomment-141657679) for this and other concerns!

jimmyheaddon:
@blairconrad @adamralph do you guys want me to pick this up again?

blairconrad:
@jimmyheaddon, I'd never discourage you from doing work. If you're so inclined, please do.

blairconrad:
Hey, @jimmyheaddon, are you still interested in working on this PR? We'd be very happy to see what you come up with.

blairconrad:
@jimmyheaddon, since you've gone quiet on this PR, I'm going to close it and return the issue back to the generally available pool. Thanks for your work so far. If you decide you'd like to come back to work on this or anything else, we'd welcome it.

PR #542, opened 2015-09-23
swap mspec for xbehave

adamralph:
closes #458 
closes #528

PR #543, opened 2015-09-24
swap mspec for xbehave and run the specs

### LINKED ISSUE #458: Switch from MSpec to xbehave

blairconrad:
closes #458 
closes #528
closes #542

### DIFF
--- a/Source/FakeItEasy.Specs/AssertingOnSetOnlyPropertiesSpecs.cs
+++ b/Source/FakeItEasy.Specs/AssertingOnSetOnlyPropertiesSpecs.cs
@@ -5,6 +5,15 @@
 
     public class AssertingOnSetOnlyPropertiesSpecs
     {
+        public interface ISetOnly
+        {
+            [SuppressMessage("Microsoft.Design", "CA1044:PropertiesShouldNotBeWriteOnly", Justification = "Required for testing.")]
+            int MyProperty { set; }
+
+            [SuppressMessage("Microsoft.Design", "CA1044:PropertiesShouldNotBeWriteOnly", Justification = "Required for testing.")]
+            bool MyProperty2 { set; }
+        }
+
         [Scenario]
         public void SetOnlyProperties(
             ISetOnly setOnly)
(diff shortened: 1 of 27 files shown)


blairconrad:
Bah. That didn't go so well. Not sure why.

adamralph:
I'd like to propose a slightly different take on this. The first commit, given that it didn't include the switching of the running of the tests from mspec to xbehave is actually invalid. I'd rather amend it so that it's valid in isolation, with the following commits actually just being refactoring. In that case I'd force push, updating #542

blairconrad:
quoth @adamralph:

>  I'd rather amend it so that it's valid in isolation, with the following commits actually just being refactoring. In that case I'd force push, updating #542

Love it. I would be thrilled to have this PR go away. Not only because it doesn't work.
Does this mean you want to do the squashing and whatnot? I'm not about who does it.
If it helps, grab my rakefile. Maybe it needs concurrency set (and a log path?).

adamralph:
I'm on it. Expect an update soon. BTW - the build failed because the xunit console runner isn't installed as a package.

blairconrad:
Oh? That's... odd. Make sense. But why do I have it at home !?

Jut as well, I was stopping by to make sure we'd removed MSpec from the packages.config.

adamralph:
Are you 100% sure you ran `rake` at home? With a runner inside VS it would just work, but it would definitely need xunit.runner.console installed in the solution for rake to work.

blairconrad:
#542 sorted itself out. Abandoning PR.

blairconrad:
Completely rakey.
Maybe a while back I added the packages as an experiment and they're hanging out now?
I'll check my packages dir when I get home. Sometime.

blairconrad:
I am not a packages cleaner-upper:

image

I think at least one of those Castle.Cores didn't even actually exist. At least nowhere but on my machine.

adamralph:
Aha! Mystery solved. :smile:

The occasional `git clean -fxd` is a good thing. :wink:

PR #546, opened 2015-09-25
Cleanup creation option specs

blairconrad:
So @jimmyheaddon doesn't have to when he rebases #462.

@adamralph, how's this sort of thing grab you?
I'm open to, y'know, whatever.


adamralph:
LGTM!

blairconrad:
Indeed. Let's Get This Money!

(That's the first Google result. Then I kept reading. I'm all caught up now.)

Thanks, Champ!

Issue #22, opened 2012-09-10
Fix Resharper settings to remove redundant hints & warnings


philippdolder:

jimmyheaddon:
Having performed this configuration many times in the past, I'd highlight that ReSharper's concept of configuration layers means we'll struggle to create a lightweight configuration for the solution which adequately overrides all users machine level settings, not without some decent effort and testing.

It's not insurmountable by any means, but it's never easy to get the configuration spot-on when working outside of a single team environment.

jimmyheaddon:
This is a much wider point on static analysis, but I suggest we split this issue up:

**StyleCop**
1. Update the StyleCop.MSBuild NuGet package and fix the outstanding StyleCop violations
2. Review the use of StyleCop suppressions in code (if applicable)
3. Discuss the enforcement of StyleCop (using 'Treat warnings as errors' in .csproj files)

**ReSharper**
1. Create the FakeItEasy.sln.DotSettings file, and mirror Settings.StyleCop only
2. Review the use of ReSharper suppressions in code
3. Amend the FakeItEasy.sln.DotSettings as per comments in #36

**Code Analysis**
1. Review the use of Code Analysis suppressions in code
2. Review the 3 rulesets in the solution and the severity of rules
3. Discuss the application of FakeItEasy.ruleset to projects (FakeItEasy.csproj is configured to use this, but Code Analysis is disabled for all build configurations)
4. Discuss the enforcement of Code Analysis in builds
5. Discuss the integration of Visual Studio 2015 analysis rules (changes with Roslyn)

**Misc Analysis**
1. Review outstanding build warnings - whether they should be addressed and whether 'Warnings as Errors' should be enabled
2. Review the various Roslyn analyser results

adamralph:
@jimmyheaddon thanks for chiming in. As you suggest we should split these up. This issue can be the "Resharper" part of your comment. The other bits can be other issues/PR's.

adamralph:
Since dumping Resharper from my toolkit for the second time, I'm once again shying away from this. The ubiquitous static analysis solution is promising to be Roslyn based analysers and I think we should lean on them for code consistency using whatever mechanisms they provide for rule customisation.

Resharper is a personal tooling choice and if it doesn't respect Roslyn analyser settings then it should be made to do so, using whatever extensions, etc. require it.

blairconrad:
@adamralph, I'm not sure I understand what you're saying. I think you're coming down on closing this...
does

[...]

Issue #262, opened 2014-02-24
Switch StyleCop violations to warnings for release builds


adamralph:
This is a proposal.

Judging from my twitter stream, StyleCop is either loved or hated.

In the interest of encouraging contribution or rather avoiding discouraging contribution, I propose that we allow PR's to be merged to master whilst containing StyleCop violations. I.e. we keep StyleCop analysis but we don't fail the build when violations are present by switching the output to warnings instead of errors.

@FakeItEasy/owners will still have the opportunity to fix up any violations after a given PR is merged and we can also expand the first point in our release notes to include StyleCop violations as well as code analysis violations.

Thoughts?

PR #549, opened 2015-10-07
StyleCop configuration

jimmyheaddon:
This seemed like a quick win, as I can only do small bits during my breaks.

As discussed in #22, I've reviewed the StyleCop suppression's, updated the NuGet packages for all projects, disabled the analysis of Designer files, and have enabled 'Treat Warnings as Errors' in the Release build configuration for all projects (in my experience using this in Debug configuration can be a little restricting when you're just playing with some throw-away code or an idea for a feature).

I can squash if wanted, I figured I'd submit individual commits as you may want one change and not another.  Everything should be correctly rebased this time! :laughing:

### DIFF
--- a/Source/FakeItEasy.Examples.VB/FakeItEasy.Examples.VB.vbproj
+++ b/Source/FakeItEasy.Examples.VB/FakeItEasy.Examples.VB.vbproj
@@ -48,6 +48,7 @@
     <AdditionalParameters />
     <CodeAnalysisRuleSet>..\FakeItEasy.Examples.ruleset</CodeAnalysisRuleSet>
     <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
+    <StyleCopTreatErrorsAsWarnings>true</StyleCopTreatErrorsAsWarnings>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
@@ -61,6 +62,7 @@
     <AdditionalParameters />
     <CodeAnalysisRuleSet>..\FakeItEasy.Examples.ruleset</CodeAnalysisRuleSet>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <StyleCopTreatErrorsAsWarnings>false</StyleCopTreatErrorsAsWarnings>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
--- a/Source/FakeItEasy.Examples.VB/FakeItEasy.Examples.VB.vbproj
+++ b/Source/FakeItEasy.Examples.VB/FakeItEasy.Examples.VB.vbproj
@@ -34,6 +34,8 @@
     <BootstrapperEnabled>true</BootstrapperEnabled>
     <TargetFrameworkProfile />
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\..\Source\</SolutionDir>
+    <NuGetPackageImportStamp>
+    </NuGetPackageImportStamp>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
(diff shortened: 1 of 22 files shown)


blairconrad:
Thanks for the work, @jimmyheaddon. This may not be incredibly straightforward.
I had thought we were planning on discussing some changes. In the past, we've wrangled over whether to have stylecop errors break the build and whatnot, and it's not clear to me that we want to deviate from what we've been using.
I'm really  of two minds. I enjoy the strictness, but do not want to turn any potential contributor off because it takes them 3 attempts to even get a file to compile after they've made a small change.
Then again, I may be jumping the gun in commenting before poring over the changes. I will take a peek as soon as I'm able.

blairconrad:
Ah. Found #262. Not one year ago we decided to treat StyleCop violations as warnings in Release builds.
I'm not saying that this is carved in stone. I'm always happy to have a chat.

blairconrad:
I like how the commits are broken up, BTW!

jimmyheaddon:
@blairconrad I completely understand, hence the multiple commits if you wish to cherry pick certain items and not users.

Personally I'm all for treating warnings as errors for StyleCop, they're generally easier to fix when compared to something like Code Analysis or most ReSharper violations.  I figured I'd made the change and you can cherry pick them as you see fit, I don't want to put any contributors off.

blairconrad:
I hear you with the draconian ways. Part of me wants to say "lock those suckers down". But I think that may be due to my environment at the Day Job. In the past, some people have been lax. At least at FakeItEasy International, the owners are… unlax. So even if we give the community an easier time, someone will swing by and clean up the warnings soonish.

I'm a little concerned about looser settings for debug than release, because I worry that people will change code, build debug, run unit tests in the IDE, and then submit PRs, which will not build. For that reason, I lean toward having debug and release configs the same. Of course, then we have to consider whether it's worthwhile forcing someone who's quickly hacking to use the proper spacing. I do worry about turning people off with that.
Has your experience with "warnings in debug, error in release" come from open source projects, or only in paid environments where you can stand to abuse the contributors a little more?

jimmyheaddon:
I've generally always had these static analysis tools enabled for all build configurations, it's never really been an issue as you pick things up whilst you're designing them, rather than refactoring before a commit.  The warnings in debug, error in release is something I've heard others mention and it seemed a reasonable solution with regards to not putting off contributors, but then I'm guessing the TeamCity build runs in release mode, and thus will fail unexpectedly?

blairconrad:
>  I'm guessing the TeamCity build runs in release mode, and thus will fail unexpectedly?

Exactly. That's what I was trying to explain with my "people will change code, build debug, run unit tests in the IDE, and then submit PRs, which will not build". You did a much better job.
My reasoning for favouring consistent release/debug settings was so that there'd be either no surprise or a surprise right away (when a second space breaks the build) rather than a surprise in 10 minutes when TeamCity is angry. Perhaps I'm not thinking it through properly.

adamralph:
I believe we should leave our StyleCop warnings as errors policy unchanged.

Treating warnings as errors is fine for some projects. E.g. in my previous job, all our code was private with no outside contributors and we treated warnings as errors universally because we all agreed to.

For OSS we need to balance such policies with the encouragement of contribution and, in my experience, treating warnings as errors does put off contributors. Switching this on in release mode only doesn't work in practice. I've tried this before and it very regularly ended up in red PR's since most contributors only build in debug before submitting and this is also something we should be lenient to in encouraging contribution. Given that StyleCop violations are so easy to fix, it's better that we, the maintainers, shoulder the burden of fixing them later when required.

In summary:

1. https://github.com/jimmyheaddon/FakeItEasy/commit/b5efe2235a7f2984049aab9121f237f68ad72cfc :+1: 
1. https://github.com/jimmyheaddon/FakeItEasy/commit/15650028f529b730e56b499f449bcc4084453cd3 :-1: 
1. https://github.com/jimmyheaddon/FakeItEasy/commit/41b80b439b9dac6bf111b9d578292d5de94fa00b :question: the violation fix and suppression removal is fine but the reformatting of the remarks element is under question. Pragmatically, I would advise limiting this commit to the violation fix and suppression removal only and worry about the rest separately. It can even be another PR.
1. https://github.com/jimmyheaddon/FakeItEasy/commit/ce55d28a0ec221596f7f059819ea33833ff72c2f :question: is this required? or YAGNI?

blairconrad:
> Do we have any designer files that were causing violations?

I didn't see any suppressions being removed, so I don't think so. @jimmyheaddon, is this just your standard procedure?

jimmyheaddon:
I didn't encounter any designer file violations, and given the lack of their use nowadays it's considered best practice to ignore them explicitly rather than littering your code with suppressions, ReSharper has the same concept.

As I said above, I'm happy if you're not interested in taking all of the commits, that was the intention.  I would say that in my experience StyleCop errors are easy to fix and violations won't get past a PR, therefore it's more work fixing them after PR comments than being informed at compile time, but that's just me.  The same doesn't apply for ReSharper and Code Analysis, which are generally more difficult and contentious.

(jimmyheaddon@41b80b4)[https://github.com/jimmyheaddon/FakeItEasy/commit/b5efe2235a7f2984049aab9121f237f68ad72cfc] I might be missing something, but I don't believe we can remove the suppression without changing the remarks element.  Therefore, it's just the "re-thrown" change?

adamralph:
Ah, @jimmyheaddon apologies, you're correct. I thought that "re-thrown" was the only violation but of course, it's the URL too.

I've just noticed that this is a private method. This makes the XML docs redundant since it's impossible to ever see them. I say we just convert them to regular comments and be done with it.

adamralph:
@jimmyheaddon btw - for links to commits, you don't need to formulate the text yourself, you just paste the link in and GitHub does it for you. (You'd actually linked to the wrong commit, but I fixed it.)

jimmyheaddon:
@adamralph - Doh! :laughing:

I've some time tonight, how would you like to proceed with the 4 commits, so I'm clear before making any changes.

adamralph:
@jimmyheaddon my vote would be to keep https://github.com/jimmyheaddon/FakeItEasy/commit/b5efe2235a7f2984049aab9121f237f68ad72cfc and then add another commit which convert the XML docs to regular comment lines (note - many lines each with `//`, rather than a multi-line `/*...*/` comment). The other commits can be discarded.

You can then force push to the same branch and this PR will be updated.

blairconrad:
I agree with @adamralph.

blairconrad:
@jimmyheaddon are you still interested in pursuing this PR? If so, I think a quick rebase against master and proceeding as outlined in https://github.com/FakeItEasy/FakeItEasy/pull/549#issuecomment-146473248 should do the trick. Thanks!

adamralph:
Replaced by #570, which preserves @jimmyheaddon's first commit from here.

PR #553, opened 2015-10-18
refactor: clean up rulesets

adamralph:


blairconrad:
Very nice.

blairconrad:
Did something bite you, or were you just trolling rulesets?

adamralph:
When I saw the staticking of test classes in #552 I was wondering if we still needed the 'make stuff static' rule switched off so I went and had a look, and whilst I was there I emptied out the rulesets and reconstructed them based on what was still required.

Issue #534, opened 2015-09-08
Stacking multiple results on the same call configuration instance doesn't work


thomaslevesque:
Sorry if the title is not clear, here's an example to illustrate the problem.

I expected the two following snippets to be equivalent:

```c#
// Snippet 1
A.CallTo(() => foo.Get()).Returns(42);
A.CallTo(() => foo.Get()).Returns(100).Twice();

// Snippet 2
var call = A.CallTo(() => foo.Get());
call.Returns(42);
call.Returns(100).Twice();
```

But apparently they're not. With the first snippet, successive calls to `foo.Get()` return 100, 100, 42, 42, 42, 42 ... (as expected). But with the second snippet, successive calls to `foo.Get()` return 100, 100, 0, 0, 0, 0... The first `Returns()` is ignored; the second `Returns()` apparently overwrites the first, rather than stacking upon it.

Is this the expected behavior? If it is, it's quite confusing, and goes against the principle of least surprise. I sometimes store the result of `A.CallTo()` to avoid repeating it multiple times (usually because I want to configure the call AND verify that it happened), and I certainly didn't expect this behavior.

blairconrad:
@thomaslevesque I would've had the same expectation that you did!

thomaslevesque:
Looking at the code, I understand why it works that way. Each `A.CallTo(...)` creates a new rule for the fake, and each rule has only one return value configuration. So, specifying the return value again just overwrites the previous configuration. It kind of makes sense, but it's not very intuitive...

Perhaps it would be possible to extend the API to allow something like this:

    A.CallTo(() => foo.Get())
        .Returns(100).Twice()
        .Then.Returns(42);

(note that the order is reversed to match the "natural" order, rather than the stacking order)

`Then` would actually add a new rule for the fake, based on the same call specification.

blairconrad:
@thomaslevesque, It may be possible. I think it would be a little nicer to adjust things to support your original example, though, if that didn't break everything else…

thomaslevesque:
> I think it would be a little nicer to adjust things to support your original example, though, if that didn't break everything else…

I agree, but I suspect it would be a pretty big undertaking, with high risk of regression... I'm not sure it's worth it.

blairconrad:
I've thought a little more about this.
I'm not sure that the behaviour is actually that bad. It's maybe a bit surprising, but is defensible.
In your original example:

```c#
A.CallTo(() => foo.Get()).Returns(42);
[...]

PR #557, opened 2015-10-23
Prevent redefining the behavior for a call rule

thomaslevesque:
This is in order to prevent an unsupported scenario where the user keeps
the result of A.CallTo(...) and uses it to define multiple behaviors.
Doing so would override the existing behavior, not stack the new one,
even though it's not necessarily what the user would expect. Making it
illegal prevents confusion. It remains possible to keep the result of
A.CallTo to make later calls to MustHaveHappened and friends.

See the comments in FakeItEasy/FakeItEasy#534 for more details.

### DIFF
--- a/Source/FakeItEasy.Tests/Expressions/ExpressionCallRuleTests.cs
+++ b/Source/FakeItEasy.Tests/Expressions/ExpressionCallRuleTests.cs
@@ -100,7 +100,7 @@
 
         private ExpressionCallRule CreateRule()
         {
-            return new ExpressionCallRule(this.callMatcher) { Applicator = x => { } };
+            return new ExpressionCallRule(this.callMatcher) { };
         }
     }
 }
--- a/Source/FakeItEasy/Configuration/BuildableCallRule.cs
+++ b/Source/FakeItEasy/Configuration/BuildableCallRule.cs
@@ -12,11 +12,17 @@ namespace FakeItEasy.Configuration
         : IFakeObjectCallRule
     {
         private readonly List<Tuple<Func<IFakeObjectCall, bool>, Action<IOutputWriter>>> wherePredicates;
+        private Action<IInterceptedFakeObjectCall> applicator;
+        private bool canSetApplicator;
+        private Func<IFakeObjectCall, ICollection<object>> outAndRefParametersValueProducer;
+        private bool canSetOutAndRefParametersValueProducer;
 
         protected BuildableCallRule()
         {
             this.Actions = new LinkedList<Action<IFakeObjectCall>>();
-            this.Applicator = call => call.SetReturnValue(call.Method.ReturnType.GetDefaultValue());
+            this.applicator = call => call.SetReturnValue(call.Method.ReturnType.GetDefaultValue());
+            this.canSetApplicator = true;
+            this.canSetOutAndRefParametersValueProducer = true;
             this.wherePredicates = new List<Tuple<Func<IFakeObjectCall, bool>, Action<IOutputWriter>>>();
         }
 
@@ -24,7 +30,24 @@ namespace FakeItEasy.Configuration
         /// Gets or sets an action that is called by the Apply method to apply this
         /// rule to a fake object call.
         /// </summary>
-        public Action<IInterceptedFakeObjectCall> Applicator { get; set; }
+        public Action<IInterceptedFakeObjectCall> Applicator
+        {
+            get
+            {
+                return this.applicator;
+            }
+
+            set
+            {
+                if (!this.canSetApplicator)
+                {
+                    throw new InvalidOperationException("The behavior for this call has already been defined");
+                }
+
+                this.applicator = value;
+                this.canSetApplicator = false;
+            }
+        }
 
         /// <summary>
         /// Gets a collection of actions that should be invoked when the configured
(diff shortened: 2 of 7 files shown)


Source/FakeItEasy.Tests/Expressions/ExpressionCallRuleTests.cs
```diff
@@ -100,7 +100,7 @@ public void UsePredicateToValidateArguments_should_pass_predicate_to_call_matche
 
         private ExpressionCallRule CreateRule()
         {
-            return new ExpressionCallRule(this.callMatcher) { Applicator = x => { } };
```
thomaslevesque: This didn't seem necessary (the tests still pass without it), and it made  the tests fail with the new behavior.

Source/FakeItEasy/Configuration/RecordingRuleBuilder.cs
```diff
@@ -14,8 +14,6 @@ public RecordingRuleBuilder(RecordedCallRule rule, RuleBuilder wrappedBuilder)
         {
             this.rule = rule;
             this.wrappedBuilder = wrappedBuilder;
-
-            rule.Applicator = x => { };
```
thomaslevesque: Same reason as in ExpressionCallRuleTests

Source/FakeItEasy/Configuration/BuildableCallRule.cs
```diff
@@ -12,19 +12,32 @@ internal abstract class BuildableCallRule
...
         /// Gets or sets an action that is called by the Apply method to apply this
         /// rule to a fake object call.
         /// </summary>
-        public Action<IInterceptedFakeObjectCall> Applicator { get; set; }
+        public Action<IInterceptedFakeObjectCall> Applicator
+        {
+            get { return this.applicator; }
+            set
+            {
+                if (!this.canSetApplicator) throw new InvalidOperationException("The behavior for this rule is already defined");
```
thomaslevesque: Is the exception message clear enough?

Source/FakeItEasy/Configuration/RecordingRuleBuilder.cs
blairconrad: But it worried me more, since the previous line was in test code, and this one is in production. Nevertheless, I can't see any harm in removing this. I tried to break the `NextCall.To` mechanism, but couldn't.

blairconrad:
The implementation seems generally sound for what it does.
I would've probably added specs so it would be obvious what external behaviour we're testing (the `Applicator` setting is all well and good, but even to experienced insiders (if I can consider myself such) it doesn't scream "Returns" or "Throws").
In addition, a number of StyleCop warnings have been introduced, which aren't huge deals, but if you keep an eye out in round two, I would consider it a favour.

My main concern is that this change addresses multiple `Returns`, `Throws`, and `CallsBaseMethods`, but nothing else, and would leave us with inconsistent behaviour, not that we don't have that now.
I think what I've discovered is actually best discussed in #534. Meet me back there?

blairconrad:
… meet me back there in a bit, actually. I've been pulled away.

Source/FakeItEasy/Configuration/BuildableCallRule.cs
blairconrad: I'd say probably not.
End users won't know what a "rule" is. Of course the stack trace should give a good hint at what went wrong, but I think it's preferable to be as specific as possible while using the end user API's language. And I'd be inclined to help guide the user to better behaviour. They're likely trying to set up behaviour that varies over time, and we could push them toward multiple calls to `A.CallTo` (or, unfortunately, `Fake.CallsTo`, so wording could become tricky).

blairconrad:
The new specs look good, I think. We've not yet sorted the BDD syntax within our specs since we've converted from MSpec, but you've matched the rest of the file, which is A-1 in my books.
I'd consider using the `Record.Exception` instead of the try/catch block—it matches the rest of the code a little bit better.
The real advantages are that it will catch all exceptions so they can be examined in the "it should" step, instead of having them bubble up out through the "when" and the code in the "it should" will more closely match the "it should" text.

Perhaps something like (warning: I am not a compiler)

```csharp
"when configuring a return then a throw on the same configuration"
    .x(() =>
    {
        configuration = A.CallTo(() => fake.Baz());
        configuration.Throws(new ArgumentNullException());
        exception = Record.Exception(() => configuration.Throws(new ArgumentException()));
    });

"it should throw an InvalidOperationException"
    .x(() => exception.Should().BeAnExceptionOfType<InvalidOperationException>());
```

of course in this example `exception` needs to become an `Exception`.

thomaslevesque:
Yes, it looks better with `Record.Exception`. I will change the tests.

blairconrad:
The new stuff looks pretty good, I think.
The only odd bit to get my head around was `SetOutAndRefParametersValueProducer`. Given the way we build the implicit out values in the expression matcher, though, I don't see a better way.
I would consider, though, adding a  ///-comment to the method to indicate its purpose (initial configuration of the call, intended to bypass the multiple-SetOutAndRefParametersValueProducer-setting logic, use sparingly) to help future readers along.

Other than that, I am content.
We've introduced a fair bit of churn with our discussion here, so if you're inclined, when (if) you add the comment I mentioned above, perhaps squashing the commits down to two would be appropriate, one for adding the specs, and one for everything else? Again, as an aid to future readers.

Then I think we can just give the other owners a chance to evaluate and we should be go.

Thanks for your work, @thomaslevesque!

thomaslevesque:
> I would consider, though, adding a ///-comment to the method to indicate its purpose

Good point, it's not obvious why this method exists... I'll add a comment



> We've introduced a fair bit of churn with our discussion here, so if you're inclined, when (if) you add the comment I mentioned above, perhaps squashing the commits down to two would be appropriate, one for adding the specs, and one for everything else?

Sure, I intended to squash the commits before merge anyway. I'll make a clean split between the specs and the implementation.

blairconrad:
Love it. Thanks!

thomaslevesque:
I made the requested changes. Just waiting for feedback from other @FakeItEasy/owners now ;)

blairconrad:
While we're waiting anyhow, typically the specs would come first in BDD (as unit tests do in TDD, although I less often see separate commits for adding tests): add the specs to show the gap in the product, then add implementation and the specs turn green.
I know. So picky!

blairconrad:
:shipit:

blairconrad:
@thomaslevesque, don't despair. We're still interested in this change. Other owners are just experiencing higher than usual outside activities. We appreciate your work and patience.

thomaslevesque:
@blairconrad, no problem, I'm not in a hurry ;)

thomaslevesque:
Closing this, since it's superseded by #569

PR #561, opened 2015-11-29
Exclude dynamic assemblies from Bootstrapper Locator scan

hmemcpy:
In my hacks around Sledgehammer, I am dynamically creating objects using [Impromptu-interface](https://github.com/ekonbenefits/impromptu-interface), and using reflected FakeItEasy types to construct these object. This causes the FakeItEasy assembly to be added as a reference to the dynamic assembly, that is generated with Impromptu-interface.

This causes FIE to throw an exception during the Bootstrapper Locator scan: attempting to call `assembly.GetExportedTypes()` on a dynamic assembly throws a `NotSupportedException`. Since my generated dynamic assembly does indeed include a reference to FIE, it is picked up for scanning.

This PR adds a Where, excluding Dynamic assemblies from the scan.

### DIFF
--- a/Source/FakeItEasy/Core/AssemblyExtensions.cs
+++ b/Source/FakeItEasy/Core/AssemblyExtensions.cs
@@ -5,8 +5,6 @@
 
     internal static class AssemblyExtensions
     {
-        private static readonly string FakeItEasyAssemblyName = Assembly.GetExecutingAssembly().FullName;
-
         /// <summary>
         /// Determines whether an assembly references FakeItEasy.
         /// </summary>
@@ -16,7 +14,7 @@
         {
             Guard.AgainstNull(assembly, "assembly");
 
-            return assembly.GetReferencedAssemblies().Any(r => r.FullName == FakeItEasyAssemblyName);
+            return assembly.GetReferencedAssemblies().Any(r => r.FullName == TypeCatalogue.FakeItEasyAssembly.FullName);
         }
 
         /// <summary>
@@ -27,6 +25,21 @@
         public static string Name(this Assembly assembly)
         {
             return new AssemblyName(assembly.FullName).Name;
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether a given assembly was generated dynamically by using reflection emit.
+        /// </summary>
+        /// <remarks>This extension works in .NET 3.5, where <c>Type.IsDynamic</c> does not exist.</remarks>
+        /// <param name="assembly">The assembly to check.</param>
+        /// <returns>Whether or not the assembly is dynamically generated.</returns>
+        public static bool IsDynamic(this Assembly assembly)
+        {
+#if NET40
+            return assembly.IsDynamic;
+#else
+            return assembly.ManifestModule.GetType().Namespace == "System.Reflection.Emit";
+#endif
         }
     }
 }
--- a/Source/FakeItEasy/Core/BootstrapperLocator.cs
+++ b/Source/FakeItEasy/Core/BootstrapperLocator.cs
@@ -25,7 +25,9 @@
             var bootstrapperInterface = typeof(IBootstrapper);
 
             var appDomainAssemblies = AppDomain.CurrentDomain.GetAssemblies();
-            var appDomainAssembliesReferencingFakeItEasy = appDomainAssemblies.Where(assembly => assembly.ReferencesFakeItEasy());
+            var appDomainAssembliesReferencingFakeItEasy = appDomainAssemblies
+                .Where(assembly => !assembly.IsDynamic())
+                .Where(assembly => assembly.ReferencesFakeItEasy());
 
             var candidateTypes = appDomainAssembliesReferencingFakeItEasy
                 .SelectMany(assembly => assembly.GetExportedTypes())
--- a/Source/FakeItEasy/Core/TypeCatalogue.cs
+++ b/Source/FakeItEasy/Core/TypeCatalogue.cs
@@ -16,8 +16,16 @@
     /// </summary>
     internal class TypeCatalogue : ITypeCatalogue
     {
-        private static readonly Assembly FakeItEasyAssembly = Assembly.GetExecutingAssembly();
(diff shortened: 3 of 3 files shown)


Source/FakeItEasy/Core/BootstrapperLocator.cs
```diff
@@ -25,7 +25,9 @@ public static IBootstrapper FindBootstrapper()
             var bootstrapperInterface = typeof(IBootstrapper);
 
             var appDomainAssemblies = AppDomain.CurrentDomain.GetAssemblies();
-            var appDomainAssembliesReferencingFakeItEasy = appDomainAssemblies.Where(assembly => assembly.ReferencesFakeItEasy());
+            var appDomainAssembliesReferencingFakeItEasy = appDomainAssemblies
+                .Where(assembly => !assembly.IsDynamic)
```
blairconrad: no Assembly.IsDynamic in .NET 3.5.
Check out https://github.com/FakeItEasy/FakeItEasy/blob/bd57a05c0de5ff37609230fd1b8bc4ec45bc9d4d/Source/FakeItEasy/Core/TypeCatalogue.cs#L64

blairconrad:
Generally looks good, @hmemcpy. At first I thought "changed tests?", but the Bootstrapper scanner's "stop at first found Bootstrapper" logic makes this harder to add than I think is worthwhile, so I'm okay without a new test.
I'd ask for two things, though, if you'd humour me.

1. squash the commits? I think adding the XML doc should've been part of 4e2e804572a6a4ddf71f69d5e9a65fc2ad85a495, and the initial build-breaking commit is nobody's friend.
2. have `TypeCatalogue.GetAllAssemblies` use the new mechanism? No sense in duplicating the logic used for finding (avoiding) dynamic assemblies

Of course, I _could_ always submit a superseding PR with these changes, but I'd rather try to convince you it's a good idea!

hmemcpy:
@blairconrad I squashed the commits. Good point.

The code in `TypeCatalogue.GetAllAssemblies` does slightly different things, it's possible to refactor, but I'd rather this be a different PR.

blairconrad:
Thanks for the squash.

> The code in TypeCatalogue.GetAllAssemblies does slightly different things

I'm not following. The context is different, but the check for dynamic assembly appears to be the same. I'm sure I'm missing something, but I'm missing what I'm missing.

> it's possible to refactor, but I'd rather this be a different PR.

I'd think separate commit in the same PR would be adequate separation, but that's possibly because I'm still missing something…

blairconrad:
>  it's a bit difficult to write a deterministic test for, unfortunately.

The only thing that comes to mind is a separate test assembly that creates its own bootstrapper in a dynamic assembly. Then with that and the DefaultBootstrappers the only available, if we don't throw an exception, we know that the dynamic assembly was avoided. It's a little weak, but not super difficult. Just a HUGE pain for the benefit.

blairconrad:
@hmemcpy, a while back you said 

> The code in `TypeCatalogue.GetAllAssemblies` does slightly different things, it's possible to refactor, but I'd rather this be a different PR.

I still don't understand that. Could you explain how it's doing different things? We're using `a.ManifestModule.GetType().Namespace != "System.Reflection.Emit"` to check whether the assembly is dynamic just as we are in your new code under 3.5. I think replacing that to a call to `IsDynamic` would be just the ticket…

hmemcpy:
@blairconrad sorry, I haven't touched it since, but what I meant was, the `TypeCatalogue.GetAllAssemblies` does *additional* things as well. So it excludes dynamic assemblies but also does some other checks. Again, it's possible to extract the check to a single extension method, but this can be done later perhaps?

hmemcpy:
I tweaked the code a bit to use the new extension method in `TypeCatalogue` as well.

blairconrad:
> what I meant was, the TypeCatalogue.GetAllAssemblies does additional things as well. So it excludes dynamic assemblies but also does some other checks. Again, it's possible to extract the check to a single extension method, but this can be done later perhaps?

Oh, absolutely. The change you've made in the new version (just incorporating your `IsDynamic` into  `TypeCatalogue` is all I was hoping for at this point. That's perfect. thanks.

blairconrad:
Thanks, @hmemcpy. I'm happy to merge this as is.

Source/FakeItEasy/Core/AssemblyExtensions.cs
```diff
@@ -5,7 +5,7 @@
 
     internal static class AssemblyExtensions
     {
-        private static readonly string FakeItEasyAssemblyName = Assembly.GetExecutingAssembly().FullName;
+        internal static readonly Assembly FakeItEasyAssembly = Assembly.GetExecutingAssembly();
```
adamralph: I'd prefer to wrap this field in a `public` property rather than expose it.

adamralph: Actually, it shouldn't be a property on this type, since this type ought to remain strictly an extension method container.

I think I'd prefer to just leave the field repeated in `TypeCatalogue`. Or alternatively, it could be exposed as a static property on `TypeCatalogue` to be reused by `AssemblyExtensions`.

adamralph:
LGTM, but I'd like to address my comments regarding the `FakeItEasyAssembly` field.

Other than that, just need a rebase and it's good to go.

hmemcpy:
@adamralph what should I expose? The assembly itself? Using a backing field?

adamralph:
@hmemcpy yes, as you've done in the last commit, but I think it would make more sense to be on `TypeCatalogue` rather than `AssemblyExtensions`.

blairconrad:
Good comment on the `FakeItEasyAssembly` field, @adamralph. I should've thought to move it as well.

hmemcpy:
Done!

adamralph:
Thanks @hmemcpy!

And even a squash... :ok_hand:

blairconrad:
Thanks, @hmemcpy. Look for your name in the release notes!
:trophy:

blairconrad:
This issue has been fixed in [release 2.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/2.0.0).

PR #564, opened 2015-12-13
Switch FakeItEasy.Tests to .Net 4.5

blairconrad:
For discussion.
Making all tests target .NET 4.5.
Had to muck about with the .NET 3.5 shims so the `Tuple` definition wouldn't fight with the one from mscorlib.

Followed `ConditionalWeakTable`s lead and moved the shims to the FakeItEasy namespace. Although FakeItEasy.Compatibility or .Shims may be a better choice.
I can certainly tidy everything up if we think this is a reasonable approach.


Source/FakeItEasy.Net35/Compatibility.cs
```diff
@@ -51,28 +51,7 @@ where entry.Item1.IsAlive
...
-    }
-
-    [SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes", Justification = "Mimicks net40 BCL type.")]
-    [SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:FileMayOnlyContainASingleClass", Justification = "Tidier.")]
-    [AttributeUsage(AttributeTargets.Property | AttributeTargets.Field | AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
-    public class ImportManyAttribute
-        : Attribute
-    {
-    }
-}
```
blairconrad: Not necessary to remove these in this step, but they are not used. A real PR would do this separately.

Source/FakeItEasy.Net35/Compatibility.cs
```diff
@@ -100,13 +79,14 @@ public Tuple(T1 item1, T2 item2)
...
-namespace System.Linq
+namespace FakeItEasy
 {
+    using System;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
 
     [SuppressMessage("StyleCop.CSharp.MaintainabilityRules", "SA1402:FileMayOnlyContainASingleClass", Justification = "Tidier.")]
-    public static class EnumerableExtensions
+    internal static partial class EnumerableExtensions
```
blairconrad: Internal to agree with the rest of the class, now that they have the same namespace. In real life, I'd look a little closer to see if the class should be combined into one file or maybe whether we actually have 2 classes here.

Source/FakeItEasy.Net35/Compatibility.cs
blairconrad: I think that these became unnecessary in or around 2b33fa75b18950e751552f87bf03af050974ae8a.
I'm going to spin off a new issue for removal. Issue mostly because I'm _slightly_ nervous about removing any public class, and I'd like to be able to point at the change.

blairconrad:
> Followed `ConditionalWeakTable`'s lead and moved the shims to the `FakeItEasy` namespace. Although `FakeItEasy.Compatibility` or `.Shims` may be a better choice.

I'm backtracking on this. Keeping them in `FakeItEasy` means we won't need to conditionally `using FakeItEasy.Compatibility` or similar. It's a weak reason, but enough for me to prefer just putting the shims in `FakeItEasy`.

Source/FakeItEasy.Net35/Compatibility.cs
blairconrad: As threatened, made #565.

PR #566, opened 2015-12-15
remove .NET 3.5 project shims from the public API

### LINKED ISSUE #565: remove .NET 3.5 project shims from the public API
The shims were first added when the 3.5 project was created.
Later on `ImportManyAttribute` and `InheritedExportAttribute` were never removed when we stopped using MEF internally (in 2b33fa75b18950e751552f87bf03af050974ae8a).

The other shims in Compatibility.cs could've been internal all long.
These should've been hidden as part of #374, but I'd analyzed the .NET 4.0 DLL, not the 3.5 one.

blairconrad:
Fixes #565.
Extracted from #564.


Source/FakeItEasy/Configuration/BuildableCallRule.cs
```diff
@@ -113,7 +113,7 @@ public void WriteDescriptionOfValidCall(IOutputWriter writer)
 
         public virtual void ApplyWherePredicate(Func<IFakeObjectCall, bool> predicate, Action<IOutputWriter> descriptionWriter)
         {
-            this.wherePredicates.Add(Tuple.Create(predicate, descriptionWriter));
+            this.wherePredicates.Add(new Tuple<Func<IFakeObjectCall, bool>, Action<IOutputWriter>>(predicate, descriptionWriter));
```
adamralph: I think `Tuple.Create()` is the idiomatic method of constructing a `Tuple<..>`. It may be worth preserving the static shim which provides it.

Source/FakeItEasy.Net35/InternalsVisibleTo.cs
```diff
@@ -0,0 +1,3 @@
+﻿﻿using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("FakeItEasy.Net35.Tests, PublicKey=00240000048000009400000006020000002400005253413100040000010001002f50d82092713be482c885861c984334606da0f54c78738a3dd0862fb2dfc6080f0780132cc65d88f0f0c70af74e8a53430962395bfc1a36fab08b7a2549d387e805c13cc84acd884447ec8c4dcfb6216df720f0998380f9c906b5de8141798d64661f036d47274e6ecb76c9cde5f4cf2b521040601e44b3914fbeb9f39127f9")]
```
adamralph: No line ending? Tut tut :wink:

Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
```diff
@@ -693,6 +696,7 @@
       <Link>Properties\FakeItEasy.Dictionary.xml</Link>
     </CodeAnalysisDictionary>
   </ItemGroup>
+  <ItemGroup />
```
adamralph: Can be reverted.

adamralph:
@blairconrad changes look good. I left just a couple of very minor comments.

Source/FakeItEasy/Configuration/BuildableCallRule.cs
blairconrad: Blow me down. I should've looked for this. We were using it in two out of three places. Should I, as part of a final cleanup commit, make the 3rd (`ZipEnumerable`) use it as well?

Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
blairconrad: Man. I've no idea how I managed that. :blush:

Source/FakeItEasy/Configuration/BuildableCallRule.cs
adamralph: > Should I, as part of a final cleanup commit, make the 3rd (ZipEnumerable) use it as well?

:+1:

Source/FakeItEasy.Net35/InternalsVisibleTo.cs
blairconrad: Oh? Lots of files don't have line endings. Didn't know we cared. Will add.

Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
adamralph: I believe it's a helpful Visual Studio feature...

Source/FakeItEasy/Configuration/BuildableCallRule.cs
blairconrad: Also, the remove/add could be considered messy. Do you prefer a complete rework and force push, or just an additional commit?

Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
blairconrad: Still, I should've noticed in the diff and reverted.

Source/FakeItEasy/Configuration/BuildableCallRule.cs
adamralph: I don't mind a force push. It's often better than keeping a series of commits which all have a single ultimate aim.

Source/FakeItEasy.Net35/InternalsVisibleTo.cs
adamralph: I imagine we probably have a lot of inconsistency already.

I usually like to add a http://editorconfig.org/ file in the root which enforces it.

blairconrad: The would only work if I had the plugin, no?

blairconrad: Sounds like not such a bad idea. Maybe in a different issue. :wink:

blairconrad:
Thanks for the great comments, @adamralph. I think I've addressed everything. I ditched the offending "inline Tuple" commit, reworked the other two, and then added the obligatory cleanup commit which fixed some readonlys and removed some not needed EnumerableExtension stuff, including the copy of `ZipEnumerable`. (Which in the process let me get rid of the extra usage of the `Tuple<T1, T2>` constructor.)

Source/FakeItEasy.Net35/InternalsVisibleTo.cs
adamralph: Yeah, you need the plug-in. It does indeed sounds like a good candidate for a new issue.

adamralph:
:+1: great stuff!

blairconrad:
You're kind. Thanks for the constructive comments and merge.
We'll still need a little work to use xUnit 2.0 in the 3.5 unit test package, but it's looking better than it was.

Issue #545, opened 2015-09-24
Switch from NUnit to xUnit.net


adamralph:
Now that the specs have switched from mspec to xBehave.net, it probably makes sense to convert the rest of the tests from NUnit to xUnit.net so that we're using a single underlying test framework for all. Given that xUnit.net also has .NET core support in the works, it would also ease the passage to solving #531.

Note that this may also solve #231.

blairconrad:
I'm not sure what "solve da097217cf6bd2470442314cfab6ac2cb34e2f2a" means. That's a commit.

blairconrad:
Also,  how do you feel about holding off on this until after 2.0? :grinning:

adamralph:
Oops! I meant to say "solve #231". Fixed.

No urgency around this at all. Can definitely wait until after 2.0.

adamralph:
Note that this can be done bit by bit, a PR for a test here, another PR for a test there, etc.

One thing that will be required before converting the first test from NUnit to xUnit.net is the introduction of a second pass at executing each test DLL in the rakefile, using the xUnit.net console. I.e. each test DLL will need two passes, one with NUnit console and one with xUnit.net console.

mat-mcloughlin:
That sounds fine to me. Happy to pick up the rake file, also happy to start shifting some across If you like.

Also would you want to move to using fluent assertions as well?

adamralph:
@mat-mcloughlin that would be awesome.

Yeah, for each test you touch, you may as well switch to FA too, that's generally been our policy. Thanks!

mat-mcloughlin:
np, I feel the need to pick up something as this to payback for the fact I haven't contributed to an OSS in months :D

blairconrad:
In order to head off PR comments from the future, how do we feel about having 2 commits per changed file: one for the NUnit Asserts→FluentAssertions (if necessary, many won't be), and one for NUnit→xUnit conversion? It may make things a little easier to read.

adamralph:
:+1: to separating the commits

blairconrad:
@mat-mcloughlin, I'm inclined to remove the **up-for-grabs** label and mark it as **taken** by you.
Suits?

mat-mcloughlin:
sure. planning to start hitting some of theme tomorrow morning

blairconrad:
Thanks, sir!

mat-mcloughlin:
[...]

PR #568, opened 2015-12-23
Make tests target 4.5

blairconrad:
Supports #545: we can't use xUnit 2.0+ without targeting .NET 4.5.

I couldn't fathom how to keep our 3.5 `Tuple` from being confused with the built-in 4.5 version. So I removed ours.
Truth be told, I prefer the result. I'm not bright enough to remember what's `Item1` and what's `Item2`.


blairconrad:
Thanks, @adamralph!

PR #569, opened 2015-12-31
Prevent redefine rule behavior

### LINKED ISSUE #534: Stacking multiple results on the same call configuration instance doesn't work

blairconrad:
Fixes #534.
Closes #557, which this PR supersedes.
I just rebased @thomaslevesque's commits after I caused a conflict with #568.
All I had to do was resolve the conflicting types for `wherePredicates` in BuildableCallRule.cs.

#557's original description:

> This is in order to prevent an unsupported scenario where the user keeps
> the result of A.CallTo(...) and uses it to define multiple behaviors.
> Doing so would override the existing behavior, not stack the new one,
> even though it's not necessarily what the user would expect. Making it
> illegal prevents confusion. It remains possible to keep the result of
> A.CallTo to make later calls to MustHaveHappened and friends.
> 
> See the comments in FakeItEasy/FakeItEasy#534 for more details.

### DIFF
--- a/Source/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
+++ b/Source/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
@@ -1,6 +1,8 @@
 ﻿namespace FakeItEasy.Specs
 {
+    using System;
     using System.Diagnostics.CodeAnalysis;
+    using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
     using Xbehave;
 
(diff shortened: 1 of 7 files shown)


thomaslevesque:
Well obviously I don't have much to say about the code :wink: 
Looks good to me :+1:

PR #574, opened 2016-01-10
replace IFakeConfigurator with IFakeOptionsBuilder

### LINKED ISSUE #461: Replace IFakeConfigurator with IFakeOptionsBuilder

blairconrad:
Fixes #461 by replacing `IFakeConfigurator` with `IFakeOptionsBuilder` (and matching parameterized) type.
This will allow the creation of options builders that can configure fakes using any of the methods available on `IFakeOptions<TFake>`.


blairconrad:
You thought I was kidding about the [22 commits](https://github.com/FakeItEasy/FakeItEasy/pull/529#issuecomment-170185815), didn't you?

blairconrad:
I'm off to bed. Will note this up a bit and/or respond to comments later. But this should satisfy @adamralph's curiosity for now.

blairconrad:
I like a clean merge line. Lemme rebase.

Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
```diff
@@ -344,11 +344,11 @@
     <Compile Include="..\FakeItEasy\Core\IFakeObjectCallRule.cs">
       <Link>Core\IFakeObjectCallRule.cs</Link>
     </Compile>
-    <Compile Include="..\FakeItEasy\Core\IFakeObjectConfigurator.cs">
-      <Link>Core\IFakeObjectConfigurator.cs</Link>
-    </Compile>
     <Compile Include="..\FakeItEasy\Core\IFakeObjectContainer.cs">
-      <Link>Core\IFakeObjectContainer.cs</Link>
+      <Link>IFakeObjectContainer.cs</Link>
```
thomaslevesque: Did you move it out of Core intentionally?

Source/FakeItEasy.Tests/Core/DynamicContainerTests.cs
```diff
@@ -106,22 +111,52 @@ public void Should_not_fail_when_more_than_one_configurator_exists_for_a_given_t
 
         private DynamicContainer CreateContainer()
         {
-            return new DynamicContainer(this.availableDummyFactories, this.availableConfigurators);
+            return new DynamicContainer(this.availableDummyFactories, this.availableOptionsBuilders);
         }
 
-        public class ConfigurationForTypeWithDummyFactory : FakeConfigurator<TypeWithDummyFactory>
+        public class OptionsBuilderForTypeWithDummyFactory : IFakeOptionsBuilder
```
thomaslevesque: Why implement the interface directly instead of inheriting `FakeOptionsBuilder<TypeWithDummyFactory>`?

Source/FakeItEasy.Tests/Core/FakeManagerProviderTests.cs
```diff
@@ -67,18 +64,6 @@ public void Fetch_should_tag_the_proxy_with_the_manager()
         }
 
         [Test]
-        public void Fetch_should_configure_the_proxy()
```
thomaslevesque: Is this test no longer relevant?

Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
blairconrad: … great question.

Source/FakeItEasy.Tests/Core/DefaultSutInitializerTests.cs
```diff
@@ -57,17 +57,27 @@ public void Should_use_same_instance_when_multiple_arguments_are_of_same_type()
         }
 
         [Test]
-        public void Should_pass_empty_fake_options_to_fake_manager()
+        public void Should_pass_impotent_options_builder_to_fake_manager_for_each_constructor_argument()
```
blairconrad: I did not!

Source/FakeItEasy.Tests/Core/FakeManagerProviderTests.cs
thomaslevesque: Ok, I get it now that I saw the changes in FakeManagerProvider

Source/FakeItEasy.Tests/Core/DynamicContainerTests.cs
blairconrad: Because `FakeOptionsBuilder<TFake>` didn't exist yet!
If you prefer, I can update this to use `FakeOptionsBuilder`.

thomaslevesque: If you don't mind, I think it would make the tests cleaner and more readable

Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
blairconrad: Oh. Sorry, rereading. I did not! Nice catch.

Source/FakeItEasy.Tests/Core/DynamicContainerTests.cs
blairconrad: No problemo. Will tack on as additional cleanup commit.

Source/FakeItEasy/Core/IFakeObjectOptionsBuilder.cs
```diff
@@ -0,0 +1,18 @@
+namespace FakeItEasy.Core
+{
+    using System;
+    using Creation;
+
+    /// <summary>
+    /// Handles global configuration of fake objects by building fake creation options.
+    /// </summary>
+    public interface IFakeObjectOptionsBuilder
```
thomaslevesque: Probably off-topic for this PR, but I'm curious... what's the difference between a `IFakeObjectOptionsBuilder` and a `IFakeOptionsBuilder`?

Source/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
```diff
@@ -29,9 +33,11 @@ public object CreateDummy(Type typeOfDummy)
             return result;
         }
 
-        public object CreateFake(Type typeOfFake, IProxyOptions options)
+        public object CreateFake(Type typeOfFake, Action<IFakeOptions> optionsBuilder)
         {
-            return this.fakeCreator.CreateFake(typeOfFake, options, this.session, throwOnFailure: true);
+            var fakeOptions = this.BuildFakeOptions(typeOfFake, optionsBuilder);
```
thomaslevesque: Shouldn't it be called `proxyOptions`?

Source/FakeItEasy/FakeOptionsBuilder.cs
```diff
@@ -0,0 +1,53 @@
...
+            return type == typeof(TFake);
+        }
+
+        /// <summary>
+        /// Manipulates <paramref name="options"/>, which will later be used to 
+        /// create a Fake.
+        /// </summary>
+        /// <param name="typeOfFake">The type the fake object represents.</param>
+        /// <param name="options">The fake options to manipulate.</param>
+        public void BuildOptions(Type typeOfFake, IFakeOptions options)
```
thomaslevesque: Should check that `typeOfFake == typeof(TFake)`

Source/FakeItEasy/Core/IFakeObjectOptionsBuilder.cs
blairconrad: I have also wondered this.

blairconrad: You probably know as much as we do. It seems to be something that's used to build options. Users can supply a new kind when creating a fake scope. I've wondered if this is ever done. Had we reliable feature usage statistics, I'd be asking "Should we drop the `CreateScope` calls that take an `IFakeObjectContainer`?".
I'm especially of this mindset after reading 

> My criteria there was “do I groan anytime a user asks me a question about this feature…” If the answer was “yes”, I killed it.

in [Deleting Code](http://jeremydmiller.com/2016/01/13/deleting-code/).

Source/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
blairconrad: Yes. Well spotted.

Source/FakeItEasy/FakeOptionsBuilder.cs
blairconrad: I was torn about this. We'll only execute `BuildOptions` if we `CanBuildOptionsForFakeOfType`.
Of course, that's a kind of temporal coupling.
Certainly it's safer to add the check. I will do so. Actually, I'd prefer to delegate to `CanCanBuildOptionsForFakeOfType` if that's acceptable.

blairconrad: Oh, and keep in mind there will be no such protection for user-implmented `IFakeOptionsBuilder`s.

blairconrad:
We're going to need a rebase. And I guess I'll need to redo all the Specs after https://github.com/FakeItEasy/FakeItEasy/pull/576#discussion_r49584676.

I'm afraid of losing things I'll need to do, so I'll put a checklist here so comments don't get lost when I rebase. @thomaslevesque, unless I hear otherwise, I'm leaving out things that seemed like answered questions to me. If they aren't settled, please chastise me!
Likewise if I've forgotten something.

- [ ] move the 3.5 project's `IFakeObjectContainer` back into `Core` (by amending commit that took it out)
- [ ] look for more opportunities to use `FakeOptionsBuilder`, as @thomaslevesque suggests for `OptionsBuilderForTypeWithDummyFactory`. (as part of a new cleanup commit)
- [ ] consider pursuing what a `IFakeObjectOptionsBuilder` and whether we should keep using it with new `Scope`s, **if there's interest**
- [ ] rename `fakeOptions` to `proxyOptions` in `DefaultFakeAndDummyManager.CreateFake` (by amending the commit that changed how we make those options)
- [ ] check that we `CanBuildOptionsForFakeOfType` before applying `FakeOptionsBuilder.BuildOptions`
- [ ] rework all the Specs, according to https://github.com/FakeItEasy/FakeItEasy/pull/576#discussion_r49584676 (by amending the commits)

Source/FakeItEasy.Tests/Core/DefaultSutInitializerTests.cs
blairconrad: Oh, and about this. @thomaslevesque, I speak English as my first (only) language and have what's probably a larger than average vocabulary.
If you're so inclined, do point out when I use terms (especially in code, comments, or documentation) that would give a non-native speaker difficulties.
For example, would the method name be better with "do-nothing" or "no-op" or "empty" in place of "impotent"? I can make such a change.

thomaslevesque: Well, the use of the term "impotent" surprised me, but I understand what you meant by it. Perhaps "empty" would have a more obvious meaning, but I don't really care either way.

blairconrad:
added 

> replace "impotent" with "empty" when referring to options builder actions (amending)

to the list, and changed spec reworking to be in its own commit, rather than amending the original

blairconrad:
I've pushed a new branch. It (I believe) addresses

- move the 3.5 project's IFakeObjectContainer back into Core (by amending commit that took it out)
- replace "impotent" with "empty" when referring to options builder actions (amending)
- rename fakeOptions to proxyOptions in DefaultFakeAndDummyManager.CreateFake (by amending the commit that changed how we make those options)

I'll continue with the other changes later, although my next two days are quite full, so if someone has an urge to peek at the changes and see if actually _have_ resolved the 3 problems, do go ahead.

thomaslevesque:
Looking good so far, waiting for the rest of the changes ;)

blairconrad:
They'll come. While you wait, does anyone think removing the `CreateScope` overloads that take the container is a good step?

thomaslevesque:
> does anyone think removing the `CreateScope` overloads that take the container is a good step?

I'd rather not remove them until we really understand what they're here for... perhaps @patrik-hagne can enlighten us?

blairconrad:
The use would be to customize how new fake objects are configured, and dummies created, rather than the default `DynamicContainer`.

To me it feels like a cool feature that would demonstrate the flexibility of FakeItEasy, but that probably nobody uses…

thomaslevesque:
> The use would be to customize how new fake objects are configured, and dummies created, rather than the default `DynamicContainer`.
>
> To me it feels like a cool feature that would demonstrate the flexibility of FakeItEasy, but that probably nobody uses…

I had a closer look and it seems that `DelegateFakeObjectContainer` _might_ be useful... I'm not sure why anyone would want to use `NullFakeObjectContainer`, though.

blairconrad:
It's interesting. There are ways to provide custom Dummies and to configure fakes without using this mechanism, but using Scopes alongside *ObjectContainers allow one to change the behaviour just while in the Scope.
I'm not really trying to argue for removing the functionality, just pointing out that if it's confusing, and nobody's benefiting, we _could_ remove it. Hard to tell if anyone's benefiting.

thomaslevesque:
I think I'd rather keep the feature. The code doesn't seem too complex to maintain, and there might be people using it. If we remove it, there's no other way to achieve the same effect, as far as I can tell.

blairconrad:
You'll find no argument here.

blairconrad:
Thanks for entertaining the idea.

blairconrad:
That last push was only so big because I'd forgotten to pull before my previous rebase. Only f3e35037baadc769abcade8ab0e040a2e0d2b29f and 611131dd592963ebac933df0eea46979da184b9d are "new" this time 'round.

Source/FakeItEasy.Tests/FakeOptionsBuilderTests.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace FakeItEasy.Tests
+{
+    using System;
+    using FakeItEasy.Creation;
+    using FluentAssertions;
+    using NUnit.Framework;
+    using TestHelpers;
+
+    public class FakeOptionsBuilderTests
```
thomaslevesque: Missing `[TestFixture]` attribute. ReSharper doesn't seem to care and detects the tests anyway, but I'm not sure about the NUnit console runner.

blairconrad: From [the docs](http://www.nunit.org/index.php/index.php?p=testFixture&r=2.6.3):
> Beginning with NUnit 2.5, the TestFixture attribute is optional for non-parameterized, non-generic fixtures.

And you can see it in [the CI server's records](http://teamcity.codebetter.com/viewLog.html?currentGroup=test&scope=%23teamcity%23FakeItEasy.Tests%23teamcity%23FakeOptionsBuilderTests&pager.currentPage=1&order=DURATION_DESC&recordsPerPage=20&filterText=FakeOptionsBuilder&status=&buildTypeId=bt929&buildId=208045&tab=testsInfo).

So I think we're safe without the tag. We've some classes with, some without. I think recently-added ones have omitted. If you're very concerned, we can publish a standard and update all classes one way or another. Then again, @mat-mcloughlin is currently converting _all_ the tests to XUnit, so the point may soon be moot.

thomaslevesque: Ah, I wasn't aware of that! No, I'm not concerned. If it works without the attribute, it's fine by me.

Source/FakeItEasy.Tests/Core/DynamicContainerTests.cs
blairconrad: As it turns out, making the change does make the options builder simpler, but the test itself becomes more complicated:

```diff
         [Test]
         public void ConfigureFake_should_apply_configuration_from_registered_options_builders()
         {
             this.availableOptionsBuilders.Add(new OptionsBuilderForTypeWithDummyFactory());

             var container = this.CreateContainer();

-            var fakeOptions = A.Fake<IFakeOptions>();
+            var fakeOptions = A.Fake<IFakeOptions<TypeWithDummyFactory>>(options=>options.Implements<IFakeOptions>());
             var fakeTypeWithDummyFactory = A.Fake<TypeWithDummyFactory>();

-            A.CallTo(() => fakeOptions.ConfigureFake(A<Action<object>>._))
-                .Invokes((Action<object> action) => action(fakeTypeWithDummyFactory));
+            A.CallTo(() => fakeOptions.ConfigureFake(A<Action<TypeWithDummyFactory>>._))
+                .Invokes((Action<TypeWithDummyFactory> action) => action(fakeTypeWithDummyFactory));

-            container.BuildOptions(typeof(TypeWithDummyFactory), fakeOptions);
+            container.BuildOptions(typeof(TypeWithDummyFactory), (IFakeOptions) fakeOptions);

             Assert.That(fakeTypeWithDummyFactory.WasConfigured, Is.True, "configuration was not applied");
         }
```

This is because the `FakeOptionsBuilder<TFake>` has strict requirements about what time of options object is passed to it, and so on.
In my opinion, this makes the code more complicated than just implementing the `IFakeOptionsBuilder` by hand, but if you like it better, I'll go forward with the change.

Either way, `RobotRunsAmokEventFakeOptionsBuilder` can be converted easily. The other options builders all derive from `ConventionBasedOptionsBuilder` in `FakeOptionsBuilderSpecs`, and I think leaving them as is certainly saves code, but if you'd like to see them derive from `FakeOptionsBuilder<TFake>`, I can whip up a diff.

thomaslevesque: Ah, I see. Well, I guess it's better to leave it as is, then.

blairconrad:
In case you're worried, I've not abandoned the PR. I'm working on rewording the Specs, but it's slow and tedious. I'm slightly consoled knowing that _reading_ the change will be so much worse!

blairconrad:
Well, for good or ill, I've updated the specs that didn't conform to the language suggested in #576.

thomaslevesque:
> Well, for good or ill, I've updated the specs that didn't conform to the language suggested in #576.
> I think

I didn't follow #576 closely, so I don't really know how to evaluate those changes... Maybe @adamralph could have a look?

blairconrad:
> I didn't follow #576 closely, so I don't really know how to evaluate those changes... Maybe @adamralph could have a look?

I'm not sure I followed all that well either, but as with anything, the last part of the PR had a goal to update the affected specs to make sure they were

1. correct,
2. intention-revealing, and
3. readable

So even though @adamralph was the driver of the changes there, I think it's important to make sure that we're all satisfied with the kinds of approach we'll take toward the specs in the future.
I found that converting the specs that I'd touched was tedious (partly because there were _a lot_ this time), but ultimately, I think that the newer version exposed some hidden assumptions and ended up reading fairly well. I'm still not _entirely_ sure about all the `.x(() => {})`s, but they were convenient places to hang the descriptions of the requirements.
If you'd like to wait for @adamralph to weigh in before merging, I completely understand, but I'm interested in your overall opinion of the format. And if nothing else, you've demonstrated a keen eye earlier in the review. Maybe you'll see a place where I've introduced a lie through lazy editing. :disappointed:

thomaslevesque:
> I'm still not entirely sure about all the `.x(() => {})`s

Me neither... perhaps we should introduce an extension method `.x()` to denote empty steps (I guess it could be added to xBehave, but in the meanwhile we can use our own)

blairconrad:
That's not a bad idea. 
I'd been toying with the idea that the "see" comments could instead be an extension method that just references the type/method/whatever. Gotta go. Talk later!

blairconrad:
e.g.

```c#
public static class SpecExtensions
{
    public static IStepBuilder See<T>(this string text)
    {
        return text.x(() => { });
    }
}

…

"Given a type that has an implicit options builder defined"
    .See<RobotActivatedEvent>();
```

instead of

```c#
"Given a type that has an implicit options builder defined"
    .x(() => { }); // see RobotActivatedEvent
```

thomaslevesque:
Oh, clever ;)
Perhaps a bit too specific, though?

blairconrad:
Too specific for xBehave, definitely.
For our own specs where we have about 60 instances of this pattern? I'm not so sure.
An overload could be added to point at methods/constructors as well, or we could make this more generic so we can point at almost any object that we can describe with code…

patrik-hagne:
I love the "See-extension"!
Den 24 jan 2016 16:11 skrev "Blair Conrad" <notifications@github.com>:

thomaslevesque:
> For our own specs where we have about 60 instances of this pattern? I'm not so sure.

In this case, I guess it makes sense

blairconrad:
@patrik-hagne, thanks. Me too.
@thomaslevesque, rereading your original empty `.x` extension comment, I see how I misinterpreted and went over-specific, that being said, I think the `See` gives a good mix or readability and (nearly no) functionality, but I feel like you're still unsure. I'm happy to reconsider. Even if I win you over, we should probably decide what to do about references to the methods (such as the constructors and the `BuildOptions` method).

The best I've come up for the `BuildOptions` is
```c#
"And the options builder updates the options to configure the fake"
    .See((DomainEventFakeOptionsBuilder ob) => ob.BuildOptions);
```

with an extension method like 

```c#
public static IStepBuilder See<T>(this string text, Func<T, Action<Type, IFakeOptions>> func)
{
    return text.x(() => { });
}
```

This is in contrast to

```c#
"And the options builder updates the options to configure the fake"
    .x(() => { }); // see DomainEventFakeOptionsBuilder.BuildOptions
```

of course, if you didn't like the _first_ `See`, that discussion is premature.

thomaslevesque:
> I feel like you're still unsure

No, I'm ok with it. Whether it's `.x()` with a comment, or `.See<T>()`, I don't really have a strong opinion on this.

blairconrad:
Cool. Although as @adamralph said, much of this may be bikeshedding, I still like having the compiler jump on me if the class goes away.

How do you feel about the (admittedly inferior) 

```c#
"And the options builder updates the options to configure the fake"
    .See((DomainEventFakeOptionsBuilder ob) => ob.BuildOptions);
```
?

We can keep looking for an alternative. We could even leave `.x()` and a comment until we find something better.

Hmm. In the future, would `nameof` give us a cheap/easy way to refer to `DomainEventFakeOptionsBuilder.BuildOptions` without too much surrounding noise?

thomaslevesque:
> How do you feel about the (admittedly inferior)

A bit verbose, but clear enough, and compiler-checked

> We could even leave .x() and a comment until we find something better.

In this case you might as well leave it as `.x(() => { })` until we find something better... I don't want to make you change all those specs more often than necessary ;)

> Hmm. In the future, would nameof give us a cheap/easy way to refer to DomainEventFakeOptionsBuilder.BuildOptions without too much surrounding noise?

You mean something like `.See(nameof(DomainEventFakeOptionsBuilder.BuildOptions))`? Yes, I guess it could work.

blairconrad:
> I don't want to make you change all those specs more often than necessary

Oh, I don't mind doing work, but I'd be sad if we had a half-baked solution that gave us a false sense of security.

Okay. I'll change the references to types at least and repush when I have time.
Thanks, sir!

blairconrad:
Although it now strikes me that 

```c#
"And the options builder updates the options to configure the fake"
    .x(() => { }); // see DomainEventFakeOptionsBuilder.BuildOptions
```

doesn't really provide more information than

```c#
"And the options builder updates the options to configure the fake"
    .x(() => { }); // see DomainEventFakeOptionsBuilder
```

because what other part of `DomainEventFakeOptionsBuilder` are we going to see?

Maybe just use the original `See<T>` for those steps as well?

blairconrad:
A late-breaking thought. We could spin off the "tidy the specs" portion into another issue for a later time (I won't let it linger too long) and just consider merging up to 
"using FakeOptionsBuilder to build RobotRunsAmokEventFakeOptionsBuilder" - 6eede85 at this time.

thomaslevesque:
[**@blairconrad**](https://github.com/blairconrad) commented on [25 janv. 2016 02:45 UTC+1](https://github.com/FakeItEasy/FakeItEasy/pull/574#issuecomment-174369345 "2016-01-25T01:45:48Z - Replied by Github Comment Enhancer"):
> A late-breaking thought. We could spin off the "tidy the specs" portion into another issue for a later time (I won't let it linger too long) and just consider merging up to 
> "using FakeOptionsBuilder to build RobotRunsAmokEventFakeOptionsBuilder" - 6eede85 at this time.

Agreed. The PR is pretty big already ;)

Somehow I missed the point where you fixed the items on your [checklist](https://github.com/FakeItEasy/FakeItEasy/pull/574#issuecomment-171621987) above. I'll review those changes ASAP.

blairconrad:
Thanks. I've put the extra "spec cleanup" commits aside for now.

thomaslevesque:
Just reviewed the last few commits. I just have one last comment (inline), otherwise everything looks good.

thomaslevesque:
OK, looks like it's ready to merge. @blairconrad, do you want another pair of eyes to review it first?

blairconrad:
Thanks, @thomaslevesque. I'm more than content with your assessment. "At least 2 owners on a merge", we say, and we're at least 2 owners.
Besides, @adamralph said he's still busy. I say merge if and when you're ready.

blairconrad:
Thanks for the comments throughout the review. I enjoyed them.

thomaslevesque:
Here we go then!

adamralph:
:+1: nice work guys!

Issue #336, opened 2014-07-15
2.0.0 release


blairconrad:
**Ready** when all other issues on this milestone are **Done** and closed.

- [x] check integration package compatibility
   - [x] AutoFixture.AutoFakeItEasy
     - stable *incompatible* with 2.0.0-rc2, but [restricted to less than 2.0](https://github.com/AutoFixture/AutoFixture/pull/617#issuecomment-217612774)
   - [x] Machine.Fakes.FakeItEasy
     - stable *incompatible* with 2.0.0-rc1
     - pre-release compatible with  2.0.0-rc2
   - [x] Autofac.Extras.FakeItEasy
     - stable *incompatible* with 2.0.0-rc1 but [restricted to less than 2.0](https://github.com/autofac/Autofac.Extras.FakeItEasy/pull/1)
     - pre-release compatible with  2.0.0-rc2
- [x] run code analysis in VS in *Release* mode and address violations (send a regular PR which must be merged before continuing) - #336
- [x] release RC https://www.nuget.org/packages/FakeItEasy/2.0.0-rc2 (1st May 2016)
- [x] ensure the RC has been available for *some period of time* (box checked on 20th May, almost 3 weeks after release)
- [x] smoke test the analyzer package. We're not releasing it, so just download the nupkg and install from local disk
- [x] if necessary, change `VERSION_SUFFIX` on [CI Server](http://teamcity.codebetter.com/admin/editBuildParams.html?id=buildType:bt929)
      to appropriate "-beta123" or "" (for non-betas) value and initiate a build
- [x] check build
-  edit draft release in [GitHub UI](https://github.com/FakeItEasy/FakeItEasy/releases):
   - [x] complete release notes, mentioning non-owner contributors, if any
   - [x] attach nupkg
   - [x] publish the release
- [x] remove https://github.com/FakeItEasy/FakeItEasy/wiki/Changes-in-version-2.0
- [x] push NuGet package
- [x] de-list pre-release or superseded buggy NuGet packages if present
- [x] tweet, mentioning contributors and post link as comment here for easy retweeting ;-)
- [x] post tweet in [Gitter ](https://gitter.im/FakeItEasy/FakeItEasy)
- [x] post links to the GitHub release in each issue in this milestone, with thanks to contributors
- [x] use `rake set_version[new_version]` to 
   - create a new branch
   - change CommonAssemblyInfo.cs to expected minor version (of form _xx.yy.zz_)
   - push to origin
[...]

PR #575, opened 2016-01-11
Make FakeScope thread-safe

IharBury:
connects to #484

FakeScope.Current modified to store the value in ExecutionContext instead of static field. This makes all fakes a bit slower but allows child contexts in parallel tests including async code support.

There is currently a problem with pull request however. It breaks FakeScopeSpecs.CreatingFakeInsideScope scenario. Xbehave executes each step as seperate async task. So the ExecutionContext (and FakeScope with it) doesn't flow between steps. Does anyone have any idea how to fix that scenario?

### DIFF
--- a/Source/FakeItEasy/Core/FakeScope.cs
+++ b/Source/FakeItEasy/Core/FakeScope.cs
@@ -2,7 +2,10 @@ namespace FakeItEasy.Core
 {
     using System;
     using System.Collections;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Runtime.Remoting.Messaging;
 
     /// <summary>
     /// Represents a scope for fake objects, calls configured within a scope
@@ -13,18 +16,47 @@ namespace FakeItEasy.Core
     internal abstract class FakeScope
         : IFakeScope
     {
+        private const string CurrentScopeLogicalCallContextKey = "FakeItEasy.FakeScope.Current";
+
+        private static readonly RootScope root;
+        private static readonly ConcurrentDictionary<Guid, FakeScope> scopes;
+
+        [SuppressMessage(
+            "Microsoft.Performance", 
+            "CA1810:InitializeReferenceTypeStaticFieldsInline", 
+            Justification = "RootScope constructor requires scopes to be initialized first.")]
         static FakeScope()
         {
-            Current = new RootScope();
+            scopes = new ConcurrentDictionary<Guid, FakeScope>();
+            root = new RootScope();
         }
 
         private FakeScope()
         {
+            this.Id = Guid.NewGuid();
+            FakeScope.scopes.TryAdd(this.Id, this);
         }
 
-        internal static FakeScope Current { get; set; }
+        internal static FakeScope Current
+        {
+            get
+            {
+                var id = (Guid?)CallContext.LogicalGetData(CurrentScopeLogicalCallContextKey);
+                FakeScope scope;
+                return id.HasValue && FakeScope.scopes.TryGetValue(id.Value, out scope) ? scope : FakeScope.root;
+            }
+            set
+            {
+                if (value == null)
+                    throw new ArgumentNullException("value");
+
+                CallContext.LogicalSetData(CurrentScopeLogicalCallContextKey, value.Id);
+            }
+        }
 
         internal abstract IFakeObjectContainer FakeObjectContainer { get; }
+
+        internal Guid Id { get; private set; }
 
         /// <summary>
         /// Creates a new scope and sets it as the current scope.
@@ -157,6 +189,8 @@ namespace FakeItEasy.Core
             {
                 this.RemoveRulesConfiguredInScope();
                 FakeScope.Current = this.parentScope;
+                FakeScope removedScope;
+                FakeScope.scopes.TryRemove(this.Id, out removedScope);
             }
 
             protected override void OnAddInterceptedCall(FakeManager fakeManager, ICompletedFakeObjectCall call)
--- a/Source/FakeItEasy.Net35/Compatibility/ConcurrentDictionary.cs
+++ b/Source/FakeItEasy.Net35/Compatibility/ConcurrentDictionary.cs
@@ -51,5 +51,27 @@
                 return this.dictionary.TryGetValue(key, out value) && this.dictionary.Remove(key);
             }
         }
+
+        public bool TryAdd(TKey key, TValue value)
+        {
+            lock (this.dictionary)
+            {
+                if (this.dictionary.ContainsKey(key))
+                {
+                    return false;
+                }
+
+                this.dictionary.Add(key, value);
+                return true;
+            }
+        }
+
+        public bool TryGetValue(TKey key, out TValue value)
+        {
+            lock (this.dictionary)
+            {
+                return this.dictionary.TryGetValue(key, out value);
+            }
+        }
     }
 }
--- a/Source/FakeItEasy.Tests/Core/FakeScopeTests.cs
+++ b/Source/FakeItEasy.Tests/Core/FakeScopeTests.cs
@@ -5,6 +5,8 @@
     using System.Linq;
     using FakeItEasy.Core;
     using NUnit.Framework;
+    using System.Threading;
+    using System.Threading.Tasks;
 
     [TestFixture]
     public class FakeScopeTests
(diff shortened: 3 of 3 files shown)


thomaslevesque:
The changes look good, but I have two worries:
- As you mentioned, XBehave doesn't flow the context between steps; this is already an issue in one test, and it's likely to cause issues for anyone using scopes in XBehave tests. I think it might be possible to flow the context, but it would require a change in XBehave.
- The logical call context uses a copy-on-write behavior, but only as of .NET 4.5 (see [this blog post](http://blog.stephencleary.com/2013/04/implicit-async-context-asynclocal.html) by @StephenCleary). For async code targeting .NET 4.0 with Microsoft.Bcl.Async, this is probably going to cause trouble.

blairconrad:
I fear I'm not following, but I am not sure how much of a problem this is. I mean, obviously it's a problem for that test as written, but is the xBehave behaviour to create a new logical thread for each step, so it will

1. create a scope in some logical thread `A` during `and a fake scope using that container` and
1. create a fake in _some other logical thread `B` that does not have `A` as an ancestor_ during `and a fake scope using that container`?

In that case, it sounds like the FakeItEasy may be behaving properly and the test is flawed.

I take @thomaslevesque's point that people will probably write tests in the way that I wrote that one, but perhaps the proper action is to discourage that.

blairconrad:
(I messed up the name of the second step when I originally made the previous comment. Fixed.)

blairconrad:
Ah! I reread @thomaslevesque's comment "I think it might be possible to flow the context, but it would require a change in XBehave". I understand now. Yes, that makes more sense than my "don't do that" admonishment. If it's feasible and reasonable general xBehave behaviour.

Source/FakeItEasy/Core/FakeScope.cs
```diff
@@ -13,19 +16,48 @@ namespace FakeItEasy.Core
...
+                var id = (Guid?)CallContext.LogicalGetData(CurrentScopeLogicalCallContextKey);
+                FakeScope scope;
+                return id.HasValue && FakeScope.scopes.TryGetValue(id.Value, out scope) ? scope : FakeScope.root;
+            }
+            set
+            {
+                if (value == null)
+                    throw new ArgumentNullException("value");
+
+                CallContext.LogicalSetData(CurrentScopeLogicalCallContextKey, value.Id);
```
TimLovellSmith: Why use LogicalSetData and not SetData? I would not think you would want the current fake scope to propagate with remoting calls, since the fake scope cannot be shared across app domains.

Source/FakeItEasy/Core/FakeScope.cs
```diff
@@ -13,19 +16,48 @@ namespace FakeItEasy.Core
     internal abstract class FakeScope
         : IFakeScope
     {
+        private const string CurrentScopeLogicalCallContextKey = "FakeItEasy.FakeScope.Current";
+
+        private static readonly RootScope root;
+        private static readonly ConcurrentDictionary<Guid, FakeScope> scopes;
```
TimLovellSmith: Why, on the context, pass GUIDs that point to fakescopes inside a dictionary instead of just passing object pointers to the fakescopes?

Source/FakeItEasy/Core/FakeScope.cs
```diff
@@ -13,19 +16,48 @@ namespace FakeItEasy.Core
...
         {
-            Current = new RootScope();
+            scopes = new ConcurrentDictionary<Guid, FakeScope>();
+            root = new RootScope();
         }
 
         private FakeScope()
         {
+            this.Id = Guid.NewGuid();
+            FakeScope.scopes.TryAdd(this.Id, this);
```
blairconrad: nit: scopes.TryAdd ? (I am not at a compiler, so can't check). Similarly elsewhere.

blairconrad: My mistake. Scopes, I think. While I'm complaining about trivialities, if you end up revising these commits, consider addressing the style warnings:
http://teamcity.codebetter.com/viewLog.html?buildId=207272&buildTypeId=bt929&tab=buildLog#_state=123&focus=124
Thanks.

Source/FakeItEasy/Core/FakeScope.cs
thomaslevesque: @TimLovellSmith, I think this is necessary to support async code; only the logical call context is flowed to the async continuation (which can run on a different thread).

TimLovellSmith: @thomaslevesque Ah, you're right.

adamralph:
I'll chime in ASAP regarding xbehave...

thomaslevesque:
[**@adamralph**](https://github.com/adamralph) commented on [11 janv. 2016 22:18 UTC+1](https://github.com/FakeItEasy/FakeItEasy/pull/575#issuecomment-170692665 "2016-01-11T21:18:22Z - Replied by Github Comment Enhancer"):
> I'll chime in ASAP regarding xbehave...

Ah, I forgot you were also the owner of xbehave... how convenient :wink:

Apparently, xBehave is executing each step in a threadpool thread. So any change to the call context that is done in a step isn't propagated to the subsequent steps... I think it would need to capture the `ExecutionContext` after each step, and pass it to the next step with `ExecutionContext.Run`. I did a quick POC in LinqPad and it seems to work:

https://gist.github.com/thomaslevesque/8a43ef3f0db3bd6061dd

Source/FakeItEasy/Core/FakeScope.cs
IharBury: @TimLovellSmith, logical call context values should be serializable when crossing remote boundaries. (And they do cross remote boundaries during testing -- between the test runner AppDomain and the tested code AppDomain.)

TimLovellSmith: @IharBury Makes sense now @thomaslevesque pointed out we have to use logical call context.

adamralph:
Thanks @thomaslevesque 

I'll pick up the thread over at https://github.com/xbehave/xbehave.net/issues/286

adamralph:
See conversation from https://github.com/FakeItEasy/FakeItEasy/issues/336#issuecomment-181328907 to https://github.com/FakeItEasy/FakeItEasy/issues/336#issuecomment-181367805

TL;DR: we're getting cold feet on using logical call context due to

1. the expectation of all testing scenarios to be flowing it correctly
1. possible bad behaviour in .net 4.0

We may be worrying too much about point 1) - if we manage to find a solution for xbehave then at least we'll know more about how difficult it is too support generally.

Regarding point 2) - we need to determine what the issue would be in .net 4.0 before deciding if it's acceptable or, if not, deciding to drop .net 4.0 support in favour of 4.5+.

thomaslevesque:
> We may be worrying too much about point 1) - if we manage to find a solution for xbehave then at least we'll know more about how difficult it is too support generally.

Regarding this point, I think it's unusual to run steps of the same logical scenario in different threads. Is there a specific reason for xBehave to do that ?

> Regarding point 2) - we need to determine what the issue would be in .net 4.0 before deciding if it's acceptable or, if not, deciding to drop .net 4.0 support in favour of 4.5+ only.

Do you mean completely dropping 4.0 support in FakeItEasy, or just the FakeScope feature? Abandoning 4.0 completely seems a bit extreme...

blairconrad:
> Do you mean completely dropping 4.0 support in FakeItEasy, or just the FakeScope feature? Abandoning 4.0 completely seems a bit extreme...

Or just threadsafety of Scopes? That would be less drastic, but also confusing.

adamralph:
> Regarding this point, I think it's unusual to run steps of the same logical scenario in different threads. Is there a specific reason for xBehave to do that ?

It's actually xunit which is doing this. It sets up a sync context which uses worker threads to run each test and, in the context of xbehave, it sees each each step as a separate test. Given that steps are all async (even if you don't mark them as such) this doesn't actually make any difference, since each step is effectively liable to be run on a different thread anyway, but it would make sense for them to be run in the same logical call context.

Regarding scopes and 4.0/4.5 - I initially meant dropping .net 4 completely, but the idea of only making scopes threadsafe 4.5 is also a possibility...

blairconrad:
> Regarding scopes and 4.0/4.5 - I initially meant dropping .net 4 completely, but the idea of only making scopes threadsafe 4.5 is also a possibility…

I'm not savvy with this sort of thing, but if I do understand things, and read the article correctly, then it may not even be as dire as "FakeItEasy for .NET 4.0 doesn't have thread-safe Scopes". It may be "FakeItEasy for .NET 4.0 doesn't have thread-safe scopes when run in a .NET 4.0 context, but would when run under 4.5+", so even if there's a problem, users might have the option of having their tests run under 4.5 (via configuration, even), and they'd get threadsafe Scopes.

thomaslevesque:
> It may be "FakeItEasy for .NET 4.0 doesn't have thread-safe scopes when run in a .NET 4.0 context, but would when run under 4.5+", so even if there's a problem, users might have the option of having their tests run under 4.5 (via configuration, even), and they'd get threadsafe Scopes, no?

You're right, I keep forgetting that. It's likely that most developer machines have .NET 4.5 installed. However it's not necessarily true for CI servers...

blairconrad:
Good point about CI servers. We need more than .NET 4.5 _installed_, though. If 4.0 were also present, it would be preferred, without config to use 4.5, no?

thomaslevesque:
> If 4.0 were also present, it would be preferred, without config to use 4.5, no?

No, because 4.5 is an in-place upgrade to 4.0

blairconrad:
Oh, man. I should just stop talking. All I do is embarrass myself.

TimLovellSmith:
@blairconrad Not everything can be obvious to everyone all the time!

blairconrad:
Thanks, @TimLovellSmith. You made my day.

blairconrad:
I've not been following all of the finer points of the exploration of this PR and the ramifications for xBehave, but let me try to sum up where I think we are so we can decide what to do next.

1. This PR attempts to make Fake Scopes threadsafe by recording the current scope in a logical call context, which is propagated to async operations.
2. The PR breaks a test, which reveals an underlying assumption we had, that all test steps would be enrolled in the same logical call context. As we've seen, xBehave does not work that way. It's not clear that xBehave can be changed to ensure that the logical call context is propagated. And even if it were, we've no guarantee that other frameworks would do so.
3. Even if the test frameworks propagated the logical call context, there would still be a problem under .NET 4.0, because the logical call context in that version does not have copy-on-write semantics, so various threads can pollute one another's call contexts.

Is this accurate? Complete?

IharBury:
@blairconrad, seems accurate & complete

adamralph:
@blairconrad points 1 and 2 mirror my understanding exactly. I'm still not sure about point 3. It seems fairly certain that this approach [won't work](https://github.com/xbehave/xbehave.net/issues/286#issuecomment-182260467) under .NET 4.0 but I'm not sure I understand exactly why. Perhaps it is indeed as simple as you state.

Regarding point 2, I polled opinion both on Twitter and GitHub on this, and the results are varied.

Damian Hickey points out that [use of ambient state](https://twitter.com/randompunter/status/697361241636003842) is the root problem here and suggests that this is a scenario [not worth supporting](https://twitter.com/randompunter/status/697363039033651201). Daniel Marbach agrees that this is a [can of worms](https://github.com/xbehave/xbehave.net/issues/286#issuecomment-182330436).

Brad Wilson, on the other hand, seems to think that [we can support this](https://github.com/xbehave/xbehave.net/issues/286#issuecomment-182638989) without getting too much involved in the manipulation of the call context but I'm not convinced. Clearly, it doesn't work in xbehave, and [*something* needs doing](https://github.com/xbehave/xbehave.net/issues/286#issuecomment-184550951).

adamralph:
@IharBury we've decided to take a different route here and remove scopes all together, see https://github.com/FakeItEasy/FakeItEasy/issues/484#issuecomment-186327238

We appreciate the effort you put into this, and it was the driving force for us to investigate the implications of trying to keep scopes using logical call context (the official replacement for thread statics in the async await world). Thank you very much.

PR #576, opened 2016-01-13
Make properties return dummies

### LINKED ISSUE #156: Make unconfigured get properties return Dummies, as unconfigured methods do
I would've expected (and I assume other users would too) unconfigured methods that return values and unconfigured get properties on fakes to return the same kinds of values. Not so!

If a getter's type is fakeable, we'll invoke the `AutoFakePropertyRule` (if not, then `DefaultReturnValueRule`). For a method, we fall back to `DefaultReturnValueRule`.

AutoFakePropertyRule will return a value obtained from `DefaultFakeAndDummyManager.CreateFake`, but DefaultReturnValueRule will return one from `DefaultFakeAndDummyManager.TryCreateDummy`.

In a lot of circumstances, the values will be the same, but if there's custom Dummy creation configured (either by a user or FakeItEasy itself), there will be different results. Let's look at an example!

```C#
[TestFixture]
public class DefaultPropertyMethodBehaviorTests
{
    public interface ILibrary
    {
        Movie FirstMovieOnShelfProperty { get; }
        Movie FirstMovieOnShelfMethod();
    }

    [Test]
    public void Should_return_same_type_for_unconfigured_method_and_property_getter()
    {
        var fakeLibrary = A.Fake<ILibrary>();

        var movieFromProperty = fakeLibrary.FirstMovieOnShelfProperty;
        var movieFromMethod = fakeLibrary.FirstMovieOnShelfMethod();

        Assert.That(movieFromProperty.GetType(), Is.EqualTo(movieFromMethod.GetType()));
    }

    public class Movie
    {
        public virtual string Title { get; set; }
    }

    public class DummyMovie: DummyDefinition<Movie>
    {
        protected override Movie CreateDummy()
        {
            return new Movie {Title = "Scott Pilgrim vs. the World"};
        }
    }
}
```

Gives us

```
Expected: <MethodPropertyTests.DefaultPropertyMethodBehaviorTests+Movie>
  But was:  <Castle.Proxies.MovieProxy>

at NUnit.Framework.Assert.That(Object actual, IResolveConstraint expression, String message, Object[] args)
at MethodPropertyTests.DefaultPropertyMethodBehaviorTests.Should_return_same_type_for_unconfigured_method_and_property_getter() in DefaultPropertyMethodBehaviorTests.cs: line 36 
```

The property gave us a Fake and the method gave us a Dummy (I think is preferable).

Note that I resorted to using a `DummyDefinition` to show the behaviour, but the same kind of thing happens when the return type is a `Task` or `Task<T>` (due to custom Tasky dummy handling added a while ago).

(I came across this when I was trying to write the the [Default fake behavior](../wiki/Default-fake-behavior) article and was experimenting. The bug is a big part of the reason it still says _"The actual rules for generating the return values are quite complicated, but here are some sample behaviours"_. It'd be convenient to be able to say "returns a Dummy".)

blairconrad:
Fixes #156.

### DIFF
--- a/Source/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
+++ b/Source/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
@@ -7,44 +7,115 @@
     public static class UnconfiguredFakeSpecs
     {
         [Scenario]
-        public static void VirtualMethod(
+        public static void VirtualMethodCalledDuringConstruction(
             MakesVirtualCallInConstructor fake)
         {
-            "when faking a class with a virtual method"
+            "Given a type with a default constructor"
+                .x(() => { }); // see MakesVirtualCallInConstructor
+
+            "And the constructor calls a virtual method"
+                .x(() => { }); // see MakesVirtualCallInConstructor.ctor()
+
+            "And the method returns a non-default value"
+               .x(() => { }); // see MakesVirtualCallInConstructor.VirtualMethod()
+
+            "When I create a fake of the type"
                 .x(() => fake = A.Fake<MakesVirtualCallInConstructor>());
 
-            "it should return a default value when the method is called during the constructor"
+            "Then the method will return a default value"
                 .x(() => fake.VirtualMethodValueDuringConstructorCall.Should().Be(string.Empty));
 
-            "it should return a default value when the method is called after the constructor"
-                .x(() => fake.VirtualMethod("call after constructor").Should().Be(string.Empty));
-
-            "it should record the method call during the constructor"
+            "And the method call will be recorded"
                 .x(() => A.CallTo(() => fake.VirtualMethod("call in constructor")).MustHaveHappened());
+        }
 
-            "it should record the method call after the constructor"
+        [Scenario]
+        public static void VirtualMethodCalledAfterConstruction(
+            MakesVirtualCallInConstructor fake,
+            string result)
+        {
+            "Given a type with a virtual method"
+                .x(() => { }); // see MakesVirtualCallInConstructor
+
+            "And the method returns a non-default value"
+                .x(() => { }); // see MakesVirtualCallInConstructor.VirtualMethod
+
+            "And a fake of that type"
+                .x(() => fake = A.Fake<MakesVirtualCallInConstructor>());
+
+            "When I call the method"
+                .x(() => result = fake.VirtualMethod("call after constructor"));
+
+            "Then it will return a default value"
+                .x(() => result.Should().Be(string.Empty));
+
+            "And the method call will be recorded"
                 .x(() => A.CallTo(() => fake.VirtualMethod("call after constructor")).MustHaveHappened());
         }
 
         [Scenario]
-        public static void VirtualProperties(
+        public static void VirtualPropertiesCalledDuringConstruction(
             FakedClass fake)
         {
-            "when faking a class with virtual properties"
+            "Given a type with a default constructor"
+                .x(() => { }); // see FakedClass
+
+            "And the constructor calls some virtual properties"
+                .x(() => { }); // see FakedClass.ctor()
+
+            "And the properties return non-default values"
+               .x(() => { }); // see FakedClass.StringProperty, FakedClass.ValueTypeProperty
+
+            "When I create a fake of the type"
                 .x(() => fake = A.Fake<FakedClass>());
 
-            "it should return a default value when a reference type property is called during the constructor"
+            "Then the reference-type property will return a default value"
                 .x(() => fake.StringPropertyValueDuringConstructorCall.Should().Be(string.Empty));
 
-            // The test for the default value of a value type is a regression test for https://github.com/FakeItEasy/FakeItEasy/issues/368:
-            "it should return a default value when a value type property is called during the constructor"
+            "And the value-type property will return a default value"
                 .x(() => fake.ValueTypePropertyValueDuringConstructorCall.Should().Be(0));
+        }
 
-            "it should return the value assigned during constructor when a reference type property is gotten after the constructor"
-                .x(() => fake.StringProperty.Should().Be("value set in constructor"));
+        [Scenario]
+        public static void VirtualReferenceTypeProperty(
+            FakedClass fake,
+            string result)
+        {
+            "Given a type with a default constructor"
+                .x(() => { }); // see FakedClass
 
-            "it should return the value assigned during constructor when a value type property is gotten after the constructor"
-                .x(() => fake.ValueTypeProperty.Should().Be(123456));
+            "And the constructor assigns a value to a virtual reference-type property"
+                .x(() => { }); // see FakedClass.ctor()
+
+            "And a fake of that type"
+                .x(() => fake = A.Fake<FakedClass>());
+
+            "When I fetch the property value"
+                .x(() => result = fake.StringProperty);
+
+            "Then it will be the value assigned during construction"
+                .x(() => result.Should().Be("value set in constructor"));
+        }
+
+        [Scenario]
+        public static void VirtualValueTypeProperty(
+            FakedClass fake,
+            int result)
+        {
+            "Given a type with a default constructor"
+                .x(() => { }); // see FakedClass
+
+            "And the constructor assigns a value to a virtual value-type property"
+                .x(() => { }); // see FakedClass.ctor()
+
+            "And a fake of that type"
+                .x(() => fake = A.Fake<FakedClass>());
+
+            "When I fetch the property value"
+                .x(() => result = fake.ValueTypeProperty);
+
+            "Then it will be the value assigned during construction"
+                .x(() => result.Should().Be(123456));
         }
 
         public class FakedClass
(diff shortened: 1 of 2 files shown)


Source/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
```diff
@@ -6,23 +6,28 @@
...
         [Scenario]
         public static void VirtualMethod(
             MakesVirtualCallInConstructor fake)
         {
-            "when faking a class with a virtual method"
+            "When faking a class with a virtual method"
                 .x(() => fake = A.Fake<MakesVirtualCallInConstructor>());
 
-            "it should return a default value when the method is called during the constructor"
+            "It should return a default value when the method is called during the constructor"
```
adamralph: Whilst we're here, shall we change the language of this test to "Given, When, Then"?

Source/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
```diff
@@ -34,17 +28,11 @@ public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
...
-                {
-                    theValue = DefaultReturnValue(fakeObjectCall);
-                }
-
                 var newRule = new CallRuleMetadata
                                   {
                                       Rule = new PropertyBehaviorRule(fakeObjectCall.Method, FakeManager)
                                       {
-                                          Value = theValue,
+                                          Value = DefaultReturnValue(fakeObjectCall),
```
adamralph: Wow, so simple! :heart:

blairconrad: It did my :heart: good too.

Source/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
blairconrad: Oh, bother. My intent, but I really dropped the ball!
Thanks. Stay tuned.

blairconrad: Actually, I struggle with this when some of the steps appear to be implicit.

* Given an unconfigured fake that calls a virtual method during its constructor
* _I don't know that I have a `When`_.
* Then the method will return a default value
* And the method call will be recorded

* Given an unconfigured fake
* When I call a virtual method
* Then the method will return a default value
* And the method call will be recorded
?

blairconrad: Okay. I tried a thing.

adamralph: I agree that this should be split into two scenarios.

In the case of the first scenario, I think we have three choices:

1. Construct the type via reflection (or Roslyn!) in a "Given" and "And"s (and create the fake similarly)
2. Declare the type and have an empty "Given" and "And"s
3. Declare the type and skip the "Given" and "And"s

In C#, it is way more convenient to declare the types up front rather than construct them reflectively,  and the whole scenario will look very different from the usage of the API we want to test, so we probably want to skip option 1 (this may not be the case in other languages). Option 2 is best in terms of the narrative but some people may baulk at having empty steps. Option 3 omits the empty step(s), but can result in a rather wordy When.

E.g.

```Gherkin
# Options 1 and 2
Scenario: A constructor calls a virtual method
  Given a type with a default constructor
    And the constructor calls a virtual method
    And the method returns a non-default value
   When I create a fake of the type
   Then the method returns a default value
    And the call is recorded

# Option 3
Scenario: A constructor calls a virtual method
   When I create a fake of a type with a default constructor which calls a virtual method which returns a non-default value
   Then the method returns a default value
    And the call is recorded
````

For the second scenario, I'd probably go with something like
```Gherkin
Scenario: A virtual method
  Given a fake with a virtual method
   When I call the method
   Then a default value is returned
    And the call is recorded
```

Forming these narratives can be quite an art form. I find it one of the most challenging, nevertheless interesting and valuable exercises in development. It really forces you drive out what it is you are testing and often yields invaluable feedback which can have a huge influence on the user story/feature itself.

adamralph:
@blairconrad did you see this comment? https://github.com/FakeItEasy/FakeItEasy/pull/576#discussion_r49584676

Our comments crossed each other in transit and the were marked as outdated by the updated commits.

blairconrad:
@adamralph I did not see https://github.com/FakeItEasy/FakeItEasy/pull/576#discussion_r49584676. Glad you pointed it out. I'll skim now, but at work. And who wrote the book this time? :grin:

Source/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
blairconrad: For the second scenario, I think my `VirtualMethodCalledAfterConstruction` case is something like your example. The grammar is slightly different, and I don't mind rewording to exactly what you have if you don't like what's there now.

blairconrad: Oh, and for the first scenario, I nearly had your "Option 3" (although I never specified that the implementation of the virtual method returns a non-default value). 

Do you have a preference between Option 2 and 3? (I was going to argue against Option 1 for the reasons you did).

blairconrad:
I continued commenting inline, for good or for ill…

Source/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
adamralph: Your second scenario looks fine, no need to reword.

I don't really have a preference between options 2 and 3. One hand, a bunch of empty steps is verbose and weird, on the other hand a When which so long that it scrolls off the side of the GitHub UI isn't very desirable.

adamralph:
Me too :smile:

Source/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
blairconrad: > on the other hand a When which so long that it scrolls off the side of the GitHub UI isn't very desirable

Oh. Is that your concern?
"Given an unconfigured fake that assigns values to a virtual reference type property during construction"

adamralph: @blairconrad no, it's not the primary concern, just one effect of it being so long :smile: 

I don't mind the one long step, but I'll bike shed this once more and say that I think it should be a "When".

"Given" sets up preconditions, "When" invokes the system under test and "Then" assesses the outcome. In this case I''d argue the preconditions are provided by type declaration and our invocation is creating the fake.

blairconrad: Yeah, I was kidding about the length/breaks.

I understand your points about G/W/T.
If we didn't mind empty steps, though, I'd consider keeping everything above in the Given and then the When could be `"When I make an unconfigured Fake of the type"`…

adamralph: I don't mind too much either way. Toss a coin :wink: 

Why do you feel the need to include "unconfigured"? Because there is no `FakeConfigurator` hiding somewhere for it?

blairconrad: Not just the lack of `FakeOptionsBuilder` (see what I did there?), but there's also no `options => options.ConfigureFake` in there. I'm attempting to highlight the fact that this is the default behaviour. If you think that's better assumed, then I'm happy to omit.

adamralph: No strong feeling about it but I think it can be left out. If the step just says "When I create a fake of the type" then expect it to do exactly that and nothing more. I don't think we need to call out what we are *not* doing.

blairconrad: Cool. I'll push something again tonight. With luck I'll be able to resolve everything in this discussion into a coherent whole.

adamralph: how about

```C#
"Given a type with a default constructor"
   .x(() => { }); // MakesVirtualCallInConstructor

"And the constructor calls a virtual method"
   .x(() => { }); // MakesVirtualCallInConstructor.ctor()

"And the method returns a non-default value"
   .x(() => { }); // MakesVirtualCallInConstructor.VirtualMethod()
```

adamralph: Updated. Added "see ...".

blairconrad: :shipit:

blairconrad:
I've updated the specs according to the conversation with @adamralph. Now that I see it, I worry that it's a little wordy. The empty steps seem a little silly. Still, nothing's flowing off the side of my screen…

adamralph:
We can revisit the silliness again later if we want :smile:

blairconrad:
Yeah, I'm content to try the specs out like this for a while and see how we feel.

PR #581, opened 2016-01-23
Added EditorBrowsable attributes to block Equals from showing in inte…

### LINKED ISSUE #580: Make That hide object members
(Raised by @dealproc in a private chat.)

… probably by extending `IHideObjectMembers`?

Otherwise we see this in the IDE:

image

Then you accidentally pick `Equals` instead of `IsEqualTo`, and, to quote @dealproc without his permission

> that's not cool the first 20 times you do it.

dealproc:
…llisense.  

Fixes #580


blairconrad:
Updated description to read "Fixes #580" so merging will auto-close #580.

Source/FakeItEasy/Core/DefaultArgumentConstraintManager.cs
```diff
@@ -23,6 +23,12 @@ public T Matches(Func<T, bool> predicate, Action<IOutputWriter> descriptionWrite
             return default(T);
         }
 
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
thomaslevesque: Putting the attribute in this class will have no effect, because it's internal. Putting it in the interface should be enough.

blairconrad:
Thanks, @dealproc, this does hide `Equals`. At the risk of looking a gift horse in the mouth, and then asking for a slightly nicer horse, I'd expected to see `IArgumentConstraintManager<T>` extend `IHideObjectMembers`, which would've hidden 
* `Equals`, 
* `ToString`,
* `GetHashCode`, and
* `GetType`

I know that `Equals` was the only one that was tripping you up, but hiding the others does make for a nicer fluent-interface-typing experience. Have you any interest in amending the commit?

blairconrad:
As a bonus, I think using `IHideObjectMembers` means we only have to update the one file…

thomaslevesque:
There's already a `IHideObjectMembers` interface for this purpose. All you need to do is make `IArgumentConstraintManager<T>` inherit from it.

thomaslevesque:
Ah, @blairconrad beat me to it ;)

blairconrad:
@thomaslevesque, duelling reviews. Fun, eh? Although in a sense, I got to it 6 hours before you did, since I'm in the west. Of course, that means that I got to it slightly before @dealproc opened the PR. How's that for service?

thomaslevesque:
> Of course, that means that I got to it slightly before @dealproc opened the PR. How's that for service?

Not bad ;)

dealproc:
Crap.  I didn't realize that there was an interface that did just this.  give me a few, as I'm setting up another dev image.  once that's complete, i'll review and update the system to do as you have doc'd.  Just thought since I was the :cry: on this, that I'd offer a bit of time to relieve it :)

thomaslevesque:
> Just thought since I was the :cry: on this, that I'd offer a bit of time to relieve it :)

And it's much appreciated, thanks!

dealproc:
ok... so not paying attention, there is an ugly check-in to ignore.  the last one uses the IHideObjectMembers in ~/Configuration and not another incantation of the interface.

Source/FakeItEasy/Core/DefaultArgumentConstraintManager.cs
```diff
@@ -1,7 +1,7 @@
 ﻿namespace FakeItEasy.Core
 {
     using System;
-
+    using System.ComponentModel;
```
thomaslevesque: No longer necessary, should be rolled back

Source/FakeItEasy/FakeItEasy.csproj
```diff
@@ -87,6 +87,7 @@
       <SubType>Code</SubType>
     </Compile>
     <Compile Include="A.of.T.cs" />
+    <Compile Include="Configuration\IHideObjectMembers.cs" />
```
thomaslevesque: Please roll this back as well (since the change doesn't actually do anything)

thomaslevesque:
@dealproc since there's only one useful change, the best way to clean up the PR is probably to reset your branch to its original state (`git reset --hard master`) and reapply the changes in `IArgumentConstraintManager<T>` (then force push the branch to replace the old commits)

dealproc:
see what i can do

Source/FakeItEasy/IArgumentConstraintManager.cs
```diff
@@ -1,7 +1,7 @@
 ﻿namespace FakeItEasy
 {
+    using Configuration;
```
blairconrad: Introducing this `using` before the "System" ones has created a warning about using order:

http://teamcity.codebetter.com/viewLog.html?buildId=208583&buildTypeId=bt929&tab=buildLog#_state=50,122&focus=125

(you can see it when you build in the IDE or command line as well).

If you're not sick of us nitpicking yet, would you consider moving this using below line 5?
If you are, I can submit a superseding PR that would do it. You'd still get credit for your commit.

blairconrad:
Thanks, @dealproc. I think it looks very good, except for one build warning that was introduced that I commented on as a "line note".

dealproc:
Oops... no worries.  I'll get that in place.  I usually do organize things in alpha order, so if there's a way to better setup the project to respect these settings, I'm all ears.

blairconrad:
Oh, I don't know of any in-IDE settings that help with this. It's StyleCop that gives us [the rule](http://www.stylecop.com/docs/SA1208.html), although it sounds like it would fight with [another rule](http://www.stylecop.com/docs/SA1210.html), they seem to have worked it out.

I think Resharper orders things this way as well. Not sure what other tools can help. I can't find a CodeMaid setting right off...

blairconrad:
@dealproc, how much will you hate me if I point out that you've [introduced tabs instead of spaces](http://teamcity.codebetter.com/viewLog.html?buildId=208723&buildTypeId=bt929&tab=buildLog#_state=50,122&focus=125) into the source?

Also, in such a case where a second commit exists only to correct an error in a previous commit, it's better practice to squash the two commits into 1?

And (and this is our doing), `master` shifted underneath you when #574 was merged. There's no conflict, so we could still merge your eventual change, but since I'm already complaining, maybe the best thing would be to follow @thomaslevesque's path earlier and

1. switch to master
2. pull from upstream
3. switch back, and then
4. "reset your branch to its original state (`git reset --hard master`) and reapply the changes in IArgumentConstraintManager<T> (then force push the branch to replace the old commits)"

An alternative would be just to fix the tabs on your branch, make a new commit (or amend the previous one) and then `rebase -i master` on your branch and [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) everything into the first commit before force pushing.

Again, I'm not trying to drive you away with my nitpicking. Just think how smoothly #583 will go once we're through this one!

adamralph:
:+1: @blairconrad

thomaslevesque:
Uh.. I'm not sure what happened, but the history is all messed up and now the PR contains all the changes from #574 (check the "Files changed" tab)... The history looks like this:

gitk screenshot

Did you try to reset the branch as I explained above?

adamralph:
A messed up rebase would have caused this to happen.

adamralph:
For such a small change it would be easier to just discard the branch and start again.

adamralph:
Thanks @dealproc, these commits were cherry-picked and combined into #595 

Look out for your name in the release notes! :trophy:

Issue #279, opened 2014-03-27
rename Source folder to src


adamralph:
Any objections? 'src' is the common convention I see in almost all OSS projects including others which I own.

At the same time I'd like to change the casing of the transient 'Build' folder to 'build'.

blairconrad:
Closed via #598.

PR #587, opened 2016-01-27
Add .editorconfig file

### LINKED ISSUE #586: Add a .editorconfig
for use with http://editorconfig.org/

Some potential contributors struggle with formatting issues, specifically around whitespace.
In some cases, these are easily worked around, and sometimes not.

The presence of a `.editorconfig` file may help avoid some problems. Not everybody is using an editor (or has the plugin) that will respect these settings, but if even a few do, it's no great hardship to add and maintain the file.

Thoughts?

thomaslevesque:
Fix #586

### DIFF
--- a/Source/.editorconfig
+++ b/Source/.editorconfig
@@ -1,0 +1,23 @@
+﻿# EditorConfig is awesome: http://EditorConfig.org
+
+# Top-most EditorConfig file
+root = true
+
+[*]
+# Windows-style newlines
+end_of_line = crlf
+
+# 4-column space indentation
+indent_style = space
+indent_size = 4
+
+charset = utf-8
+trim_trailing_whitespace = true
+
+max_line_length = 160
+
+insert_final_newline = true
+
+[*.rb]
+# 2-column space indentation
+indent_size = 2
--- a/Source/FakeItEasy.sln
+++ b/Source/FakeItEasy.sln
@@ -5,6 +5,7 @@ VisualStudioVersion = 12.0.31101.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{A6A7B553-5359-4FB9-B19C-8A23004F49DB}"
 	ProjectSection(SolutionItems) = preProject
+		.editorconfig = .editorconfig
 		CommonAssemblyInfo.cs = CommonAssemblyInfo.cs
 		FakeItEasy.Dictionary.Tests.xml = FakeItEasy.Dictionary.Tests.xml
 		FakeItEasy.Dictionary.xml = FakeItEasy.Dictionary.xml


blairconrad:
https://github.com/FakeItEasy/FakeItEasy/blob/22796eb6f110857d24483f8858fac6960722bcf9/CONTRIBUTING.md#spaces-not-tabs

mentions a 2-space indent for the rakefile. You know, while we're here.

Source/.editorconfig
```diff
@@ -0,0 +1,12 @@
+﻿; Top-most EditorConfig file
```
blairconrad: I'm new to this, but I gotta say—I was excited about seeing

```
# EditorConfig is awesome: http://EditorConfig.org
```

Source/.editorconfig
```diff
@@ -0,0 +1,12 @@
+﻿; Top-most EditorConfig file
+root = true
+
+[*.cs]
+; Windows-style newlines
+end_of_line = crlf
+; 4-column space indentation
```
blairconrad: super nitpicky, but would anybody mind a newline between 6 and 7?

blairconrad:
Generally I like it. Nice job!

Source/.editorconfig
blairconrad: :stuck_out_tongue_winking_eye:

thomaslevesque:
> https://github.com/FakeItEasy/FakeItEasy/blob/22796eb6f110857d24483f8858fac6960722bcf9/CONTRIBUTING.md#spaces-not-tabs
> 
> mentions a 2-space indent for the rakefile. You know, while we're here.

It also says that it should be 4 spaces for all file types (except Ruby), so I should probably apply this to `[*]`

Source/.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+# 4-column space indentation
+indent_style = space
+indent_size = 4
+
+charset = utf-8
+trim_trailing_whitespace = true
+
+[*.cs]
+# Windows-style newlines
+end_of_line = crlf
```
blairconrad: A while back, @adamralph commented about me missing a newline in a file. 
Ah: https://github.com/FakeItEasy/FakeItEasy/pull/566#discussion_r47742056

which was my introduction to .editorconfig. 

Do we want to take that step as well? I've no objection.

thomaslevesque: I don't mind either. Will add it.

thomaslevesque:
I also set the `max_line_length` to 160, as mentioned in the contribution guidelines

Source/.editorconfig
```diff
@@ -0,0 +1,25 @@
+﻿# EditorConfig is awesome: http://EditorConfig.org
```
blairconrad: Ha! You did it! I was joking. Mostly. I do think it's fun. But I'm quite silly.

thomaslevesque: Well, I guess it can't hurt to include the link, in case someone sees this file and wonders what it is ;)

blairconrad:
Looks like that's only supported by vi and emacs, but shouldn't do any harm. I wouldn't think.

blairconrad:
Looks good to me. I'm for merging, but I've never used editorconfig, so wouldn't mind a chime-in from an experienced user.

Nice job, @thomaslevesque!

Source/.editorconfig
blairconrad: True. But even aside from that, it makes me smile.

thomaslevesque:
I just squashed the commits; no point in having 4 distinct commits for something that's new...

blairconrad:
Yeah, I noticed the superfluous commits, but I assumed you were going to squash. Because you're like that. :+1:

Source/.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+charset = utf-8
+trim_trailing_whitespace = true
+
+max_line_length = 160
+
+insert_final_newline = true
+
+[*.cs]
+# Windows-style newlines
+end_of_line = crlf
```
adamralph: I would probably move this line into the `[*]` section and kill the `[*.cs]` section. If we ever add any file types which need LF, such as `.sh` then we can add a section for them.

Source/.editorconfig
```diff
@@ -0,0 +1,25 @@
...
+
+[*.rb]
+# 2-column space indentation
+indent_style = space
```
adamralph: This line is redundant and can be removed. It is covered in the `[*]` section.

adamralph:
@thomaslevesque I've reviewed the PR and added a couple of comments. If you agree, I'd be happy if you amended the current commit.

Source/.editorconfig
thomaslevesque: OK, it makes sense

blairconrad:
It just occurs to me. Should .editorconfig not have gone at the root level (not in `Source`)?

adamralph:
`.editorconfig` has to be at the same level as the solution in order for the VS add-in to recognise it. 

Incidentally, I think we should move to this layout - https://gist.github.com/davidfowl/ed7564297c61fe9ab814

thomaslevesque:
> Incidentally, I think we should move to this layout - https://gist.github.com/davidfowl/ed7564297c61fe9ab814

I like it! I don't really have an issue with the current layout, but the one you propose looks cleaner

blairconrad:
> .editorconfig has to be at the same level as the solution in order for the VS add-in to recognise it.

Oh? 

> To load the settings for a particular source file, the plugin searches for a file named ".editorconfig" in the same directory as the source file and in each parent directory.

from the [extension's description](https://visualstudiogallery.msdn.microsoft.com/c8bccfe2-650c-4b42-bc5c-845e21f96328).

Not doubting you, but I'd certainly not expected a requirement to be sibling to the .sln.

adamralph:
I was under the impression that it stops walking up the directory path when hits the .sln level but I could be wrong.

blairconrad:
> Incidentally, I think we should move to this layout - https://gist.github.com/davidfowl/ed7564297c61fe9ab814

Perhaps an update to #279?

Also, I don't love the `.sln` at root, at least not without explanation. What's the reasoning? So it's the `.editorconfig`'s friend?

blairconrad:
> I was under the impression that it stops walking up the directory path when hits the .sln level but I could be wrong.

To the lab!

blairconrad:
I'm back. I moved the .editorconfig to the root directory, set the number of spaces per indent to 7, restarted the IDE, and…
the new setting was respected!

Issue #154, opened 2013-08-02
switch off code analysis during build


adamralph:
Given that code analysis doesn't run on the build server (and requires jumping through several ugly stress inducing hoops to get working in that context), I'm wondering if we should bother running this as part of the build in release config.

The current risk is that we accept a pull request (from owners as well as contributors), everything builds find on the build server and then we update our local master and can't build using rake, since a code analysis violation is present.

The other problem with the current approach is that code analysis slows down the build a lot, and ever time we run with rake, we have to pay this price.

The alternative would be to switch off code analysis in the csproj for Release config and run manually on an ad-hoc basis. This could also be done via a specific rake task which is not part of the default task.

PR #588, opened 2016-01-29
Make unconfigured methods that return type Lazy<T> and Task<T> behave more like methods that return T

### LINKED ISSUE #560: Make unconfigured methods that return type Lazy<T> and Task<T> behave more like methods that return T
Version: 1.25.3
So I noticed today when I tried this framework for the first time that it hangs when the code calls a method with the signature:
`Task<QueueDescription> CreateQueueAsync(string queueName);`
Checking the output I can see:
```
Exception thrown: 'System.MissingMethodException' in mscorlib.dll
Exception thrown: 'System.MissingMethodException' in mscorlib.dll
Exception thrown: 'Castle.DynamicProxy.InvalidProxyConstructorArgumentsException' in FakeItEasy.dll
Exception thrown: 'System.MissingMethodException' in mscorlib.dll
Exception thrown: 'System.MissingMethodException' in mscorlib.dll
Exception thrown: 'Castle.DynamicProxy.InvalidProxyConstructorArgumentsException' in FakeItEasy.dll
Exception thrown: 'System.ArgumentException' in mscorlib.dll
Exception thrown: 'Castle.DynamicProxy.ProxyGenerationException' in FakeItEasy.dll
Exception thrown: 'Castle.DynamicProxy.Generators.GeneratorException' in FakeItEasy.dll
Exception thrown: 'System.ArgumentException' in FakeItEasy.dll
Exception thrown: 'Castle.DynamicProxy.Generators.GeneratorException' in FakeItEasy.dll
Exception thrown: 'Castle.DynamicProxy.Generators.GeneratorException' in FakeItEasy.dll
Exception thrown: 'System.ArgumentException' in FakeItEasy.dll
Exception thrown: 'System.ArgumentException' in FakeItEasy.dll
```

After wasting some time on it the problem seems to be because QueueDescription has an internal constructor, and for some reason FakeItEasy is trying to instantiate an instance rather than just returning null.
Here is a simple way of reproducing the issue:
```c#
[TestMethod]
public async Task Test_BreakIt()
{
    var breaker = A.Fake<IFakeItEasyBreaker>();
    await breaker.BreakIt();
}

public interface IFakeItEasyBreaker
{
    Task<ClassWithInternalConstructor> BreakIt();
}

public class ClassWithInternalConstructor
{
    internal ClassWithInternalConstructor()
    {
    }
}
```

In the steps to reproduce I could of course add:
`[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]`
But it does not help with my actual issue, since QueueDescription is a part of Microsoft.ServiceBus.dll which I don't have much power over.
Then I can of course setup so that i tell it to explicitly return null, but coming from Moq this has never been an issue. The test I got this problem with was a previous Moq test which worked fine with Moq, it returned null as expected.

thomaslevesque:
Fix #560

For `Task<T>` and `Lazy<T>`, if a dummy `T` cannot be created, the task result (or the lazy value) should be `default(T)`


thomaslevesque:
I'm afraid Git didn't realize that I had just renamed AsyncAwaitSpecs.cs to TaskSpecs.cs, so the diff doesn't know what changed. Here's a summary:
- renamed AsyncAwaitSpecs to TaskSpecs
- moved the test interfaces and classes inside the specs class (to match the pattern used elsewhere)
- added two scenarios

blairconrad:
I've taken a first pass, and think the fix (as well as the code in the specs) look good. As noted, I'm just a little unsure about the step descriptions…

thomaslevesque:
While I'm at it, would you mind if I dropped the `FooAsyncAwait` class from `TaskSpecs`? It serves no purpose, it just calls the fake methods, so it only makes the specs more verbose...

blairconrad:
Yeah, I don't see any point in it…

Source/FakeItEasy.Specs/LazySpecs.cs
```diff
@@ -50,5 +70,12 @@ protected override IFoo Create()
                 return instance;
             }
         }
+
+        public class Bar
+        {
+            internal Bar()
```
adamralph: I know this is really picky, specially in a test lib, but we generally avoid `internal` on members and only use it on entire types.

blairconrad: Prepared to be disappointed in my next PR, then!

thomaslevesque: In this case, I can make it private. But in TaskSpecs, it has to be internal, because I'm explicitly calling it in one of the scenarios. However I'm not sure the scenario with a configured method is really useful, I added it only for consistency with the existing specs. It's already covered extensively in other specs and tests.

Source/FakeItEasy.Specs/TaskSpecs.cs
```diff
@@ -0,0 +1,184 @@
...
+
+            Task<int> QueryAsync();
+
+            Task<Foo> GetFooAsync();
+
+            Task<Bar> GetBarAsync();
+        }
+
+        [Scenario]
+        public static void ConfiguredAsyncMethodWithReturnValueOfTypeInt(
```
thomaslevesque: I'm leaning toward removing this scenario and the other `Configured*` ones, because they just test that the method returns the configured value, which is already tested in many other places. What do you think?

thomaslevesque:
I change the wording in the specs, let me know if it's what you expected.

Source/FakeItEasy.Specs/TaskSpecs.cs
blairconrad: I agree.

Source/FakeItEasy.Specs/LazySpecs.cs
```diff
@@ -23,19 +25,37 @@ public interface IFoo
             ILazyFactory fake,
             Lazy<IFoo> lazy)
         {
-            "establish"
+            "Given a fake"
                 .x(() => fake = A.Fake<ILazyFactory>());
 
-            "when calling a method that returns a lazy"
+            "When calling an unconfigured method that returns a lazy of a fakeable type"
```
blairconrad: Technically it's "dummyability" that's the determining factor, not fakeability. That's a bit of a quibble here, since every fakeable type is dummyable. But meet me below.

Source/FakeItEasy.Specs/LazySpecs.cs
```diff
@@ -23,19 +25,37 @@ public interface IFoo
...
 
+        [Scenario]
+        public static void LazyReturnValueOfNonFakeableType(
+            ILazyFactory fake,
+            Lazy<Bar> lazy)
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<ILazyFactory>());
+
+            "When calling a method that returns a lazy of a non-fakeable type"
```
blairconrad: Lots of types aren't fakeable, but you can make a Dummy out of them, so this description is a little misleading, or at least weaker than the real requirement.

blairconrad:
I think the wording is good, although I've picked some nits around the distinction between when types can be faked and when they can be made into dummies.

Source/FakeItEasy.Specs/LazySpecs.cs
thomaslevesque: OK. Shall I write "non-dummyable" then ? It sounds a bit weird ;)

thomaslevesque: Apparently "dummyable" is a real word... at least, it is recognized by ReSpeller. The English language is amazing ;)

thomaslevesque: I made the change to "dummyable", let me know if you prefer something different

Source/FakeItEasy.Specs/TaskSpecs.cs
thomaslevesque: done

Source/FakeItEasy.Specs/LazySpecs.cs
blairconrad: Sorry, I saw this comment earlier and didn't get to the code. I figured you'd already pushed the new commit.
I think dummyable sounds great. Well, at least I think it conveys the intention. StyleCop may think differently.

Yes, English is… quite welcoming of words. You wouldn't see Académie française allowing such an uncouth word!

blairconrad: Oh. You did make the change. Below? You may have missed it on line 49 because of all the exciting conversation?

thomaslevesque: Oops, I made the change in TaskSpecs, and forgot to do it in LazySpecs... I amended the commit.

blairconrad:
I just ran code analysis on the solution for the first time after pulling new master and got a number of violations, largely around "dummyable". We may have to rephrase.
:disappointed:

blairconrad:
"type that cannot be made a Dummy" is descriptive, but hardly brief…

adamralph:
"type that cannot be dummied"?

thomaslevesque:
Ah, I forgot to run the analysis (perhaps we should run it during CI?)

I could rephrase, or just add "dummyable" to `FakeItEasy.Dictionary.Tests.xml`, which seems to be there for that purpose and already contains words like "fakeable". It fixes the spelling warnings.

There are also warnings on `GetFooAsync` and `GetBarAsync`, apparently FxCop wants me to change them to properties if appropriate. Well, I don't think it's appropriate, so I'll just suppress the warnings if you don't mind ;)

blairconrad:
There are apparently difficulties running FxCop during CI (although even if we ran it, if we treated the warnings as warnings, as we do with StyleCop violations, I don't know how much it would help: someone would still have to think to look for the warnings). This the reason we have the item in the release checklist to run the Code Analysis before releasing.

I don't mind owning "dummyable" along with "fakeable".

blairconrad:
All else being equal, I tend to go along with the FxCop changes, because I think the suppressions add a little bit of noise to the files, but `async` properties just don't feel right to me, so keeping them as methods and suppressing the warning sounds good.

blairconrad:
I found the discussion about the difficulties with FxCop on TeamCity: https://github.com/FakeItEasy/FakeItEasy/issues/154#issue-17567738

PR #589, opened 2016-01-29
Apply .editorconfig rules

thomaslevesque:
Using [eclint](https://github.com/jedmao/eclint).

I ran it only on C# files, and I temporarily disabled the `max_line_length` rule during the process, so only these rules were fixed:
- indentation style
- end of line style
- trailing whitespace
- final newline
- encoding

The only issue I encountered was that eclint choked on the `ä` in @patrik-hagne's name, so I fixed it manually. Other than that, the tool looks solid and can be used again in the future (if you don't mind installing Node.js)

Also renamed `Fake(T)Tests.cs` to `Fake.of.T.Tests.cs`, because the command line had trouble with the parentheses, and for consistency with similar files


blairconrad:
We had generated files that were running afoul of StyleCop settings?
Did applying the editorconfig settings exacerbate this problem?

thomaslevesque:
> We had generated files that were running afoul of StyleCop settings?
> Did applying the editorconfig settings exacerbate this problem?

Actually, the problem appeared only after applying the editorconfig settings... I don't know where these files come from (they're in the obj directory), or why StyleCop doesn't ignore them by default. Also, they're all empty, except for a blank line... so StyleCop complains that a code file cannot start with a blank line.

thomaslevesque:
btw, this should probably be merged before #588, which is small and easier to rebase.

blairconrad:
Oddly, I can't replicate the StyleCopGenerated file problem, but I believe you!
I see we're waiting on build results from TC, but I am happy for this to merge. I'll check back later.

thomaslevesque:
> I see we're waiting on build results from TC, but I am happy for this to merge. I'll check back later.

We're also waiting for the build on #588, but I think there's an issue with TC, because the build _is_ complete... it just didn't notify GitHub

adamralph:
ahhh... my StyleCop OCD has never felt so satisfied... :relaxed:

PR #592, opened 2016-01-31
Rationalize priority values

### LINKED ISSUE #525: Standardize extension points' Priority property values and policy
Each of `IArgumentValueFormatter`, `IDummyFactory`, and `IFakeOptionsBuilder` has an integer `Priority` property that's used to choose the appropriate class when more than one such extension point applies to a particular argument/Dummy/Fake type.

The higher `Priority` wins.

`FakeConfigurator` and  `DummyFactory` return `0` as a default value, but `ArgumentValueFormatter` returns `int.MinValue`.

Currently, only `IArgumentValueFormatter` provides concrete implementations which are used by FakeItEasy, but with recent increases in the power of `IDummyFactory` and `IFakeOptionsBuilder`, we may want to implement some of FakeItEasy's features using these extension points.

To avoid confusion, and collisions with user's implementations, I propose that:

1. `FakeOptionsBuilder`, `DummyFactory`, and `ArgumentValueFormatter` all return `0` as default values for `Priority` (users can always override if they need to)
2. FakeItEasy's concrete implementations of the interfaces return very low negative values, in the `int.MinValue` range. **update: we decided on -1**
3. We document the default value of the abstract helper classes and reserve all negative `Priority` values for FakeItEasy's internal use.

blairconrad:
Fixes #525.
Using a custom Priority type to ensure clients can only make non-negative priorities.

### DIFF
--- a/Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
+++ b/Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
@@ -154,7 +154,7 @@ namespace FakeItEasy.Tests.Core
 
             // Act
             var typesWithNonNegativePriority = allArgumentValueFormatters
-                .Where(f => f.Priority >= 0)
+                .Where(f => f.Priority >= new Priority(0))
                 .Select(f => f.GetType());
 
             // Assert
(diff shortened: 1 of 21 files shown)


Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
```diff
@@ -139,7 +141,28 @@ public void Should_prefer_exact_type_formatter_to_interface_formatter()
             var result = this.formatter.GetArgumentValueAsString("string value");
 
             // Assert
-            Assert.That(result, Is.EqualTo("a string"));
+            result.Should().Be("a string");
+        }
+
+        [Test]
+        public void Provided_formatters_should_have_negative_priority()
```
blairconrad: I would've used a test case source, but if there are no defined formatters, the test registers as inconclusive.
Same for the dummy factory and options builder.

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
```diff
@@ -149,16 +172,16 @@ private void AddTypeFormatter<T>(string formattedValue)
...
 
-        private void AddTypeFormatter(Type type, string formattedValue, int priority)
+        private void AddTypeFormatter(Type type, string formattedValue, short priority)
         {
-            var formatter = A.Fake<IArgumentValueFormatter>();
-            A.CallTo(() => formatter.ForType).Returns(type);
-            A.CallTo(() => formatter.GetArgumentValueAsString(A<object>._)).Returns(formattedValue);
-            A.CallTo(() => formatter.Priority).Returns(priority);
-            this.registeredTypeFormatters.Add(formatter);
+            var newFormatter = A.Fake<IArgumentValueFormatter>();
```
blairconrad: it conflicted with a field name

blairconrad:
spotted a problem. Will be back.

PR #593, opened 2016-01-31
Rationalize priority values

### LINKED ISSUE #525: Standardize extension points' Priority property values and policy

blairconrad:
Fixes #525.
Using a custom Priority type to ensure clients can only make non-negative priorities.

### DIFF
--- a/Source/FakeItEasy.Specs/ArgumentValueFormatterSpecs.cs
+++ b/Source/FakeItEasy.Specs/ArgumentValueFormatterSpecs.cs
@@ -9,7 +9,7 @@
         [Scenario]
         public static void GenericArgumentValueFormatterDefaultPriority(
             IArgumentValueFormatter formatter,
-            int priority)
+            Priority priority)
         {
             "Given an argument value formatter that does not override priority"
                 .x(() => formatter = new SomeArgumentValueFormatter());
@@ -17,8 +17,8 @@
             "When I fetch the Priority"
                 .x(() => priority = formatter.Priority);
 
-            "Then it should be 0"
-                .x(() => priority.Should().Be(0));
+            "Then it should be the default priority"
+                .x(() => priority.Should().Be(Priority.Default));
         }
 
         [SuppressMessage("Microsoft.Performance", "CA1812:AvoidUninstantiatedInternalClasses", Justification = "Required for testing.")]
--- a/Source/FakeItEasy.Specs/ArgumentValueFormatterSpecs.cs
+++ b/Source/FakeItEasy.Specs/ArgumentValueFormatterSpecs.cs
@@ -1,0 +1,37 @@
+﻿namespace FakeItEasy.Specs
+{
+    using System.Diagnostics.CodeAnalysis;
+    using FluentAssertions;
+    using Xbehave;
+
+    public static class ArgumentValueFormatterSpecs
+    {
+        [Scenario]
+        public static void GenericArgumentValueFormatterDefaultPriority(
+            IArgumentValueFormatter formatter,
+            int priority)
+        {
+            "Given an argument value formatter that does not override priority"
+                .x(() => formatter = new SomeArgumentValueFormatter());
+
+            "When I fetch the Priority"
+                .x(() => priority = formatter.Priority);
+
+            "Then it should be 0"
+                .x(() => priority.Should().Be(0));
+        }
+
+        [SuppressMessage("Microsoft.Performance", "CA1812:AvoidUninstantiatedInternalClasses", Justification = "Required for testing.")]
+        private class SomeClass
+        {
+        }
+
+        private class SomeArgumentValueFormatter : ArgumentValueFormatter<SomeClass>
+        {
+            protected override string GetStringValue(SomeClass argumentValue)
+            {
+                return "formatted SomeClass";
+            }
+        }
+    }
+}
--- a/Source/FakeItEasy.Specs/DummyFactorySpecs.cs
+++ b/Source/FakeItEasy.Specs/DummyFactorySpecs.cs
@@ -31,7 +31,7 @@ namespace FakeItEasy.Specs
         [Scenario]
         public static void GenericDummyFactoryDefaultPriority(
             IDummyFactory formatter,
-            int priority)
+            Priority priority)
         {
             "Given an argument value formatter that does not override priority"
                 .x(() => formatter = new SomeDummyFactory());
@@ -39,8 +39,8 @@ namespace FakeItEasy.Specs
             "When I fetch the Priority"
                 .x(() => priority = formatter.Priority);
 
-            "Then it should be 0"
-                .x(() => priority.Should().Be(0));
+            "Then it should be the default priority"
+                .x(() => priority.Should().Be(Priority.Default));
         }
 
         private class SomeClass
@@ -60,9 +60,9 @@ namespace FakeItEasy.Specs
     {
         private int nextID = 1;
 
-        public int Priority
+        public Priority Priority
         {
-            get { return -3; }
+            get { return Priority.Default; }
         }
 
         public bool CanCreate(Type type)
@@ -80,6 +80,11 @@ namespace FakeItEasy.Specs
 
     public class RobotRunsAmokEventDummyFactory : DummyFactory<RobotRunsAmokEvent>
     {
+        public override Priority Priority
+        {
+            get { return new Priority(3); }
+        }
+
         protected override RobotRunsAmokEvent Create()
         {
             return new RobotRunsAmokEvent { ID = -17 };
--- a/Source/FakeItEasy.Specs/DummyFactorySpecs.cs
+++ b/Source/FakeItEasy.Specs/DummyFactorySpecs.cs
@@ -10,10 +10,10 @@ namespace FakeItEasy.Specs
         public static void DummyFactoryUsage(
             RobotActivatedEvent dummy)
         {
-            "when a dummy factory is defined for a set of types"
+            "When I create a Dummy of a type that has a dummy factory defined"
                 .x(() => dummy = A.Dummy<RobotActivatedEvent>());
 
-            "it should create a dummy from the factory"
+            "Then it should be created by the factory"
                 .x(() => dummy.ID.Should().BeGreaterThan(0));
         }
 
@@ -21,10 +21,13 @@ namespace FakeItEasy.Specs
         public static void DummyFactoryPriority(
             RobotRunsAmokEvent dummy)
         {
-            "when two dummy factories apply to the same type"
+            "Given two dummy factories which apply to the same type"
+                .x(() => { }); // see DomainEventDummyFactory and RobotRunsAmokEventDummyFactory
+
+            "When I create a Dummy of the type"
                 .x(() => dummy = A.Dummy<RobotRunsAmokEvent>());
 
-            "it should use the one with higher priority"
+            "Then it should be created by the factory with the higher priority"
                 .x(() => dummy.ID.Should().Be(-17));
         }
 
--- a/Source/FakeItEasy.Specs/DummyFactorySpecs.cs
+++ b/Source/FakeItEasy.Specs/DummyFactorySpecs.cs
@@ -27,6 +27,33 @@ namespace FakeItEasy.Specs
             "it should use the one with higher priority"
                 .x(() => dummy.ID.Should().Be(-17));
         }
+
+        [Scenario]
+        public static void GenericDummyFactoryDefaultPriority(
+            IDummyFactory formatter,
+            int priority)
+        {
+            "Given an argument value formatter that does not override priority"
+                .x(() => formatter = new SomeDummyFactory());
+
+            "When I fetch the Priority"
+                .x(() => priority = formatter.Priority);
+
+            "Then it should be 0"
+                .x(() => priority.Should().Be(0));
+        }
+
+        private class SomeClass
+        {
+        }
+
+        private class SomeDummyFactory : DummyFactory<SomeClass>
+        {
+            protected override SomeClass Create()
+            {
+                return new SomeClass();
+            }
+        }
     }
 
     public class DomainEventDummyFactory : IDummyFactory
--- a/Source/FakeItEasy.Specs/FakeOptionsBuilderSpecs.cs
+++ b/Source/FakeItEasy.Specs/FakeOptionsBuilderSpecs.cs
@@ -158,7 +158,7 @@ namespace FakeItEasy.Specs
         [Scenario]
         public void GenericFakeOptionsBuilderDefaultPriority(
             IFakeOptionsBuilder builder,
-            int priority)
+            Priority priority)
         {
             "Given a fake options builder that does not override priority"
                 .x(() => builder = new SomeClassOptionsBuilder());
@@ -166,8 +166,8 @@ namespace FakeItEasy.Specs
             "When the default priority is fetched"
                 .x(() => priority = builder.Priority);
 
-            "Then it should be 0"
-                .x(() => priority.Should().Be(0));
+            "Then it should be the default priority"
+                .x(() => priority.Should().Be(Priority.Default));
         }
 
         [Scenario]
@@ -232,9 +232,9 @@ namespace FakeItEasy.Specs
 
     public abstract class ConventionBasedOptionsBuilder : IFakeOptionsBuilder
     {
-        public int Priority
+        public Priority Priority
         {
-            get { return 0; }
+            get { return Priority.Default; }
         }
 
         public bool CanBuildOptionsForFakeOfType(Type type)
@@ -401,9 +401,9 @@ namespace FakeItEasy.Specs
     {
         private int nextID = 1;
 
-        public int Priority
+        public Priority Priority
         {
-            get { return -1; }
+            get { return Priority.Default; }
         }
 
         public bool CanBuildOptionsForFakeOfType(Type type)
@@ -439,6 +439,11 @@ namespace FakeItEasy.Specs
     public class RobotRunsAmokEventFakeOptionsBuilder : FakeOptionsBuilder<RobotRunsAmokEvent>
     {
         public static readonly DateTime ConfiguredTimestamp = new DateTime(1997, 8, 29, 2, 14, 03);
+
+        public override Priority Priority
+        {
+            get { return new Priority(5); }
+        }
 
         protected override void BuildOptions(IFakeOptions<RobotRunsAmokEvent> options)
         {
--- a/Source/FakeItEasy.Specs/FakeOptionsBuilderSpecs.cs
+++ b/Source/FakeItEasy.Specs/FakeOptionsBuilderSpecs.cs
@@ -160,7 +160,7 @@ namespace FakeItEasy.Specs
             IFakeOptionsBuilder builder,
             int priority)
         {
-            "Given an options builder that extends the generic base"
+            "Given a fake options builder that does not override priority"
                 .x(() => builder = new SomeClassOptionsBuilder());
 
             "When the default priority is fetched"
--- a/Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
+++ b/Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
@@ -5,6 +5,7 @@ namespace FakeItEasy.Tests.Core
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.IO;
+    using System.Linq;
     using FakeItEasy.Core;
     using NUnit.Framework;
 
@@ -140,6 +141,27 @@ namespace FakeItEasy.Tests.Core
 
             // Assert
             Assert.That(result, Is.EqualTo("a string"));
+        }
+
+        [Test]
+        public void Built_in_formatters_should_have_negative_priority()
+        {
+            // Arrange
+            var allArgumentValueFormatters = typeof(A).Assembly.GetTypes()
+                .Where(t => t.CanBeInstantiatedAs(typeof(IArgumentValueFormatter)))
+                .Select(Activator.CreateInstance)
+                .Cast<IArgumentValueFormatter>();
+
+            // Act
+            var typesWithNonNegativePriority = allArgumentValueFormatters
+                .Where(f => f.Priority >= 0)
+                .Select(f => f.GetType());
+
+            // Assert
+            Assert.That(
+                typesWithNonNegativePriority,
+                Is.Empty,
+                "because no built-in formatters should have non-negative priority");
         }
 
         private void AddTypeFormatter<T>(string formattedValue)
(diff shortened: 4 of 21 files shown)


Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
```diff
@@ -139,7 +141,25 @@ public void Should_prefer_exact_type_formatter_to_interface_formatter()
             var result = this.formatter.GetArgumentValueAsString("string value");
 
             // Assert
-            Assert.That(result, Is.EqualTo("a string"));
+            result.Should().Be("a string");
+        }
+
+        [Test]
+        public void Provided_formatters_should_have_negative_priority()
```
blairconrad: Would've used TestCaseSource, but when there are no formatters, the test returns inconclusive. Same for the Dummy factories and options builders.

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
```diff
@@ -149,16 +169,16 @@ private void AddTypeFormatter<T>(string formattedValue)
...
 
-        private void AddTypeFormatter(Type type, string formattedValue, int priority)
+        private void AddTypeFormatter(Type type, string formattedValue, short priority)
         {
-            var formatter = A.Fake<IArgumentValueFormatter>();
-            A.CallTo(() => formatter.ForType).Returns(type);
-            A.CallTo(() => formatter.GetArgumentValueAsString(A<object>._)).Returns(formattedValue);
-            A.CallTo(() => formatter.Priority).Returns(priority);
-            this.registeredTypeFormatters.Add(formatter);
+            var newFormatter = A.Fake<IArgumentValueFormatter>();
```
blairconrad: "formatter" fought with the member of the same name

Source/FakeItEasy.Specs/ArgumentValueFormatterSpecs.cs
```diff
@@ -0,0 +1,37 @@
...
+            IArgumentValueFormatter formatter,
+            Priority priority)
+        {
+            "Given an argument value formatter that extends the generic base"
+                .x(() => formatter = new SomeArgumentValueFormatter());
+
+            "When I fetch the Priority"
+                .x(() => priority = formatter.Priority);
+
+            "Then it should be 0"
```
thomaslevesque: Only  if the formatter doesn't override Priority (should probably mention it in the "Given")

Source/FakeItEasy.Specs/DummyFactorySpecs.cs
```diff
@@ -27,15 +27,42 @@ public static class DummyFactorySpecs
...
+            IDummyFactory formatter,
+            Priority priority)
+        {
+            "Given a dummy factory that extends the generic base"
+                .x(() => formatter = new SomeDummyFactory());
+
+            "When I fetch the Priority"
+                .x(() => priority = formatter.Priority);
+
+            "Then it should be 0"
```
thomaslevesque: Same remark as for the value formatter

Source/FakeItEasy.Specs/FakeOptionsBuilderSpecs.cs
```diff
@@ -167,7 +167,7 @@ public class FakeOptionsBuilderSpecs
                 .x(() => priority = builder.Priority);
 
             "Then it should be 0"
```
thomaslevesque: Same remark as above

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
```diff
@@ -139,7 +141,25 @@ public void Should_prefer_exact_type_formatter_to_interface_formatter()
...
+                .Select(Activator.CreateInstance)
+                .Cast<IArgumentValueFormatter>();
+
+            // Act
+            var typesWithNonNegativePriority = allArgumentValueFormatters
+                .Where(f => f.Priority >= new Priority(0))
+                .Select(f => f.GetType());
+
+            // Assert
+            typesWithNonNegativePriority.Should().BeEmpty("because no formatters should have non-negative priority");
```
thomaslevesque: nitpicking: no _built-in_ formatters should have a non-negative priority

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
thomaslevesque: But we do have internal formatters, so why is it a problem?

Source/FakeItEasy.Tests/Core/DummyFactoryTests.cs
```diff
@@ -37,6 +38,24 @@ public void Create_should_guard_against_bad_type_argument()
                 .And.ParamName.Should().Be("type");
         }
 
+        [Test]
+        public void Provided_factories_should_have_negative_priority()
```
thomaslevesque: I assume this test is intended for hypothetical factories we would add in the future?

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
blairconrad: It's a fair point. Mostly to be consistent with the tests for FakeOptionsBuilders and DummyFactories. If you're prefer cooler a TestCaseSource here, with the caveat that the test would turn into an "inconclusive" if we removed the formatters at some time in the future, I'm happy to comply.

Alternatively, I could experiment with a test case source that _if there were no formatters (dummy factories, options builders)_, adds a test-provided source that would pass.

Source/FakeItEasy.Tests/Core/DynamicContainerTests.cs
```diff
@@ -116,9 +116,9 @@ private DynamicContainer CreateContainer()
 
         public class OptionsBuilderForTypeWithDummyFactory : IFakeOptionsBuilder
         {
-            public int Priority
+            public Priority Priority
             {
-                get { return 0; }
+                get { return new Priority(0); }
```
thomaslevesque: Perhaps we should add a `Priority.Default` property instead of writing `new Priority(0)` every time? And we could also have an internal `Priority.Internal` property for `new Priority(-1)`. Not really important, but it looks more elegant IMO...

Source/FakeItEasy.Tests/Core/DummyFactoryTests.cs
blairconrad: Exactly.

Source/FakeItEasy.Tests/Core/DynamicContainerTests.cs
blairconrad: I do like `Priority.Default`, and nearly had it.
Will do!

I am not so sure about `Priority.Internal`. Right now, we only use the one value (-1), but who's to say that we won't add others? If we're content for that to be a problem for the future, I can do.

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
thomaslevesque: Ideally, it would be better, because if it fails for one formatter, at least it will tell us which one. But in practice, if you add a formatter and the test suddenly fails, it's not hard to guess which formatter broke it... I don't have a really strong opinion one way or another.

blairconrad: Oh. FluentAssertions doesn't report what's in the sequence when it's not empty! I'm used to NUnit, which does.
Hmm. Does anyone know if FluentAssertions from the future has better error reporting when a sequence is not empty?
And now I wonder about how XUnit's... whatever TestCaseSource is works, since we're moving to that, it seems.

thomaslevesque: Actually, it does report what's in the sequence, I forgot about that.

    new[] {"hello"}.Should().BeEmpty();

> Expected collection to be empty, but found {"hello"}.

So my remark no longer stands. We can keep the test as it is now, as far as I'm concerned.

Source/FakeItEasy.Tests/PriorityTests.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            return new Priority(value);
+        }
+
+        private static Priority ShortPriority(short value)
+        {
+            return new Priority(value);
+        }
+    }
+}
```
thomaslevesque: The coverage looks pretty extensive, but dotCover reports 0%... looks like a bug in dotCover.

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
blairconrad: Hmm. I was off reversing the condition, to force a failure, and got

> Expected collection to be empty because no formatters should have non-negative priority, but found 2.

Did you run your example in Linqpad, using future FluentAssertions?

thomaslevesque: In LinqPad, yes, but with the latest stable release.

thomaslevesque: Looks like FIE is using a very old version (2.2.0.0). The latest stable version is 4.2.1.

thomaslevesque: I can confirm that FluentAssertions 4.2.1 reports the content of the collection, whereas 2.2.0.0 does not.

blairconrad: So we could.

1. update to FA 4.2.1, or
2. change the test so that it reports the failing types, perhaps by using a test case source, adding an always-good formatter (dummy factory, options builder) if no test targets are found
3. something that I can't think of right now

If you have a preferred approach, I'm happy to take it.

thomaslevesque: I think we might as well update FA. I just tried, it breaks one test, but it's easy to fix

blairconrad: I just did a trial run of upgrading to FluentAssertions 4.2.1. There's one test that fails all of a sudden, but other than that, it looks good.

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
blairconrad: How did I not see your note 50 minutes before I made mine? Sigh.
I like the idea.

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
blairconrad: Okay, we're on nearly the same train of thought it seems, although I updated to `new object[] { 1, 2, 3,}` to match the test one above. (I prefer your approach, for what it's worth).

Do you want to PR, or should I?

thomaslevesque: > Do you want to PR, or should I?

To update FA to 4.2.1? I can do it. Let's merge your PR first, once it's ready, then I'll send a new PR.

blairconrad: Oh, I figured we'd want FluentAssertions first, since it would make my PR valid. Otherwise, I have weak tests!

blairconrad: Also, I won't get the other stuff fixed up for a few hours...

thomaslevesque: > Oh, I figured we'd want FluentAssertions first, since it would make my PR valid. Otherwise, I have weak tests!

As you like

blairconrad: Oh, it wasn't a demand or anything!
'sides, I was going to have to force-push anyhow. Might as well rebase while I'm at it.
(This is really just a ploy to get to use [my new toy](https://twitter.com/Hippopottoman/status/693810163329978368).)

Source/FakeItEasy.Specs/ArgumentValueFormatterSpecs.cs
adamralph: Generally I tend to steer away from stating things which are *not* being done, but in this case I'd be happy with "Given an argument value formatter that extends the generic base without overriding the priority".

Incidentally the initial wording is probably overkill. The base class is already the SUT for the test class, and it is abstract so inheritance from it implies "is a" (as opposed to interface inheritance "can do"), so I think it would be enough to say "Given an argument value formatter that does not override priority".

Source/FakeItEasy/Priority.cs
```diff
@@ -0,0 +1,169 @@
...
+    using System;
+
+    /// <summary>
+    /// Indicates precedence between otherwise indistinguishable options, for example when deciding
+    /// which <see cref="IDummyFactory"/> should be used to create a Dummy of a given type.
+    /// In making such decisions, the higher-valued <c>Priority</c> will be used.
+    /// </summary>
+    public struct Priority : IComparable<Priority>, IEquatable<Priority>
+    {
+        private readonly short value;
```
adamralph: We should probably use `int` rather than `short`. See http://stackoverflow.com/a/131354/49241

Source/FakeItEasy.Specs/ArgumentValueFormatterSpecs.cs
blairconrad: I like this. I'll incorporate it. If you'd like a refinement, @thomaslevesque, keep chatting. I'll make as many changes as we need!

Source/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
blairconrad: Thanks, @thomaslevesque. I'll leave the tests as is, now that we have a modern FluentAssertions.

Source/FakeItEasy/Priority.cs
```diff
@@ -0,0 +1,193 @@
...
+            return this.value.GetHashCode();
+        }
+
+        /// <summary>
+        /// Returns a textual description of this <see cref="Priority"/>.
+        /// </summary>
+        /// <returns>A textual description of this <see cref="Priority"/>, indicating its value.</returns>
+        public override string ToString()
+        {
+            return "Priority<" + this.value + ">";
```
adamralph: This should really use `string.Concat()` to avoid an extra string on the heap (result of the first `+`) `</micro-optimisation>`.

Source/FakeItEasy/Priority.cs
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        /// <summary>
+        /// Indicates whether the current object is equal to another object.
+        /// </summary>
+        /// <returns> True if the current object is equal to the <paramref name="obj"/> parameter; otherwise, false.</returns>
+        /// <param name="obj">An object to compare with this object.</param>
+        public override bool Equals(object obj)
+        {
+            return (obj is Priority) && this.Equals((Priority)obj);
```
adamralph: I think this is performing  two casts, one for the `is` and one for the explicit cast. Would this be better?

```C#
return obj != null && obj.GetType() == typeof(Priority) && this.Equals((Priority)obj);
```

Source/FakeItEasy/Priority.cs
blairconrad: Will do.

Source/FakeItEasy/Priority.cs
blairconrad: I have no idea. I'm happy to make the change, though.

Source/FakeItEasy/Priority.cs
```diff
@@ -0,0 +1,193 @@
...
+        /// <summary>
+        /// Compares <see paramref="left"/> and <see paramref="right"/>, returning <c>true</c> if
+        /// and only if the value of <c>left</c> is less than that of <c>right</c>.
+        /// </summary>
+        /// <param name="left">One priority to compare.</param>
+        /// <param name="right">The other priority to compare.</param>
+        /// <returns><c>true</c> if and only if the value of <c>left</c> is less than that of <c>right</c>.</returns>
+        public static bool operator <(Priority left, Priority right)
+        {
+            return left.CompareTo(right) < 0;
```
adamralph: For this and other operators, would it be more readable to delegate straight down to the `value` of each operand?

E.g.

```C#
    return left.value < right.value;
```

Source/FakeItEasy/Priority.cs
blairconrad: Hmm. ReSharper is recommending `is` after the change.…

Source/FakeItEasy/Priority.cs
blairconrad: FWIW, the IL produced is identical:

```IL
.method public hidebysig virtual 
	instance string ToString () cil managed 
{
	// Method begins at RVA 0x3184
	// Code size 27 (0x1b)
	.maxstack 8

	IL_0000: ldstr "Priority<"
	IL_0005: ldarg.0
	IL_0006: ldfld int32 FakeItEasy.Priority::'value'
	IL_000b: box [mscorlib]System.Int32
	IL_0010: ldstr ">"
	IL_0015: call string [mscorlib]System.String::Concat(object, object, object)
	IL_001a: ret
} // end of method Priority::ToString
```

But it _is_ good to keep in practice.

adamralph: So the compiler re-writes it using `string.Concat()`? Wow! :dizzy_face:

blairconrad: Indeed. I think this is common practice for a small number of +es together. But don't quote me on that. Maybe it does it for medium numbers of +es.
I'm going to the lab again, aren't I?

blairconrad: Well, that was a bust, as I expected it might be:

```c#
return "a" + "b" + "c" + "d" + "e" + "f" + "g" + "h" + "i" + "j" + "k" + "l" + "m" + "n" + "o" + "p" + "q" +
                   "r" + "s" + "t" + "u" + "v" + "w" + "x" + "y" + "z";
```

```
method public hidebysig virtual 
	instance string ToString () cil managed 
{
	// Method begins at RVA 0x3184
	// Code size 6 (0x6)
	.maxstack 8

	IL_0000: ldstr "abcdefghijklmnopqrstuvwxyz"
	IL_0005: ret
} // end of method Priority::ToString
```

blairconrad:
```c#
return "a" + this.value + "b" + this.value + "c" + this.value + "d" + this.value + "e" + this.value + "f" +
       this.value + "g" + this.value + "h" + this.value + "i" + this.value + "j" + this.value + "k" +
       this.value + "l" + this.value + "m" + this.value + "n" + this.value + "o" + this.value + "p" +
       this.value + "q" + this.value + "r" + this.value + "s" + this.value + "t" + this.value + "u" +
       this.value + "v" + this.value + "w" + this.value + "x" + this.value + "y" + this.value + "z" +
       this.value;
```

gives.... some horrible IL. Ending in

```
	IL_0257: ldloc.0
	IL_0258: ldc.i4.s 50
	IL_025a: ldstr "z"
	IL_025f: stelem.ref
	IL_0260: ldloc.0
	IL_0261: ldc.i4.s 51
	IL_0263: ldarg.0
	IL_0264: ldfld int32 FakeItEasy.Priority::'value'
	IL_0269: box [mscorlib]System.Int32
	IL_026e: stelem.ref
	IL_026f: ldloc.0
	IL_0270: call string [mscorlib]System.String::Concat(object[])
	IL_0275: ret
```

blairconrad: I suspected (and am feeling brave enough to posit now) that if the compiler _knows how many_ `+`s we have, it may just decide to use `Concat` on its own.

Source/FakeItEasy/Priority.cs
blairconrad: If you like.

blairconrad: Would you like me to do the same in `Equals(object)`? Otherwise, I can continue to delegate to `Equals(Priority)`.

Source/FakeItEasy/Priority.cs
blairconrad: And ultimately, we're never going to use `Equals` (except for the tests), so performance is not so much an issue. In line with the "operator" comment, let's stick with the more readable option.

Source/FakeItEasy/Priority.cs
blairconrad: Keeping as-is in `Equals(object)`. I think everyone's expecting to see delegation to `Equals(Priority)` at some point in here.

blairconrad:
Force-pushing to hide all @adamralph's comments!
Or because I'm out the door soon for work, and that's as far as the changes can go for now.

Source/FakeItEasy/Priority.cs
thomaslevesque: > I suspected (and am feeling brave enough to posit now) that if the compiler knows how many +s we have, it may just decide to use Concat on its own.

Yes, the compiler does its best to optimize that kind of code. I expected it to be mentioned in the C# spec, but couldn't find it, so it's probably implementation-specific.

Source/FakeItEasy/Priority.cs
thomaslevesque: > I think this is performing two casts, one for the is and one for the explicit cast. Would this be better?

I don't think so. Calling `GetType` is probably not cheaper than a cast, and `is` already takes care of the null-check.

blairconrad:
(but I think I've addressed everything that was raised so far)

thomaslevesque:
> (but I think I've addressed everything that was raised so far)

I think so too. I'll wait a few hours before merging in case @adamralph has anything to add.

blairconrad:
… after implementing all the required methods and operators, I'd almost wished I'd sucked a `Comparable<T>` base class out so we'd only have to define `CompareTo`, `ToString`, and `GetHashCode` here. but it felt too YAGNI.

blairconrad:
Thanks for the great comments, folks!

PR #584, opened 2016-01-27
#583 - Added IHideObjectMembers interface to hide members from intellisense that are inappropriate in the context.

dealproc:
Fix #583

PR #595, opened 2016-02-01
Object members

### LINKED ISSUE #580: Make That hide object members

### LINKED ISSUE #583: make IFakeOptions of T hide object members.
Very much like #580.
Otherwise we get offered `Equals` and `ToString` and so on, for no good reason.

adamralph:
fixes #580 
fixes #583
replaces #581
replaces #584


thomaslevesque:
Looking good
Thanks @adamralph and @dealproc!

PR #596, opened 2016-02-01
585 ihideobjectmembers namespace

### LINKED ISSUE #585: Move IHideObjectMembers to top level namespace
It's currently in `FakeItEasy.Configuration` because, I can only guess, it was originally only used for types in that namespace. We're already using it for several types not in that namespace, e.g. `IFakeOptions`. It should be in the namespace `FakeItEasy`.

adamralph:
fixes #585

### DIFF
--- a/Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
+++ b/Source/FakeItEasy.Net35/FakeItEasy.Net35.csproj
@@ -146,9 +146,6 @@
     <Compile Include="..\FakeItEasy\Configuration\IFakeConfigurationManager.cs">
       <Link>Configuration\IFakeConfigurationManager.cs</Link>
     </Compile>
-    <Compile Include="..\FakeItEasy\Configuration\IHideObjectMembers.cs">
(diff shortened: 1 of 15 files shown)


thomaslevesque:
Everything seems to be in order. Thanks @adamralph!

blairconrad:
We're a fan of fully-qualified namespaces, then? I'll keep an eye out.

blairconrad:
Bah. I was hoping ReSharper would help me out here, but the setting "Redundant name qualifier" doesn't just apply to namespaces.

thomaslevesque:
Yes, I have the same problem... it always wants so simplify `FakeItEasy.Something` into just `Something`. I don't know if there's a way to change that behavior.

thomaslevesque:
ah, found it: _ReSharper > Options > Code Editing > C# > Code Style > Reference qualification > Prefer fully qualified name at nested scope_

blairconrad:
Toggled. Thanks @thomaslevesque

PR #597, opened 2016-02-01
Move .editorconfig to root

blairconrad:
So editors other than Visual Studio can take advantage of it, e.g. when editing
Ruby files.


blairconrad:
Per https://github.com/FakeItEasy/FakeItEasy/pull/587#issuecomment-177678067 and subsequent discussion.

Issue #537, opened 2015-09-17
support build on MSBuild v14-only machines?


blairconrad:
Per https://github.com/FakeItEasy/FakeItEasy/pull/536#issuecomment-141205040.
As a convenience to contributors, should the rakefile fall back to MSBuild version 14 when there's no version 12?

In theory the newer tools should be able to build the older projects just fine, no?

jimmyheaddon:
:+1:

adamralph:
:+1: I'd probably go 14 first, falling back to 12.

blairconrad:
Sure. I'm easy.

blairconrad:
Should we switch this task to a general "use VS2015 / MSBuild 14" for builds?
I'm interested in exploring some of the alternative ways of structuring the projects, specifically using shared folders to manage our linked source (and as a way to potentially make it easier for us to run all the tests for each framework we target), and think that's a first step.

thomaslevesque:
> In theory the newer tools should be able to build the older projects just fine, no?

Yes, I only use VS2015 and I have no trouble building.

> I'm interested in exploring some of the alternative ways of structuring the projects, specifically using shared folders to manage our linked source (and as a way to potentially make it easier for us to run all the tests for each framework we target), and think that's a first step.

Shared projects would indeed make it easier to share code between projects. But they're only supported as of VS2013 update 1 IIRC; not sure if that's an issue.

blairconrad:
> But they're only supported as of VS2013 update 1 IIRC; not sure if that's an issue.

Not if we make VS2015 the blessed IDE?

adamralph:
With the recent updates to VS2015 (chiefly the return of some sanity to the NuGet UI) it mostly useable so I'd be happy to make MSBuild 14 the required version.

I'm still a little hesitant around shared projects. I know that the xunit guys went down that route initially for v2 but then abandoned it due to the IDE playing very badly with them in practice.

blairconrad:
We can still explore other options. Earlier you'd had reservations about the scheme Castle.Core uses as well (building a single project with many configurations), no?

adamralph:
> Earlier you'd had reservations about the scheme Castle.Core uses as well (building a single project with many configurations), no?

Yep, I find that very awkward too. It requires multiple build passes over the solution/projects to produce the final output.

[...]

PR #598, opened 2016-02-07
Switch to standard .NET layout

adamralph:
As per https://gist.github.com/davidfowl/ed7564297c61fe9ab814

As discussed at https://github.com/FakeItEasy/FakeItEasy/pull/587#issuecomment-177689315


adamralph:
One thing that I didn't change yet was to put NuGet.config in the root. This is because we are using the old "solution level" package feature in https://github.com/FakeItEasy/FakeItEasy/blob/614ac91f1bb9e91622db364e5b27b1564b9fda56/Source/.nuget/packages.config and AFAIK the version of NuGet we are using needs this to be in a `.nuget` folder so it seemed messy to separate NuGet.config and packages.config for now. I think this is something we can address later.

thomaslevesque:
> Sorry, we could not display the entire diff because too many files (608) changed. 

Oops!

thomaslevesque:
A quick pass showed no obvious issues, but the PR is pretty big, so I could have missed something

FakeItEasy.sln
```diff
@@ -1,26 +1,32 @@
-﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
```
blairconrad: I'm surprised at the switch up to 14. Perhaps this should be part of a separate PR?

FakeItEasy.sln
```diff
@@ -1,26 +1,32 @@
...
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{A6A7B553-5359-4FB9-B19C-8A23004F49DB}"
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
-		CommonAssemblyInfo.cs = CommonAssemblyInfo.cs
-		FakeItEasy.Dictionary.Tests.xml = FakeItEasy.Dictionary.Tests.xml
+		.gitattributes = .gitattributes
+		.gitignore = .gitignore
+		.mailmap = .mailmap
+		CONTRIBUTING.md = CONTRIBUTING.md
```
blairconrad: I've always found it weird to include files that don't contribute to the solution build in the .sln.
But if that's what people like, I'll not stand in the way.

rakefile.rb
```diff
@@ -1,30 +1,30 @@
 require 'albacore'
 
 msbuild_command = "C:/Program Files (x86)/MSBuild/12.0/Bin/MSBuild.exe"
-nuget_command  = "Source/packages/NuGet.CommandLine.2.8.0/tools/NuGet.exe"
-nunit_command  = "Source/packages/NUnit.Runners.2.6.3/tools/nunit-console.exe"
-xunit_command = "Source/packages/xunit.runner.console.2.0.0/tools/xunit.console.exe"
+nuget_command  = "packages/NuGet.CommandLine.2.8.0/tools/NuGet.exe"
+nunit_command  = "packages/NUnit.Runners.2.6.3/tools/nunit-console.exe"
+xunit_command = "packages/xunit.runner.console.2.0.0/tools/xunit.console.exe"
```
blairconrad: I didn't realize our `=`s were so ragged. Kudos on preserving!

FakeItEasy.sln
adamralph: In my understanding, the idea is that the solution is effectively the "root" of the entire thing, so if you open the solution, you can effectively edit anything and everything from inside VS.

FakeItEasy.sln
adamralph: Yeah, I did hover over this for a few seconds before deciding to let it past. It's because I edited the solution inside VS2015 and then hit "save". Thing is, `MinimumVisualStudioVersion = 10.0.40219.1` (below) hasn't changed. AFAIK the first line `# Visual Studio 14` is just a comment which says which version saved the file, and the second line `VisualStudioVersion = 14.0.24720.0` is just a hint to the Visual Studio Version Selector as to which version to prefer when the sln is shelled out.

blairconrad:
I've made a few notes. The upgrading of the `sln` to VS2015 is the only thing I'm actually concerned about. Aside from that, I'm happy to see a merge.

adamralph:
> The upgrading of the sln to VS2015 is the only thing I'm actually concerned about. 

@blairconrad see https://github.com/FakeItEasy/FakeItEasy/pull/598#discussion_r52159258

It still opens and builds in VS2013 (I just tried it).

FakeItEasy.sln
blairconrad: … assuming you'd want to. *shudder*

:shipit:

FakeItEasy.sln
blairconrad: But if people have VS2015, then open the solution, edit a file, and start using `nameof`, they'll be all excited and happy and then later on `rake` will fail (or worse the CI build for the PR), and then frustration…

adamralph: Sure, but this is also the current situation. The current sln on master can be opened by anyone using VS2015. Indeed, on my main dev machine I *only* have VS2015.

If we want to restrict people to C#5 then that is done in the project file like so - https://github.com/Particular/NServiceBus/commit/ab95b4f5a00fe9c2866721756671b1aa6d82e000#diff-f94eb20da3f228e8c172d40dd6cab771

We can address that in another PR.

thomaslevesque: > If we want to restrict people to C# 5 then that is done in the project file like so

I don't think it makes sense if we intend to switch to VS2015 completely...

adamralph: Sure, but we can discuss that elsewhere. It's irrelevant to this PR.

blairconrad: >  The current sln on master can be opened by anyone using VS2015. Indeed, on my main dev machine I only have VS2015.

But I have VS2013 and VS2015. Up 'til now, the solution has only opened in VS2013. Today (after I checked out your PR branch) it opened in VS2015, so that's new.

A while ago, we were discussing moving to MSBuild 14 only and you came down against it, [until there was a demonstrable benefit](https://github.com/FakeItEasy/FakeItEasy/issues/537#issuecomment-177367158), such as using shared projects. That made sense to me.
So why not stick with what we have for now?

You've at least hinted that things work fine for you on your VS2015-only machine, yes? Or at least no worse than they would with this change?
And there's a potentially small degradation (or at least change) for those of us who have VS2013 and VS2015…

thomaslevesque: > You've at least hinted that things work fine for you on your VS2015-only machine, yes?

Works fine for me as well, with only VS2015 installed.

adamralph: OK, fair enough. The fact that it opens for you in VS2015 now instead of VS2013 is indeed a change. I'll revert these lines. Would you prefer a commit amend and force push? Or a new commit?

Incidentally, I prefer the fact that it opens in VS2015 by default with this change. It continually annoyed me when it opened in VS2013 by default. But anyway, we can discuss the VS versioning aspect elsewhere - it shouldn't hold up this PR.

blairconrad: Thanks for reverting.
I don't mind a force push or an additional commit. As pleases you.

And I'm not against making VS2015 the default. Sorry if I didn't make that clear. I'm just not _for_ it sneaking in with "move .sln to the root". If we'd like to consciously make 2015 the default, I'm happy to discuss. And probably will go along with it, even thought 2015 seems to be super sluggish for me. What's that all about? Scratch that. Let's talk about it later if at all.

blairconrad: > If we'd like to consciously make 2015 the default, I'm happy to discuss.

It occurs to me that we've discussed. I'd be content with an update to the description of the PR to indicate that that change came along with the others, if we want to just move ahead.

adamralph: I'd rather keep this PR focused on thing. I'll push the amend so we can merge this and we'll tackle restricting to VS2015+ separately.

adamralph:
OK, I pushed the amended commit which leaves the VS version unchanged. This should be good to go now.

blairconrad:
Thanks, @adamralph!

PR #599, opened 2016-02-09
refactor: upgrade to NuGet.CommandLine.3.3.0

adamralph:
I upgraded to 3.3.0 in the hope that is would recognise `packages.config` which sits next to the `.sln` file rather than under `.nuget`, but alas, it doesn't. (I've continued that thread of discussion over at https://github.com/NuGet/Home/issues/1521#issuecomment-181721352 .)

Still, upgrading to 3.3.0 is not a bad thing (it yields nicer console output at least), so here it is, FWIW.


blairconrad:
Surely **build**, not **refactoring**?

blairconrad:
:+1:

adamralph:
> Surely **build**, not **refactoring**?

`¯\_(ツ)_/¯` grey area?

blairconrad:
I'm not fussy.

PR #606, opened 2016-02-22
Remove container from scope

### LINKED ISSUE #603: Remove custom object containers (currently used by Scopes)
From https://github.com/FakeItEasy/FakeItEasy/issues/484#issuecomment-184679735.

When creating a new Fake Scope, it's possible to provide a custom `IFakeObjectContainer` that the scope will use to create new Dummies and to build up Fake Options. With recent improvements in Dummy Factories and Fake Option Builders, this power is no longer necessary. And considering that Fake Scopes aren't safe across threads, now seems to be the time to remove the custom object containers from Scopes (and FakeItEasy).

blairconrad:
Fixes #603.


src/FakeItEasy/Core/DynamicDummyFactory.cs
```diff
@@ -0,0 +1,50 @@
...
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
+
+    /// <summary>
+    /// Tries to create Dummies by invoking an appropriate <see cref="IDummyFactory"/>.
+    /// </summary>
+    public class DynamicDummyFactory
```
thomaslevesque: It doesn't need to be public, does it?

src/FakeItEasy/Core/DynamicOptionsBuilder.cs
```diff
@@ -0,0 +1,44 @@
...
+    using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Linq;
+    using FakeItEasy.Creation;
+
+    /// <summary>
+    /// Builds fake options by invoking an appropriate <see cref="IFakeOptionsBuilder"/>.
+    /// </summary>
+    public class DynamicOptionsBuilder
```
thomaslevesque: Same here, should probably be internal

tests/FakeItEasy.IntegrationTests/ContainerTests.cs
```diff
@@ -1,34 +1,13 @@
...
 {
     using System;
-    using System.Collections.Generic;
-    using Creation;
-    using FakeItEasy.Core;
+    using FluentAssertions;
     using NUnit.Framework;
 
     [TestFixture]
     public class ContainerTests
```
thomaslevesque: Does this fixture still make sense? It only contains a test that covers dummy factories. Perhaps this that test should be moved somewhere else? In any case, it has nothing to do with containers, since containers don't exist anymore.

tests/FakeItEasy.IntegrationTests/ContainerTests.cs
```diff
@@ -1,34 +1,13 @@
...
-
-            using (Fake.CreateScope(container))
-            {
-                Assert.That(A.Dummy<string>(), Is.EqualTo("fake string"));
-                Assert.That(A.Dummy<int>(), Is.EqualTo(100));
-            }
-        }
-
-        [Test]
         public void Should_load_configurations_from_executing_assembly_using_default_container()
```
thomaslevesque: This test could probably be moved to DummyTests

tests/FakeItEasy.IntegrationTests/FakingClassesTests.cs
```diff
@@ -12,10 +11,7 @@ public class FakingClassesTests
         [Test]
         public void Should_be_able_to_get_a_fake_value_of_uri_type()
         {
-            using (Fake.CreateScope(new NullFakeObjectContainer()))
-            {
-                A.Fake<Uri>();
-            }
+            A.Fake<Uri>();
```
thomaslevesque: No assertion? What's the point of this test?

tests/FakeItEasy.Tests/Core/FakeScopeTests.cs
```diff
@@ -14,12 +16,15 @@ public void Dispose_sets_outside_scope_as_current_scope()
...
+
             using (Fake.CreateScope())
             {
-                Assert.That(FakeScope.Current, Is.Not.SameAs(scope));
+                newCurrentScope = FakeScope.Current;
             }
 
-            Assert.That(FakeScope.Current, Is.SameAs(scope));
+            newCurrentScope.Should().NotBeSameAs(scope, "new scopes should not be the original scope");
+            FakeScope.Current.Should().BeSameAs(scope, "current scope should reviert to original scope");
```
thomaslevesque: typo: "revert"

src/FakeItEasy/Core/DynamicDummyFactory.cs
blairconrad: *gasp*. Of course not! I forgot to internalize when I split the class!

src/FakeItEasy/Core/DynamicOptionsBuilder.cs
blairconrad: Indeed. How embarrassing.

tests/FakeItEasy.IntegrationTests/ContainerTests.cs
blairconrad: Good point. Really, creating a Dummy from a Dummy Factory is so well covered in the Specs we could delete this class with no ill effects, I think.

tests/FakeItEasy.IntegrationTests/FakingClassesTests.cs
blairconrad: My guess is that the lack of blowing up indicated successful creation of the Fake.
We could assert that it's not null. Or that it's a Fake.
Unfortunately, we don't have any clue about why `Uri`s are interesting.

tests/FakeItEasy.Tests/Core/FakeScopeTests.cs
blairconrad: Well spotted!

tests/FakeItEasy.Tests/FakeItEasy.Tests.csproj
```diff
@@ -273,4 +272,4 @@
     </PropertyGroup>
     <Error Condition="!Exists('..\..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets')" Text="$([System.String]::Format('$(ErrorText)', '..\..\packages\StyleCop.MSBuild.4.7.49.1\build\StyleCop.MSBuild.Targets'))" />
   </Target>
-</Project>
+</Project>
```
blairconrad: I didn't see this in the diff at home, but it looks like the final end of line was removed. I guess `.editorconfig` wasn't respected by the IDE…
I'll have to check my plugin once I get home.

blairconrad:
Enjoying your comments, @thomaslevesque. Thanks for catching the foolish mistakes!

tests/FakeItEasy.IntegrationTests/FakingClassesTests.cs
adamralph: It should really be done in this kind of way:

```C#
var exception = Record.Exception(() => A.Fake<Uri>());

exception.Should().BeNull("because we can create fakes of type 'System.Uri'");
```

tests/FakeItEasy.Tests/FakeItEasy.Tests.csproj
adamralph: VS doesn't respect `.editorconfig` when it plays with .csproj files. AFAIK the plug-in only respects files that pass through the text editor window.

tests/FakeItEasy.IntegrationTests/FakingClassesTests.cs
thomaslevesque: > "because we can create fakes of type 'System.Uri'"

Although it doesn't make a lot of sense, since these fakes won't be configurable and they won't represent valid URIs :wink: (not that it really matters for the purpose of this test)

adamralph: I can only guess that, at some point, FIE failed create fakes of type `System.Uri` and that is why this test was added.

adamralph: https://github.com/FakeItEasy/FakeItEasy/commit/062c4454a4d122fcc3c17f42907ac675f449e792

blairconrad: Thanks for the commit.
You prefer null-checking the exception to checking that we got a valid Fake back? I can go either way, but I was surprised.

tests/FakeItEasy.Tests/FakeItEasy.Tests.csproj
blairconrad: Ah. So it'll keep doing this. Shall I leave as-is, then?

thomaslevesque: > Ah. So it'll keep doing this. Shall I leave as-is, then?

Yes, no point fighting VS over this...

tests/FakeItEasy.IntegrationTests/FakingClassesTests.cs
adamralph: > You prefer null-checking the exception to checking that we got a valid Fake back?

We're not really checking that we got a valid Fake back, we're just letting the exception that may be thrown during its creation bubble out to the test framework and therefore implicitly fail the test rather than explicitly failing it using an assertion.

tests/FakeItEasy.Tests/FakeItEasy.Tests.csproj
adamralph: I usually try and catch these locally before committing, but I won't block a PR over it.

blairconrad: Don't worry about blocking my PR. As always, I'm happy to make any change, no matter how small, if it makes sense.
But if we have to manually revert these changes every time we have VS edit a .csproj, I'm with @thomaslevesque—I think we should go with its flow.

tests/FakeItEasy.IntegrationTests/FakingClassesTests.cs
blairconrad: > We're not really checking that we got a valid Fake back,

Indeed. I'm suggesting that's a better approach than verifying that no exception was thrown.

blairconrad:
I've:

1. reverted the whitespace changes to the .csprojs, for now. Although if VS is going to keep making these changes, I wonder if we shouldn't update all the .csproj files and the .editorconfig to match
2. made `DynamicDummyFactory` and `DynamicOptionsBuilder` internal
3. removed `CA1007:UseGenericsWhereAppropriate` suppression from `DynamicDummyFactory` - it seemed not to be needed once the class was internal
4. removed `ContainerTests`
5. made `Should_be_able_to_get_a_fake_value_of_uri_type` check that we actually make a fake `Uri`
6. "reviert" → "revert" in `FakeScopeTests`

new commits coming. I think that was everything.

blairconrad:
Huh. The PR listed my new commits _before_ my note. Odd.

thomaslevesque:
Looks good to me :+1:

blairconrad:
Thanks!

tests/FakeItEasy.IntegrationTests/FakingClassesTests.cs
```diff
@@ -12,10 +11,13 @@ public class FakingClassesTests
...
-            {
-                A.Fake<Uri>();
-            }
+            // Arrange
+
+            // Act
+            var fake = A.Fake<Uri>();
+
+            // Assert
+            Fake.GetFakeManager(fake).Should().NotBeNull("because we should be able to create a fake Uri");
```
adamralph: I'm not sure about this.

Is there a possible code path for `A.Fake<T>()` to return `null`?

As I understood it, this test was added because there was some infinite recursion for certain constructors. I guess in the case of `System.Uri` the `(Uri, Uri)` constructor would have made it happen and a `StackOverflowException` would be thrown.

In that case we should really change this test to be `Should_be_able_to_get_a_fake_value_for_a_type_with_a_constructor_taking_arguments_of_that_type` and the assertion should revert to asserting that no exception is thrown. We can also switch to a custom test type instead of using `System.Uri` to avoid misleading people again.

thomaslevesque: > Is there a possible code path for A.Fake<T>() to return null?

I don't think so. It would throw an exception rather than returning null.

> the assertion should revert to asserting that no exception is thrown

The problem is that you can't (usually) catch a `StackOverflowException`... so it would be a bit pointless.

adamralph: > The problem is that you can't (usually) catch a StackOverflowException... so it would be a bit pointless.

True, which, I suppose, is why the original test didn't attempt to catch and assert an exception, but just tried to create the fake and let any exception bubble up to the test runner.

I think I'd prefer that to the not-null assertion, which I don't think can *ever* fail.

blairconrad:
> Is there a possible code path for `A.Fake<T>()` to return `null`?

Not that I can see, no. It'll make a fake or throw.

> we should really change this test to be `Should_be_able_to_get_a_fake_value_for_a_type_with_a_constructor_taking_arguments_of_that_type`

I've no objection to this. I think it's outside of the scope of this change, since the argument could be made whether we'd kept the container injection or not, but I'm happy to submit a PR.

> and the assertion should revert to asserting that no exception is thrown

I am not sure why the weaker "no exception thrown" is preferable to the stronger "we got an object back that is actually a Fake", since that's what we'd ultimately want.
Suppose a `null` were returned? That's undesirable. But just checking that no exception was thrown would not detect the problem.

tests/FakeItEasy.IntegrationTests/FakingClassesTests.cs
blairconrad: > Is there a possible code path for `A.Fake<T>()` to return `null`?

Not that I can see, no. It'll make a fake or throw.

> we should really change this test to be `Should_be_able_to_get_a_fake_value_for_a_type_with_a_constructor_taking_arguments_of_that_type`

I've no objection to this. I think it's outside of the scope of this change, since the argument could be made whether we'd kept the container injection or not, but I'm happy to submit a PR.

> and the assertion should revert to asserting that no exception is thrown

I am not sure why the weaker "no exception thrown" is preferable to the stronger "we got an object back that is actually a Fake", since that's what we'd ultimately want.
Suppose a `null` were returned? That's undesirable. But just checking that no exception was thrown would not detect the problem.

thomaslevesque: > I think I'd prefer that to the not-null assertion, which I don't think can ever fail.

But the assertion isn't on `A.Fake<Uri>()`, it's on `Fake.GetFakeManager(fake)`, which _would_ return null if `fake` wasn't actually a fake.

blairconrad: Exactly.

adamralph: Oh, I see now. :blush: 

OK, I agree that the assertion can stay as is.

I do think we should change the test name (and probably use a custom type rather than `System.Uri`) and I also agree that that is out of scope here and should be a separate PR. I also think that the assertion reason should change to "because the object should be a fake".

I'll send the PR.

blairconrad: Oh!
I didn't see "I'll send the PR" in my e-mail client.
I was making an issue, which I was going to take, as I had no objection to doing the work, now that we've agreed what the right work is. Will close my issue. Sorry/thanks.

PR #611, opened 2016-02-24
Make A.That throw when used outside of A.CallTo

### LINKED ISSUE #559: Make A<T>.Ignored throw an exception when it is invoked
I know, that sounds crazy... but it's not ;)

`A<T>.Ignored` (or its shortcut `A<T>._`) is used in call specifications to denote that any value of this argument should be a match. If you have long type names (e.g. `IEnumerable<CustomerInvoiceDetailLine>` or something like that), it can be tempting to do this:

```
var any = A<IEnumerable<CustomerInvoiceDetailLine>>.Ignored;
A.CallTo(() => x.PrintInvoiceDetails(any))....
```

However, this doesn't work, because FIE analyzes the expression in `A.CallTo` to detect calls to `A<T>.Ignored`, and in this case the expression doesn't contain it. `A<T>.Ignored` is not meant to actually be called, only to be analyzed as part of the call specification.

This can be very confusing to many users, so I think `A<T>.Ignored` should throw an exception (with a message explaining the problem) when it is invoked. It would be better to be able to detect it at compile time, but I don't think it's possible.

(this was inspired by [this question about Moq](http://stackoverflow.com/q/33669722/98713), which has the same issue)

thomaslevesque:
Fixes #559

### DIFF
--- a/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
+++ b/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
@@ -9,6 +9,7 @@ namespace FakeItEasy.Core
         [ThreadStatic]
         private static List<IArgumentConstraint> trappedConstraints;
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "CallTo", Justification = "It's an identifier")]
         public static void ReportTrappedConstraint(IArgumentConstraint constraint)
         {
             if (trappedConstraints == null)
--- a/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
+++ b/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
@@ -11,10 +11,12 @@ namespace FakeItEasy.Core
 
         public static void ReportTrappedConstraint(IArgumentConstraint constraint)
         {
-            if (trappedConstraints != null)
+            if (trappedConstraints == null)
             {
-                trappedConstraints.Add(constraint);
+                throw new InvalidOperationException("A<T>.Ignored, A<T>._, and A<T>.That can only be used in the context of a call specification with A.CallTo()");
             }
+
+            trappedConstraints.Add(constraint);
         }
 
         public IEnumerable<IArgumentConstraint> TrapConstraints(Action actionThatProducesConstraint)
--- a/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -303,6 +303,8 @@ namespace FakeItEasy.Specs
                              .Should().BeFalse());
         }
 
+        [SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "InvalidOperationException", Justification = "It's an identifier")]
+        [SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "CallTo", Justification = "It's an identifier")]
         [Scenario]
         public static void IgnoredArgumentConstraintOutsideCallSpec(
             Exception exception)
@@ -317,6 +319,8 @@ namespace FakeItEasy.Specs
                 .x(() => exception.Message.Should().Be("A<T>.Ignored, A<T>._, and A<T>.That can only be used in the context of a call specification with A.CallTo()"));
         }
 
+        [SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "InvalidOperationException", Justification = "Because it's the way it should be written")]
+        [SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "CallTo", Justification = "Because it's the way it should be written")]
         [Scenario]
         public static void ThatArgumentConstraintOutsideCallSpec(
             Exception exception)
--- a/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -4,9 +4,9 @@ namespace FakeItEasy.Specs
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.Runtime.InteropServices;
+    using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
     using Xbehave;
-    using Xunit;
 
     public static class CallMatchingSpecs
     {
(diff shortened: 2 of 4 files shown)


tests/FakeItEasy.Tests/Expressions/ExpressionArgumentConstraintFactoryTests.cs
```diff
@@ -186,7 +188,7 @@ public void OnAfterCallIntercepted(ICompletedFakeObjectCall call, IFakeObjectCal
 
                 if (call.Method.Name.Equals("TrapConstraints"))
                 {
-                    call.GetArgument<Action>(0).Invoke();
```
thomaslevesque: This was causing the constraint to be executed outside the trap; I just used a real trap instead.

src/FakeItEasy/Core/ArgumentConstraintTrap.cs
```diff
@@ -11,10 +11,12 @@ internal class ArgumentConstraintTrap
 
         public static void ReportTrappedConstraint(IArgumentConstraint constraint)
         {
-            if (trappedConstraints != null)
+            if (trappedConstraints == null)
             {
-                trappedConstraints.Add(constraint);
+                throw new InvalidOperationException("A.Ignored, A._ and A.That can only be used in the context of a call specification with A.CallTo()");
             }
```
blairconrad: Super nitpick: I think we are in favour of the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma). I've been using it in comments and exception messages.

thomaslevesque: Done!

src/FakeItEasy/Core/ArgumentConstraintTrap.cs
```diff
@@ -11,10 +11,12 @@ internal class ArgumentConstraintTrap
 
         public static void ReportTrappedConstraint(IArgumentConstraint constraint)
         {
-            if (trappedConstraints != null)
+            if (trappedConstraints == null)
             {
-                trappedConstraints.Add(constraint);
+                throw new InvalidOperationException("A.Ignored, A._, and A.That can only be used in the context of a call specification with A.CallTo()");
```
thomaslevesque: Reading this again, I realize it should be `A<T>.Ignored` rather than `A.Ignored`... makes the message harder to read, though. What do you think?

thomaslevesque:
I wonder if this change deserves a spec... the new behavior is covered in `Should_fail_when_reporting_trapped_constraint_outside_call_to_trap_constraints`, but it isn't obvious to the reader that it's related to `A<T>.Ignored`, `A<T>.That` etc.

thomaslevesque:
I'm adding some specs, but I have a BDD-noob question: what do I do when there is nothing to put in the "Given" step?

Currently I have this:

        [Scenario]
        public static void IgnoredArgumentConstraintOutsideCallSpec(
            Exception exception)
        {
            "When A<T>.Ignored is used outside a call specification"
                .x(() => exception = Record.Exception(() => A<string>.Ignored));

            "Then it should throw an InvalidOperationException"
                .x(() => exception.Should().BeAnExceptionOfType<InvalidOperationException>());

            "And the exception message should explain why it's invalid"
                .x(() => exception.Message.Should().Be("A.Ignored, A._, and A.That can only be used in the context of a call specification with A.CallTo()"));
        }

Is it ok to just skip the "Given" step?

src/FakeItEasy/Core/ArgumentConstraintTrap.cs
blairconrad: Oh, that's a good point. I think the `<T>`s are a good idea.

thomaslevesque:
> Is it ok to just skip the "Given" step?

According to [this answer](http://stackoverflow.com/a/9032636/98713), it's OK to omit any of the step. I can't find any other source to confirm, but in this case I clearly don't need any initial context, so I'll go with it.

tests/FakeItEasy.Tests/Expressions/ExpressionArgumentConstraintFactoryTests.cs
blairconrad: Clever.

blairconrad:
> Is it ok to just skip the "Given" step?

It's okay with me! My understanding is as you say - any step can be omitted.

blairconrad:
I think the specs were definitely the way to go. Heck, I'd be happy with them _instead_ of the unit tests. But I wouldn't ask you to undo the unit test work.

blairconrad:
I've taken a pass and am generally happy, with the following notes:

1. You've introduced a number of code analysis violations, specifically in the Specs, with words ("CallTo", "InvalidOperationException") in some test names.
2. I like the names for the new specs. Our practice has been to update files to new conventions when we touch them. How do you feel about taking a pass at the rest of the CallMatching specs in a cleanup commit?
3. Similarly for `ExpressionArgumentConstraintFactoryTests`, how do you feel about an additional commit to make use of FluentAssertions?

thomaslevesque:
> Our practice has been to update files to new conventions when we touch them. How do you feel about taking a pass at the rest of the CallMatching specs in a cleanup commit?

I _knew_ I should have put the new spec in their own file! Just kidding, I'll update the rest of the specs of course.

thomaslevesque:
> Similarly for ExpressionArgumentConstraintFactoryTests, how do you feel about an additional commit to make use of FluentAssertions?

OK

thomaslevesque:
> You've introduced a number of code analysis violations, specifically in the Specs, with words ("CallTo", "InvalidOperationException") in some test names.

Not sure what I should do about that... Add the terms to the dictionary? Suppress the violations? 
Adding _CallTo_ to the dictionary introduces many new warnings, because now _CallTo_ "exists as a discrete term", so it should be written "Callto" in identifiers :confused:

I guess I'll just suppress the violations then...

(Sometimes I hate these dumb tools)

blairconrad:
> (Sometimes I hate these dumb tools)

But it's so nice when they catch our legitimate mistakes, like my "reviert" the other day. Oh, wait… :wink:

thomaslevesque:
All done

(don't worry about the failing AppVeyor build; it was just an experience, but I didn't manage to make it work yet, because of an obscure Ruby error)

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -46,22 +46,22 @@ public interface IHaveAnOutParameter
         public static void ParameterArrays(
             ITypeWithParameterArray fake)
         {
-            "establish"
+            "Given a fake"
                 .x(() => fake = A.Fake<ITypeWithParameterArray>());
 
-            "when matching calls with parameter arrays"
+            "When a call with a parameter array made on this fake"
```
blairconrad: grammar nit: "array _is_ made"

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -169,61 +185,50 @@ public interface IHaveAnOutParameter
...
-            "it should assign the constraint value to the out parameter"
-                .x(() =>
-                    {
-                        string outString = "a different string";
-
-                        subject.TryGetValue("any key", out outString);
-
-                        outString.Should().Be("a constraint string");
-                    });
+            "Then it should assign the constraint value ot the out parameter"
```
blairconrad: "ot" → "to"

blairconrad: Also, should it return `true`?

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -234,46 +239,64 @@ public interface IHaveAnOutParameter
...
-                    {
-                        string refString = "a different string";
+            "Then it should return the configured value"
+                .x(() => result.Should().BeTrue());
 
-                        subject.CheckYourReferences(ref refString)
-                            .Should().BeFalse();
-                    });
+            "And it should assign the constraint value to the ref parameter"
+                .x(() => value.Should().Be(constraintValue));
```
blairconrad: value was already constraintValue, from line 260. Perhaps  use a local temp on that line? Or  reset value before line 261?

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
blairconrad: Oh. This may be covered near line 324.

blairconrad:
I complained a little more, as is my wont, but it's looking good!

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
thomaslevesque: > value was already constraintValue, from line 260

I know, but that's what the original spec was doing...

> Perhaps use a local temp on that line? Or reset value before line 261?

If I do that, it won't match the configured call

thomaslevesque:
I fixed the two typos you mentioned (amended the commit)

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
blairconrad: You're right. I was confusing the `out` semantics. Withdrawn. Sorry.

adamralph:
LGTM, can we merge? I don't think it's *necessary* to rebase, as in #619

adamralph:
I.e. I don't mind which one we merge, but I think raising a separate PR only for a trivial rebase is a bit OTT.

thomaslevesque:
> LGTM, can we merge? I don't think it's necessary to rebase, as in #619

Not necessary, but it doesn't hurt, and makes the history cleaner. I would have rebased anyway, but I forgot... it's done now.

Issue #600, opened 2016-02-09
FakeAsserter.AssertWasCalled is not thread-safe


tillig:
I have some fakes set up that raise events which run a handler on a separate thread. My test executes and verifies that the handler has executed and done its work. This async execution is causing `MustHaveHappened()` assertions to fail.

Probably best explained with an example...

I have a _cache_ and a _cache proxy_. Think distributed cache with a proxy wrapped around it holding a copy of things in memory. When someone _writes_ to the cache proxy, it updates its in-memory set of information and then sends a request to the distributed cache to update its remote info. When the distributed cache has any updates, it notifies the cache proxy that it should refresh its local data from the distributed cache.

It does end up doing a bit of circular-ish logic like:

1. User writes to the cache proxy.
2. Cache proxy writes to local storage.
3. Cache proxy writes to distributed cache.
4. Distributed cache gets new value and fires change event.
5. Cache proxy receives change event (on a separate/background thread) and refreshes the local contents.

The fake is set up like this:

``` csharp
var cache = A.Fake<ICache>();
var proxy = new CacheProxy(cache);
A.CallTo(() => cache.Set(A<Key>.Ignored, A<Value>.Ignored))
  .Invokes(call => {
    var thread = new Thread(() => proxy.Reload());
    thread.IsBackground = true;
    thread.Start();
  });
```

This actually executes perfectly - the cache proxy "Set" method calls the cache "Set" method and on a background thread the proxy gets a signal to Reload.

Depending on the threading timing, I get an exception like this:

```
System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Collections.Generic.List`1.Enumerator.MoveNext()
   at System.Linq.Enumerable.Count[TSource](IEnumerable`1 source, Func`2 predicate)
   at FakeItEasy.Core.FakeAsserter.AssertWasCalled(Func`2 callPredicate, String callDescription, Func`2 repeatPredicate, String repeatDescription)
   at FakeItEasy.Configuration.RuleBuilder.MustHaveHappened(Repeated repeatConstraint)
   at FakeItEasy.Configuration.RuleBuilder.ReturnValueConfiguration`1.MustHaveHappened(Repeated repeatConstraint)
   at FakeItEasy.AssertConfigurationExtensions.MustHaveHappened(IAssertConfiguration configuration)
[...]

PR #613, opened 2016-02-25
Fix for Issue #600 - Additional Thread Safety

tillig:
This is a stab at fixing some of the challenges raised in issue #600. It probably isn't a 100% solution, but I think it's a good step forward.

Items of note:
- For lists that appeared to get modified by the API, I switched to use `System.Collections.Generic.SynchronizedCollection<T>` which is available from .NET 3.5 in the System.ServiceModel assembly. This adds a sort of odd reference, but it means no shim is needed for 3.5 compatibility.
- `SynchronizedCollection<T>` isn't marked `[Serializable]` but `FakeManager` is (for whatever reason) so I had to implement a bit of custom serialization to work around the challenges there.
- I had a problem building where NuGet package restore was trying to restore `NuGet.CommandLine` at the solution level... but was using a version of that already checked in. I was getting an error as it tried to restore the package over the top of itself (file locking) so I removed that from the `packages.config`. It's checked into the repo anyway, so I don't think it needs to be restored each time.

Open to questions / comments. Again, I can't guarantee it'll 100% solve the issue, but it should be a bit closer.

### DIFF
--- a/src/FakeItEasy/Configuration/BuildableCallRule.cs
+++ b/src/FakeItEasy/Configuration/BuildableCallRule.cs
@@ -11,7 +11,7 @@ namespace FakeItEasy.Configuration
     internal abstract class BuildableCallRule
         : IFakeObjectCallRule
     {
-        private readonly List<WherePredicate> wherePredicates;
+        private readonly SynchronizedCollection<WherePredicate> wherePredicates;
         private Action<IInterceptedFakeObjectCall> applicator;
         private bool canSetApplicator;
         private Func<IFakeObjectCall, ICollection<object>> outAndRefParametersValueProducer;
@@ -23,7 +23,7 @@ namespace FakeItEasy.Configuration
             this.applicator = call => call.SetReturnValue(call.Method.ReturnType.GetDefaultValue());
             this.canSetApplicator = true;
             this.canSetOutAndRefParametersValueProducer = true;
-            this.wherePredicates = new List<WherePredicate>();
+            this.wherePredicates = new SynchronizedCollection<WherePredicate>();
         }
 
         /// <summary>
--- a/src/FakeItEasy/Configuration/BuildableCallRule.cs
+++ b/src/FakeItEasy/Configuration/BuildableCallRule.cs
@@ -11,7 +11,7 @@ namespace FakeItEasy.Configuration
     internal abstract class BuildableCallRule
         : IFakeObjectCallRule
     {
-        private readonly SynchronizedCollection<WherePredicate> wherePredicates;
+        private readonly List<WherePredicate> wherePredicates;
         private Action<IInterceptedFakeObjectCall> applicator;
         private bool canSetApplicator;
         private Func<IFakeObjectCall, ICollection<object>> outAndRefParametersValueProducer;
@@ -23,7 +23,7 @@ namespace FakeItEasy.Configuration
             this.applicator = call => call.SetReturnValue(call.Method.ReturnType.GetDefaultValue());
             this.canSetApplicator = true;
             this.canSetOutAndRefParametersValueProducer = true;
-            this.wherePredicates = new SynchronizedCollection<WherePredicate>();
+            this.wherePredicates = new List<WherePredicate>();
         }
 
         /// <summary>
--- a/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
+++ b/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
@@ -7,7 +7,7 @@ namespace FakeItEasy.Core
         : IArgumentConstraintTrapper
     {
         [ThreadStatic]
-        private static List<IArgumentConstraint> trappedConstraints;
+        private static SynchronizedCollection<IArgumentConstraint> trappedConstraints;
 
         public static void ReportTrappedConstraint(IArgumentConstraint constraint)
         {
@@ -19,7 +19,7 @@ namespace FakeItEasy.Core
 
         public IEnumerable<IArgumentConstraint> TrapConstraints(Action actionThatProducesConstraint)
         {
-            trappedConstraints = new List<IArgumentConstraint>();
+            trappedConstraints = new SynchronizedCollection<IArgumentConstraint>();
             var result = trappedConstraints;
 
             actionThatProducesConstraint.Invoke();
--- a/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
+++ b/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
@@ -7,7 +7,7 @@ namespace FakeItEasy.Core
         : IArgumentConstraintTrapper
     {
         [ThreadStatic]
-        private static SynchronizedCollection<IArgumentConstraint> trappedConstraints;
+        private static List<IArgumentConstraint> trappedConstraints;
 
         public static void ReportTrappedConstraint(IArgumentConstraint constraint)
         {
@@ -19,7 +19,7 @@ namespace FakeItEasy.Core
 
         public IEnumerable<IArgumentConstraint> TrapConstraints(Action actionThatProducesConstraint)
         {
-            trappedConstraints = new SynchronizedCollection<IArgumentConstraint>();
+            trappedConstraints = new List<IArgumentConstraint>();
             var result = trappedConstraints;
 
             actionThatProducesConstraint.Invoke();
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -5,7 +5,6 @@ namespace FakeItEasy.Core
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
     using System.Reflection;
-    using System.Runtime.Serialization;
     using FakeItEasy.Configuration;
 
     /// <summary>
@@ -14,12 +13,12 @@ namespace FakeItEasy.Core
     /// by using the AddRule-method.
     /// </summary>
     [Serializable]
-    public partial class FakeManager : IFakeCallProcessor, ISerializable
+    public partial class FakeManager : IFakeCallProcessor
     {
         private readonly LinkedList<CallRuleMetadata> allUserRulesField;
         private readonly CallRuleMetadata[] postUserRules;
         private readonly CallRuleMetadata[] preUserRules;
-        private readonly SynchronizedCollection<ICompletedFakeObjectCall> recordedCallsField;
+        private readonly ConcurrentList<ICompletedFakeObjectCall> recordedCallsField;
         private readonly LinkedList<IInterceptionListener> interceptionListeners;
         private readonly WeakReference objectReference;
 
@@ -49,30 +48,8 @@ namespace FakeItEasy.Core
                                          new CallRuleMetadata { Rule = new DefaultReturnValueRule() }
                                      };
 
-            this.recordedCallsField = new SynchronizedCollection<ICompletedFakeObjectCall>();
+            this.recordedCallsField = new ConcurrentList<ICompletedFakeObjectCall>();
             this.interceptionListeners = new LinkedList<IInterceptionListener>();
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="FakeManager"/> class
-        /// for serialization.
-        /// </summary>
-        /// <param name="info">The <see cref="SerializationInfo"/> containing the object data.</param>
-        /// <param name="context">The source for this deserialization.</param>
-        protected FakeManager(SerializationInfo info, StreamingContext context)
-        {
-            this.allUserRulesField = (LinkedList<CallRuleMetadata>)info.GetValue("allUserRulesField", typeof(LinkedList<CallRuleMetadata>));
-            this.postUserRules = (CallRuleMetadata[])info.GetValue("postUserRules", typeof(CallRuleMetadata[]));
-            this.preUserRules = (CallRuleMetadata[])info.GetValue("preUserRules", typeof(CallRuleMetadata[]));
-            this.recordedCallsField = new SynchronizedCollection<ICompletedFakeObjectCall>();
-            var recordedCallsToAdd = (ICompletedFakeObjectCall[])info.GetValue("recordedCallsField", typeof(ICompletedFakeObjectCall[]));
-            foreach (var call in recordedCallsToAdd)
-            {
-                this.recordedCallsField.Add(call);
-            }
-
-            this.interceptionListeners = (LinkedList<IInterceptionListener>)info.GetValue("interceptionListeners", typeof(LinkedList<IInterceptionListener>));
-            this.objectReference = (WeakReference)info.GetValue("objectReference", typeof(WeakReference));
         }
 
         /// <summary>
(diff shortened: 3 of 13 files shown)


thomaslevesque:
Hi @tillig, thanks for the PR!



> For lists that appeared to get modified by the API, I switched to use `System.Collections.Generic.SynchronizedCollection<T>` which is available from .NET 3.5 in the System.ServiceModel assembly. This adds a sort of odd reference, but it means no shim is needed for 3.5 compatibility.

I didn't even knew this class existed! Looks neat, but when you look at the code, it doesn't seem to be doing a very good job at ensuring thread safety... [Enumeration is definitely not thread-safe](http://referencesource.microsoft.com/#System.ServiceModel/System/ServiceModel/SynchronizedCollection.cs,127): it gets the enumerator under the lock, but enumeration _won't_ happen under the lock, so it's still possible to get an error because the collection is modified during enumeration. To make it safe, it would have to either make a copy of the list, or `yield return` all items while under the lock.

Regarding NuGet.CommandLine, @blairconrad ran into the same issue recently. I'm not sure what the solution is... we'll have to check if we actually need that dependency.

src/FakeItEasy/Configuration/BuildableCallRule.cs
```diff
@@ -11,7 +11,7 @@ namespace FakeItEasy.Configuration
     internal abstract class BuildableCallRule
         : IFakeObjectCallRule
     {
-        private readonly List<WherePredicate> wherePredicates;
+        private readonly SynchronizedCollection<WherePredicate> wherePredicates;
```
thomaslevesque: Not sure it's necessary here... I think this collection is only modified during configuration of the fake, which is unlikely to be made concurrently from multiple threads

src/FakeItEasy/Core/ArgumentConstraintTrap.cs
```diff
@@ -7,7 +7,7 @@ internal class ArgumentConstraintTrap
         : IArgumentConstraintTrapper
     {
         [ThreadStatic]
-        private static List<IArgumentConstraint> trappedConstraints;
+        private static SynchronizedCollection<IArgumentConstraint> trappedConstraints;
```
thomaslevesque: This collection will only ever be used from a single thread, since it's `ThreadStatic`

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -48,11 +49,33 @@ internal FakeManager(Type fakeObjectType, object proxy)
...
+            this.recordedCallsField = new SynchronizedCollection<ICompletedFakeObjectCall>();
+            var recordedCallsToAdd = (ICompletedFakeObjectCall[])info.GetValue("recordedCallsField", typeof(ICompletedFakeObjectCall[]));
+            foreach (var call in recordedCallsToAdd)
+            {
+                this.recordedCallsField.Add(call);
+            }
+
+            this.interceptionListeners = (LinkedList<IInterceptionListener>)info.GetValue("interceptionListeners", typeof(LinkedList<IInterceptionListener>));
+            this.objectReference = (WeakReference)info.GetValue("objectReference", typeof(WeakReference));
+        }
```
thomaslevesque: This makes me a bit uneasy... how to be sure that it will actually behave the same way? I think it would be better to write our own thread-safe collection, make it serializable, and use it instead of `SynchronizedCollection` (which would also solve the enumeration issue I mentioned earlier)

src/FakeItEasy/Core/FakeScope.cs
```diff
@@ -110,11 +110,11 @@ public override IEnumerator<ICompletedFakeObjectCall> GetEnumerator()
 
             internal override IEnumerable<ICompletedFakeObjectCall> GetCallsWithinScope(FakeManager fakeObject)
             {
-                List<ICompletedFakeObjectCall> calls;
+                IList<ICompletedFakeObjectCall> calls;
 
                 if (!this.recordedCallsGroupedByFakeManager.TryGetValue(fakeObject, out calls))
                 {
-                    calls = new List<ICompletedFakeObjectCall>();
+                    calls = new SynchronizedCollection<ICompletedFakeObjectCall>();
```
thomaslevesque: Actually, we should probably return `Enumerable.Empty<T>()` here... the method returns `IEnumerable<T>`, so the result is not meant to be modified by the caller.

src/FakeItEasy/Core/OrderedFakeAsserter.cs
```diff
@@ -2,15 +2,13 @@ namespace FakeItEasy.Core
...
-    using System.Globalization;
-    using System.IO;
 
     internal class OrderedFakeAsserter : IFakeAsserter
     {
         private readonly IEnumerable<IFakeObjectCall> originalCallList;
         private readonly CallWriter callWriter;
         private readonly Queue<IFakeObjectCall> calls;
-        private readonly List<AssertedCall> assertedCalls = new List<AssertedCall>();
+        private readonly SynchronizedCollection<AssertedCall> assertedCalls = new SynchronizedCollection<AssertedCall>();
```
thomaslevesque: This collection is only used when making assertions, so it probably doesn't need to be thread-safe

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
```diff
@@ -45,7 +45,7 @@ public T CreateFake<T>(Action<IFakeOptions<T>> optionsBuilder)
         /// </returns>
         public IList<T> CollectionOfFake<T>(int numberOfFakes)
         {
-            var result = new List<T>();
+            var result = new SynchronizedCollection<T>();
```
thomaslevesque: The list won't be used by FakeItEasy, only by the user, so I don't think it needs to be thread-safe. The user has no reason to assume that it is.

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: To be honest, I don't know exactly why `FakeManager` has to be serializable... I think it's related to cross-AppDomain calls.

tillig: There are unit tests around the serialization already that failed without this and passed after adding. However, I think you're on to something with the custom collection type. I'll see what I can do.

src/FakeItEasy/Configuration/BuildableCallRule.cs
tillig: Is that a reason to _not_ make the change? If it's not going to hurt anything, it'd be nice to cover as much as possible.

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
tillig: Again, is that a reason to _not_ make the fix?

src/FakeItEasy/Configuration/BuildableCallRule.cs
thomaslevesque: Well, with that reasoning, all frameworks would only use thread-safe collections everywhere... But thread-safety isn't free, so it should only be applied where it makes sense.

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
thomaslevesque: I think it is. FakeItEasy builds the list and returns it, it doesn't keep track of it or make any change to it afterwards. It's the caller's responsibility to use the list in a thread-safe way. To make an analogy, you wouldn't expect `Directory.GetFiles` to return a thread-safe list, and it doesn't.

src/FakeItEasy/Configuration/BuildableCallRule.cs
tillig: I won't be "that guy" and argue it with you; I'll revert and we can agree to disagree on this one. I think there is value in certain types of framework going a little further than the minimum required - unit testing and mocking being among those.

src/FakeItEasy/Core/OrderedFakeAsserter.cs
tillig: Making assertions was where I ran into #600 in the first place.

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
tillig: Fair enough. As long as you can guarantee in the future that nothing internal will ever use this for any reason and cast it to `IEnumerable<T>` then, no, you probably don't need that safety. I'll revert.

src/FakeItEasy/Core/OrderedFakeAsserter.cs
thomaslevesque: Yes, but the problem was not on this collection; it was on the recorded calls.

tillig: Another we can agree to disagree on. I'll roll back though I think it's valuable.

thomaslevesque: Why? Do you have reason to think the problem occurred when accessing the `assertedCalls` collection?

tillig: I have reason to believe that #600 goes beyond just _literally that one exception that I reported_ and I'm not looking forward to run into another problem that's hard to troubleshoot. We obviously disagree on the need for _proactively_ solving these things, which is okay, but is a pretty fundamental difference of opinion given the changes _I, as the PR submitter and issue filer_, am interested in. I recognize not every change is interesting to the _project_ and while that's disappointing, I'll defer to your judgment as a project member.

thomaslevesque: Well, in any case, I won't take the decision alone, so don't rush to revert your changes. I'd like to have the opinion of other @FakeItEasy/owners too.

tillig:
Updated per comments including a custom serializable, thread-safe-including-enumeration list.

adamralph:
Would [`ConcurrentBag<T>`](https://msdn.microsoft.com/en-us/library/dd381779(v=vs.110).aspx) fit the bill for the >= .net 4 case? It's serializable. We could convert the `ConcurrentList<T>` added here to a `ConcurrentBag<T>` shim for use in the .net 3.5 project. `ConcurrentBag<T>` doesn't implement `IList<T>`, but does it need to?

tillig:
`ConcurrentBag<T>` is unordered. I went custom to ensure you could count on the enumeration of the list to be consistent. If that's not important, yeah, it could be switched.

adamralph:
@tillig good point. Although `List<T>` (and `IList<T>`) make no official guarantees about order, the .NET FCL implementation has always done so, and people rely no it.

blairconrad:
I've not quite had time to go through the code, but the discussion is moving along, so I figure I should jump in sooner than later.

First, @tillig, thanks for working on this. We appreciate the contribution. And I know we're stifling your exuberance in asking you to scale back the scope of the changes, but the flip side of "the existing code may cause a problem in the future" is "any changes we make may cause a problem in the future", and we tend to favour fewer changes for that reason.
If nothing else, we really do prefer smaller targeted changes at one time. I've not had time to examine all the lists that you were changing (and now aren't), so there's no value judgement about the changes, but I think we'd all agree that if there's a demonstrable benefit to a change, we'd be in favour of it, but _not all the changes have to go in at once_, and the phased approach is easier for readers to understand and easier to debug and/or rollback in the very unlikely event of a problem being introduced. 

Now, if I can make some specific comments about what I've seen in the PR:

I think instead of a List replacement, we'd be better off using a `ConcurrentQueue` (perhaps exposed as some sort of `IProducerConsumerCollection` - I haven't looked into that) for storage.

Advantages:

1. when we're on .NET 4+, we can use the native implementation, which is probably superior to anything we'd write ourselves. (and if we ever drop 3.5, as @tillig hinted at before, then we get to delete some of our custom code - deleting code is always a win)
2. there's a more targeted interface/contract - we're making promises about how we'll add items to the collection (this is an extension of the "list better than bag" discussion above).

Also, I see we have no added tests to ensure that we're fixing the problem we intend to. I've not dug in deeply enough to see if it's possible to write a consistent, repeatable test to make sure that we've fixed the problem (and more to the point, that it stays fixed in the face of future changes), but we should definitely have such test(s) if it's possible. Sometimes writing such tests can be tricky, and if it becomes problematic, I'm happy to take a look and work with you on it, @tillig.

Thanks for listening, everyone.

adamralph:
BTW @tillig please ignore

> Merge the latest changes from master into this branch.

Please [rebase instead](https://github.com/FakeItEasy/FakeItEasy/blob/master/CONTRIBUTING.md#handling-updates-from-upstreammaster). (This is currently my favourite complaint to GitHub.)

Also, if we've exhausted your enthusiasm with our requests, please let us know and we'll just cherry pick your commits and build on them.

tillig:
The test to ensure enumeration works even when the list is changing is on the custom list type. As long as that's the list type you're getting during assertion, there's no reason to believe it wouldn't fix the problem.

I don't think _adding_ synchronized collections is going to cause you problems in the future. We can argue the potential performance impact (which is where I'd wager the difference of opinion comes in - I'm not sure a mocking framework has "performance requirements" like an MVC or DI framework might) but making stuff thread-safe that may or may not need to be thread-safe doesn't sound like a risk to me.

I appreciate everyone taking the time to humor me as I muddle through the various challenges. It's good stuff, and I'm glad to be helping.

Unfortunately, while I want to be a good community member, I'm running out of time to be able to do too terribly much more here. Getting the fairly intricate build environment set up, trying to understand the code / organization / style / customs being used, and iterating here so far has been several hours in the making. Rather than me guessing at what you want, I'm going to have to wait until I get a _very specific directive_ so I can make one more batch of changes and call it good.

And I do apologize if that sounds pointed. It's not meant to be, I just want to level-set expectations. I'm kinda out of time, but I also do want to be helpful. (As it stands, it's really hard for me to even get the myriad Autofac issues addressed, current day job requirements being what they are. You gang know how that is, I'm sure.)

I'm also happy if, as @adamralph says, you want to just start here and cherry-pick. (I did notice the note about rebasing in the CONTRIBUTING.md, but thanks for the reminder. Different projects do like to handle history and merge commits differently. I normally don't rebase.)

**Let me know how to proceed.**

adamralph:
@tillig 

> I appreciate everyone taking the time to humor me as I muddle through the various challenges. It's good stuff, and I'm glad to be helping.

This is what we enjoy, and we're glad you enjoy it too :smile: 

I fully understand your points about the amount time you've put in (which is hugely appreciated), and the limitations on putting more time in. Please don't have any concerns about not being a good community member, everything you've done so far is positive and constructive and we're grateful.

At this point, given you're busy with other things and our penchant for tinkering until we get things "just right", I think the best course of action is for us to cherry pick your commits, which solve the bulk of the problem, and add the various bits of lipstick that we want ourselves as further commits. This doesn't in anyway detract from our appreciation of your work compared with a scenario where we just merged the PR.

Thanks again.

tillig:
Sounds like a plan. Thanks! :+1:

blairconrad:
@tillig, as you know, we've decided to take a different route here, splitting the PR into #618 and  #620.

We appreciate the effort you put into this. Thank you very much.

PR #614, opened 2016-02-25
Use MSBuild v12 if no MSBuild v14

### LINKED ISSUE #537: support build on MSBuild v14-only machines?

blairconrad:
Fixes #537.

### DIFF
--- a/rakefile.rb
+++ b/rakefile.rb
@@ -1,6 +1,10 @@
 require 'albacore'
 
-msbuild_command = "C:/Program Files (x86)/MSBuild/12.0/Bin/MSBuild.exe"
+msbuild_command = [
+  "C:/Program Files (x86)/MSBuild/14.0/Bin/MSBuild.exe",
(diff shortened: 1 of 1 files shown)


adamralph:
Isn't it the other way round? Use MSBuild v12 if no MSBuild v14?

blairconrad:
Oh, drat. I had it that way and reversed in the rakefile after your commit, but used --no-edit with --amend. Will fix.

adamralph:
@blairconrad the code change looks fine (if I understand it correctly), only the commit message and PR title look wrong.

blairconrad:
Fixed, I think. Thanks!

PR #618, opened 2016-02-25
Remove NuGet command line tool from packages

### LINKED ISSUE #616: Remove NuGet.Commandline from packages.config
Per @tillig in https://github.com/FakeItEasy/FakeItEasy/pull/613#issue-136232159:

> I had a problem building where NuGet package restore was trying to restore NuGet.CommandLine at the solution level... but was using a version of that already checked in. I was getting an error as it tried to restore the package over the top of itself (file locking) so I removed that from the packages.config. It's checked into the repo anyway, so I don't think it needs to be restored each time.

I recently had this problem as well.

blairconrad:
Fixes #616


blairconrad:
I took @tillig's change and made a small addition, as it didn't seem quite right to have NuGet.exe live alongside all the other packages.

adamralph:
The changes look good, but the `.gitignore` file also needs trimming down, similar to https://github.com/xbehave/xbehave.net/commit/48a9d72041a9b13a874b114d8598fd058e551ac5#diff-a084b794bc0759e7a6b77810e01874f2

blairconrad:
Oh, good catch.

PR #619, opened 2016-02-26
Make A.That throw when used outside of A.CallTo

### LINKED ISSUE #559: Make A<T>.Ignored throw an exception when it is invoked

blairconrad:
Fixes #559.
Closes #611.
I just rebased #611 onto master.

### DIFF
--- a/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
+++ b/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
@@ -9,6 +9,7 @@ namespace FakeItEasy.Core
         [ThreadStatic]
         private static List<IArgumentConstraint> trappedConstraints;
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "CallTo", Justification = "It's an identifier")]
         public static void ReportTrappedConstraint(IArgumentConstraint constraint)
         {
             if (trappedConstraints == null)
--- a/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
+++ b/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
@@ -11,10 +11,12 @@ namespace FakeItEasy.Core
 
         public static void ReportTrappedConstraint(IArgumentConstraint constraint)
         {
-            if (trappedConstraints != null)
+            if (trappedConstraints == null)
             {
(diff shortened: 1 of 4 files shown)


adamralph:
@blairconrad what's the reason for this? #611 doesn't have conflicts.

blairconrad:
No, but it would've been a nicer commit line. I'm happy to to merge #611 instead.

blairconrad:
I merged #611 instead. (as it turns out, @thomaslevesque rebased in the meantime)

adamralph:
:+1:

blairconrad:
(I'd thought @thomaslevesque was awake at the time, and I'd give him the option of hitting a button rather than rebasing and then harassing me to hit the "merge" button, thereby speeding things along so I could rebase another PR before bed. Not that there was a big rush. I just thought it would save him a little bit of menial work, which I didn't mind doing while watching mindless TV. ([Castle](https://en.wikipedia.org/wiki/Castle_(TV_series)), because I knew you were curious. The episode was better than usual, actually.)

adamralph:
:smile: all good. (I haven't that seen that one.)

blairconrad:
It's okay. Like many "police procedurals", most weeks the episodes are eminently skippable. But the lead, Nathan Fillion, can be hilarious, so when they do a funny/quirky episode, they're quite fun…

adamralph:
Ah, Nathan Fillion... I loved his role in Firefly and Serenity.

PR #620, opened 2016-03-01
Make FakeAsserter threadsafe

### LINKED ISSUE #600: FakeAsserter.AssertWasCalled is not thread-safe

blairconrad:
Fixes #600.

### DIFF
--- a/src/FakeItEasy.Net35/Compatibility/ConcurrentQueue.cs
+++ b/src/FakeItEasy.Net35/Compatibility/ConcurrentQueue.cs
@@ -1,0 +1,34 @@
+namespace System.Collections.Concurrent
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    [Serializable]
+    internal class ConcurrentQueue<T> : IEnumerable<T>
+    {
+        private readonly Queue<T> queue = new Queue<T>();
+
+        public void Enqueue(T item)
+        {
+            lock (this.queue)
+            {
+                this.queue.Enqueue(item);
+            }
+        }
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            lock (this.queue)
+            {
+                return this.queue.ToList().GetEnumerator();
+            }
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return this.GetEnumerator();
+        }
+    }
+}
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -19,7 +19,7 @@ namespace FakeItEasy.Core
         private readonly LinkedList<CallRuleMetadata> allUserRulesField;
         private readonly CallRuleMetadata[] postUserRules;
         private readonly CallRuleMetadata[] preUserRules;
-        private readonly ConcurrentQueue<ICompletedFakeObjectCall> recordedCallsField;
+        private readonly ConcurrentQueue<ICompletedFakeObjectCall> recordedCalls;
         private readonly LinkedList<IInterceptionListener> interceptionListeners;
         private readonly WeakReference objectReference;
 
@@ -49,7 +49,7 @@ namespace FakeItEasy.Core
                                          new CallRuleMetadata { Rule = new DefaultReturnValueRule() }
                                      };
 
-            this.recordedCallsField = new ConcurrentQueue<ICompletedFakeObjectCall>();
+            this.recordedCalls = new ConcurrentQueue<ICompletedFakeObjectCall>();
             this.interceptionListeners = new LinkedList<IInterceptionListener>();
         }
 
@@ -92,11 +92,6 @@ namespace FakeItEasy.Core
         public virtual IEnumerable<ICompletedFakeObjectCall> RecordedCallsInScope
         {
             get { return FakeScope.Current.GetCallsWithinScope(this); }
-        }
-
-        internal ConcurrentQueue<ICompletedFakeObjectCall> AllRecordedCalls
-        {
-            get { return this.recordedCallsField; }
         }
 
         internal LinkedList<CallRuleMetadata> AllUserRules
@@ -184,6 +179,24 @@ namespace FakeItEasy.Core
                     listener.OnAfterCallIntercepted(readonlyCall, ruleToUse.Rule);
                 }
             }
+        }
+
+        /// <summary>
+        /// Returns a list of all calls on the managed object.
+        /// </summary>
+        /// <returns>A list of all calls on the managed object.</returns>
+        internal IEnumerable<ICompletedFakeObjectCall> GetRecordedCalls()
+        {
+            return this.recordedCalls;
+        }
+
+        /// <summary>
+        /// Records that a call has occurred on the managed object.
+        /// </summary>
+        /// <param name="call">The call to remember.</param>
+        internal void RecordCall(ICompletedFakeObjectCall call)
+        {
+            this.recordedCalls.Enqueue(call);
         }
 
         /// <summary>
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -1,6 +1,7 @@
 namespace FakeItEasy.Core
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
@@ -18,7 +19,7 @@ namespace FakeItEasy.Core
         private readonly LinkedList<CallRuleMetadata> allUserRulesField;
         private readonly CallRuleMetadata[] postUserRules;
         private readonly CallRuleMetadata[] preUserRules;
-        private readonly List<ICompletedFakeObjectCall> recordedCallsField;
+        private readonly ConcurrentQueue<ICompletedFakeObjectCall> recordedCallsField;
         private readonly LinkedList<IInterceptionListener> interceptionListeners;
         private readonly WeakReference objectReference;
 
@@ -48,7 +49,7 @@ namespace FakeItEasy.Core
                                          new CallRuleMetadata { Rule = new DefaultReturnValueRule() }
                                      };
 
-            this.recordedCallsField = new List<ICompletedFakeObjectCall>();
+            this.recordedCallsField = new ConcurrentQueue<ICompletedFakeObjectCall>();
             this.interceptionListeners = new LinkedList<IInterceptionListener>();
         }
 
@@ -93,7 +94,7 @@ namespace FakeItEasy.Core
             get { return FakeScope.Current.GetCallsWithinScope(this); }
         }
 
-        internal List<ICompletedFakeObjectCall> AllRecordedCalls
+        internal ConcurrentQueue<ICompletedFakeObjectCall> AllRecordedCalls
         {
             get { return this.recordedCallsField; }
         }
--- a/src/FakeItEasy.Net35/FakeItEasy.Net35.csproj
+++ b/src/FakeItEasy.Net35/FakeItEasy.Net35.csproj
@@ -681,6 +681,7 @@
       <Link>WhereConfigurationExtensions.cs</Link>
     </Compile>
     <Compile Include="Compatibility\ConcurrentDictionary.cs" />
+    <Compile Include="Compatibility\ConcurrentQueue.cs" />
     <Compile Include="Compatibility\ConditionalWeakTable.cs" />
     <Compile Include="Compatibility\EnumerableExtensions.cs" />
     <Compile Include="InternalsVisibleTo.cs" />
--- a/src/FakeItEasy/Core/FakeScope.cs
+++ b/src/FakeItEasy/Core/FakeScope.cs
@@ -58,7 +58,7 @@ namespace FakeItEasy.Core
         /// <param name="call">The call that is intercepted.</param>
         internal void AddInterceptedCall(FakeManager fakeManager, ICompletedFakeObjectCall call)
         {
-            fakeManager.AllRecordedCalls.Add(call);
+            fakeManager.AllRecordedCalls.Enqueue(call);
             this.OnAddInterceptedCall(fakeManager, call);
         }
 
--- a/src/FakeItEasy/Core/FakeScope.cs
+++ b/src/FakeItEasy/Core/FakeScope.cs
@@ -58,7 +58,7 @@ namespace FakeItEasy.Core
         /// <param name="call">The call that is intercepted.</param>
         internal void AddInterceptedCall(FakeManager fakeManager, ICompletedFakeObjectCall call)
         {
-            fakeManager.AllRecordedCalls.Enqueue(call);
+            fakeManager.RecordCall(call);
             this.OnAddInterceptedCall(fakeManager, call);
         }
 
@@ -178,7 +178,7 @@ namespace FakeItEasy.Core
 
             internal override IEnumerable<ICompletedFakeObjectCall> GetCallsWithinScope(FakeManager fakeObject)
             {
-                return fakeObject.AllRecordedCalls;
+                return fakeObject.GetRecordedCalls();
             }
 
             protected override void OnAddRule(FakeManager fakeObject, CallRuleMetadata rule)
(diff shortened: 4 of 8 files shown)


src/FakeItEasy.Net35/Compatibility/ConcurrentQueue.cs
```diff
@@ -0,0 +1,34 @@
+namespace FakeItEasy
```
blairconrad: I'm not wild about the namespace, but if I use System.Collection.Concurrent, the compiler can't tell it from the real one when it builds the .NET 3.5 unit tests. Same as for `ConditionalWeakTable`.
An alternative would be to not have any .NET 3.5 tests for this class and add tests that would verify that the bug is fixed (like `Should_not_throw_when_asserting_while_calls_are_being_made_on_the_fake` below, but to run against the 3.5 library). I actually kind of prefer that, but if we're going to take that step would really like to pull in _all_ the tests (and specs) to be run against all versions of the library, and that (should) get easier with shared projects in VS2015, so wonder if we shouldn't just drag our feet a little.
Regardless, as always, I'll do whatever you like!

thomaslevesque: > if I use System.Collection.Concurrent, the compiler can't tell it from the real one when it builds the .NET 3.5 unit tests.

Isn't it just because `FakeItEasy.Net35.Tests` actually targets .NET 4.5? It *should* target 3.5... There are just two small gotchas:
- it references `FakeItEasy.Tests`, which targets 4.5. Easily fixed by just linking the source files of the helper classes
- it uses FluentAssertions, which requires at least 4.0. But there are very few tests in `FakeItEasy.Net35.Tests`, so it would probably be OK to not use FluentAssertions for those tests.

blairconrad: It _is_ because it targets .NET 4.5. Yes, it should target 3.5, but we made the change in #568 in order to prepare for moving the tests to xUnit 2.0 (#545). We made the change in target a little earlier so it would be something that @mat-mcloughlin wouldn't have to worry about as he converted the test innards.

thomaslevesque: Ah, I see. Another option is to disambiguate with namespace aliases.

thomaslevesque: (I mean reference aliases, actually)

blairconrad: I worked pretty hard on this in #568, and was unable to make it happen.
Although with your inspiration, I may have done. Don't merge, but like the latest commit?

thomaslevesque: Just tried it, it works fine with a reference alias. See this commit: https://github.com/thomaslevesque/FakeItEasy/commit/ec736a8b87b2ff22d504f80e4dd1c4896d480a3d

thomaslevesque:
Ah, looks like you did the exact same thing I was suggesting... right at the same time

blairconrad:
When I have time, I'll squash the new commit in with the second one and repush.
Thanks for the inspiration. I wonder if last time I wasn't trying to alias the System reference. Which didn't go so well. It certainly didn't today.

Oh, and later I'll submit another PR for the other shims that are in "ugly" namespaces.

thomaslevesque:
> Oh, and later I'll submit another PR for the other shims that are in "ugly" namespaces.

:+1:

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -93,7 +94,7 @@ public virtual IEnumerable<ICompletedFakeObjectCall> RecordedCallsInScope
             get { return FakeScope.Current.GetCallsWithinScope(this); }
         }
 
-        internal List<ICompletedFakeObjectCall> AllRecordedCalls
+        internal IEnumerable<ICompletedFakeObjectCall> AllRecordedCalls
         {
             get { return this.recordedCallsField; }
```
adamralph: This is a minor point, bordering on a nitpick, but with the addition of `RecordCall()` below, and the switch to `IEnumerable<T>` from `List<T>`, it looks like we're shifting more towards hiding the backing field as an implementation detail.

For completeness, I'd probably change this to:

```C#
internal IEnumerable<ICompletedFakeObjectCall> GetRecordedCalls()
{
    return this.recordedCallsField.ToList();
}
```

The call to `ToList()` is to prevent access to the backing field, ensuring that all mutation happens via `RecordCall()`.

As part of this refactoring I'd probably also rename `recordedCallsField` to `recordedCalls` (presumably the suffix is something left over from some old code state).

blairconrad: :+1: for renaming the field.

To make sure I understand, can you explain why you prefer a method call to a property for accessing the recorded calls.

(And for the `ToList`, you're just worried that callers will cast the enumerable back to a queue and mutate? I think we've not been protecting internal members in this way so far. I'm not against doing so, wouldn't've worried about the internal clients being so ill-behaved.)

adamralph: I told you it bordered on nit-pick :stuck_out_tongue: 

I don't feel strongly about this specific case, although I do think that generally, this is where stuff gets messy with properties.

Properties have field semantics, even though they are disguised methods under the hood. I don't expect a property to mutate, but a call to `RecordedCalls()` effectively does that. So, if the object state is mutating, that's fine but the mutation of a property via another method call feels like a side effect. If it is a method that is mutating the internal state, I prefer a method to retrieve that mutated state.

The `ToList()` really is a nitpick. As you say, not so important for internals, but the cost is low (incidentally `ToList()` has been demonstrated as the most efficient way among the other options of `ToArray()`, `Select(_ => _)`, etc. to do this for collection less than thousands of items large). I'd probably insist on the `ToList()` if the method were public.

blairconrad: Slightly lost. Possibly due to typo.
Lemme try to sum up.

It feels weird for the `RecordCall` (`RecordedCalls` in your comment?) _method_ to make a change that's visible in the `AllRecordedCalls` _property_.

adamralph: That's the essence of it.

thomaslevesque: > The call to ToList() is to prevent access to the backing field, ensuring that all mutation happens via RecordCall().

I'm not so sure it's a good idea. `ConcurrentQueue<T>.GetEnumerator()` already does a `ToList()`. So, when calling `GetRecordedCalls()`, this would cause 2 copies of the list to be created... isn't that overkill? If it was public, it might make sense (to prevent users from tampering with the FakeManager), but it's internal, so users can't access it anyway. Since the return type is `IEnumerable<T>`, we can't accidentally add something to the list (we would have to explicitly cast it, and we know we shouldn't do that).

thomaslevesque: And if we just return the field directly, it's no longer necessary to change the property to a method

blairconrad: @thomaslevesque so, you're saying "leave it as it is, except maybe rename the horribly-named field"?

Also, one note:

> I'm not so sure it's a good idea. ConcurrentQueue<T>.GetEnumerator() already does a ToList()

I was going to say that our shim does. But it turns out the real one does too! I'm forever learning things from you.

blairconrad: And I'm not sure I understand

> And if we just return the field directly, it's no longer necessary to change the property to a method

We had been returning the field directly, and @adamralph still suggested the property→method change. Or am completely missing your point?

thomaslevesque: > @thomaslevesque so, you're saying "leave it as it is, except maybe rename the horribly-named field"?

Exactly

> I was going to say that our shim does. But it turns out the real one does too! I'm forever learning things from you.

Oh! Actually, I was thinking only about our shim... I lost sight of the fact that it was only a shim :flushed:
But anyway, why not use the shim all the time? Is there any benefit in using the "real" `ConcurrentQueue<T>`?

thomaslevesque: > We had been returning the field directly, and @adamralph still suggested the property→method change. Or am completely missing your point?

Perhaps I misread what @adamralph was saying. My understanding was that it would have made sense to make it a method because we would have returned a different list every time. But now that it's just a wrapper for the field, it doesn't really matter. But if you want to make it a method, I don't mind :wink:

blairconrad: @thomaslevesque, I think the "real" `ConcurrentQueue<T>` has better performance characteristics. It's supposed to be [super fancy and lockless and all kinds of stuff](https://msdn.microsoft.com/en-us/library/dd997305%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396). For our purposes, it probably doesn't matter. We could use all "shim". Me, I like the idea of, if ever we drop 3.5, just deleting the project and then we have less of our own code.

I think the `ToList` and the property/method concerns were separate in @adamralph's mind. I think the property/method concern comes down to this:

> If it is a method that is mutating the internal state, I prefer a method to retrieve that mutated state.

Currently I have a method mutating internal state that is then accessed via a property. I see the point, but feel a little less strongly about it than @adamralph, it seems. However, like you, I don't mind it being a method (in fact, @adamralph has swayed me, and I think it's a little better, but I'd go with the prevailing flow).

However, I am with you on the `ToList` front. I think it's excessive for a property (soon method) that exposes the field as an `IEnumerable` and that is `internal`. So I think omit it. @adamralph, please continue trying to convince us otherwise if you think it's warranted.

adamralph:
To continue the (now) academic conversation thread, the API was my first concern, and the exposure of the field reference second. The conversation was always largely academic since it's an internal API. If it were public I'd be more insistent since we'd be locking ourselves in until 3.0, but with an internal we can always change in any minor release.

Perhaps this is crossing into aesthetics, but I'd prefer this:

```C#
fakeManager.RecordCall(foo);
fakeManager.RecordCall(bar);
var recordedCalls = fakeManager.GetRecordedCalls();
recordedCalls.Count().Should().Be(2);
fakeManager.RecordCall(baz);
recordedCalls.Count().Should().Be(2);
```
to
```C#
fakeManager.RecordCall(foo);
fakeManager.RecordCall(bar);
var recordedCalls = fakeManager.RecordedCalls;
recordedCalls.Count().Should().Be(2);
fakeManager.RecordCall(baz);
recordedCalls.Count().Should().Be(3);
```

Firstly, the second example looks asymmetrical. If I'm calling

```C#
var recordedCalls = fakeManager.RecordedCalls;
```

And `fakeManager` is mutable with respect to it's `RecordedCalls` property, I'd expect to be able to

```C#
fakeManager.RecordedCalls.Add(baz);
```
rather than have to swap over to
```C#
fakeManager.RecordedCall(baz);
```
The semantics I see in the class are "I can record calls" and "at any time you can get the calls I've recorded so far", so the `GetRecordedCalls()` method which gives the *current snapshot* of the recorded calls seems more intuitive to me.

With regard to whether `ToList()` needs to be called, this less important, and can be subject to "insider" knowledge of the type used for field. If the enumerator does take a snapshot each time using `ToList()` internally then that's fine. For a public API, I still feel *slightly* uncomfortable giving out the reference itself, since as @thomaslevesque says, the cunning but bad-practicing developer can inspect the type, say in a debug session, see that's it's actually a queue, and right some code to cast and mutate. I've seen much more horrific things done in code during my career. However, again, this is highly academic in this case since it's internal and even it were public it would a be pathological thing to do. This is more a general practice consideration.

Regarding the use of the shim, I agree with @blairconrad that we should *always* favour the BCL type. The .NET 3.5 shim is a necessary evil, with less guarantees and support than the BCL type. Indeed, we've thrown around the idea of dropping 3.5 support in the not too distant future, so it would great if these shim types could disappear at that point.

thomaslevesque:
@adamralph, you convinced me. A method is better than a property here.

Regarding `ToList()`, another way to hide the original field is to do this:

    return recordedCalls.Select(x => x);

adamralph:
@thomaslevesque yep, that's the way I used to do it, but see here - https://github.com/FakeItEasy/FakeItEasy/pull/620#discussion_r54579420

> (incidentally `ToList()` has been demonstrated as the most efficient way among the other options of `ToArray()`, `Select(_ => _)`, etc. to do this for collection less than thousands of items large)

However, I made a mistake. `ToArray()` is the one I was thinking of , see http://geekswithblogs.net/BlackRabbitCoder/archive/2010/11/04/c.net-fundamentals-returning-an-immutable-collection.aspx

Unfortunately that post doesn't include `ToList()` in its comparison. But, anyway, micro-optimisation.

`¯\_(ツ)_/¯`

blairconrad:
Just to muddy the micro-optimisation waters, if we really cared about performance, we'd have to worry about 

1. what's going to be done with the returned enumerable - if every item is going to be processed, then `ToArray` and similar have an advantage over a `Select` (or `Skip(0)`, apparently).
2. size - if we have to allocate an array or list, then we have to allocate an array or list 
3. the source of the collection. I believe `ToArray` will optimize when it knows the original count of the source. Otherwise I think it tries copying to larger and larger arrays until it succeeds. Then again, maybe that's what a `List` constructor does, just it's inside the `List` class

But ultimately, it won't matter for us.

adamralph:
If forced to make a guess, I'd say the number of calls will rarely go into
the tens, and almost never into the hundreds, which would make `ToArray()`
the best choice. If we cared.

Issue #626, opened 2016-03-11
move documentation from wiki to git


blairconrad:
Having the docs in the wiki is convenient, and allows for convenient updates by anyone in the world, but introduces problems keeping the documentation synchronized with the code. This was most apparent during the current long-running 2.0.0 release cycle, but I've noticed it other times as well.
If we moved the documentation into the code, and used something like [Read the Docs](https://readthedocs.org/) to host our documentation, we'd be able to 

1. Update the documentation at the same time the code was changing. Ideally whoever was contributing a fix or enhancement would also update the documentation, but we could do so in a superseding PR if needed. This would ease our cognitive load, and help reduce the work needed to release a new version - no more scrambling at the last minute to make sure the docs are correct.
2. Make available different versions of the documentation, for those stuck using older versions of the library
3. Not have to worry so much about footnotes or the fancy "[Version 1.x | Version  2.x]" selectors on pages.

I'd be inclined to put product documentation in the source. You know, things that relate directly to using FakeItEasy and may change release by release.
I think keeping the contributor / owner docs in the wiki makes more sense. 

Changes I'd like to make before publishing Please push back if there's anything that doesn't make sense

- [x] tidy everything up as best I can
- [x] create separate branches for working on the [1.25.3](http://bfakeiteasy.readthedocs.org/en/read-the-docs-1.25.3/) and [2.0](http://bfakeiteasy.readthedocs.org/en/read-the-docs-2.0/) versions
- [x] in the 1.25.3 branch, move all the "in 1.x" topics to replace their "unversioned" twin (see [Raising Events](https://github.com/FakeItEasy/FakeItEasy/wiki/Raising-Events) for a sample, if you don't know what I'm talking about)
- [x] remove from the 2.0.0 documentation all the notes that refer to how things changed between 1.x versions. Like the footnotes on the Bootstrapper page. We'd be making the 2.0.0 documents a clean break from the 1.x version. The 1.25.3 documentation could retain these
- [x] reference to "Running multithreaded unit tests with FakeItEasy" from the 2.0.0 version of the "External Resources" page. We shouldn't need it.
- [x] remove "Slow startup time" from the 2.0.0 documentation. The issues we knew about have been addressed, and there have been no complaints for a long time. With no reason to assume that this will be a problem again, why devote space to it?
- [x] remove "Upgrading from older versions" from the 2.0.0 documentation? It handles one very specific task that will need to be done to upgrade from an old version, ignoring nearly everything else.
[...]

PR #627, opened 2016-03-21
Put 1.25.3 docs in the repo

blairconrad:
Connects to #626.

Docs can be previewed at http://bfakeiteasy.readthedocs.org/en/read-the-docs-1.25.3/

Imported the documentation from the wiki, in a MkDocs-style layout.
Adjusted the markdown to conform to MkDocs/ReadTheDocs requirements,
removed references to features in 2.0, and
corrected some minor errors.

### DIFF
--- a/docs/scanning-for-extension-points.md
+++ b/docs/scanning-for-extension-points.md
@@ -1,0 +1,40 @@
+# Scanning for Extension Points
+
+On initialization, essentially as soon as a FakeItEasy type is
+accessed, FakeItEasy uses reflection to look for internal and
+user-supplied extension points. In most cases, there is no _need_ for
+users to define any extensions, but they may be used to enhance the
+power and usability of FakeItEasy.
+
+There are currently three kinds of extension points defined:
+
+* [Custom Dummy Creation](custom-dummy-creation.md) rules,
+* Fake Configurators
+* [Argument Value Formatters](formatting-argument-values.md)
+
+Please see their individual documentation pages to learn how each of these is used.
+
+## The scanning process
+
+On startup, FakeItEasy searches:
+* its own assembly,
+* assemblies already loaded in the current AppDomain and
+* additional assemblies identified by the [Bootstrapper](bootstrapper.md)'s
+  `GetAssemblyFileNamesToScanForExtensions` method<sup>1</sup>
+  
+for classes that implement the various extensions points.
+Any such classes found are added to a catalogue and used at need.
+
+**Note: this does not apply to the Silverlight version of the DLL,
+  which does not load externally-supplied extension points. Under
+  Silverlight, only extensions defined in the FakeItEasy assembly are
+  used.**
+
+----
+1. In FakeItEasy 1.17.0 or earlier, there was no
+  [Bootstrapper](bootstrapper.md), and all DLLs in the current working
+  directory were considered as sources for extension points. This lead
+  to some problems, notably slow startup. If you use an old version of
+  FakeItEasy and have these problems,
+  [upgrade now](https://nuget.org/packages/FakeItEasy/).
+  
--- a/docs/argument-constraints.md
+++ b/docs/argument-constraints.md
@@ -1,0 +1,97 @@
+# Argument constraints
+
+When configuring and [asserting](assertion.md) calls in FakeItEasy, the arguments of the call can be constrained so that only calls to the configured method where the arguments matches the constraint are selected.
+
+# Matching values exactly
+
+Assume the following interface exists:
+```csharp
+public interface IFoo
+{
+    void Bar(string s, int i);
+}
+```
+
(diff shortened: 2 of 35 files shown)


blairconrad:
Oh, ordinarily I'd've preferred to have a series of commits in this PR, but I was undisciplined in my conversion/fixup process, and didn't think I could separate out commits without essentially doing the whole thing again.
If you prefer, I can do the whole thing again.

docs/scanning-for-extension-points.md
```diff
@@ -0,0 +1,40 @@
...
+On initialization, essentially as soon as a FakeItEasy type is
+accessed, FakeItEasy uses reflection to look for internal and
+user-supplied extension points. In most cases, there is no _need_ for
+users to define any extensions, but they may be used to enhance the
+power and usability of FakeItEasy.
+
+There are currently three kinds of extension points defined:
+
+* [Custom Dummy Creation](custom-dummy-creation.md) rules,
+* Fake Configurators
```
thomaslevesque: Wasn't there a wiki page about fake configurators? Can't find it now, but I seem to remember it was there at some point...

thomaslevesque:
Looks good to me. Of course, I didn't read everything in detail, since most of it is existing content from the wiki. I just checked that there was no reference to 2.0 features, and saw no obvious mistake.

docs/scanning-for-extension-points.md
blairconrad: Actually, there never was! Hard to believe, I know, but we didn't have one.
In fact, I had only the most tenuous of ideas of what a "Fake Configurator" was until, well, pretty near the time we started on 2.0 work, I think.
I'm lazy and figured that everyone had gotten along okay without such a page up 'til now, so I didn't bother making one for the conversion.
If you'd like, I can write something up.

blairconrad:
Thanks, yeah. I can't even say that I read _everything in detail_ either. I read it, but it's pretty easy to gloss over things. There may be mistakes, but at least I don't think there are any places that confuse 1.x functionality with 2.x. And probably most of the mistakes were already there…

We can always keep this open a bit longer in case @adamralph wants to take a look. If nothing else, you've given me the confidence I need to proceed with the 2.0 PR!

docs/scanning-for-extension-points.md
thomaslevesque: > If you'd like, I can write something up.

I don't think it's worth the trouble. Most users will probably want to use the 2.0 version anyway.

blairconrad: And even if not, I think _we_ want most uses to use the 2.0 version! :wink:

adamralph:
LGTM, but I don't think we should be merging this into a branch named `1.25.3-docs`. Effectively, we are performing a hack 1.x and if, we ever wanted to do an actual patch 1.25.3, we'd want to do this on the same branch, so I think we should call this branch `support-1` (borrowing from gitlfow).

blairconrad:
I'm not really fussy about the naming, and we can certainly rename, but could you point me to a reference to the `support-1` naming scheme? I couldn't find it at [A successful Git Branching Model](http://nvie.com/posts/a-successful-git-branching-model/).

Also, I'm not entirely sure why we need to keep the branch around after we move the 1.25.3 tag to the commit that has the merge, so am not sure the name matters so much…

adamralph:
The `support-` branch concept has never made it past "experimental" in gitflow. We use it at Particular for NServiceBus to support patching, and back porting features to, older versions. There is some info on it around the interwebs, e.g. https://groups.google.com/forum/#!topic/gitflow-users/I9sErOSzYzE

We have to keep the branch around. If we delete it, then we delete the commits that only it contains, including the one that is tagged.

blairconrad:
> We have to keep the branch around. If we delete it, then we delete the commits that only it contains, including the one that is tagged.

Oh? That doesn't jive with the behaviour I've seen.
Upstream has only these branches:

* 1.25.3-docs
* AssertionExceptions
* NSubstituteSpike
* master

And I created 1.25.3-docs from the 1.25.3 tag last week…

adamralph:
All the current tags live on one or more branches. The 1.25.3 tag lives on the master branch and branches which originate from it. If you were to delete those branches, you'd lose all the commits and hence the tags.

Note that tags in got are very different to tags in, say, SVN. In SVN there is effectively no difference, but in git, a tag belongs to a commit, and it's existence depends on the existence of that commit.

blairconrad:
Aha! When the 1.25.3 tag was created, it wasn't "on master". It was on some other crazy hotfix branch whose name I've fogotten.
Now, 1.25.3 is "on master" because we merged the commit that was tagged with 1.25.3 into master.
I get it!

blairconrad:
Closing in favour of #634, which attempts to merge into the (new) support-1 branch.

Issue #490, opened 2015-05-19
Diagnostic for unused call specification


thomaslevesque:
The analyzer should warn about unused call specifications.

A call to `A.CallTo(...)` (or to any other method of the fluent API used to specify the call) should always be followed by another call, or the result should be assigned to a variable. Otherwise, it means that the call specification is not used, which is almost certainly a mistake.

The diagnostic following shown for the following calls:

```
IAnyCallConfigurationWithNoReturnTypeSpecified.WithReturnType<TMember>
IArgumentValidationConfiguration<TInterface>.WhenArgumentsMatch
ArgumentValidationConfigurationExtensions.WithAnyArguments<TInterface>
IWhereConfiguration<T>.Where
WhereConfigurationExtensions.Where
A.CallTo // (all 3 overloads)
Fake<T>.CallsTo // both overloads
Fake<T>.AnyCall
```

Related to #489 

---

TODO:

- [ ] replace attributes with member to diagnostic map

drauch:
I would also like R# analyzer support for this. I know #505 has been closed, however, R# 10 now features a better attribute describing exactly the situation at hand:

https://www.jetbrains.com/help/resharper/10.0/Reference__Code_Annotation_Attributes.html?origin=old_help#MustUseReturnValueAttribute

Please add `MustUseReturnValueAttribute` to the `A.CallTo()` method!

drauch:
I've created a pull request for R# in https://github.com/FakeItEasy/FakeItEasy/pull/629

blairconrad:
@drauch, thanks for your interest. Your point about moving to another issue is a good one, but because I think it'll be of interest, let me ask this here:

What benefit do the ReSharper annotations provide over the code analyzer? (I know, annotations can be used with older VS, but only if have ReSharper 10. Aside from that.)

If the code analyzer were available right now, what impetus would there be to add the annotations?

drauch:
Of course the main advantage is the immediate availability. Nobody really knows when the Roslyn Analyzer is going to be "done done & released".

Other than that, it is nice to have R# InspectCode available on build agents where MSBuild 14 is not yet installed (this is of personal interest for us). In the long run, we're fine with a released version of the analyzer as well, however, I'm smelling there will be no analyzer for FakeItEasy 1.x and our pull request #629  is immediately mergeable (and is easily removed later on as well). So... :-)

thomaslevesque:
[...]

PR #629, opened 2016-03-22
Added ReSharper MustUseReturnValue attribute to A.CallTo()

### LINKED ISSUE #633: add MustUseReturnValue to `A.CallTo` and `NextCall.To`
Proposed by @drauch in https://github.com/FakeItEasy/FakeItEasy/issues/490#issuecomment-199825874.

We hadn't cared for the idea of putting `Pure` on the methods back when discussing #505, but `MustUseReturnValue` has exactly the semantics we want.

@thomaslevesque has determined that this would present a duplicate warning to one that the Analyzer discussed in #490 would, but there's no real harm.

drauch:
Fixes #633.


src/FakeItEasy/A.cs
```diff
@@ -80,6 +81,7 @@ public static T Dummy<T>()
         /// </summary>
         /// <param name="callSpecification">An expression where the configured member is called.</param>
         /// <returns>A configuration object.</returns>
+        [MustUseReturnValue("Must specify an assertion as well, e.g., .MustHaveHappened()")]
```
ulrichb: Justification text: Maybe "CallTo() makes no sense without following _setup or_ assertion."?

drauch: Improved messages (force pushed changes -> see new commits).

thomaslevesque:
Thanks for the PR @drauch.
Nitpicking: rather than repeating the same message every time in the attribute, could you introduce a constant? (I don't think it's possible to use resources for this)
Otherwise, everything looks good to me

blairconrad:
If we're going to provide this service for `A.CallTo`, I think we should for `NextCall.To` as well.

drauch:
@thomaslevesque : I've extracted a constant (own class, so that it can be reused within `NextCall.To`).

@blairconrad : Done.

src/FakeItEasy/JetBrainsAnnotationsJustificationMessages.cs
```diff
@@ -0,0 +1,15 @@
+namespace FakeItEasy
+{
+    using JetBrains.Annotations;
+
+    /// <summary>
+    /// Common justification messages for JetBrains ReSharper annotations.
+    /// </summary>
+    public static class JetBrainsAnnotationsJustificationMessages
```
blairconrad: Should probably be `internal`, like the attribute. There's no need to expose this.

blairconrad:
@drauch, thanks for the adjustment. I've made one more comment about the visibility of `JetBrainsAnnotationsJustificationMessages`. How do you feel about making the change and squashing the commits? There's been a little bit of back and forth what with our comments and the standalone commit for the 3.5 build and whatnoat.

adamralph:
I've been away on vacation so chiming in later here, but I'm still :-1: on us shipping R# annotations. I'll add my reasoning over at #490

drauch:
@blairconrad : Sure can do, however, not before you've come to a final verdict whether to merge this branch or not. Please, keep me posted.

adamralph:
@FakeItEasy/owners I propose that we close this, citing the [Resharper Gallery](https://github.com/FakeItEasy/FakeItEasy/issues/490#issuecomment-202840961) as the solution for R# annotations.

blairconrad:
@adamralph, based on https://github.com/FakeItEasy/FakeItEasy/issues/490#issuecomment-199825874, I don't think there's anything else to do.

adamralph:
@drauch I'm closing this as a `wontfix` since @FakeItEasy/owners have come to the consensus that we prefer that the community to use the [Resharper Gallery](https://resharper-plugins.jetbrains.com/packages) as the solution for Resharper annotations (more at https://github.com/FakeItEasy/FakeItEasy/issues/490#issuecomment-202840961).

We are grateful for your contribution and we appreciate the time you put into it. It is always with some reluctance that we close a community PR without merging it. However, having commits merged into master is by no means the only method of contribution. Your proposal has driven out a valuable conversation which is a significant contribution in itself.

blairconrad:
As @adamralph said, @drauch, thanks very much for your work, and I'm additionally sorry that it took us a while to reach the decision we did.

I would still be happy to collaborate on an external annotations pack, if you're interested.

blairconrad:
Hey, @drauch! I've some promising results from an external annotations pack.

image

PR #634, opened 2016-03-29
Read the docs 1.25.3

blairconrad:
Contributes to #626. 

Docs can be previewed at http://bfakeiteasy.readthedocs.org/en/read-the-docs-1.25.3/

Imported the documentation from the wiki, in a MkDocs-style layout.
Adjusted the markdown to conform to MkDocs/ReadTheDocs requirements,
removed references to features in 2.0, and
corrected some minor errors.

Replaces #627, and contains the same initial commit, e2c0680bf391008c761ba97eb463c29dfa4da4e0.

Connects to #626

### DIFF
--- a/docs/argument-constraints.md
+++ b/docs/argument-constraints.md
@@ -1,0 +1,97 @@
+# Argument constraints
+
+When configuring and [asserting](assertion.md) calls in FakeItEasy, the arguments of the call can be constrained so that only calls to the configured method where the arguments matches the constraint are selected.
+
+# Matching values exactly
+
+Assume the following interface exists:
+```csharp
+public interface IFoo
+{
+    void Bar(string s, int i);
+}
+```
+
+Then the arguments to Bar can be constrained used to limit call matching:
+```csharp
+var foo = A.Fake<IFoo>();
+
+A.CallTo(() => foo.Bar("hello", 17).MustHaveHappened();
+```
+
+Then FakeItEasy will look _only_ for calls made with the arguments `"hello"` and `17` - no other calls will match the rule.
+
+When checking for argument equality, FakeItEasy uses `object.Equals`. If the type to be checked does not provide an adequate `Equals` method, you may have to use the `That.Matches` method described in [Custom Matching](#custom-matching). Be particularly careful of types whose `Equals` methods perform reference equality rather than value equality. In that case, the objects have to be _the same object_ in order to match, and this sometimes produces unexpected results. When in doubt, verify the type's `Equals` behaviour manually.
+
+# Other matchers
+## Ignoring arguments
+Suppose the value of the integer in the `Bar` call wasn't important, but the string was. Then the following constraint could be used:
+```csharp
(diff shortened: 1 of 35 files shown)


thomaslevesque:
You just changed the target branch, right?

blairconrad:
@thomaslevesque, I did just change the target branch. Sorry that wasn't more clear. Well, I squashed the last 2 commits for the code-highlighting into 1 as well. But the content is all the same as the previous one.

I couldn't find a way to re-target the PR without removing and recreating. Do you know if there is one?

thomaslevesque:
> I couldn't find a way to re-target the PR without removing and recreating. Do you know if there is one?

I don't think there is.

Well, I have no more comments than on #627, so it's still :+1: for me.

blairconrad:
Thanks, @thomaslevesque.

adamralph:
I've moved the 1.25.3 tag to the tip of support-1.

blairconrad:
So sudden. Exciting!

blairconrad:
… and http://fakeiteasy.readthedocs.org/ looks good!
Well, pretty good. Still need to style the 2.0 version and remove the "Changes in 2.0".

blairconrad:
Thanks for the merge.

adamralph:
BTW - I also switched on `1.25.3` in the readthedocs and did a build. That way, when `2.0.0` is released, we can still refer back to it. (`stable` is currently `1.25.3`, but when we tag `2.0.0` then `stable` will point to that.)

blairconrad:
Oh, I thought `stable` would show all the tags. The things I don't know.

blairconrad:
Will we have to switch on each tag as it's released?

adamralph:
AFAIK each time we make a release, the `stable` readthedocs build will get triggered and point to it.

So when we release 3.0.0, we'll have to switch on the last 2.x tag so that people can still browse to it.

Incidentally, I can't see any path for the users to click through to other versions of the docs from a given version of the docs. Am I missing something?

adamralph:
I was missing something. There's a little drop down arrow in the box in the bottom left of the page.

blairconrad:
> AFAIK each time we make a release, the `stable` readthedocs build will get triggered and point to it.

I think so too!

> So when we release 3.0.0, we'll have to switch on the last 2.x tag so that people can still browse to it.

I'm not sure why this is the case. Or if so, why it wouldn't be also true for 2.1, 2.2, …
Let's just keep an eye on it?

adamralph:
We could choose to build docs for previous minors, i.e. 2.1, 2.2, etc. but I'm not sure there will be much need for someone to go back and browse the 2.2 docs when the latest are for 2.3.

As you say, let's see how it goes...

blairconrad:
I think when we release, `stable` will move, but the builds for older tagged versions will stick around!
But I'm just guessing.

adamralph:
> If you are part of a company that uses Read the Docs to host documentation for a commercial product, we offer paid plans that offer email support, additional build resources, and features like CDN support and private documentation.

Perhaps there's something those "additional build resources" that does it.

PR #636, opened 2016-03-30
Remove Changes in 2.0 topic

blairconrad:
As requested in https://github.com/FakeItEasy/FakeItEasy/issues/626#issuecomment-201889165.

I've put the content in the [2.0.0 release notes](https://github.com/FakeItEasy/FakeItEasy/releases/tag/untagged-34c8cd037f49eccbfdb4), with documentation topic links pointing to ReadTheDocs' `latest` branch (eventually should be changed to 2.0.0, I'd think).


thomaslevesque:
:+1:

adamralph:
I think we can change the release notes to point to stable right now since, when we publish them (which creates the tag), stable will (almost) immediately be 2.0.0.

blairconrad:
Thanks for the merge.

> I think we can change the release notes to point to stable right now

So will `latest`. Of course, `stable` will remain 2.0.0 for a little big longer than `latest` will, but if we're worried about that, we'll want to change it to 2.0.0 anyhow.
I'm not bothered one way or another whether it's `stable` or `latest` at this point, but am not interested in making the change. Go ahead, if you think there's value.

adamralph:
> Of course, stable will remain 2.0.0 for a little bit longer than latest will, but if we're worried about that, we'll want to change it to 2.0.0 anyhow.

Good point. Let's change it to `2.0.0` after we switch on and build that tag. No point changing it to `stable` now and then changing again later.

PR #638, opened 2016-04-03
Switch the solution's minimum version to VS2015

### LINKED ISSUE #615: Switch the solution's minimum version to VS2015
From https://github.com/FakeItEasy/FakeItEasy/issues/537#issuecomment-188925534

Would allow us to use fancy C#6 features. Also supports other issues, like #489.

thomaslevesque:
Fixes #615

### DIFF
--- a/FakeItEasy.sln
+++ b/FakeItEasy.sln
@@ -1,7 +1,7 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
-MinimumVisualStudioVersion = 10.0.40219.1
+# Visual Studio 14
+VisualStudioVersion = 14.0.24720.0
+MinimumVisualStudioVersion = 14.0.0.0
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{A6A7B553-5359-4FB9-B19C-8A23004F49DB}"
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
--- a/rakefile.rb
+++ b/rakefile.rb
@@ -1,9 +1,9 @@
 require 'albacore'
 
-msbuild_command = [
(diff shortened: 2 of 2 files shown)


FakeItEasy.sln
```diff
@@ -1,7 +1,7 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
-MinimumVisualStudioVersion = 10.0.40219.1
+# Visual Studio 14
```
blairconrad: Is it called "14" or should we have VS2015 here? Or did the tools choose this name?

rakefile.rb
```diff
@@ -1,9 +1,9 @@
 require 'albacore'
 
-msbuild_command = [
-  "C:/Program Files (x86)/MSBuild/14.0/Bin/MSBuild.exe",
-  "C:/Program Files (x86)/MSBuild/12.0/Bin/MSBuild.exe"
-].find(lambda{ raise "MSBuild not found"}) { |msbuild| File.file?(msbuild) }
+msbuild_command = "C:/Program Files (x86)/MSBuild/14.0/Bin/MSBuild.exe"
+if !File.file?(msbuild_command)
+  raise "MSBuild not found"
+end
```
blairconrad: And I was so proud of having learned the fancy 'find in array or blow up' construction. Ah well.

blairconrad:
It looks very good to me, except for my minor question about "VS 14/VS 2015". Assuming we like "14", I don't see a reason not to merge before 2.0. Does anyone else?

FakeItEasy.sln
thomaslevesque: Yes, that's weird. But I didn't choose to put "14" there, Visual Studio did ;)

rakefile.rb
adamralph: We'll probably want it back once VS 15 goes RTM :wink:

adamralph:
"Visual Studio 14" is indeed chosen by VS when saving the solution file.

thomaslevesque:
Thanks for the merge @adamralph

PR #640, opened 2016-04-04
Add diagnostic for unused call specification

### LINKED ISSUE #490: Diagnostic for unused call specification

thomaslevesque:
There's only one analyzer so far; it checks for unused call
specifications, i.e. calls to A.CallTo(...) that are not used.
At this point the analysis doesn't try to be very smart, it just check if
the call to A.CallTo(...) is the only thing in the statement.

Closes #490


adamralph:
@thomaslevesque nice work!

One thing that immediately leaps out at me is the vsix. Is this required to write an analyzer?

I'm a complete analyzer noob so be prepared for some dumb questions.

thomaslevesque:
> One thing that immediately leaps out at me is the vsix. Is this required to write an analyzer?

No, but it's created by default, so I thought I'd leave it. It's supposed to be an alternative to installing the NuGet package. But now I realize that it doesn't seem to work... I had it working at some point, so I must have done something wrong.

Anyway, we could probably remove it and keep only the NuGet package.

adamralph:
> Anyway, we could probably remove it and keep only the NuGet package.

:+1: I feel very nervous about taking on the maintenance of a VSIX.

FakeItEasy.sln
```diff
@@ -74,6 +74,14 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "samples", "samples", "{2EED
 		samples\FakeItEasy.Examples.ruleset = samples\FakeItEasy.Examples.ruleset
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "analyzer", "analyzer", "{86990EDA-25B4-4BFC-BC44-14D1B754261A}"
```
adamralph: Do we want this folder, or should the analyzer project live under `src`? The analyzer test project lives under `tests`.

thomaslevesque:
> :+1: I feel very nervous about taking on the maintenance of a VSIX.

Yes, I see what you mean... even though it doesn't contain any code, I'm still having trouble making it work. I'll remove it.

thomaslevesque:
Don't look too closely yet, I'm still changing a few things

FakeItEasy.sln
thomaslevesque: I'm changing the project structure now, will make more sense

FakeItEasy.sln
```diff
@@ -74,6 +74,14 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "samples", "samples", "{2EED
 		samples\FakeItEasy.Examples.ruleset = samples\FakeItEasy.Examples.ruleset
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "analyzer", "analyzer", "{86990EDA-25B4-4BFC-BC44-14D1B754261A}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "FakeItEasy.Analyzer", "src\analyzer\FakeItEasy.Analyzer\FakeItEasy.Analyzer.csproj", "{F97E3B4B-080D-414A-9F00-EE20F4451E36}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "FakeItEasy.Analyzer.Test", "src\analyzer\FakeItEasy.Analyzer.Test\FakeItEasy.Analyzer.Test.csproj", "{38F3191B-1CE0-4EE0-930F-794A2F2F4CDB}"
+EndProject
```
adamralph: The solution and disk locations seem to be at odds (assuming the VSIX is going to be removed):

```
(Solution)
  analyzer
    FakeItEasy.Analyzer
  tests
    Unit
      FakeItEasy.Analyzer.Test

(Disk)
  src
    analyzer
      FakeItEasy.Analyzer    
      FakeItEasy.Analyzer.Test
```

I think it should be probably look like this:

```
(Solution)
  src
    FakeItEasy.Analyzer
  tests
    Unit
      FakeItEasy.Analyzer.Tests

(Disk)
  src
      FakeItEasy.Analyzer
  tests
      FakeItEasy.Analyzer.Tests
```

Note that `FakeItEasy.Analyzer.Tests` should also be pluralised.

adamralph:
OK. Let me know when it's ready for a review.

Whilst you're there, there a few standard things missing from the new projects, such as the `TreatWarningsAsErrors` settings, StyleCop.MSBuild, CodeAnalysis, etc. Probably easiest to compare with one of the currrent csproj files.

Also, the rake file should be altered so that the new tests are run and the new package is created.

thomaslevesque:
OK, made a few fixes:
- moved the Analyzer and Analyzer.Tests projects to their proper locations
- removed the VSIX
- filled the nuspec file for the analyzer, which I had forgotten

thomaslevesque:
> Whilst you're there, there a few standard things missing from the new projects, such as the TreatWarningsAsErrors settings, StyleCop.MSBuild, CodeAnalysis, etc. Probably easiest to compare with one of the currrent csproj files.

Ah, yes, forgot about that. Some of the code (the test helpers, mostly) comes from the project template; should it be excluded from StyleCop analysis, or fixed to conform to StyleCop rules?

> Also, the rake file should be altered so that the new tests are run and the new package is created.

OK, will look at it. I don't know Ruby at all, so it's going to be fun :wink:

Probably not today, though...

The rest is ready for review now, I think.

adamralph:
I think we should fix the code to conform to our rules. The template is just code to get us started, but ultimately we own and maintain the code.

src/FakeItEasy.Analyzer/Diagnostic.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <licenseUrl>https://raw.githubusercontent.com/FakeItEasy/FakeItEasy/master/License.txt</licenseUrl>
+    <copyright>Copyright (c) FakeItEasy contributors. (fakeiteasyfx@gmail.com)</copyright>
+    <tags>FakeItEasy, analyzers</tags>
+  </metadata>
+  <!-- The convention for analyzers is to put language agnostic dlls in analyzers\portable50 and language specific analyzers in either analyzers\portable50\cs or analyzers\portable50\vb -->
+  <files>
+    <file src="*.dll" target="analyzers\dotnet\cs" exclude="**\Microsoft.CodeAnalysis.*;**\System.Collections.Immutable.*;**\System.Reflection.Metadata.*;**\System.Composition.*" />
+    <file src="tools\*.ps1" target="tools\" />
+  </files>
+</package>
```
adamralph: Can we rename this file to FakeItEasy.Analyzer.nuspec?

src/FakeItEasy.Analyzer/Diagnostic.nuspec
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+  <metadata>
+    <id>FakeItEasy.Analyzer</id>
+    <version>1.0.0.0</version>
```
adamralph: This should be 0.0.0 so it works with our versioning strategy in the rake file.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.csproj
```diff
@@ -0,0 +1,133 @@
...
+      <HintPath>..\..\packages\Microsoft.Composition.1.0.30\lib\portable-net45+win8+wp8+wpa81\System.Composition.TypedParts.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="System.Reflection.Metadata, Version=1.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\System.Reflection.Metadata.1.2.0\lib\portable-net45+win8\System.Reflection.Metadata.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+  </ItemGroup>
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\v4.5\Microsoft.Portable.CSharp.targets" />
+  <Target Name="AfterBuild">
```
adamralph: We should remove this target and rely on the rake file to create the package, using the same versioning strategy as the main package.

src/FakeItEasy.Analyzer/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Reflection;
```
adamralph: Most of this file should come from CommonAssemblyInfo.cs.

Note that that CommonAssemblyInfo.cs has too much in it right now, e.g. the `SuppressMessage` and `InternalsVisibleTo` bits. It should be trimmed down to whatever is common across both the FakeItEasy and FakeItEasy.Analyzer packages and the rest should sit in local AssemblyInfo.cs files for the specific projects (FakeItEasy and FakeItEasy.Net35 would share a single AssemblyInfo.cs).

src/FakeItEasy.Analyzer/ReadMe.txt
```diff
@@ -0,0 +1,33 @@
+﻿
```
adamralph: Most, if not all, of this file should be nuked.

src/FakeItEasy.Analyzer/packages.config
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.CodeAnalysis.Analyzers" version="1.1.0" targetFramework="portable45-net45+win8" />
+  <package id="Microsoft.CodeAnalysis.Common" version="1.2.1" targetFramework="portable45-net45+win8" />
+  <package id="Microsoft.CodeAnalysis.CSharp" version="1.2.1" targetFramework="portable45-net45+win8" />
+  <package id="Microsoft.CodeAnalysis.CSharp.Workspaces" version="1.2.1" targetFramework="portable45-net45+win8" />
+  <package id="Microsoft.CodeAnalysis.Workspaces.Common" version="1.2.1" targetFramework="portable45-net45+win8" />
+  <package id="Microsoft.Composition" version="1.0.30" targetFramework="portable45-net45+win8" />
+  <package id="NuGet.CommandLine" version="3.3.0" targetFramework="portable45-net45+win8" />
```
adamralph: This can be removed when we address https://github.com/FakeItEasy/FakeItEasy/pull/640/files#r58458364

src/FakeItEasy.Analyzer/tools/install.ps1
```diff
@@ -0,0 +1,49 @@
...
+    {
+        foreach ($analyzerFilePath in Get-ChildItem $languageAnalyzersPath -Filter *.dll)
+        {
+            if($project.Object.AnalyzerReferences)
+            {
+                $project.Object.AnalyzerReferences.Add($analyzerFilePath.FullName)
+            }
+        }
+    }
+}
```
adamralph: I didn't know that analyzer packages required PowerShell cruft :cry: 

I guess they're not going to be quite as ubiquitous as I hoped...

tests/FakeItEasy.Analyzer.Tests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Reflection;
```
adamralph: This file can be removed.

tests/FakeItEasy.Analyzer.Tests/Helpers/CodeFixVerifier.Helper.cs
```diff
@@ -0,0 +1,85 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.Formatting;
+using Microsoft.CodeAnalysis.Simplification;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+
+namespace TestHelper
```
adamralph: Can we change the namespace to `FakeItEasy.Analyzer.Tests.Helpers`? Same applies to all the other templated code.

adamralph:
@thomaslevesque I've finished my first pass over the changes.

thomaslevesque:
> @thomaslevesque I've finished my first pass over the changes.

Thanks @adamralph, I'll make the changes ASAP

tests/FakeItEasy.Analyzer.Tests/Helpers/CodeFixVerifier.Helper.cs
thomaslevesque: Sure

src/FakeItEasy.Analyzer/Properties/AssemblyInfo.cs
thomaslevesque: Agreed.

src/FakeItEasy.Analyzer/Diagnostic.nuspec
thomaslevesque: OK

blairconrad:
Um. Wow. A lot going on here. Good work guys, and I'll jump in when I can. Not tonight (Monday), but I'll look at what I can on Tuesday. I'm keen to learn, if nothing else.

src/FakeItEasy.Analyzer/Diagnostic.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <description>Provides diagnostic analyzers to warn about incorrect usage of FakeItEasy.</description>
+    <language>en-US</language>
+    <projectUrl>http://fakeiteasy.github.io/</projectUrl>
+    <iconUrl>http://fakeiteasy.github.io/img/fakeiteasy_logo_256_square_white.png</iconUrl>
+    <licenseUrl>https://raw.githubusercontent.com/FakeItEasy/FakeItEasy/master/License.txt</licenseUrl>
+    <copyright>Copyright (c) FakeItEasy contributors. (fakeiteasyfx@gmail.com)</copyright>
+    <tags>FakeItEasy, analyzers</tags>
+  </metadata>
+  <!-- The convention for analyzers is to put language agnostic dlls in analyzers\portable50 and language specific analyzers in either analyzers\portable50\cs or analyzers\portable50\vb -->
+  <files>
```
blairconrad: I do not understand how this comment relates to the following `file` tag. is dotnet the new portable50?

blairconrad:
>  I'll look at what I can on Tuesday

I did not look at it on Tuesday. I took a superficial pass, mostly reading @adamralph's comments and looking at the fringe stuff. I'm still keen to read the core at some time, but the new 2.0 work is preventing me.

src/FakeItEasy.Analyzer/Diagnostic.nuspec
thomaslevesque: I'm not sure... the comment was already there in the template, I didn't write it. See [this page](http://blog.nuget.org/20150729/Introducing-nuget-uwp.html) for an explanation of `dotnet` .

src/FakeItEasy.Analyzer/UnusedCallSpecificationAnalyzer.cs
```diff
@@ -0,0 +1,65 @@
...
+            if (expr == null)
+            {
+                return;
+            }
+
+            var member = memberAccess.Name as IdentifierNameSyntax;
+            if (member == null)
+            {
+                return;
+}
```
blairconrad: very nitpicky, but this moustache is hard up against the left margin

adamralph: <kbd>Ctrl</kbd>+<kbd>K</kbd>,<kbd>Ctrl</kbd>+<kbd>D</kbd> FTW :wink:

src/FakeItEasy.Analyzer/SymbolExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+            while (!string.IsNullOrEmpty(ns?.Name))
+            {
+                stack.Push(ns.Name);
+                ns = ns.ContainingNamespace;
+            }
+
+            return string.Join(".", stack);
+        }
+    }
+}
```
blairconrad: I'm sure this code was template code, but the variable names leave something to be desired.

adamralph: Agreed. I think we should make sure this code looks like "code that we would write" before merging.

tests/FakeItEasy.Analyzer.Tests/Helpers/DiagnosticVerifier.cs
```diff
@@ -0,0 +1,282 @@
...
+        /// Get the CSharp analyzer being tested - to be implemented in non-abstract class.
+        /// </summary>
+        /// <returns>The diagnostic analyzer being tested.</returns>
+        protected virtual DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
+        {
+            return null;
+        }
+
+        /// <summary>
+        /// Get the Visual Basic analyzer being tested (C#) - to be implemented in non-abstract class.
```
blairconrad: I suspect that everything in `Helpers` was provided, but what's this comment mean? "**Visual Basic** analyzer being tested **(C#)**"

tests/FakeItEasy.Analyzer.Tests/UnusedCallSpecificationAnalyzerTests.cs
```diff
@@ -0,0 +1,113 @@
...
+                    Locations = new[] { new DiagnosticResultLocation("Test0.cs", 9, 13) }
+                });
+        }
+
+        protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
+        {
+            return new UnusedCallSpecificationAnalyzer();
+        }
+    }
+}
```
blairconrad: :+1:

src/FakeItEasy.Analyzer/UnusedCallSpecificationAnalyzer.cs
thomaslevesque: Uh... I don't even know how it happened

src/FakeItEasy.Analyzer/SymbolExtensions.cs
thomaslevesque: Actually, it's not template code... I wrote this myself :confounded:. I wrote this myself.

The variable names make sense to me, but if you don't like them, I'll change them.

tests/FakeItEasy.Analyzer.Tests/Helpers/DiagnosticVerifier.cs
thomaslevesque: Yes, template code again... I'll fix it

src/FakeItEasy.Analyzer/SymbolExtensions.cs
adamralph: Oh! Sorry :worried: 

BTW - I had a play to see if I could boil the method a bit and this is what I came up with:

```C#
        public static string GetFullName(this ISymbol symbol)
        {
            if (symbol == null)
            {
                throw new ArgumentNullException(nameof(symbol));
            }

            var names = new Stack<string>();
            names.Push(symbol.Name);
            while ((symbol = symbol.ContainingType) != null)
            {
                names.Push(symbol.Name);
            }

            while ((symbol = symbol.ContainingNamespace) != null)
            {
                names.Push(symbol.Name);
            }

            return string.Join(".", names);
        }
```

src/FakeItEasy.Analyzer/UnusedCallSpecificationAnalyzer.cs
adamralph: It's probably worth doing a quick <kbd>Ctrl</kbd>+<kbd>K</kbd>, <kbd>Ctrl</kbd>+<kbd>D</kbd> on all the generated files. Including app.config, which looks a little odd.

src/FakeItEasy.Analyzer/SymbolExtensions.cs
blairconrad: @thomaslevesque, now I feel bad. (And at the risk of compounding things, I liked the original version of your comment better. I thought you were very funny!)

The code is certainly understandable, and very terse names work well enough given the limited scope of the method. It's just that my inclination would've been to choose names that were a little more descriptive. (Which is in the eye of the beholder, of course. It's all style.)

As an example of the sort of name I might've picked, and I'm not saying "Take these! Take these!", I'd probably go with something more like

```c#
var nameParts = new Stack<string>();
nameParts.Push(type.Name);
var containingType = type.ContainingType;
while (containingType != null)
{
    nameParts.Push(containingType.Name);
    type = containingType;
    containingType = containingType.ContainingType;
}

var typeNamespace = type.ContainingNamespace;
while (!string.IsNullOrEmpty(typeNamespace?.Name))
{
    nameParts.Push(typeNamespace.Name);
    typeNamespace = typeNamespace.ContainingNamespace;
}

return string.Join(".", nameParts);
```

although I don't love `typeNamespace`. I'd had `@namespace`, but I try to avoid `@`s. They feel dirty to me.

Maybe this is no better.

As an aside, I also moved the assignment `type = containingType` down one line because it felt better to me to have the 2 "walking to the outside of the code" lines next to each other.

I'd wondered whether you even needed to keep track of `type` inside the first loop, since I would have guessed that the namespace would be the same. I tried to experiment with a call of the form `OuterClass.InnerClass.Method()`, but was not smart enough to get to the point where `GetFullName` would be called. :confounded:

adamralph: @blairconrad have a look at my snippet. You don't need the namespace variable if you widen the parameter to `ISymbol` you can just assign to it on both loops as you recurse upwards.

`nameParts` isn't a bad idea.

adamralph: Although come to think of it, it's difficult to justify the method being relevant to any instance `ISymbol`. It only really has relevance for a type, so probably ought to stay as `INamedTypeSymbol `.

thomaslevesque: > (And at the risk of compounding things, I liked the original version of your comment better. I thought you were very funny!)

I was looking for an "embarrassed" emoji, but couldn't find a suitable one :wink:

blairconrad: @adamralph, sorry for posting after you. I'd _started_ writing before, and I guess that keeps the page from updating, so I never saw yours. Or something.

@thomaslevesque, I often look for an "embarrassed" emoji. I think I will just take a selfie and use it in the future.

thomaslevesque:
I think I fixed everything, except code analysis (FxCop) warnings. Working on it now.

thomaslevesque:
All done

adamralph:
Thanks @thomaslevesque, I'll do another pass over it ASAP.

FakeItEasy.sln
```diff
@@ -23,6 +23,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{59466C4D-C407-4DD9-B77B-D99162B9FC7C}"
 	ProjectSection(SolutionItems) = preProject
+		AssemblyInfo.cs = AssemblyInfo.cs
```
adamralph: This file should live in `src\`.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.csproj
```diff
@@ -0,0 +1,139 @@
...
+    <EmbeddedResource Include="Resources.resx">
+      <Generator>ResXFileCodeGenerator</Generator>
+      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
+    </EmbeddedResource>
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="app.config" />
+    <None Include="FakeItEasy.Analyzer.nuspec">
+      <SubType>Designer</SubType>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
```
adamralph: I don't think this line should be here. The nuspec shouldn't be copied to the output directory.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <authors>Thomas Levesque, FakeItEasy contributors</authors>
+    <description>Provides diagnostic analyzers to warn about incorrect usage of FakeItEasy.</description>
+    <language>en-US</language>
+    <projectUrl>http://fakeiteasy.github.io/</projectUrl>
+    <iconUrl>http://fakeiteasy.github.io/img/fakeiteasy_logo_256_square_white.png</iconUrl>
+    <licenseUrl>https://raw.githubusercontent.com/FakeItEasy/FakeItEasy/master/License.txt</licenseUrl>
+    <copyright>Copyright (c) FakeItEasy contributors. (fakeiteasyfx@gmail.com)</copyright>
+    <tags>FakeItEasy, analyzers</tags>
+  </metadata>
+  <!-- The convention for analyzers is to put language agnostic dlls in analyzers\portable50 and language specific analyzers in either analyzers\portable50\cs or analyzers\portable50\vb -->
```
adamralph: Shall we remove this comment? I presume it comes from the template?

src/FakeItEasy.Analyzer/packages.config
adamralph: This line still needs to be removed. Now that we use the rakefile to package the analyzer, the project doesn't need a dependency on Nuget.CommandLine.

tests/FakeItEasy.Analyzer.Tests/Helpers/DiagnosticResult.cs
```diff
@@ -0,0 +1,93 @@
...
+
+    /// <summary>
+    /// Struct that stores information about a Diagnostic appearing in a source.
+    /// </summary>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1815:OverrideEqualsAndOperatorEqualsOnValueTypes", Justification = "Not necessary")]
+    public struct DiagnosticResult
+    {
+        private DiagnosticResultLocation[] locations;
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1819:PropertiesShouldNotReturnArrays", Justification = "It's not intended to be a public API, so it doesn't matter")]
```
adamralph: I guess we could make this type `internal` but since it's only part of the test lib and not part of the package, it doesn't really matter.

adamralph:
@thomaslevesque I've done another pass over the changes. A few more comments, but all very minor.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.nuspec
thomaslevesque: Yes, it does. I'll remove it.

tests/FakeItEasy.Analyzer.Tests/Helpers/DiagnosticResult.cs
thomaslevesque: It's used by `DiagnosticVerifier`, which is public, and has to be public, since it's intended to be the base class for test fixtures (*); IIRC NUnit requires test fixture classes to be public.

(*) admittedly, it's a very questionable design... I guess we could revisit it at some point

thomaslevesque: Actually, forget what I just said. I can make the methods where it's exposed internal as well.

FakeItEasy.sln
thomaslevesque: Ah, good catch. It was in the `src` folder in the solution, and I didn't realize it wasn't in the physical `src` folder...

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.csproj
thomaslevesque: Fair enough. It made sense when the NuGet package was built my MSBuild, but not anymore

src/FakeItEasy.Analyzer/packages.config
thomaslevesque: Ah, yes, forgot about it

thomaslevesque:
> @thomaslevesque I've done another pass over the changes. A few more comments, but all very minor.

Thanks @adamralph!

I made the changes and rebased my branch. I'll squash everything once you have reviewed the last commit.

adamralph:
@thomaslevesque :+1: LGTM.

Regarding the squashing, the only thing I'd like to have in a separate commit is the upgrade to StyleCop.MSBuild 4.7.51. The easiest way to do that is probably to merge https://github.com/FakeItEasy/FakeItEasy/pull/654 first, and then rebase onto master and squash everything into a single commit.

thomaslevesque:
> Regarding the squashing, the only thing I'd like to have in a separate commit is the upgrade to StyleCop.MSBuild 4.7.51. The easiest way to do that is probably to merge #654 first, and then rebase onto master and squash everything into a single commit.

OK, I'll merge #654 first

adamralph:
@thomaslevesque I did a smoke test of the NuGet package by installing it in a sample console app and I got this:

> An instance of analyzer FakeItEasy.Analyzer.UnusedCallSpecificationAnalyzer cannot be created from c:\users\adam\documents\visual studio 2015\Projects\ConsoleApplication1\packages\FakeItEasy.Analyzer.2.0.0\analyzers\dotnet\cs\FakeItEasy.Analyzer.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=1.2.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' or one of its dependencies. The system cannot find the file specified..

I guess it's because `FakeItEasy.Analyzer.nuspec` needs to list all the packages in `FakeItEasy.Analyzer/packages.config` (apart from StyleCop.MSBuild) as dependencies. This wasn't a problem in the template since it was packing against the `.csproj` which automatically pulls out the dependencies from `packages.config`, but we are packing against a `.nuspec`. I can't remember why we went down the `.nuspec` route instead of `.csproj`. We could potentially switch, but I'd rather deal with that in a separate refactoring PR. For the purposes of this PR it's probably easier to list the dependencies.

thomaslevesque:
Oops!

> This wasn't a problem in the template since it was packing against the .csproj which automatically pulls out the dependencies from packages.config

I'm not sure what you mean by that... There was already a .nuspec in the template, and it was used to generate the package:

```
  <Target Name="AfterBuild">
    <GetAssemblyIdentity AssemblyFiles="$(OutDir)\$(AssemblyName).dll">
      <Output TaskParameter="Assemblies" ItemName="AnalyzerAssemblyInfo" />
    </GetAssemblyIdentity>
    <Exec Command="&quot;$(SolutionDir)packages\NuGet.CommandLine.3.3.0\tools\NuGet.exe&quot; pack Diagnostic.nuspec -NoPackageAnalysis -Ve
rsion %(AnalyzerAssemblyInfo.Version) -OutputDirectory ." WorkingDirectory="$(OutDir)" LogStandardErrorAsError="true" ConsoleToMSBuild="true
">
      <Output TaskParameter="ConsoleOutput" PropertyName="OutputOfExec" />
    </Exec>
  </Target>
```

I'll try to add the dependencies to the package

adamralph:
> There was already a .nuspec in the template, and it was used to generate the package:

Yes, but in the template, `nuget.exe` was invoked against the `.csproj` instead of the `.nuspec`. In that case, it still uses the nuspec for some metadata, but it gets other metadata from the `.csproj`, such as the version (from `AssemblyInfo.cs`) and the dependencies (from `packages.config)`, etc.

adamralph:
Sorry, ignore my last comment. I didn't read your snippet carefully enough. I'm not really sure how the original package could have worked then.

thomaslevesque:
> I'm not really sure how the original package could have worked then.

I don't know either... I had it working at some point, but now I get the same error as you

thomaslevesque:
OK, it doesn't seem to be the right way to do it... I'm getting the same error, and it adds the references to the project itself, which we don't want.

At some point I updated the analyzer's dependencies to the latest version; looks like it was a mistake. I'll look at another project (Code Cracker) to see how they did it.

thomaslevesque:
OK, managed to fix it. From what I understand, the code analysis packages I was referencing were too recent for that version of Visual Studio. The documentation leaves a lot to be desired, because I didn't find any explanation about that anywhere :confused: 

I downgraded everything to 1.1.0. Only thing I'm not sure of is whether it will work on VS2015 without update 1. I had to use 1.1.0 rather than 1.0.0, because the meta-analyzer had a bug that caused the analyzer to fail to build in VS2015.1 (see https://github.com/dotnet/roslyn/issues/7276). If that's an issue, I can remove the meta-analyzer.

(I added a new commit so that you can see the changes, I'll squash again before the merge)

thomaslevesque:
> Only thing I'm not sure of is whether it will work on VS2015 without update 1.

OK, apparently it won't. VS2015.0 comes with Roslyn 1.0.

> If that's an issue, I can remove the meta-analyzer (that's what Code Cracker did, apparently).

Actually I was mistaken; Code Cracker still uses the meta-analyzer (but it doesn't work on VS2015.1, gives error AD0001, which is configured to be a warning rather than an error)

adamralph:
Does this mean each release of our analyzer has to be tied to a particular VS update? With no forwards compatibility?

What is the meta-analyzer?

thomaslevesque:
> Does this mean each release of our analyzer has to be tied to a particular VS update? With no forwards compatibility?

No, if we target Roslyn 1.0 it will work with all releases of Visual Studio (starting with VS2015 of course). It is forward-compatible.

> What is the meta-analyzer?

It's an analyzer that analyzes analyzers :wink:. Basically, it makes sure you write the analyzer correctly, follow the conventions, etc

I think the best course is to:
- target Roslyn 1.0, so that all VS2015 users can use the analyzer
- keep the meta-analyzer, and configure error AD0001 to be a warning ([like this](https://github.com/code-cracker/code-cracker/blob/ece9e3915ff53e5ea5f952502694d1c5d5577bb3/src/Common/CodeCracker.Common/CodeCracker.Common.csproj#L18))

adamralph:
OK, thanks for the explanation.

If we were to target Roslyn 1.1 (i.e. VS2015.1 upwards), would that remove the meta-analyzer error/warning?

thomaslevesque:
> If we were to target Roslyn 1.1 (i.e. VS2015.1 upwards), would that remove the meta-analyzer error/warning?

Yes. That's what I did in my last commit, and I'm not getting the error.

thomaslevesque:
(Actually I was confused about the AD0001 error/warning. It is a warning, not an error, but since we have enabled "warnings as errors", it's reported as an error)

adamralph:
In that case I vote for us to support VS2015.1 upwards. Update 1 has been around for 6 months now and has proven its stability. I think it's fair to call VS2015.0 obsolete and not worry about supporting it.

tests/FakeItEasy.Analyzer.Tests/FakeItEasy.Analyzer.Tests.csproj
```diff
@@ -0,0 +1,148 @@
...
+      <Name>FakeItEasy.Analyzer</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Properties\" />
+    <Folder Include="Verifiers\" />
+  </ItemGroup>
```
adamralph: I know you're going to hate me for this nitpick :grin:, but both of these folders are empty and redundant, so this `ItemGroup` can be removed.

adamralph:
Apart from that one last nitpick, I think this is ready to go (assuming we all agree to support only VS2015.1 upwards).

blairconrad:
I've not taken the time to look over the new changes, but got the gist of the analyzer core and was happy with it. I trust both of you to dot your tees and cross your eyes, so don't hold up the merge for me.
Supporting VS2015.1+ sounds fine to me.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.nuspec
```diff
@@ -0,0 +1,20 @@
+﻿<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+  <metadata>
+    <id>FakeItEasy.Analyzer</id>
+    <version>0.0.0</version>
+    <title>FakeItEasy.Analyzer</title>
+    <authors>Thomas Levesque, FakeItEasy contributors</authors>
+    <description>Provides diagnostic analyzers to warn about incorrect usage of FakeItEasy.</description>
```
adamralph: Perhaps we should add something like "Works in VS 2015 Update 1 and later versions."?

adamralph: rewording: "Works in Visual Studio 2015 Update 1 or later."

tests/FakeItEasy.Analyzer.Tests/FakeItEasy.Analyzer.Tests.csproj
thomaslevesque: Oops! I'll remove them

thomaslevesque:
> In that case I vote for us to support VS2015.1 upwards. Update 1 has been around for 6 months now and has proven its stability. I think it's fair to call VS2015.0 obsolete and not worry about supporting it.

OK, nothing to change then ;)

I removed the empty folders and ItemGroup, added the remark about VS2015 Update 1 to the nuspec, and squashed everything

thomaslevesque:
Thanks for the feedback and the merge @adamralph!

adamralph:
Thanks for all the hard work @thomaslevesque!

Issue #175, opened 2013-09-26
New syntax for configuration of property setters


dskalski:
I have simple class
```c#
public class Simple
{
    public virtual int VirtualProperty { get; set; }
}
```
When i run (FakeItEasy.1.13.1)
```c#
var strict = A.Fake<Simple>(options => options.Strict());
A.CallTo(() => strict.VirtualProperty).CallsBaseMethod();
strict.VirtualProperty = 999;
```
I get an error
Call to non configured method "set_VirtualProperty" of strict fake.

And I have to
```c#
var strict = A.Fake<Simple>(options => options.Strict());
A.CallTo(strict).Where(a => a.Method.Name == "get_VirtualProperty").CallsBaseMethod();
A.CallTo(strict).Where(a => a.Method.Name == "set_VirtualProperty").CallsBaseMethod();
strict.VirtualProperty = 999;
```
Does CallBaseMethod () works for virtual property? What am I doing wrong

---
*Update*

The proposed syntax is:

```C#
A.CallToSet(() => foo.Bar)…
A.CallToSet(() => foo.Bar).To(42)…
A.CallToSet(() => foo.Bar["the answer"]).To(42)…
A.CallToSet(() => foo.Bar).To(() => A<int>.That.IsGreaterThan(0)))…
A.CallToSet(() => …).WithAnyArguments()…
A.CallToSet(() => …).WhenArgumentsMatch(call => …)…
```

These methods would allow configuration of the call in the same way that is currently provided by

```C#
A.CallTo(foo).Where(a => a.Method.Name == "set_Foo" ... )...
```

adamralph:
Are we seeing any light at the end of the tunnel for this?

blairconrad:
Oh, man. I just reread. I still wish we had a better syntax, but am not sure that anything has come along that helps.
Unless we can Roslyn it up, but that's currently beyond me…

adamralph:
I don't see how Roslyn can help...

blairconrad:
That's probably because I don't yet understand how Roslyn works!
I figured we'd be able to use the compiler as a service to notice when we had a set action (as a lambda, not an expression) inside an `A.CallTo` and, using the fancy fancy syntax tree inspection powers, rewrite the darn thing so it was something that our current rule builders would understand, maybe even directly to `A.CallTo(fake).Where(x => x.Method.Name == "set_Foo")` and so on.
[...]

Issue #637, opened 2016-03-31
Chained ordered assertions


adamralph:
E.g.

```C#
A.SequenceOfCalls(
    A.CallTo(() => fake.Bar(2)).MustHaveHappened(),
    A.CallTo(() => fake.Bar(1)).MustHaveHappened());
```

...with, perhaps, a better method name.

@blairconrad has [proved that this can work](https://github.com/blairconrad/FakeItEasy/tree/different-ordered-assertions) simply by adding syntactic sugar to the `A` class.

---
*update*

The new syntax will be:

```C#
A.CallTo(() => fake.Bar(3)).MustHaveHappened(Repeated.Exactly.Twice)
    .Then(A.CallTo(() => fake.Bar(2)).MustHaveHappened())
    .Then(A.CallTo(() => fake.Bar(1)).MustHaveHappened());
```

- [ ] Remove #602 from the 2.0.0 milestone and label `wontfix`

adamralph:
let's compare to 1.x:

```C#
using (var scope = Fake.CreateScope())
{
    ...
    using (scope.OrderedAssertions())
    {
        ...
    }
}
```

there's no A in there, which means this would be no worse for 2.x:

```C#
In.Order(
    A.CallTo(() => fake.Bar(2)).MustHaveHappened(),
    A.CallTo(() => fake.Bar(1)).MustHaveHappened());
```
...or anything else not beginning with `A`.

However, bringing this API in "from the wild" and onto A is compelling...

adamralph:
https://twitter.com/adamralph/status/715424646653456384

thecodejunkie:
I do like `A.SequenceOfCalls` to be honest +1

M-Zuber:
I'd have to agree with @TheCodeJunkie .
Its descriptive and not even that long

cecilphillip:
A.OrderedSequence ?

adamralph:
@cecilphillip we did consider variants of `A.OrderXXX` but the problem is the vowel. It should be "an orderXXX"

adamralph:
@thecodejunkie @M-Zuber @cecilphillip thanks for chiming in.

I've been having some more thought about this myself. One of the main drivers of the API design of FakeItEasy has always been "fluency" which, love it or hate it, is the crafting of statements which read very similarly to English sentences:

```C#
var foo = A.Fake<IFoo>();                     // "foo" is a fake "IFoo".
[...]

PR #643, opened 2016-04-06
Remove NextCall

### LINKED ISSUE #642: Remove NextCall.To()
Only known uses are

* "Specifying a call by example" - http://fakeiteasy.readthedocs.org/en/stable/specifying-a-call-to-configure/#specifying-a-call-by-example
* a way to configure property setters, which will be made redundant by #175
* VB.NET syntactic sugar http://fakeiteasy.readthedocs.org/en/stable/assertion/#vbnet - an edge case feature not worth supporting

blairconrad:
Fixes #642.

If you prefer to wait until #175 is merged fixed, I completely understand.


blairconrad:
Also, I gotta say, I _loved_ being able to change the docs at the same time.

blairconrad:
> Also, I gotta say, I _loved_ being able to change the docs at the same time.

… if only the release notes were in the source too. Le sigh. :grinning:

thomaslevesque:
I think you can also remove the following classes:
- `RecordingCallBuilder`
- `RecordingCallRule`
- `RecordedCallRule`

They exist only to support `NextCall.To`, as far as I can tell.

blairconrad:
Oh, for the love of Pete, @thomaslevesque. You're absolutely right.
Last night I was thinking about what could go with `NextCall`, and I _completely_ forgot to do anything about it this morning.
So glad you're here.

thomaslevesque:
> If you prefer to wait until #175 is merged fixed, I completely understand.

I see no reason to wait. It will make it easier to implement #637

blairconrad:
I see no reason to wait either. I just didn't want to be difficult.

blairconrad:
Anybody see a need to keep `FakeManager.DoNotRecordCall` and `IgnoreCallInRecording`? They're public, but probably were only ever used by `RecordingCallRule`.

thomaslevesque:
> Anybody see a need to keep FakeManager.DoNotRecordCall and IgnoreCallInRecording? They're public, but probably were only ever used by RecordingCallRule.

I'm :+1: for removing them

adamralph:
:fire: 'em

blairconrad:
Don't merge! I haven't removed `DoNotRecordCall` yet.
Also, once that's gone, `IInterceptedFakeObjectCall` is no longer pulling any weight. Not sure if it's worth keeping in order to distinguish it from parent `IWritableFakeObjectCall`. How far do we want to take this? The name `IInterceptedFakeObjectCall` is a nice, intention-revealing one.

blairconrad:
I removed `DoNotRecordCall` in a separate commit, in case we decide not to bother. Ditto `IWritableFakeObjectCall`, which was no different from `IInterceptedFakeObjectCall` once `DoNotRecordCall` was removed.
If you'd rather see another approach, or a squash, I'm more than happy to change anything.

thomaslevesque:
> Ditto IWritableFakeObjectCall, which was no different from IInterceptedFakeObjectCall once DoNotRecordCall was removed.

:+1:

adamralph:
@blairconrad if you're all done now, I'd vote for squashing the last 3 "remove" commits.

blairconrad:
Oh, I probably should've commented. Squashed. Thanks!

blairconrad:
Thanks for the merge, @thomaslevesque. And for doing my work in spotting the extra classes that could be removed!

blairconrad:
@jeremymeng, thanks for picking up on the leftover files. And @adamralph for cleaning up.

PR #605, opened 2016-02-22
Implement new ordered call assertion system

thomaslevesque:
Fixes #602

- [x] New `.InOrder(context)` syntax after `MustHaveHappened()`
- [x] Specs for basic cases
- [x] More test cases
  - [x] Success cases
  - [x] With repeat constraint
  - [x] With `NextCall.To` (uses `RecordedCallRule`)
- [x] Remove code related to the old ordered assertion system

PR #646, opened 2016-04-09
Chained ordered assertions

### LINKED ISSUE #637: Chained ordered assertions

blairconrad:
Fixes #637.


blairconrad:
There's some unpleasantness around `AfterMustHaveHappenedConfiguration` and `IAfterMustHaveHappenedConfiguration`. The latter is empty, and nearly useless, except that it allows us to ensure that the `Then` calls can't be nested inside one another. Also, we tend to return interfaces from our methods.
As implemented, the interface requires a downcast to `AfterMustHaveHappenedConfiguration` in `SubsequentOrderedCallAssertion.Then`. It all feels a little gross, but I couldn't figure out another way to access `CheckCallHappenedInOrder` without exposing it, and there's no reasons for clients to see that.

We _could_ remove the interface and just return an `AfterMustHaveHappenedConfiguration` from `MustHaveHappened`. Then the cast goes away and we don't have an empty interface not pulling very much weight. What do you guys think? Or if you see an even better way around the problem, I'm all ears.

docs/ordered-assertions.md
```diff
@@ -65,15 +65,13 @@ public void Should_start_work_within_unit_of_work()
 
     // Assert
     var sequentialCallContext = A.SequentialCallContext();
```
thomaslevesque: This line should also be removed

docs/ordered-assertions.md
```diff
@@ -65,15 +65,13 @@ public void Should_start_work_within_unit_of_work()
...
+    A.CallTo(() => unitOfWorkFactory.BeginWork()).MustHaveHappened()
+        .Then(A.CallTo(() => usefulCollaborator.JustDoIt()).MustHaveHappened())
+        .Then(A.CallTo(() => unitOfWork.Dispose()).MustHaveHappened());
 }
 ```
 
-In the assertion phase of the test we need to create a "sequential call context" that records the order of the calls that it's examined while verifying that they occurred.
-
-Then to do the assertions we use the regular `MustHaveHappened` call, but chain on `InOrder(sequentialCallContext)` so the assertion verifies not only that that call happened, but that it occurred in the right order _relative other calls that have been asserted against the same sequential call context_.
+The regular `MustHaveHappened` calls are made, but the results are chained together using `Then`, which verifies not only that that call happened, but that it occurred in the right order relative other calls that have been asserted in the same chain.
```
thomaslevesque: relative _to_ other calls
(I think)

thomaslevesque:
> There's some unpleasantness around `AfterMustHaveHappenedConfiguration` and `IAfterMustHaveHappenedConfiguration`. [...]

I can't see an easy way around it, but it doesn't bother me too much.

tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
```diff
@@ -173,27 +222,40 @@ public interface IHaveOneGenericParameter
             "And a call on the Fake, passing argument 2"
                 .x(() => fake.Bar(2));
 
-            "And a call-ordering context"
-                .x(c => sequentialCallContext = A.SequentialCallContext());
+            "When I expect a call with argument 1 to have been made twice"
+                .x(() => exception = Record.Exception(() => lastAssertion = A.CallTo(() => fake.Bar(1)).MustHaveHappened(Repeated.Exactly.Twice)));
```
thomaslevesque: Is it really necessary to record the exception here, since you don't expect one to be thrown?

tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
```diff
@@ -173,27 +222,40 @@ public interface IHaveOneGenericParameter
...
+            "When I expect a call with argument 1 to have been made twice"
+                .x(() => exception = Record.Exception(() => lastAssertion = A.CallTo(() => fake.Bar(1)).MustHaveHappened(Repeated.Exactly.Twice)));
+
+            "And then a call with argument 2"
+                .x(() => exception = exception ?? Record.Exception(() => lastAssertion = lastAssertion.Then(A.CallTo(() => fake.Bar(2)).MustHaveHappened())));
 
-            "When I expect the call with argument 1 to have been made at least twice"
-                .x(() => A.CallTo(() => fake.Bar(1)).MustHaveHappened(Repeated.AtLeast.Twice).InOrder(sequentialCallContext));
+            "And then a call with argument 3"
+                .x(() => exception = exception ?? Record.Exception(() => lastAssertion.Then(A.CallTo(() => fake.Bar(3)).MustHaveHappened(Repeated.Exactly.Once))));
```
thomaslevesque: same here

thomaslevesque:
I have a few minor remarks, but on the whole the PR looks great!

docs/ordered-assertions.md
blairconrad: Whoops! Yes.

docs/ordered-assertions.md
blairconrad: Yes, I do think that's better. Thanks.

blairconrad: Yes, I think it's better with "to".

tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
blairconrad: I see your point. For some reason, I was thinking that storing the first exception was a good idea, but just letting the unexpected ones fly is probably better. Will make the change throughout.

docs/ordered-assertions.md
blairconrad: I'll make the same change in IOrderableCallAssertion.cs and AfterMustHaveHappenedConfiguration.cs

blairconrad:
@thomaslevesque, thanks for the comments. Excellent catch for the problems in the docs.
I think I've addressed everything (by amending the previous commits - it makes it harder for you to rereview, I think. In the future, would you prefer additional commits which are later squashed?).

thomaslevesque:
> In the future, would you prefer additional commits which are later squashed?

Yes, I think it would be better (except for trivial changes)

thomaslevesque:
The changes look good to me :+1:

I'll wait a bit before merging in case @adamralph has other remarks.

samples/FakeItEasy.Examples.VB/AssertingCalls.vb
```diff
@@ -13,4 +13,11 @@ Public Class AssertingCalls
 
         A.CallTo(Function() factory.CreateWithColor(A(Of String).Ignored)).MustHaveHappened()
     End Sub
+
+    Public Sub Asserting_ordered_function_calls()
+        Dim factory = A.Fake(Of IWidgetFactory)()
+
+        A.CallTo(Function() factory.CreateWithColor(A(Of String).Ignored)).MustHaveHappened().Then(
```
adamralph: A nitpick, and I'm not sure what's idiomatic in VB, but I would expect the line break to be before `.Then(...`

src/FakeItEasy/Configuration/AfterMustHaveHappenedConfiguration.cs
```diff
@@ -17,11 +17,38 @@ public AfterMustHaveHappenedConfiguration(FakeManager fakeManager, ICallMatcher
...
+            private readonly SequentialCallContext context;
+
+            public SubsequentOrderedCallAssertion(SequentialCallContext context)
+            {
+                this.context = context;
+            }
+
+            public IOrderableCallAssertion Then(IAfterMustHaveHappenedConfiguration nextCall)
+            {
+                ((AfterMustHaveHappenedConfiguration)nextCall).CheckCallHappenedInOrder(this.context);
```
adamralph: Something's off here. Between the public invocation point (`IAfterMustHaveHappenedConfiguration.Then()`) and this cast, there's nothing which guarantees that this cast will work. I guess `CheckCallHappenedInOrder` isn't appropriate to be public on `IAfterMustHaveHappenedConfiguration` which suggests that some other object should have that responsibility.

Incidentally, is there any reason for the parameter name to differ in these signatures?

* `IAfterMustHaveHappenedConfiguration.Then(IAfterMustHaveHappenedConfiguration nextAssertion)`
* `IOrderableCallAssertion.Then(IAfterMustHaveHappenedConfiguration nextCall)`

tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
```diff
@@ -1,169 +1,210 @@
...
+                    A.CallTo(() => fake.Bar(1)).MustHaveHappened(Repeated.Exactly.Twice)
+                        .Then(A.CallTo(() => fake.Bar(2)).MustHaveHappened())
+                        .Then(A.CallTo(() => fake.Bar(3)).MustHaveHappened())));
 
-            "And I expect the call with argument 2 to have been made next"
-                .x(() => exception = Record.Exception(() => A.CallTo(() => fake.Bar(2)).MustHaveHappened().InOrder(sequentialCallContext)));
-
-            "Then the assertion should succeed"
-                .x(() => exception.Should().BeNull("because the assertion should have succeeded"));
+            "Then the assertion should pass"
```
adamralph: For consistency of language between the steps, I would expect either "When I assert..." or "Then the expectation...".

tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
```diff
@@ -1,169 +1,210 @@
...
-
-            "And I expect the call with argument 1 to have been made next"
-                .x(() => exception = Record.Exception(() => A.CallTo(() => fake.Bar(1)).MustHaveHappened().InOrder(sequentialCallContext)));
+            "When I expect a call with argument 1 to have been made twice and then a call with argument 2 and then a call with argument 3"
+                .x(() => exception = Record.Exception(() =>
+                    A.CallTo(() => fake.Bar(1)).MustHaveHappened(Repeated.Exactly.Twice)
+                        .Then(A.CallTo(() => fake.Bar(2)).MustHaveHappened())
+                        .Then(A.CallTo(() => fake.Bar(3)).MustHaveHappened())));
 
             "Then the assertion should fail"
```
adamralph: Same as https://github.com/FakeItEasy/FakeItEasy/pull/646/files#r59128761

tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
```diff
@@ -1,169 +1,210 @@
...
+            "When I expect a call with argument 1 to have been made on the first Fake, then on the second, and then a call with argument 2 to be made on the first Fake"
+                .x(() => exception = Record.Exception(() =>
+                    A.CallTo(() => fake1.Bar(1)).MustHaveHappened()
+                        .Then(A.CallTo(() => fake2.Bar(1)).MustHaveHappened())
+                        .Then(A.CallTo(() => fake1.Bar(2)).MustHaveHappened())));
 
-            "Then the assertion should succeed"
-                .x(() => exception.Should().BeNull("because the assertion should have succeeded"));
+            "Then the assertion should pass"
+                .x(() => exception.Should().BeNull());
```
adamralph: Same as https://github.com/FakeItEasy/FakeItEasy/pull/646/files#r59128761

src/FakeItEasy/Configuration/AfterMustHaveHappenedConfiguration.cs
adamralph: Aha, this is what you referred to in https://github.com/FakeItEasy/FakeItEasy/pull/646#issuecomment-207685665

adamralph:
@blairconrad I've finished my first pass over the changes. It's great work!

My comments are all nitpicks apart from the casting issue. I'd like us to investigate a way around that before merging.

tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
```diff
@@ -204,33 +258,52 @@ public interface IHaveOneGenericParameter
...
+                .x(() => baz.BazMethod());
+
+            "When I expect the qux call to have been made"
+                .x(() => lastAssertion = A.CallTo(() => qux.QuxMethod()).MustHaveHappened());
+
+            "And I express the same expectation again"
+                .x(() => exception = Record.Exception(() => lastAssertion.Then(A.CallTo(() => qux.QuxMethod()).MustHaveHappened())));
+
+            "Then the expectation should fail"
+                .x(() => exception.Should().BeAnExceptionOfType<ExpectationException>());
```
adamralph: I stopped adding references to my "expect vs assert" comment about step naming after a couple of cases :smile: 

The language is consistent in this scenario, but I think I'd prefer "Then the expectation should not be met", an expectation can't really "fail".

src/FakeItEasy/Configuration/AfterMustHaveHappenedConfiguration.cs
thomaslevesque: > Something's off here. Between the public invocation point (IAfterMustHaveHappenedConfiguration.Then()) and this cast, there's nothing which guarantees that this cast will work. I guess CheckCallHappenedInOrder isn't appropriate to be public on IAfterMustHaveHappenedConfiguration which suggests that some other object should have that responsibility.

It's the same issue as with `ISequentialCallContext.CheckNextCall`. We don't want to make it public, but it has to be... The alternative is to expose the `AfterMustHaveHappenedConfiguration ` class directly, but it's consistent with the rest of the API where we usually expose interfaces.

A possible solution is to expose the method in the interface, but hide it with `[EditorBrowsable(Never)]` and document it to be "conceptually internal", as [in #605](https://github.com/FakeItEasy/FakeItEasy/pull/605/files#diff-00e69038fb19194983b078aa099c04ceR13)

samples/FakeItEasy.Examples.VB/AssertingCalls.vb
blairconrad: VB doesn't require line termination characters, so can't always tell when a statement ends. Breaking the line before `.Then` confuses the compiler, so I hoisted the `(` (and preceding characters) up to this line. I am not sure what is more idiomatic in VB, but will switch to the [line confinuation character `_`](https://msdn.microsoft.com/en-us/library/ba9sxbw4.aspx?f=255&MSPPError=-2147217396) instead.

src/FakeItEasy/Configuration/AfterMustHaveHappenedConfiguration.cs
blairconrad: @adamralph, the parameter value differs because I'd had all `nextCall`s and then saw your multi-step example where you called these items "assertions" and I missed one class in the conversion. Well spotted. Will make them all assertions.

blairconrad: About the downcast, exposing `CheckCallHappenedInOrder` and asking people not to use it is my least favourite of the options. It makes for a dirty API (even though we would be going to lengths to keep the method from being _easily_ accessible). If there's going to be an irregularity, and notes like "don't use this, I mean it", I'd rather they were confined to the internals.

We have liked to return interfaces, but I don't see a compelling reason to when we have a single implementation and there's an advantage to not returning an interface. We could claim it's for extensibility, but I'm not sure there's a huge advantage there. While we _may_ want to add another method to the class, that can be done with a class as well as an interface (and is technically breaking for an interface, but not for a class, so long as the method isn't abstract), I don't think there's much we can do to the ordered assertions feature without requiring a rework of the API (something that would allow us to know when the last ordered assert is made). 

As noted before, we could return the concrete class. I'm not sure what's so terrible about this. If we decide to add another method, we could. If we decide to subclass and split the implementation, we could. If making a method virtual after the fact is considered a breaking change (I don't think it should be), we could make `Then` virtual now.

A variation on "return an interface" or "return a concrete class" is to "return an abstract class", which is nearly the same interface, from the outside. Then `CheckCallHappenedInOrder` can stay on the class, internal. It doesn't bother the users, and we don't have to have a scary cast. I'd had this implemented initially, but didn't see a technical advantage over either returning an interface or a concrete class. In this case, the abstract class would take the place of `IAfterMustHaveHappenedConfiguration`.

I prefer returning some sort of class to returning an interface.

However, to make returning an interface and casting, more palatable, if that's a front-runner for you guys, we could do the following:

1. comments explaining why we're casting and not to add additional production-ready implementations of `IAfterMustHaveHappenedConfiguration`
2. to augment the comments, we could even add a unit test to ensure that nobody adds an implementation unawares

tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
blairconrad: Yes. I did not do a good job with this. The documentation calls things "assertions". I will make everything like this.

tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
blairconrad: "not be met" is better with an expectation, but I think we should adopt the ubiquitous language and switch to "assertion" everywhere. Assuming we are all in agreement, then I think saying that "The assertion should fail" is acceptable, yes? (Likewise "pass" when things go better.)

src/FakeItEasy/Configuration/AfterMustHaveHappenedConfiguration.cs
thomaslevesque: > As noted before, we could return the concrete class. I'm not sure what's so terrible about this.

It's not terrible; it's just inconsistent with what we do elsewhere in the API, which bothers me slightly, but isn't a very serious problem.

An abstract class with a protected constructor would be fine. I would suggest a different name, though; `AfterMustHaveHappenedConfiguration` doesn't fit very well IMO, because it's not used to configure anything, but only to chain assertions. The way I see it, what this class really represents is a successful assertion (because if the assertion fails regardless of the order, `MustHaveHappened()` will throw). So we could call it `AssertedCall`, or `SuccessfullCallAssertion`, or something along those lines.

> 1. comments explaining why we're casting and not to add additional production-ready implementations of IAfterMustHaveHappenedConfiguration
> 2. to augment the comments, we could even add a unit test to ensure that nobody adds an implementation unawares

This doesn't really help; it could prevent _us_ from accidentally adding other implementations in FakeItEasy itself, but it wouldn't prevent the user from doing so. Since the interface is public, anyone can implement it and pass an instance of their implementation to `Then`. I hadn't realized that initially, but now that I see it, I agree that the cast is an issue; it's a violation of the [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

blairconrad:
Uploaded changes for everything but the downcast. Before merge, I'd be inclined to squash the new commit into the two previous ones (or all into one, if that's preferable).

src/FakeItEasy/Configuration/AfterMustHaveHappenedConfiguration.cs
blairconrad: @thomasleveque, I'd also not considered people extending the interface. I'm glad you brought up Liskov. It's an excellent point.
So, unless we want to broaden the interface, it does seem that returning a class it the way to go. It's not common throughout FakeItEasy's API, but we have an uncommon situation here. Although I could try to play around with the code a little more. Maybe I've just not found the best implementation. In the meantime:

Your idea of returning an abstract class with protected constructor (in order to allow clients to extend the class, I assume) should resolve the problem.
At least if someone _did_ extend the class and pass an instance to `Then`, then we'd end up calling the internal, non-virtual `CheckCallHappenedInOrder`, which is fine, but in this case, I wonder why we'd provide an extensible type anyhow. There seems to be no benefit for users to provide an extension. Or do you see one?

I'm happy to the name. I just kept `AfterMustHaveHappenedConfiguration` from the old implementation because it didn't bother me, and the class did about the same thing - it seemed that providing a `Then` method is not so different from providing an `InOrder`. 

Your suggested names are serviceable, but I'd consider a different tack. One reason for the class to be exposed is to distinguish it from the result of the `Then` to avoid the nested `Then`s. While thinking of that aspect, I thought of `UnorderedCallAssertion`, which a raw `MustHaveHappened` is. It's only by wrapping one in a `Then` or using the `Then` method on one that we come up with an ordered assertion. Then the class that I've called `SubsequentOrderedCallAssertion` could drop the `Subsequent`: once we start returning from `Then`, we've imposed an ordering.

blairconrad:
I [made a comment](https://github.com/FakeItEasy/FakeItEasy/pull/646#discussion_r59147325) that I think will be collapsed due to being on an outdated diff.

src/FakeItEasy/Configuration/AfterMustHaveHappenedConfiguration.cs
blairconrad: I cannot think of a less gross alternative to returning a class, either abstract or concrete.

thomaslevesque: > Your idea of returning an abstract class with protected constructor (in order to allow clients to extend the class, I assume) should resolve the problem.

Actually, no, I didn't intend to allow the client to extend the class (which wouldn't make much sense). I meant to say an _internal_ constructor, not a protected one.



> Your suggested names are fine, but I'd considered a different tack. One reason for the class to be exposed is to distinguish it from the result of the `Then` to avoid the nested `Then`s. While thinking of that aspect, I thought of `UnorderedCallAssertion`, which a raw `MustHaveHappened` is. It's only by wrapping one in a `Then` or using the `Then` method on one that we come up with an ordered assertion. Then the class that I've called `SubsequentOrderedCallAssertion` could drop the `Subsequent`: once we start returning from `Then`, we've imposed an ordering.

Sounds good!

blairconrad: Ah. I do prefer an internal constructor. I'm glad I asked.

Thanks for the name support. I'll get started on those changes.

blairconrad:
Gents, I've made `UnorderedCallAssertion` abstract, removing `IAfterMustHaveHappenedConfiguration`.
I am still of the opinion that we don't gain anything by having `UnorderedCallAssertion` abstract, and it requires an additional internal concrete class (`InitialOrderableCallAssertion`), but I'm not bothered by it so much, if you're content with the result.

thomaslevesque:
> I am still of the opinion that we don't gain anything by having UnorderedCallAssertion abstract

Perhaps you're right. I don't remember why I wanted to make it abstract. Perhaps to prevent the user from creating instances of it, but it can also be achieved by making the constructor internal. I don't mind if you prefer to make it concrete.

blairconrad:
As it turns  out, I forgot to make the constructor internal anyhow after I extracted the concrete class . Will collapse. Thanks.

blairconrad:
Collapsed.

adamralph:
:+1: LGTM. I :heart: that the downcast has gone away.

I was about to hit the merge button, but I though't just ask first: is there any downside to us returning the concrete type rather than an interface? Nothing springs to my mind.

thomaslevesque:
> I was about to hit the merge button, but I though't just ask first: is there any downside to us returning the concrete type rather than an interface? Nothing springs to my mind.

I don't think so. The only thing that annoys me is the inconsistency with what we do elsewhere in the API (we usually return interfaces), but it's better than the alternatives.

blairconrad:
I wish I'd just returned a class to begin with, as I'd been inclined to do. Sorry for the runaround.
I was worried about deviating from returning an interface for the same reason, "is there are downside?", I wondered.

In this case, where we really want to be the source of the assertion, and there's no benefit to the clients creating an implementation, I think the class is the better choice.

> we usually return interfaces

And I don't want to start something, but I wonder if we took a hard look at many of those places we might not say, "we really don't need for the clients to muck about with things, so a class might do as well", especially in the fluent API. But in general, I don't mind defaulting to interfaces.

Using an interface probably lets us be a little more flexible in case we want to redistribute responsibilities. Although in most cases, I don't worry so much about changing the actual types used to provide the fluent API.
Of course in this case, people could be saving away the `UnorderedCallAssertion` for multi-step assertions, and a change to the type would break them (unless they're using `var` or `IOrderableCallAssertion`.

So... do you want me to resquash into two commits, or just hit the big "squash it all" button?
I'm not fussy.
Oh. If you hit the "squash it all" button, do you have to make up a new commit message? Which will only have been seen by one person!? :scream:

adamralph:
> If you hit the "squash it all" button, do you have to make up a new commit message? Which will only have been seen by one person!? :scream: 

It takes the PR title as the first line of the commit message and then takes the commit message of each squashed commit and adds it as a bullet point.

If we're happy with that, and we all promise not to alter it, then we keep the option switched on. Otherwise, we'll have to switch it off.

adamralph:
I'll keep thinking about the interface vs class thing a little longer but I probably won't find an argument against it. If @thomaslevesque wants to push the merge (and squash) button in the meantime then I won't complain.

thomaslevesque:
> If @thomaslevesque wants to push the merge (and squash) button in the meantime then I won't complain.

I don't see a squash button...

adamralph:
image

This assumes we are all happy with https://github.com/FakeItEasy/FakeItEasy/pull/646#issuecomment-208400662

If not,  @blairconrad will have to squash.

blairconrad:
re squash button:

> It takes the PR title as the first line of the commit message and then takes the commit message of each squashed commit and adds it as a bullet point.
> If we're happy with that, and we all promise not to alter it, then we keep the option switched on. Otherwise, we'll have to switch it off.

I'm content for it to work that way, and for the option to stay, although I bet I'd see very few PRs whose title and individual commit messages I'd want to see treated that way. E.g., I don't think this one is a good candidate.

blairconrad:
I can squash. Do we want the original pair of commits where I set up the specs/docs, and then the second for the implementation (my preference) or would you like one, which I'm also fine with?

adamralph:
I think one commit is fine. I don't really see a need to separate the docs. Indeed, it's probably even nicer to have them in the same commit, otherwise the docs are effectively lying for at least one commit. WRT to the specs, I do like red commits followed by green, but IMHO this only makes sense when the altered specs can at least compile. For the acceptance tests for an API, this is often not the case.

---

> I bet I'd see very few PRs whose title and individual commit messages I'd want to see treated that way

I tend to agree, which means that the features has little use for us. Given that, and in the interest of preserving our four eyes rules, perhaps it's better to switch off the option.

blairconrad:
Yay! My first instinct appears to have been right, although I was just knee-jerking it.

adamralph:
@thomaslevesque any opposition to switching it off?

thomaslevesque:
One commit is fine by me.

> image

Ah, I see. It only appears after you click the "Merge pull request" button, and before you confirm. It's a nice feature, but it's a shame that it doesn't let us change the message.

> @thomaslevesque any opposition to switching it off?

No, switch it off if you like. We managed without it so far, so we won't really miss it.

blairconrad:
Pushed.

Thanks for the review!

adamralph:
> it's a shame that it doesn't let us change the message

It does. What I described is just the pre-population of the edit boxes. But changing the message would violate our four eyes rule.

I've switched the option off in our 3 of our repos.

PR #647, opened 2016-04-11
Fix invalid code in implicit creation options doc

thomaslevesque:
There were several errors in the example code

### DIFF
--- a/docs/implicit-creation-options.md
+++ b/docs/implicit-creation-options.md
@@ -61,21 +61,21 @@ example, if you wanted all Fakes to be Strict, you might write
 something like this:
 
 ```csharp
-class MakeEverythingStrictOptionsBuilder: IFakeOptionsBuilder
+class MakeEverythingStrictOptionsBuilder : IFakeOptionsBuilder
 {
     public bool CanBuildOptionsForFakeOfType(Type type)
     {
         return true;
     }
 
-    public void BuildOptions(Type typeOfFake, IFakeOptions options);
+    public void BuildOptions(Type typeOfFake, IFakeOptions options)
     {
-        options => options.Strict();
+        options.Strict();
     }
 
-    public int Priority
+    public Priority Priority
     {
-        get { Priority.Default; } // equivalent to value 0
+        get { return Priority.Default; } // equivalent to value 0
     }
 }
 ```


thomaslevesque:
Ah, I didn't realize editing the file directly on GitHub would create a branch in the FakeItEasy repo, instead of my fork... I'll recreate the PR if that's an issue.

docs/implicit-creation-options.md
```diff
@@ -61,21 +61,21 @@ example, if you wanted all Fakes to be Strict, you might write
...
     {
-        options => options.Strict();
+        options.Strict();
     }
 
-    public int Priority
+    public Priority Priority
     {
-        get { Priority.Default; } // equivalent to value 0
+        get { return Priority.Default; }
```
blairconrad: I think it's a good idea to retain the "equivalent to 0", or to note that up somewhere in else in the article, as was done in [Formatting Argument Values](http://fakeiteasy.readthedocs.org/en/latest/formatting-argument-values/) (but not, I see in [Custom Dummy Creation](http://fakeiteasy.readthedocs.org/en/latest/custom-dummy-creation/)) so users know what they're overriding, and that supplying a value of `0` won't help them if they are competing with another custom implementation that's using `Priority.Default`.
If you agree, we should probably fix Custom Dummy Creation. If not, we should fix Formatting Argument Values.

blairconrad:
Oh my goodness! It was so bad. Thank you for fixing it.
I'm not bothered by the branch at all. It's probably what community members would do, and I'd be happy to merge that from them. I've had many of my similar changes merged into other projects.

docs/implicit-creation-options.md
thomaslevesque: Ah, yes. Don't know why I removed the comment, I didn't intend to. I put it back (amended the commit).

thomaslevesque: > If you agree, we should probably fix Custom Dummy Creation. If not, we should fix Formatting Argument Values.

Instead of explaining how `Priority` works on every extension point page, perhaps we should created a Priority page?

blairconrad: @thomaslevesque, great idea!

adamralph:
> I'm not bothered by the branch at all. It's probably what community members would do

Other community members can't create branches in the FakeItEasy repo. Only we can do that as owners.

FYI the way to avoid this is to make the edit from your fork, and then PR back to the main repo, although the UI is clunky. You have to make your edit, then "propose change", which creates the branch and opens the "new PR" page. Then you have to cancel that, e.g. by going back to your fork's home page, then click "new PR" for the new branch. Alternatively you can create a new branch first in the UI, then edit that and push directly to it, then create the PR.

blairconrad:
> > I'm not bothered by the branch at all. It's probably what community members would do

> Other community members can't create branches in the FakeItEasy repo. Only we can do that as owners.

By "what community members would do" I meant that they'd use the GitHub UI in the same way. I don't think owners should be penalized due their having extra power. I'd be against a commit directly on master (as I accidentally made happen long ago before GitHub introduced this behaviour), but I don't mind these kinds of edits coming in on short-lived branches in FakeItEasy's repo.

Sorry for the confusion.

PR #649, opened 2016-04-11
removed redundant using and partial namespace

adamralph:
connects to #637

Just a couple of tiny things I noticed whilst poking around in the implementation of #637 for my own curiosity.


thomaslevesque:
:+1:

blairconrad:
Odd. Usually Resharper's all over me for those sorts of things. But I'm busy ignoring some of the "switch to C#6" warnings, so maybe my sensitivity is turned down.

PR #650, opened 2016-04-11
add docs items to solution

adamralph:
connects to #626

I realised that `mkdocs.yml` was the only top level item missing from the "Solution Items" folder, and then realised that, for consistency, we ought to include the docs in the solution too.


blairconrad:
Oh, man. You just bought us a lot of future work, adding all the new files. I know I'm going to forget.
I guess if this is how people like to operate, I've no objection, and the change appears to be sound.

thomaslevesque:
:+1:, it will make it easier to access the docs when working on the code

PR #652, opened 2016-04-11
added entry for @thomaslevesque to mailmap

adamralph:
...plus a commit to sort the entries.


blairconrad:
Always going that extra mile, @adamralph!

Issue #651, opened 2016-04-11
Consider removing SequenceNumber from ICompletedFakeObjectCall


blairconrad:
It's new, so nobody will miss it.
Is there any benefit to exposing it? Detriment?

adamralph:
:+1: to remove it. It's an internal implementation detail.

thomaslevesque:
If you can find an easy way to remove it, sure, it would certainly be better. But I don't think it will be easy... I already thought about how to do it, and couldn't come up with anything satisfying.

adamralph:
Ah, sorry, I thought it was something trivial to do.

So I humbly went and had a look at the code rather than commenting from afar: given that it's in the `FakeItEasy.Core` namespace and that no test writer will ever touch it directly, does it really matter?

blairconrad:
The `Invokes` and `ReturnsLazily` users will see it, is all.

blairconrad:
I see `Method`, `Arguments`, and `FakedObject` being useful to them. `SequenceNumber` less so.
I'm not saying there's that much harm, but once people see it, they may ask questions, and…

adamralph:
Ah yes, I see what you mean

![image](https://cloud.githubusercontent.com/assets/677704/14469459/f3e4d9d0-00e4-11e6-9df0-a427faa59524.png)

adamralph:
I suppose it *could* be useful. E.g. I might use `Invokes()` to record the sequence number of all my fake calls and then do some kind of fancy advanced ordered assertion. As usual, it's a need I'd always try to avoid having, but still...

adamralph:
In fact, this is a nice escape hatch for us for when someone asks how to do anything other than the very basic ordered assertion capability which we've propagated from 1.x to 2.x. It could be worth documenting as such and would be a good reason to keep the property.

blairconrad:
If people want to use Invokes to record sequence numbers, they can make up their own sequence numbers!

adamralph:
Hmmm.. true.

blairconrad:
I'd rather do something really helpful to support better ordering assertion (i.e. implement better ordering assertions) or say "this is what we have" than say "This is what we have. Plus this other thing."

Gimme a few minutes.

blairconrad:
Unsurprisingly, @thomaslevesque is right. It seems tough to get rid of. The old "make it a class trick" would be applicable, but I don't want to go to that well too often.

[...]

Issue #374, opened 2014-10-25
Review code with an eye to reducing type and member visibility (or existence!)


blairconrad:
While we have nothing to hide, having too many publicly visible types and members can be a drawback for a framework. For example:

1. too many public items can present too many options to clients, confusing them, and 
1. there is always a danger in changing public items, as clients may be relying on types continuing to exist, or on particular member signatures

This came to light during discussions around several changes in #373, where a new public type was added and some public constructors (for public types) were changed. In this case, the changes were discussed and deemed to be (likely) safe, since the affected members were of no use to clients. But if they're of no use to clients, why expose them? If they had been internal to begin with, we could've avoided some angst and confusion.

Thus, we should perhaps examine FakeItEasy's public types and public (and protected? probably not as a big a concern) on those types, with an eye to reducing the library's surface area without negatively affecting clients. And then we could act on the findings.

Suggested actions that could improve the library's API, in approximate descending order of preference:

1. removing items from the public API if they're no longer desirable (#410)
1. changing public types that are of no interest to the outside world from public to internal
1. changing public or protected members that are of no interest to the outside world to internal
1. for types that are used by the outside, considering providing a public interface or more focused type and changing the original to be internal
1. moving types that need to stay public into a specialized namespace (`Sdk`?) so they won't hit clients in the face when their IntelliSense kicks in after they type `FakeItEasy.`

We may want to make breaking changes here, but should not do so carelessly. If it's reasonable to preserver the user experience, we should do so.

_This text appeared when the issue was created, before we decided to make it part of the 2.0.0 milestone_:

> We may consider breaking changes for users, but in that case we should
> 
> 1. increase the major version number of the library
> 1. probably make as many [breaking changes](https://github.com/FakeItEasy/FakeItEasy/issues?q=is%3Aopen+is%3Aissue+label%3Abreaking) as we can in one release so we don't have to shock users more than once

I suggest starting a checklist in this issue to track what we've looked at, but spinning off individual issues for each change (or related set of changes) in order to increase discoverability in case someone is adversely affected by a change.

Okay. I've added [a Google Spreadsheet with all the visible named types exported from FakeItEasy](https://docs.google.com/spreadsheets/d/1fD83ph65YQwnzB0XxzjeNrOcZlfriYCC0syYDJIANvQ/edit?usp=sharing) (excluding those that come from Castle.Core). With notes.

blairconrad:
[...]

PR #658, opened 2016-04-12
[WIP]make FakeManager internal

adamralph:
*Currently for discussion only.*

So here's a thing. Again, whilst poking around wrt #651, I noticed that `FakeManager`, apart from one dubious call in the specs, can be internal.

Call the :cop:s! Am I committing heresy here? It is designed to be public?

### DIFF
--- a/src/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
@@ -4,7 +4,7 @@ namespace FakeItEasy.Core
     using System.Linq;
 
     /// <content>Auto fake property rule.</content>
-    public partial class FakeManager
+    internal partial class FakeManager
     {
         [Serializable]
         private class AutoFakePropertyRule
--- a/src/FakeItEasy/Core/FakeManager.EventRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.EventRule.cs
@@ -7,7 +7,7 @@ namespace FakeItEasy.Core
     using System.Reflection;
 
     /// <content>Event rule.</content>
-    public partial class FakeManager
+    internal partial class FakeManager
     {
         [Serializable]
         [SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable", Justification = "Would provide no benefit since there is no place from where to call the Dispose-method.")]
--- a/src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
@@ -5,7 +5,7 @@ namespace FakeItEasy.Core
     using FakeItEasy.Creation;
 
     /// <content>Object member rule.</content>
-    public partial class FakeManager
+    internal partial class FakeManager
     {
         [Serializable]
         private class ObjectMemberRule
--- a/src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
@@ -5,7 +5,7 @@ namespace FakeItEasy.Core
     using System.Reflection;
 
     /// <content>Property behavior rule.</content>
-    public partial class FakeManager
+    internal partial class FakeManager
     {
         private class PropertyBehaviorRule
             : IFakeObjectCallRule
--- a/src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
@@ -4,7 +4,7 @@ namespace FakeItEasy.Core
     using System.Linq;
 
     /// <content>Property setter rule.</content>
-    public partial class FakeManager
+    internal partial class FakeManager
     {
         [Serializable]
         private class PropertySetterRule
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -14,7 +14,7 @@ namespace FakeItEasy.Core
     /// by using the AddRule-method.
     /// </summary>
     [Serializable]
-    public partial class FakeManager : IFakeCallProcessor
+    internal partial class FakeManager : IFakeCallProcessor
     {
         private readonly LinkedList<CallRuleMetadata> allUserRulesField;
         private readonly CallRuleMetadata[] postUserRules;
--- a/src/FakeItEasy/Fake.cs
+++ b/src/FakeItEasy/Fake.cs
@@ -23,7 +23,7 @@ namespace FakeItEasy
         /// <returns>The fake object manager.</returns>
         [DebuggerStepThrough]
         [SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "object", Justification = "The term fake object does not refer to the type System.Object.")]
-        public static FakeManager GetFakeManager(object fakedObject)
+        internal static FakeManager GetFakeManager(object fakedObject)
         {
             return Facade.GetFakeManager(fakedObject);
         }
--- a/tests/FakeItEasy.Specs/CreationSpecs.cs
+++ b/tests/FakeItEasy.Specs/CreationSpecs.cs
@@ -71,7 +71,7 @@ namespace FakeItEasy.Specs
                 .x(() => fakes.Should().ContainItemsAssignableTo<ICollectionItem>());
 
             "And all items should be fakes"
-                .x(() => fakes.Should().OnlyContain(item => Fake.GetFakeManager(item) != null));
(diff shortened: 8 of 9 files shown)


thomaslevesque:
Looking at its public API, looks like most of it shouldn't be public. I don't imagine why a user would want to manipulate call rules explicitly. The only thing that might be useful IMO is `AddInterceptionListener`.

thomaslevesque:
Anyway, I like the idea. It would probably allow us to make lots of other things internal.

adamralph:
Tentatively added to the 2.0.0 milestone.

blairconrad:
You know how sometimes I say "don't wait for me"?
Could you wait for me this time?

blairconrad:
As a result of #374, in https://docs.google.com/spreadsheets/d/1fD83ph65YQwnzB0XxzjeNrOcZlfriYCC0syYDJIANvQ/edit#gid=0 I said this:

> Provides advanced functionality, which users may be relying upon. Constructor made internal in #428 (2.0)

Although based on your PR, assuming it builds, I can't see why we wouldn't make `FakeManager` internal.

blairconrad:
AHA!

I knew I was missing something. [`Fake.GetFakeManager`](https://github.com/FakeItEasy/FakeItEasy/pull/658/files#diff-4068e827d6df07a7b3477b60ceac125eR26). That's what means we can't make it internal. Unless we really want to take `FakeManager` away from people.

I am not in favour of most of this functionality being exposed, but we have to realize that taking the method may impact people.

I couldn't find any FakeManager usage in all of GitHub, but people [are using `GetFakeManager`](https://github.com/search?utf8=%E2%9C%93&q=GetFakeManager+-repo%3AFakeItEasy%2FFakeItEasy&type=Code&ref=searchresults).

thomaslevesque:
> I knew I was missing something. `Fake.GetFakeManager`. That's what means we can't make it internal. Unless we really want to take `FakeManager away from people`.

From the results of your search, I see only 2 use cases (one of the results has nothing to do with FakeItEasy):

- `FakeObjectType`: either to check the faked type, or check that the object is a fake (which is incorrect, because `GetFakeManager` will return null if it's not a fake)
- `AddInterceptionListener`: to record all calls made to fakes

These use cases seem reasonable to me, so I don't think we should drop support for them.

I didn't find any usage of `AddRuleFirst`, `AddRuleLast` or `RemoveRule` (except in a repo that contains code from an old version of FakeItEasy)

> I am not in favour of most of this functionality being exposed, but we have to realize that taking the method may impact people.

Well, we're already impacting people with many breaking changes in 2.0.0. But I agree that we shouldn't remove functionality that's actually being used (even if it's only by a few people).

Even though we want to make the API as easy to use as possible for most users who are just writing simple unit tests, we should keep in mind that FakeItEasy is also used by library authors to implement more advanced functionality. For instance [this repo](https://github.com/omnibs/service-ghost/blob/2e9dda30c18e5a4755fabbd3068464e2b601e58f/ServiceGhost.Core/GhostHelper.cs) seems to be doing pretty hairy stuff (using FakeItEasy to record all calls to an assembly).

blairconrad:
> `AddInterceptionListener`: to record all calls made to fakes

I gotta say, I don't understand why the `FakeManager` is needed for that. I took a quick look at the ServiceGhost code and couldn't understand what was going on that wouldn't've worked fine just with `options=>options.Wrapping(…).RecordedBy(…)`.

> `FakeObjectType`: either to check the faked type, or check that the object is a fake (which is incorrect, because GetFakeManager will return null if it's not a fake)

Seems a little odd, but harmless.

> Well, we're already impacting people with many breaking changes in 2.0.0. But I agree that we shouldn't remove functionality that's actually being used (even if it's only by a few people).

Oh, I didn't even mean to quite go that far. In general, I'd like to keep functionality that's being used, but if we decided not to, that's a possibility. I just wanted to make sure we were _deciding_ not to, not just removing it accidentally.

So it sounds like we're aiming for "make FakeManager and GetFakeManager internal, while preserving at least the ability to get a Fake's faked type, and _maybe_ a way to intercept every call, even though Blair's not sure it's necessary"?

blairconrad:
Oh, and ServiceGhost  doesn't seem to be in active development. It might be perfectly happy staying on old FakeItEasy.

thomaslevesque:
> Seems a little odd, but harmless.

Not harmless: if the object is not a fake, `Fake.GetFakeManager(obj).FakeObjectType` will throw a `NullReferenceException`. Anyway, we could easily add a `Fake.IsFake(object)` convenience method.

> I gotta say, I don't understand why the `FakeManager` is needed for that. I took a quick look at the ServiceGhost code and couldn't understand what was going on that wouldn't've worked fine just with `options=>options.Wrapping(…).RecordedBy(…)`.

I don't think it really does the same thing. `RecordedBy` takes a `ISelfInitializingFakeRecorder`, whereas `AddInterceptionListener` takes a `IInterceptionListener`. Apparently service-ghost uses the latter to time the calls: it starts a stopwatch in `OnBeforeCallIntercepted`, and stops it in `OnAfterCallIntercepted`. So it's really a different use case.

> Oh, and ServiceGhost doesn't seem to be in active development. It might be perfectly happy staying on old FakeItEasy.

True

blairconrad:
> Not harmless: if the object is not a fake, Fake.GetFakeManager(obj).FakeObjectType will throw a NullReferenceException. Anyway, we could easily add a Fake.IsFake(object) convenience method.

I was unclear. I wasn't advocating for their exact code, more that checking to see that an object is a Fake or that wanting to know what it's faking seems like a reasonable use case. (Although not one that I can imagine needing.)

Ah, I didn't find ServiceGhost's timing code. I see now how it's used. On that, I guess adding listeners could be useful, but I'd be inclined to do it as an `IFakeOptions`, were I to implement it afresh.

And I am not wild about providing the `IFakeObjectCallRule` to the `OnAfterCallIntercepted`. The only thing that I'd want a client to see in that is `NumberOfTimesToCall`, and even I'm not even sure there's value in that.

adamralph:
What's the general feeling on this? I don't want to delay 2.0 unnecessarily with a desire to "get all the breaking changes in now", but `FakeManager` is a particularly odorous part of the public API. Anything named "{something}manager" usually is. It would be nice to have the option to refactor it away during the lifetime of 2.x.

thomaslevesque:
I'm all for it, but we need to provide new ways to:
1. check that an object is a fake
2. intercept all calls made to fakes (or not... not sure how useful it really is)

adamralph:
Do we also need to provide a new way to [get the fake object type](https://github.com/jamiehumphries/FakeItEasy.Auto/blob/16f9d7b14875748490dce61bd3e21cce1c0fa7ed/FakeItEasy.Auto.Tests/TestHelpers/ObjectAssertionsExtensions.cs#L20)?

thomaslevesque:
> Do we also need to provide a new way to get the fake object type?

Maybe. I'm not sure how useful it really is.

blairconrad:
I think that getting the fake object type might be important to people. I've seen questions about it. On StackOverflow? I think it's odd, but that's me.

Of course, once we can do that, then figuring out whether an object is fake should be easy.

I'm in favour of hiding FakeManager. #175 is going to take me a little time anyhow, so if someone wants to take this issue, I say there's no better time than the present.
Otherwise, I suppose we could delay until I'm free and reassess.

thomaslevesque:
I just had an idea: couldn't we make all fakes implement an `IFake` interface (in addition to whatever type they're faking)?

Checking if an object is a fake could be done by testing the type:

    if (obj is IFake)

And we could expose `FakeObjectType` and `AddInterceptionListener` on that interface.

Regarding `AddInterceptionListener`, I'm still not sure we should keep it. The only project that seems to be using it isn't really using FakeItEasy for mocking, but rather as an interception mechanism, for which other tools (e.g. Castle.Core) are better suited.

godrose:
Check this project: https://github.com/godrose/Attest

On 15 April 2016 at 14:46, Thomas Levesque <notifications@github.com> wrote:

> I just had an idea: couldn't we make all fakes implement an IFake
> interface (in addition to whatever type they're faking)?
>
> Checking if an object is a fake could be done by testing the type:
>
> if (obj is IFake)
>
> And we could expose FakeObjectType and AddInterceptionListener on that
> interface.
>
> Regarding AddInterceptionListener, I'm still not sure we should keep it.
> The only project that seems to be using it isn't really using FakeItEasy
> for mocking, but rather as an interception mechanism, for which other tools
> (e.g. Castle.Core) are better suited.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/FakeItEasy/FakeItEasy/pull/658#issuecomment-210433118>
>



-- 
Best Regards,
Gena

adamralph:
@godrose can you be more specific?

adamralph:
I'm leaning towards closing this rather than delaying the 2.0.0 release for it. If, during the lifetime of 2.x, we manage to make `FakeManager` redundant as a public type, we can `[Obsolete]` it, and schedule it for removal in 3.0.0.

Yay or nay?

blairconrad:
I can live with the obsolete and subsequent removal plan.

PR #659, opened 2016-04-12
Add test cases for Should_be_registered_as_non_singleton

### LINKED ISSUE #632: ServiceLocatorTests.Should_be_registered_as_non_singleton tests nothing
As noted by @jeremymeng in #631.

[`NonSingletonTypes`](https://github.com/jeremymeng/FakeItEasy/blob/e39b9a23e3d2171e5ecd312903e9ad5e0b118941/tests/FakeItEasy.Tests/ServiceLocatorTests.cs#L45) is empty, providing no value as a test.
We should either add something to the list, or remove the test.

thomaslevesque:
Fixes #632

### DIFF
--- a/tests/FakeItEasy.Tests/ServiceLocatorTests.cs
+++ b/tests/FakeItEasy.Tests/ServiceLocatorTests.cs
@@ -3,7 +3,7 @@ namespace FakeItEasy.Tests
     using System;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
-    using System.Linq;
+    using FakeItEasy.Core;
     using FakeItEasy.Creation;
     using FakeItEasy.Expressions;
     using FluentAssertions;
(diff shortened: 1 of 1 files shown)


blairconrad:
Lovely.

tests/FakeItEasy.Tests/ServiceLocatorTests.cs
```diff
@@ -13,17 +13,25 @@ namespace FakeItEasy.Tests
...
-            typeof(ExpressionArgumentConstraintFactory),
-            typeof(IProxyGenerator)
-        };
+            get
+            {
+                yield return typeof(IExpressionCallMatcherFactory);
+                yield return typeof(ExpressionArgumentConstraintFactory);
+                yield return typeof(IProxyGenerator);
+            }
+        }
```
adamralph: It's not hugely important since this is just our test lib, but was there a problem with the original simple array?

`IEnumerable<T>`+`yield return` is a beautiful language construct but it has it's costs, i.e. a lot of scaffolding and extra allocation in the background, so it's only beneficial in certain cases, e.g. if you may potentially yield a lot of items and consumers often bail out early for some reason, or if it results in much simpler code.

In this case we know the test runner is always going to exhaust the enumerator and I'd argue the array is simpler code.

thomaslevesque: > It's not hugely important since this is just our test lib, but was there a problem with the original simple array?

Not really; but `SingletonTypes` was a field initialized with an array, and `NonSingletonTypes` was a property. I thought I'd make it consistent by making both of them properties.


> IEnumerable<T>+yield return is a beautiful language construct but it has it's costs, i.e. a lot of scaffolding and extra allocation in the background, so it's only beneficial in certain cases, e.g. if you may potentially yield a lot of items and consumers often bail out early for some reason, or if it results in much simpler code.

True, there was no real reason here. Since `NonSingletonTypes` was a property, it felt more natural to do it this way, then I did the same in `SingletonTypes` for consistency. I can change it to an array if you prefer.

adamralph: > I can change it to an array if you prefer.

Sure, if you have spare cycles, but as I say, it's not hugely important here. More of a conversation point.

PR #661, opened 2016-04-13
Stop using NUnit for assertions

blairconrad:
Contributes to #545.

Technically I suppose I didn't have to convert `NullGuardedConstraint.Assert` into something more FluentAssertiony, but it felt better to do so.
I broke the commits up so you as to not have major changes to both `NullGuardedConstraint` and `NullGuardedConstraintTests` in the same commit. Tests passed at each stage.


thomaslevesque:
Looks good to me.

Just one thing that bothers me a little: do we really need a _400 lines_ class just to check that a call is null-guarded ? I'm using [this code](http://www.thomaslevesque.com/2014/11/02/easy-unit-testing-of-null-argument-validation/), which, as far as I can tell, does the same thing.

blairconrad:
It does seem like a lot, doesn't it?

However, it's there, it works, and it's doing a bit more than the sample you posted, as it infers parameter names, allows for some parameters to not be null-checked, returns a message complaining about all the errors found, not just the first, and so on.

Do we need all this? To produce output that only people working on FakeItEasy will see?
Probably not. But it was there, and I didn't see the point in going through work to make it dumber…

tests/FakeItEasy.Tests/Creation/ProxyGeneratorSelectorTests.cs
```diff
@@ -92,16 +93,15 @@ public void Should_delegate_calls_to_default_generator_asking_about_non_delegate
             var getHashCode = fake.GetType().GetMethod("GetHashCode");
 
             string reason = null;
-            A.CallTo(() => this.defaultProxyGenerator.MethodCanBeInterceptedOnInstance(getHashCode, fake, out reason))
-                .Returns(true).AssignsOutAndRefParameters("reason");
+            A.CallTo(() => this.defaultProxyGenerator.MethodCanBeInterceptedOnInstance(getHashCode, fake, out reason)).Returns(true).AssignsOutAndRefParameters("reason");
```
adamralph: I'll merge as is, but I think this line break should have stayed present. Note how the old statement is completely visible in the GitHub diff UI but the new one spills over out of sight to the right.

Indeed, I've sent a few PR's with refactoring commits which limit line lengths to ~120 chars. (Our [guidelines](https://github.com/FakeItEasy/FakeItEasy/blob/master/CONTRIBUTING.md#line-width) say ~160, but I think we should shorten that to ~120 which allows some breathing room for the diff UI, which seems to spill over at around 130.)

adamralph:
:heart: 

Yes, `NullGuardedAssertion` is intricate, but this PR is doing what it needs to do, and that's to switch from NUnit assertion to FluentAssertions, which clears one obstacle to us switching from NUnit to xunit.

tests/FakeItEasy.Tests/Creation/ProxyGeneratorSelectorTests.cs
blairconrad: I'd've been happy to restore the linebreak. I prefer it as well.

blairconrad:
Thanks for the review and merge, fellows.

tests/FakeItEasy.Tests/Creation/ProxyGeneratorSelectorTests.cs
adamralph: :+1: it's just so trivial that I didn't want to hold up the PR.

blairconrad: I've no problem with the preferred line length becoming 120. I'm not that fussy.
(And I should probably also admit that for me, the whole statement is visible in the GitHub UI. Landscape mode, 1920px, 12px Consolas. I do use [the github expandizr](https://github.com/thecodejunkie/github.expandinizr), though. Maybe that's the difference.)

adamralph: I see this:

image

1920, 100% scaling, all browser (Chrome) defaults, no Github.ExpandinizR

PR #664, opened 2016-04-13
switch links from wiki to readthedocs

adamralph:
connects to #626


blairconrad:
:+1:

Now you see why I didn't rush on it. So many things to change. Because we aren't DRY.
Should just have everything point to a central registry of concepts, which will redirect.
(We should not have this.)

Thanks!

adamralph:
Yeah, the duplication bothers me.

PR #667, opened 2016-04-14
Change analyzer diagnostic id to FakeItEasy001

thomaslevesque:
I couldn't find any guidelines about what we should use for diagnostic ids, but we should probably use numbered ids with a common prefix; that's what everyone else seems to be doing.

Also, should we change the default severity of the diagnostic to `Error` rather than `Warning` ? Ignoring the result of `CallTo` will almost certainly not be intentional. And it can always be overriden in the .ruleset file if necessary.

Connects to #490

### DIFF
--- a/src/FakeItEasy.Analyzer/UnusedCallSpecificationAnalyzer.cs
+++ b/src/FakeItEasy.Analyzer/UnusedCallSpecificationAnalyzer.cs
@@ -10,7 +10,7 @@ namespace FakeItEasy.Analyzer
     [DiagnosticAnalyzer(LanguageNames.CSharp)]
     public class UnusedCallSpecificationAnalyzer : DiagnosticAnalyzer
     {
-        public const string DiagnosticId = "UnusedCallSpecification";
+        public const string DiagnosticId = "FakeItEasy001";
         internal const string Category = "Usage";
 
         internal static readonly LocalizableString Title = new LocalizableResourceString(nameof(Resources.UnusedCallSpecificationTitle), Resources.ResourceManager, typeof(Resources));
--- a/src/FakeItEasy.Analyzer/UnusedCallSpecificationAnalyzer.cs
+++ b/src/FakeItEasy.Analyzer/UnusedCallSpecificationAnalyzer.cs
@@ -17,7 +17,7 @@ namespace FakeItEasy.Analyzer
         internal static readonly LocalizableString MessageFormat = new LocalizableResourceString(nameof(Resources.UnusedCallSpecificationMessageFormat), Resources.ResourceManager, typeof(Resources));
         internal static readonly LocalizableString Description = new LocalizableResourceString(nameof(Resources.UnusedCallSpecificationDescription), Resources.ResourceManager, typeof(Resources));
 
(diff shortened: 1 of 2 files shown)


blairconrad:
I like the common prefix, but don't particularly enjoy numbers. They mean nothing to me. They ARE nice and unique, I suppose, and if everyone else is doing it, I don't have a stronger argument than "it'd be nice to understand something about the ID without going to a lookup table".

Of course, part of my reticence may be because I still have half a plan in the back of my head to convince you to switch to embedding attributes in FakeItEasy to drive the analyzer warnings, and the attributes would look better with a textual ID. :)

But other than that, I've no problem with the change.

I do agree that this diagnostic should default `Error`.

thomaslevesque:
> I like the common prefix, but don't particularly enjoy numbers. They mean nothing to me. They ARE nice and unique, I suppose, and if everyone else is doing it, I don't have a stronger argument than "it'd be nice to understand something about the ID without going to a lookup table".

It's true that numbers mean nothing, but they don't come alone, you have the diagnostic description with it. It neither better nor worse that CS0649 (C# compiler), CA1823 (Code Analysis), or SA1507 (StyleCop). At least the FakeItEasy prefix makes it clear where the warning/error comes from.

> Of course, part of my reticence may be because I still have half a plan in the back of my head to convince you to switch to embedding attributes in FakeItEasy to drive the analyzer warnings, and the attributes would look better with a textual ID. :)

I haven't explored that possibility yet. I'm not sure what you have in mind; the way I see it, the attributes wouldn't include the diagnostic id, it would be something like `[MustUseCallSpecification]`.

> I do agree that this diagnostic should default Error.

OK, I'll make the change.

blairconrad:
I agree that FakeItEasy#### is no worse than CS0649 and so on, and is even better, due to the "FakeItEasy". I was just looking for an opportunity to make them better/bend them to my will.

> I haven't explored that possibility yet. I'm not sure what you have in mind; the way I see it, the attributes wouldn't include the diagnostic id, it would be something like `[MustUseCallSpecification]`.

You don't know what I have in mind because my plan is secret.
The thought was that an unused call specification looks like lots of other unused return values, and could be detected the same way. Rather than duplicating the detection code, if the same detector was used, it could take advantage of attribute to 

1. know which unused return values are significant (as most of them would be fine), and
2. customize the message, since they won't all be call specifications. Well, maybe they will. Until we know different, there's no need to change.

If there was a common attribute that identified (e.g.) unused return values, the constructor could take the diagnostic ID to specify the kind of message to emit to the user.
In that case, since we'd embed the diagnostic ID in FakeItEasy's source, at the point of method definition, having something half-human readable might be easier.

I realize this is very theoretical. I'm not going to push for it. It was just a thought.

thomaslevesque:
> The thought was that an unused call specification looks like lots of other unused return values, and could be detected the same way. Rather than duplicating the detection code, if the same detector was used, it could take advantage of an attribute to

Good idea :+1: 

I just explored the Roslyn API a bit, it shouldn't be too hard to retrieve the attribute. So your idea seems perfectly feasible :smiley:

blairconrad:
> perfectly feasible

I'd actually spiked it. It worked. As we'll see in #668, I bet.

PR #668, opened 2016-04-15
Extend unused call specification diagnostic to remaining methods

thomaslevesque:
Prototype based on @blairconrad's [idea](https://github.com/FakeItEasy/FakeItEasy/pull/667#issuecomment-210079669) to use attributes in FakeItEasy's code to drive the analyzer.

The "unused call specification" diagnostic is emitted for any method that has the `[MustUseReturnValue("UnusedCallSpecification")]` attribute.

I tried to design the code to make it easy to add new diagnostics.

Supersedes #667, as it includes the changes made it that PR

Connects to #490


src/AssemblyInfo.cs
```diff
@@ -16,8 +16,6 @@
 // Module level suppress messages.
 [module: SuppressMessage("Microsoft.Design", "CA1020:AvoidNamespacesWithFewTypes", Justification = "Contains several internal types.", Scope = "namespace", Target = "FakeItEasy.Creation")]
 [module: SuppressMessage("Microsoft.Design", "CA1020:AvoidNamespacesWithFewTypes", Justification = "Contains several internal types.", Scope = "namespace", Target = "FakeItEasy.Expressions")]
-[module: SuppressMessage("Microsoft.Design", "CA1020:AvoidNamespacesWithFewTypes", Justification = "These types are only used by VB and should not confuse other developers.", Scope = "namespace", Target = "FakeItEasy.VisualBasic")]
-[module: SuppressMessage("Microsoft.Design", "CA1020:AvoidNamespacesWithFewTypes", Justification = "Is in a namespace of its own to allow for just using the extensions when they are explicitly requested.", Scope = "namespace", Target = "FakeItEasy.ExtensionSyntax")]
-[module: SuppressMessage("Microsoft.Design", "CA1020:AvoidNamespacesWithFewTypes", Scope = "namespace", Target = "FakeItEasy.ExtensionSyntax.Full", Justification = "Is in a namespace of its own to allow for just using the extensions when they are explicitly requested.")]
```
thomaslevesque: Unrelated to the prototype; I just removed obsolete suppressions.

src/FakeItEasy.Analyzer/DiagnosticDefinitions.cs
```diff
@@ -0,0 +1,50 @@
+namespace FakeItEasy.Analyzer
+{
+    using System;
+    using System.Collections.Immutable;
+    using System.Reflection;
+    using Microsoft.CodeAnalysis;
+
+    public static class DiagnosticDefinitions
```
thomaslevesque: Should really be internal, but it's used in the unit tests. The unit test assembly can be added as a friend assembly, but it needs to be strong-named for that. Since it's just a prototype at this point, I didn't do it yet.

src/FakeItEasy.Analyzer/DiagnosticDefinitions.cs
```diff
@@ -0,0 +1,50 @@
+namespace FakeItEasy.Analyzer
+{
+    using System;
+    using System.Collections.Immutable;
+    using System.Reflection;
+    using Microsoft.CodeAnalysis;
+
+    public static class DiagnosticDefinitions
+    {
+        public static DiagnosticDescriptor UnusedCallSpecification { get; } = CreateDiagnosticDescriptor(nameof(UnusedCallSpecification), "FakeItEasy0001", "Usage", DiagnosticSeverity.Error, true);
```
thomaslevesque: On the same line because StyleCop gets confused by the syntax otherwise.

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
```diff
@@ -0,0 +1,87 @@
...
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using static DiagnosticDefinitions;
+
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class UnusedReturnValueAnalyzer : DiagnosticAnalyzer
+    {
+        private static readonly ImmutableDictionary<string, DiagnosticDescriptor> Diagnostics =
+            GetDiagnosticsMap(
+                nameof(UnusedCallSpecification));
```
thomaslevesque: Other diagnostics to be produced by this analyzer should be added here.

src/FakeItEasy/Analysis/MustUseReturnValueAttribute.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace FakeItEasy.Analysis
+{
+    using System;
+
+    /// <summary>
+    /// Indicates that the return value of a method should not be ignored.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public sealed class MustUseReturnValueAttribute : Attribute
```
thomaslevesque: Has to be public to be seen by Roslyn (otherwise it's only accessible by reflection). It's unfortunate, because the user isn't supposed to see it... Perhaps the Roslyn assembly can be added as a friend assembly; not sure it would work, but it's worth trying.

src/AssemblyInfo.cs
adamralph: No problem, but I would prefer this to be in a separate refactoring commit.

adamralph:
It's a compelling approach (and I really like the implementation!). The presence of the attribute in the FIE public API is slightly bothersome, but I'm not sure how much, and I'm also not sure if it might even be a good thing. As @blairconrad mentioned elsewhere

> And the attribute on the member could be viewed either as clutter or as in-place docs!

I need to muse on it a little...

src/AssemblyInfo.cs
thomaslevesque: Sure, I'll do that

thomaslevesque:
Another side effect of having the attribute publicly visible is that users can use it to decorate their own methods, which will be picked up by the analyzer... I'm not sure if it's a good or a bad thing.

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
```diff
@@ -0,0 +1,87 @@
...
+            if (call.Parent is StatementSyntax)
+            {
+                var firstArg = call.ArgumentList.Arguments.FirstOrDefault();
+                var lambda = firstArg?.Expression as LambdaExpressionSyntax;
+                var callSpec = lambda?.Body?.ToString()
+                               ?? firstArg?.ToString()
+                               ?? "(unknown)";
+                var diagnostic = Diagnostic.Create(descriptor, call.GetLocation(), callSpec);
+                context.ReportDiagnostic(diagnostic);
+            }
```
thomaslevesque: I just realized that this part of the code isn't really generic; it's specific to `A.CallTo`, since it tries to extract the call specification. I could probably refactor further to have some diagnostic-specific code.

adamralph:
> Another side effect of having the attribute publicly visible is that users can use it to decorate their own methods, which will be picked up by the analyzer... I'm not sure if it's a good or a bad thing.

Hmmm.. good point. This will feed into my musing...

src/FakeItEasy/Analysis/MustUseReturnValueAttribute.cs
blairconrad: I would either try the friend approach, or failing that would not object to accessing the attribute by reflection. I did it in my spike, and it seemed to work.

blairconrad: Actually, I don't even see how you're using it now, except by reflection in `UnusedReturnValueAnalyzer.GetDiagnosticName`.

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
blairconrad: I was surprised by this, as I was thinking of the different flavours of unused return values as a single diagnostic, just with different descriptions. Shows my inexperience with the technology.

Given the multiple diagnostics, I like how you've ensured we'll need minimal changes to support a new one. :+1:

blairconrad:
@thomaslevesque, :heart_eyes:

Thanks for entertaining my crazy notion. I realize we're not done yet, but I like the look of things. Now I should run out and try to add an annotation to `WhenArgumentsMatch`...

blairconrad:
I tried `WithAnyArguments` because it was less typing. The analyzer found the problem, but the message was not quite right, probably because of @thomaslevesque's [note](https://github.com/FakeItEasy/FakeItEasy/pull/668#discussion_r59843715):

```
// Test0.cs(9,13): error FakeItEasy0001: Unused call specification for '(unknown)'; did you forget to configure or assert the call?
GetCSharpResultAt(9, 13, UnusedReturnValueAnalyzer.FakeItEasy0001)
```

I am very encouraged!

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
blairconrad: I wonder how specific things need to be. We already have the line and column from the diagnostic. Perhaps referencing the FakeItEasy call (`CallTo`, `WhenArgumentsMatch`, …) is sufficient?

src/FakeItEasy/Analysis/MustUseReturnValueAttribute.cs
thomaslevesque: > Actually, I don't even see how you're using it now, except by reflection in `UnusedReturnValueAnalyzer.GetDiagnosticName`.

I'm using the Roslyn API, not reflection :wink:

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
thomaslevesque: Well, the description is in the `DiagnosticDescriptor`, so we need several `DiagnosticDescriptor`s

thomaslevesque:
> I tried WithAnyArguments because it was less typing. The analyzer found the problem, but the message was not quite right, probably because of @thomaslevesque's note:

The problem here is that it's `WithAnyArguments()` that triggers the diagnostic, but the call specification is actually in `A.CallTo()`... We'll have to come up with a way to handle that properly.

src/FakeItEasy/Analysis/MustUseReturnValueAttribute.cs
blairconrad: Oh. Clearly I understand nothing.
I did make the attribute `internal` and it all built. I was surprised.

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
blairconrad: I assumed the description could be generated at need by keying off the attribute constructor argument. I did not think it totally through.

blairconrad:
They're both kind of call specifications, no? Or am I thinking of it wrong?
`WithAnyArguments` is definitely contributing to the specification of the call. Without it, FakeItEasy wouldn't act on the right calls.

src/FakeItEasy/Analysis/MustUseReturnValueAttribute.cs
thomaslevesque: > I did make the attribute internal and it all built.

Yes, because the analyzer has no compile-time reference to the attribute. It just uses Roslyn to look at the publicly visible attributes on the `IMethodSymbol`.

blairconrad: There is still something that I do not understand, it seems. I've tried again, making `MustUseReturnValueAttribute` internal, and not only do things compile, but all your tests pass. I thought your original comment would preclude that.

thomaslevesque: > There is still something that I do not understand, it seems. I've tried again, making MustUseReturnValueAttribute internal, and not only do things compile, but all your tests pass. I thought your original comment would preclude that.

Strange... When I tried, `GetAttributes()` wasn't returning anything. It only worked after I made the attribute public. But you're right, I just tried it again, the tests pass, and the analyzer works as expected in Visual Studio. Well, that's good news :)

adamralph: Well, that certainly changes things. If we can make the attribute internal then I think we can adopt this refactoring straight away. We can tackle the benefits of exposing them publicly in a separate issue, but the 2.0 release isn't dependent on that.

thomaslevesque: > If we can make the attribute internal then I think we can adopt this refactoring straight away.

It's not quite ready yet, there are a few issues to take care of. Hopefully I'll finish it over the weekend.

thomaslevesque:
> `WithAnyArguments` is definitely contributing to the specification of the call. Without it, FakeItEasy wouldn't act on the right calls.

Indeed. What would you expect the diagnostic message to be in this case?

blairconrad:
@thomaslevesque, I would expect exactly the same message. The return value of the call specification, which is formed by `A.CallTo` and `WithAnyArguments`, was not used.

thomaslevesque:
OK, I've been thinking a bit more about this.

In this PR I removed the `UnusedCallSpecificationAnalyzer` to replace it with a more generic `UnusedReturnValueAnalyzer`. But the analyzer needs to extract the call specification expression to include it in the diagnostic message, which will not be the case for other (non related to call specification) "unused return value" diagnostics. So it's not really a generic purpose analyzer for unused return values, and I think I'll revert it to `UnusedCallSpecificationAnalyzer` (but still rely on attributes to know which methods should be analyzed).

For cases like `A.CallTo(() => foo.Bar(0)).WithAnyArguments()`, if `WithAnyArguments` is decorated with the attribute, we still need to extract the call description from `A.CallTo(...)`. So the analyzer needs to walk back to `A.CallTo(...)`, or whatever method specifies the call expression. So I'm thinking about adding another attribute on `A.CallTo` to indicate that it specifies the expression. We will end up with something like that:

```
[ReturnsCallSpecification, CallSpecificationExpression]
public static IVoidArgumentValidationConfiguration CallTo(Expression<Action> callSpecification)

[ReturnsCallSpecification, CallSpecificationExpression]
public static IReturnValueArgumentValidationConfiguration<T> CallTo<T>(Expression<Func<T>> callSpecification)

[ReturnsCallSpecification, CallSpecificationExpression]
public static IAnyCallConfigurationWithNoReturnTypeSpecified CallTo(object fake)

[ReturnsCallSpecification]
public static TInterface WithAnyArguments<TInterface>(this IArgumentValidationConfiguration<TInterface> configuration)

[ReturnsCallSpecification]
TInterface WhenArgumentsMatch(Func<ArgumentCollection, bool> argumentsPredicate);

...
```
When the analyzer finds an unused return value from a call to method decorated with `[ReturnsCallSpecification]`, it will walk back the call chain to find the first call to a method decorated with `[CallSpecificationExpression]`, and will use that call to extract the call specification expression.

What do you think? Does it make sense?

thomaslevesque:
OK, I went ahead and implemented it. It's a bit less ambitious than the previous version (it only checks for unused call specification), but it now works correctly with `WithAnyArguments`.

So this:

    A.CallTo(() => foo.Bar()).WithAnyArguments();

gives the following message:

> Unused call specification for 'foo.Bar()'; did you forget to configure or assert the call?

I added a test for `WithAnyArguments`, and another one for `A.CallTo(foo)` (which was supported but not yet covered in tests).

If you like it, all that remains to be done is to add the `[ReturnsCallSpecification]` and `[CallSpecificationExpression]` to the relevant methods (I only did it on the 3 `A.CallTo` overloads, `WithAnyArguments` and `WhenArgumentsMatch`)

blairconrad:
Wow, @thomaslevesque, you've been quite busy. I appreciate all the hard work, but must admit I don't understand the reason for some of it.
I suspect that the disconnect lies with in my not thinking things through enough, or some aspect of analyzers that I just don't  understand.

You said

> if `WithAnyArguments` is decorated with the attribute, we still need to extract the call description from `A.CallTo(…)`. 

I'm not sure why. I mean, I guess it's so the message can include the argument to `A.CallTo`, yes?
Without this, would the message just talk about `WithAnyArguments`, or is there some more fundamental problem that I'm not seeing?

If I've figured out the problem, then to my mind, it's no problem. I think an error that complained about the result of `WithAnyArguments` not being used from the statement `A.CallTo(() => foo.Bar()).WithAnyArguments()` would be perfectly clear to the user. I don't think backtracking to include the argument to `A.CallTo` provides much (any?) benefit. Certainly not enough to warrant an additional attribute being introduced. In fact, I think it may put too much focus on the argument to `A.CallTo`. A naive user who wrote

```c#
 A.CallTo(() => foo.Bar()).WithAnyArguments();
```

and saw the diagnostic error

```
Unused call specification for 'foo.Bar()'; did you forget to configure or assert the call?"
```

might focus too much on the area around `foo.Bar()`, thinking they need to do something with the result of that call. Or maybe around the `A.CallTo`, and become confused because that result is being used.

However, if I'm alone in this, and the consensus is that adjusting the information to talk about `foo.Bar` in the diagnostic message, I wonder if a different tack wouldn't work.
I find the two attributes confusing. Partly it could be their names, as I keep having to look at how they are used to in order to understand what they mean. But even if it's that, and the names could be changed, I think one attribute would be better than two. So as an experiment, I removed `CallSpecificationExpression` from all the `A.CallTo`s. And rewrote `GetCallSpecificationExpression`:

```c#
private static string GetCallSpecificationExpression(InvocationExpressionSyntax call)
{
    var lastCall = call;
    while (call != null)
    {
        lastCall = call;
        var memberAccess = call.Expression as MemberAccessExpressionSyntax;
        call = memberAccess?.Expression as InvocationExpressionSyntax;
    }

    var firstArg = lastCall.ArgumentList.Arguments.FirstOrDefault();
    var lambda = firstArg?.Expression as LambdaExpressionSyntax;
    var callSpec = lambda?.Body?.ToString()
                   ?? firstArg?.ToString();
    return callSpec;
}
```

Now instead of looking for `CallSpecificationExpressionAttribute`, we just walk up the call chain and pick the last method call. That finds us `A.CallTo`.
All the existing tests pass.

Maybe I'm not thinking about this enough, but I believe any call specification that we have will begin with a method call on `A`, so I think this will cover all cases we have now, and probably all future ones.

thomaslevesque:
> I mean, I guess it's so the message can include the argument to A.CallTo, yes?

Yes

> Without this, would the message just talk about WithAnyArguments, or is there some more fundamental problem that I'm not seeing?

Well, you said [here](https://github.com/FakeItEasy/FakeItEasy/pull/668#issuecomment-211128278) that you would expect the same message. I assumed that meant the message should be

> Unused call specification for 'foo.Bar()'; did you forget to configure or assert the call?



> If I've figured out the problem, then to my mind, it's no problem. I think an error that complained about the result of WithAnyArguments not being used from the statement A.CallTo(() => foo.Bar()).WithAnyArguments() would be perfectly clear to the user. I don't think backtracking to include the argument to A.CallTo provides much (any?) benefit. Certainly not enough to warrant an additional attribute being introduced. In fact, I think it may put too much focus on the argument to A.CallTo. A naive user who wrote

Ah, I see your point. Yes, including the whole call chain might be better, and it's simpler to implement. _And_ it makes it possible to keep the analyzer generic, as it was in the previous commit.

> So as an experiment, I removed CallSpecificationExpression from all the A.CallTos. And rewrote GetCallSpecificationExpression:

Yes, I guess that would work too. But if we want to include the whole `A.CallTo(() => foo.Bar()).WithAnyArguments()`, it's no longer necessary.

OK, so I think I'll roll back my last commit, and just change what we include in the message.

blairconrad:
@thomaslevesque, that clears everything up!

> Well, you said [here](https://github.com/FakeItEasy/FakeItEasy/pull/668#issuecomment-211128278) that you would expect the same message. I assumed that meant the message should be

> > Unused call specification for 'foo.Bar()'; did you forget to configure or assert the call?

I'd thought that we were just talking about whether `WithAnyArguments` was really a call specification, so I made the misleading "exactly the same message" statement about the message being the same. I'm sorry this forced you to contort the code into doing something unnatural. I should've indicated that I would've expected the same message modulo stuff between the quotes.

I'm indifferent about including the whole `A.CallTo(() => foo.Bar()).WithAnyArguments()` or just `WithAnyArguments` or just indicating the general file position. I'm sure whichever you think makes most sense and is reasonable to implement will be great.

thomaslevesque:
I reverted the last commit and changed the call description in the message (this was even easier than I thought : `call.ToString()` :smile:)

blairconrad:
@thomaslevesque, I'm sorry about the tortuous route, but like the destination!

thomaslevesque:
> I'm sorry about the tortuous route

Don't be! I had strayed from the path, it's much better this way :wink:

thomaslevesque:
I added the attribute to the following methods:
```
IAnyCallConfigurationWithNoReturnTypeSpecified.WithReturnType<TMember>
IArgumentValidationConfiguration<TInterface>.WhenArgumentsMatch
IWhereConfiguration<T>.Where
ArgumentValidationConfigurationExtensions.WithAnyArguments<TInterface>
A.CallTo // (all 3 overloads)
Fake<T>.CallsTo // both overloads
Fake<T>.AnyCall
```

Let me know if I forgot something

thomaslevesque:
I had missed one:

```
WhereConfigurationExtensions.Where
```

blairconrad:
@thomaslevesque, I would've missed the `Fake<T>` ones. Nicely found!

I can't think of anything else.

adamralph:
LGTM.

One thing - this PR is no longer just a refactoring.

It's adding diagnostics for a host of other methods not mentioned in https://github.com/FakeItEasy/FakeItEasy/issues/490. That's fine, but it should be communicated somehow. Should we re-open #490 and expand the scope?

thomaslevesque:
> One thing - this PR is no longer just a refactoring.
> 
> It's adding diagnostics for a host of other methods not mentioned in #490.

Good point

> That's fine, but it should be communicated somehow. Should we re-open #490 and expand the scope?

I'd rather create a new issue, it will make things easier to track.

adamralph:
> I'd rather create a new issue, it will make things easier to track.

Seems like overkill. #490 isn't released yet. If I think about the 2.0 release notes, it will be rather odd and asymmetrical to have #490 describing just one diagnostic, and some other issue describing all the rest. Alternatively, we could have an issue for each diagnostic, but I think that's overkill too.

thomaslevesque:
I see your point. I reopened #490, will update the description.

blairconrad:
Fantastic work, @thomaslevesque, I'm happy to merge, but am leery of doing so to a PR labelled "[WIP]". Is more awesomeness coming, or are you done?

adamralph:
I think the commits could probably do with a squash. Ideally I'd like to see something like:

* refactor analyzer to use attribute
* add attributes to remaining methods

The "Remove obsolete code analysis suppressions" commit is completely unrelated, but I don't mind it being tagged on here as a separate commit.

adamralph:
I guess we could rename this PR to something like "Add UnusedCallSpecification to remaining methods"?

thomaslevesque:
> I guess we could rename this PR to something like "Add UnusedCallSpecification to remaining methods"?

Or "Extend unused call specification diagnostic to remaining methods" ?

thomaslevesque:
> The "Remove obsolete code analysis suppressions" commit is completely unrelated, but I don't mind it being tagged on here as a separate commit.

Since I'm reorganizing the commits anyway, I'll put it in a separate PR

thomaslevesque:
- I squashed everything into 2 commits, as suggested by @adamralph 
- I renamed the PR and removed the WIP

thomaslevesque:
> or are you done?

I'm done, I think :wink:

tests/FakeItEasy.Analyzer.Tests/UnusedCallSpecificationAnalyzerTests.cs
```diff
@@ -97,17 +98,115 @@ interface IFoo { int Bar(); }
                 test,
                 new DiagnosticResult
                 {
-                    Id = UnusedCallSpecificationAnalyzer.DiagnosticId,
+                    Id = DiagnosticDefinitions.UnusedCallSpecification.Id,
                     Message =
-                        "Unused call specification for 'foo.Bar()'; did you forget to configure or assert the call?",
+                        "Unused call specification for 'A.CallTo(() => foo.Bar())'; did you forget to configure or assert the call?",
```
blairconrad: A last-minute comment: I think the "for" should be struck here (well, where it's generated). `A.CallTo(() => foo.Bar())` is the call specification, it's not _for_ it.

thomaslevesque: Ah, yes. I noticed it and intended to fix it, but I was focused on something else and it slipped my mind.

thomaslevesque:
I remove the "for" in the diagnostic message, and repackaged the commits more cleanly.

adamralph:
Awesome, thanks @thomaslevesque. I'll give it a once-over ASAP

thomaslevesque:
Just rebased after #672 was merged

tests/FakeItEasy.Analyzer.Tests/UnusedCallSpecificationAnalyzerTests.cs
```diff
@@ -97,17 +98,115 @@ interface IFoo { int Bar(); }
                 test,
                 new DiagnosticResult
                 {
-                    Id = UnusedCallSpecificationAnalyzer.DiagnosticId,
+                    Id = DiagnosticDefinitions.UnusedCallSpecification.Id,
                     Message =
-                        "Unused call specification for 'foo.Bar()'; did you forget to configure or assert the call?",
+                        "Unused call specification 'A.CallTo(() => foo.Bar())'; did you forget to configure or assert the call?",
```
adamralph: It's not really a refactoring, since the functionality of the analyzer has also changed. I'll let it pass though :wink:

src/FakeItEasy.Analyzer/DiagnosticDefinitions.cs
```diff
@@ -0,0 +1,50 @@
+namespace FakeItEasy.Analyzer
+{
+    using System;
+    using System.Collections.Immutable;
+    using System.Reflection;
+    using Microsoft.CodeAnalysis;
+
+    internal static class DiagnosticDefinitions
+    {
+        public static DiagnosticDescriptor UnusedCallSpecification { get; } = CreateDiagnosticDescriptor(nameof(UnusedCallSpecification), "FakeItEasy0001", "Usage", DiagnosticSeverity.Error, true);
```
adamralph: Should "Usage" be "FakeItEasy.Usage"?

The docs for `DiagnosticDescriptor` say:

```C#
        //   category:
        //     The category of the diagnostic (like Design, Naming etc.). For example, for CA1001:
        //     "Microsoft.Design".
```

src/FakeItEasy/Analysis/MustUseReturnValueAttribute.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace FakeItEasy.Analysis
+{
+    using System;
+
+    /// <summary>
+    /// Indicates that the return value of a method should not be ignored.
```
adamralph: Perhaps this should read

> Indicates that the return value of a method must be used.

in order to tie-in with the name of the attribute?

src/FakeItEasy/Analysis/MustUseReturnValueAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <summary>
+    /// Indicates that the return value of a method should not be ignored.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    internal sealed class MustUseReturnValueAttribute : Attribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MustUseReturnValueAttribute"/> class.
+        /// </summary>
+        /// <param name="diagnosticName">Name of the diagnostic that is reported when the return value is ignored.</param>
```
adamralph: Perhaps this should read

> The name of the diagnostic that is reported when the return value is not used.

src/FakeItEasy/Analysis/MustUseReturnValueAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+        public MustUseReturnValueAttribute(string diagnosticName)
+        {
+            this.DiagnosticName = diagnosticName;
+        }
+
+        /// <summary>
+        /// Gets the name of the diagnostic that is reported when the return value is ignored.
```
adamralph: And here:

> Gets the name of the diagnostic that is reported when the return value of the method is not used.

adamralph:
In fact, given `MustUseReturnValueAttribute` is `internal`, perhaps we can just drop the XML comments entirely?

adamralph:
@thomaslevesque I've finished my pass over the PR.

Left some questions, of minor importance.

src/FakeItEasy.Analyzer/DiagnosticDefinitions.cs
thomaslevesque: Yes, you're probably right. I'll change it.

thomaslevesque:
> In fact, given MustUseReturnValueAttribute is internal, perhaps we can just drop the XML comments entirely?

fine by me!
I probably won't be able to do it this weekend, though

thomaslevesque:
I pushed the last changes. I left them in their own commit to make review easier, but I can squash them before merge.

adamralph:
@thomaslevesque thanks. LGTM. Please go ahead and squash and I'm happy to merge.

thomaslevesque:
> @thomaslevesque thanks. LGTM. Please go ahead and squash and I'm happy to merge.

Done !

adamralph:
Thanks @thomaslevesque! Awesome work again.

blairconrad:
Hear hear! Very nice, sir.

Issue #361, opened 2014-09-17
stop requiring event arguments to extend EventArgs in event raisers for .NET Standard


blairconrad:
From @bondsbw in #30:

> I know I'm coming in late to this discussion, but something to consider (I apologize if I missed this in past discussion):\
> .NET 4.5 no longer constrains EventHandler to using something that derives from EventArgs. TEventArgs can now be anything.
> 
> Of course compiling for .NET 4.5 may not make sense depending on the goals of the project.
> But I made a few alterations to the source to bump it to 4.5 and remove the constraint,
> and was able to get everything working as expected (other than Silverlight projects which I did not load).

9swampy:
Thank-you. It's taken a long time to make sense of using #30 but thanks to your last comment I got it to work with 2.0.0beta009 in a 4.5 project.

For:
public delegate void NewFrameEventHandler(object sender, NewFrameEventArgs eventArgs);
Exposed via:
event NewFrameEventHandler NewFrame;

Either of the following worked:
.NewFrame += Raise.With<NewFrameEventHandler>(new object[] { this, new NewFrameEventArgs(bitmap) });
.NewFrame += Raise.With<NewFrameEventHandler>(this, new NewFrameEventArgs(bitmap));

[Slightly off topic as in my case NewFrameEventArgs does still inherit from EventArgs]

blairconrad:
@9swampy, I'm glad you were able to raise the events. Thanks for the feedback.

adamralph:
So do we actually have to do anything? Or will 2.0 just do this already?

blairconrad:
I'll give it a whirl soonish and get back to you, @adamralph.

blairconrad:
@adamralph, the 2.0 betas do not do what we want.

If there were a delegate defined, and the event defined in terms of the delegate, we could raise the event, just as with any delegate-defined event, by using this syntax:

```c#
public delegate void DoesNotExtendEventHandler(object sender, DoesNotExtendEventArgs args);

public interface IRaiseEvents
{
    event DoesNotExtendEventHandler AnEvent;
}

…
fake.AnEvent += Raise.With<DoesNotExtendEventHandler>(this, new DoesNotExtendEventArgs());
```

If we omit the `<DoesNotExtendEventHandler>`, nothing compiles because `DoesNotExtendEventArgs` isn't convertible to `EventArgs`.

[...]

PR #669, opened 2016-04-15
refactor: move AssemblyInfo.cs to FakeItEasy folder to match all other cs files

adamralph:
It suddenly struck me that FakeItEasy.Net35 links all of its source from FakeItEasy (apart from the shims) so `AssemblyInfo.cs` is no different.

It also implies the original `CommonAssemblyInfo.cs` should also have always been `AssemblyInfo.cs` sitting inside the `FakeItEasy\Properties` folder and it's only since the analyzer that we've actually had a need for a `CommonAssemblyInfo.cs`.

The thing that highlighted this to me was when I browsed to the `src` folder and saw `CommonAssemblyInfo.cs` and `AssemblyInfo.cs` sitting next to each other, with another `AssemblyInfo.cs` sitting lower down inside `FakeItEasy.Analyzer\Properties`.


thomaslevesque:
LGTM :+1: 

Now that we're officially using VS2015, perhaps we should consider switching to a shared project. It would make it much easier to share code between FakeItEasy and FakeItEasy.Net35. Currently it's quite tedious; every time we add or move a file in a project, we have to do it again in the other.

blairconrad:
I'd interested in a shared project as well, especially since it'd help us target 4.5 for #361 (and perhaps other reasons). Would the shared project also work for .NET Core?

thomaslevesque:
> Would the shared project also work for .NET Core?

No, .NET Core uses yet another project model (project.json)

thomaslevesque:
> No, .NET Core uses yet another project model (project.json)

But I *think* it would be possible to target all frameworks with the `project.json` model

blairconrad:
> But I *think* it would be possible to target all frameworks with the `project.json` model

I thought that might be the case. Perhaps this can help with our "sharing" problem.

adamralph:
I'm fairly sure project.json won't let us target .NET 3.5. I think it's
4.5.1 and later.  It's also a continually changing space. I'm reluctant to
do much in 'the new world' before RC2 (and even then who knows what
surprises RTM is going to bring).
On 16 Apr 2016 4:16 am, "Blair Conrad" <notifications@github.com> wrote:

But I *think* it would be possible to target all frameworks with the
project.json model

I thought that might be the case. Perhaps this can help with our "sharing"
problem.

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/FakeItEasy/FakeItEasy/pull/669#issuecomment-210715729>

thomaslevesque:
> I'm fairly sure project.json won't let us target .NET 3.5. I think it's 4.5.1 and later.

Ah, yes, forgot about that

> It's also a continually changing space. I'm reluctant to do much in 'the new world' before RC2 (and even then who knows what surprises RTM is going to bring).

Agreed

PR #670, opened 2016-04-15
update how_to_build.md

adamralph:
A few things were out of date here.

### DIFF
--- a/how_to_build.md
+++ b/how_to_build.md
@@ -5,7 +5,7 @@ It also replicates the build on the Continuous Integration build server and is t
 
 *Don't be put off by the prerequisites!* It only takes a few minutes to set them up and only needs to be done once. If you haven't used [Rake](http://rake.rubyforge.org/ "RAKE -- Ruby Make") before then you're in for a real treat!
 
-You can also build the solution using Visual Studio 2012 or later, but this doesn't provide the same assurances as the Rake build.
+You can also build the solution using Visual Studio 2015 or later, but this doesn't provide the same assurances as the Rake build.
 
 At the time of writing the build is only confirmed to work on Windows using the Microsoft .NET framework.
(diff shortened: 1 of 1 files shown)


how_to_build.md
```diff
@@ -5,22 +5,24 @@ It also replicates the build on the Continuous Integration build server and is t
...
-You can also build the solution using Visual Studio 2012 or later, but this doesn't provide the same assurances as the Rake build.
+You can also build the solution using Visual Studio 2015 or later, but this doesn't provide the same assurances as the Rake build.
 
 At the time of writing the build is only confirmed to work on Windows using the Microsoft .NET framework.
 
 ## Prerequisites
 
 1. Ensure you have .NET framework 3.5 and 4.0/4.5 installed.
 
+1. Ensure you have Visual Studio 2015 or MSBuild 14 installed.
```
adamralph: This is because MSBuild no longer ships with the framework. It ships with VS releases.

how_to_build.md
```diff
@@ -41,7 +43,7 @@ Using a command prompt, navigate to your clone root folder and execute:
 
 `bundle exec rake`
 
-This executes the default build tasks. After the build has completed, the build artifacts will be located in `Build`.
```
thomaslevesque: About this, I think TeamCity is still looking for artifacts in `Build` (in addition to `artifacts`). I noticed it a few days ago but didn't have time to check if it was normal; I don't think it is.

thomaslevesque:
:+1:

how_to_build.md
adamralph: I just checked TC and it did indeed have some redundant Build/ entries in its artifacts paths. I trimmed it down to just `artifacts\**\*.*`.

PR #673, opened 2016-04-21
New property setter syntax

### LINKED ISSUE #175: New syntax for configuration of property setters

blairconrad:
Fixes #175.

### DIFF
--- a/src/FakeItEasy/Configuration/FakeConfigurationManager.cs
+++ b/src/FakeItEasy/Configuration/FakeConfigurationManager.cs
@@ -2,8 +2,11 @@ namespace FakeItEasy.Configuration
 {
     using System;
     using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
     using System.Linq.Expressions;
+    using System.Reflection;
     using Expressions;
+    using FakeItEasy.Core;
 
     internal class FakeConfigurationManager
         : IFakeConfigurationManager
@@ -29,12 +32,7 @@ namespace FakeItEasy.Configuration
 
             this.AssertThatMemberCanBeIntercepted(callSpecification);
 
-            var fake = this.expressionParser.GetFakeManagerCallIsMadeOn(callSpecification);
-            var rule = this.ruleFactory.Invoke(callSpecification);
-
-            fake.AddRuleFirst(rule);
-
-            return this.configurationFactory.CreateConfiguration(fake, rule);
+            return this.CreateVoidArgumentValidationConfiguration(callSpecification);
         }
 
         [SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "This is by design when using the Expression-, Action- and Func-types.")]
@@ -59,6 +57,123 @@ namespace FakeItEasy.Configuration
             manager.AddRuleFirst(rule);
 
             return this.configurationFactory.CreateAnyCallConfiguration(manager, rule);
+        }
+
+        public IPropertySetterAnyValueConfiguration<TValue> CallToSet<TValue>(Expression<Func<TValue>> propertySpecification)
+        {
+            Guard.AgainstNull(propertySpecification, nameof(propertySpecification));
+            this.AssertThatMemberCanBeIntercepted(propertySpecification);
+
+            var setterExpression = BuildSetterFromGetter(propertySpecification);
+
+            return new PropertySetterConfiguration<TValue>(setterExpression, this.CreateVoidArgumentValidationConfiguration);
+        }
+
+        private static string GetPropertyName(MethodCallExpression methodCallExpression)
+        {
+            if (methodCallExpression != null &&
+                HasThis(methodCallExpression) &&
+                methodCallExpression.Method.IsSpecialName)
+            {
+                var methodName = methodCallExpression.Method.Name;
+                if (methodName.StartsWith("get_", StringComparison.Ordinal))
+                {
+                    return methodName.Substring(4);
+                }
+            }
+
+            return null;
+        }
+
+        private static bool HasThis(MethodCallExpression methodCallExpression)
+        {
+            return (methodCallExpression.Method.CallingConvention & CallingConventions.HasThis) ==
+                   CallingConventions.HasThis;
+        }
+
+        private static MethodCallExpression BuildSetterFromMemberExpression<TValue>(MemberExpression memberExpression)
+        {
+            PropertyInfo propertyInfo = memberExpression.Member as PropertyInfo;
+            if (propertyInfo == null)
+            {
+                throw new ArgumentException("The expression refers to '" + memberExpression.Member.Name +
+                                            "', which is a field, not a property getter.");
+            }
+
+            var propertySetterInfo = propertyInfo.GetSetMethod(nonPublic: true);
+            if (propertySetterInfo == null)
+            {
+                throw new ArgumentException("The property '" + memberExpression.Member.Name + "' does not have a setter.");
+            }
+
+            var arguments = new[] { BuildArgumentThatMatchesAnything<TValue>() };
+
+            return Expression.Call(memberExpression.Expression, propertySetterInfo, arguments);
+        }
+
+        private static Expression BuildArgumentThatMatchesAnything<TValue>()
+        {
+            Expression<Func<TValue>> value = () => A<TValue>.Ignored;
+            return value.Body;
+        }
+
+        private IVoidArgumentValidationConfiguration CreateVoidArgumentValidationConfiguration(LambdaExpression lambda)
+        {
+            var fake = this.expressionParser.GetFakeManagerCallIsMadeOn(lambda);
+            var rule = this.ruleFactory.Invoke(lambda);
+            fake.AddRuleFirst(rule);
+
+            return this.configurationFactory.CreateConfiguration(fake, rule);
+        }
+
+        private string GetExpressionDescription<TValue>(Expression<Func<TValue>> expression)
+        {
+            var matcher = new ExpressionCallMatcher(
+                expression,
+                ServiceLocator.Current.Resolve<ExpressionArgumentConstraintFactory>(),
+                ServiceLocator.Current.Resolve<MethodInfoManager>(),
+                this.callExpressionParser);
+
+            var expressionDescription = matcher.DescriptionOfMatchingCall;
+            return expressionDescription;
+        }
+
+        private MethodCallExpression BuildSetterFromGetter<TValue>(Expression<Func<TValue>> propertySpecification)
+        {
+            var memberExpression = propertySpecification.Body as MemberExpression;
+            return memberExpression == null
+                ? BuildSetterFromMethodCall(propertySpecification)
+                : BuildSetterFromMemberExpression<TValue>(memberExpression);
+        }
+
+        private MethodCallExpression BuildSetterFromMethodCall<TValue>(Expression<Func<TValue>> propertySpecification)
+        {
+            var methodCallExpression = propertySpecification.Body as MethodCallExpression;
+            var indexerName = GetPropertyName(methodCallExpression);
+            if (indexerName == null)
+            {
+                var expressionDescription = GetExpressionDescription(propertySpecification);
+                throw new ArgumentException("Expression '" + expressionDescription +
+                                            "' must refer to a property or indexer getter, but doesn't.");
+            }
+
+            var parameterTypes = methodCallExpression.Method.GetParameters()
+                .Select(p => p.ParameterType)
+                .Concat(new[] { methodCallExpression.Method.ReturnType })
+                .ToArray();
+            var instance = methodCallExpression.Object;
+            var indexerSetterInfo = instance.Type.GetMethod("set_" + indexerName, parameterTypes);
+
+            if (indexerSetterInfo == null)
+            {
+                var expressionDescription = GetExpressionDescription(propertySpecification);
+                throw new ArgumentException("Expression '" + expressionDescription +
+                                            "' refers to an indexed property that does not have a setter.");
+            }
+
+            var arguments = methodCallExpression.Arguments.Concat(new[] { BuildArgumentThatMatchesAnything<TValue>() });
+
+            return Expression.Call(instance, indexerSetterInfo, arguments);
         }
 
         private void AssertThatMemberCanBeIntercepted(LambdaExpression callSpecification)
--- a/src/FakeItEasy/Configuration/IPropertySetterConfiguration.cs
+++ b/src/FakeItEasy/Configuration/IPropertySetterConfiguration.cs
@@ -1,13 +1,12 @@
 namespace FakeItEasy.Configuration
 {
     /// <summary>
-    /// Provides configuration methods for methods that does not have a return value.
+    /// Provides configuration for property setters and allows the user to specify validations for arguments.
     /// </summary>
-    public interface IVoidConfiguration
+    public interface IPropertySetterConfiguration
         : IExceptionThrowerConfiguration,
-          ICallbackConfiguration<IVoidConfiguration>,
+          ICallbackConfiguration<IPropertySetterConfiguration>,
           ICallBaseConfiguration,
-          IOutAndRefParametersConfiguration,
           IAssertConfiguration,
           IDoNothingConfiguration
     {
--- a/src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
+++ b/src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
@@ -1,0 +1,131 @@
+namespace FakeItEasy.Configuration
+{
+    using System;
+    using System.Linq;
+    using System.Linq.Expressions;
+    using FakeItEasy.Core;
+
+    internal class PropertySetterConfiguration<TValue>
+        : IPropertySetterAnyValueConfiguration<TValue>
+    {
+        private readonly MethodCallExpression setterExpression;
+
+        private readonly Func<LambdaExpression, IVoidArgumentValidationConfiguration> voidArgumentValidationConfigurationFactory;
+
+        public PropertySetterConfiguration(
+            MethodCallExpression setterExpression,
+            Func<LambdaExpression, IVoidArgumentValidationConfiguration> voidArgumentValidationConfigurationFactory)
+        {
+            this.setterExpression = setterExpression;
+            this.voidArgumentValidationConfigurationFactory = voidArgumentValidationConfigurationFactory;
+        }
+
+        public IPropertySetterConfiguration To(TValue value)
+        {
+            return this.To(() => value);
+        }
+
+        public IPropertySetterConfiguration To(Expression<Func<TValue>> valueConstraint)
+        {
+            var newSetterExpression = this.CreateSetterExpressionWithNewValue(valueConstraint);
+            var voidArgumentValidationConfiguration = this.CreateArgumentValidationConfiguration(newSetterExpression);
+            return AsPropertySetterConfiguration(voidArgumentValidationConfiguration);
+        }
+
+        public IAfterCallSpecifiedConfiguration Throws(Func<IFakeObjectCall, Exception> exceptionFactory)
+        {
+            return this.CreateArgumentValidationConfiguration(this.setterExpression)
+                .Throws(exceptionFactory);
+        }
+
+        public IPropertySetterConfiguration Invokes(Action<IFakeObjectCall> action)
+        {
+            var voidConfiguration = this.CreateArgumentValidationConfiguration(this.setterExpression)
+                .Invokes(action);
+            return AsPropertySetterConfiguration(voidConfiguration);
+        }
+
+        public IAfterCallSpecifiedConfiguration CallsBaseMethod()
+        {
+            return this.CreateArgumentValidationConfiguration(this.setterExpression)
+                .CallsBaseMethod();
+        }
+
+        public UnorderedCallAssertion MustHaveHappened(Repeated repeatConstraint)
+        {
+            return this.CreateArgumentValidationConfiguration(this.setterExpression)
+                .MustHaveHappened(repeatConstraint);
+        }
+
+        public IAfterCallSpecifiedConfiguration DoesNothing()
+        {
+            return this.CreateArgumentValidationConfiguration(this.setterExpression)
+                .DoesNothing();
+        }
+
+        public IPropertySetterConfiguration WhenArgumentsMatch(Func<ArgumentCollection, bool> argumentsPredicate)
+        {
+            var voidConfiguration = this.CreateArgumentValidationConfiguration(this.setterExpression)
+                .WhenArgumentsMatch(argumentsPredicate);
+            return AsPropertySetterConfiguration(voidConfiguration);
+        }
+
+        private static IPropertySetterConfiguration AsPropertySetterConfiguration(IVoidConfiguration voidArgumentValidationConfiguration)
+        {
+            return new ProperySetterAdapter(voidArgumentValidationConfiguration);
+        }
+
+        private MethodCallExpression CreateSetterExpressionWithNewValue(Expression<Func<TValue>> valueExpression)
+        {
+            var arguments = this.setterExpression.Arguments
+                .Take(this.setterExpression.Arguments.Count - 1)
+                .Concat(new[] { valueExpression.Body });
+            return Expression.Call(
+                this.setterExpression.Object,
+                this.setterExpression.Method,
+                arguments);
+        }
+
+        private IVoidArgumentValidationConfiguration CreateArgumentValidationConfiguration(
+            MethodCallExpression expression)
+        {
+            return this.voidArgumentValidationConfigurationFactory(Expression.Lambda(expression));
+        }
+
+        private class ProperySetterAdapter : IPropertySetterConfiguration
+        {
+            private IVoidConfiguration voidConfiguration;
+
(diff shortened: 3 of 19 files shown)


blairconrad:
Initial `A.CallTo` support. More coming, but I figured I'd give you an opportunity to check out the Specs and initial implementation.
No rush or pressure. Thanks.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -61,6 +65,61 @@ public IAnyCallConfigurationWithNoReturnTypeSpecified CallTo(object fakeObject)
...
+                this.callExpressionParser);
+
+            var expressionDescription = matcher.DescriptionOfMatchingCall;
+            return expressionDescription;
+        }
+
+        private LambdaExpression BuildSetterFromGetter<TValue>(Expression<Func<TValue>> propertySpecification)
+        {
+            MemberExpression memberExpression = propertySpecification.Body as MemberExpression;
+            if (memberExpression == null)
```
thomaslevesque: That won't work for indexed properties, because in that the body is a call to `get_TheProperty(index)` (of course in C# `TheProperty` is always `Item`, since named indexers are not allowed, but in VB it can be anything)

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -61,6 +65,61 @@ public IAnyCallConfigurationWithNoReturnTypeSpecified CallTo(object fakeObject)
...
+
+            var propertySetterInfo = propertyInfo.GetSetMethod(nonPublic: true);
+            if (propertySetterInfo == null)
+            {
+                throw new ArgumentException("The property '" + memberExpression.Member.Name + "' does not have a setter.");
+            }
+
+            Expression<Func<TValue>> value = () => A<TValue>.Ignored;
+
+            IEnumerable<Expression> indices = propertySpecification.Parameters.Cast<Expression>();
```
thomaslevesque: `Parameters` is the list of parameters of the lambda, not the property's indices. In this case it's always empty, because an `Func<TValue>` doesn't have any parameter.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -61,6 +65,61 @@ public IAnyCallConfigurationWithNoReturnTypeSpecified CallTo(object fakeObject)
...
+            var arguments = indices.Concat(new[] { value.Body });
+
+            return Expression.Lambda(Expression.Call(memberExpression.Expression, propertySetterInfo, arguments));
```
thomaslevesque: Shouldn't it be `Expression.Assign(memberExpression, value)` ?

thomaslevesque: I guess it depends on how we want to analyze it later

thomaslevesque:
Looking good so far :+1:

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: Yeah, I added a test for an indexer and discovered this very thing this (Thursday, as I type) morning. I started adding the conversion from indexed get to indexed set, but am not quite there (having problems finding the instance).
Also, I did not know that VB supported named indexers. I guess in that case the best way to detect them would be to look for `get_TheProperty` and assume nobody made a method that's so named… Which I've never tried, but it must be possible.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: Whoops. Good point. Something else that would've come up from the experiments with indexers. Thanks!

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: I see what you're saying, but this converts the property to a call to `set_TheProperty`, which I knew would work with all the other plumbing in the background. We can experiment with an `Expression.Assign` later, if you like, but I fear it would lead to much more work.

blairconrad:
Thanks for the comments, @thomaslevesque. They're quite helpful. And for the encouragement. I've had very little time to work on this, but am attempting to plug away.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
thomaslevesque: > We can experiment with an Expression.Assign later, if you like, but I fear it would lead to much more work.

No, forget it, it's simpler your way. Out of curiosity, I tried to manually craft an expression with a property assignment and pass it to `A.CallTo`. Of course, it failed miserably; the expression parser just doesn't support it.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
thomaslevesque: > (having problems finding the instance)

`MethodCallExpression.Object`?

> I guess in that case the best way to detect them would be to look for `get_TheProperty` and assume nobody made a method that's so named…

You can check if the `IsSpecialName` property on the `MethodInfo` to distinguish between a property getter and a "normal" method. And you can't have both a `get_TheProperty` method and a `TheProperty` property with a getter in the same class, the compiler won't let you.

blairconrad: > MethodCallExpression.Object?

Yes! I'd tried that yesterday with no luck at all, and was going crazy. I even had it sitting, commented-out, in my editor. Today I uncommented it, replaced a `.GetType()` access with `.Type`, and used the arguments to the method instead of the parameters for the lambda to construct the call and it worked _just like it was supposed to_.

Thank you!

> `IsSpecialName`

of course. I was even already using `IsSpecialName` when looking explicitly for `get_Item`. I should need only a minor change to support named indexers. And I guess a VB test project. How terrible would it be to use the VB Integration test project for this kind of thing?

thomaslevesque: > And I guess a VB test project. How terrible would it be to use the VB Integration test project for this kind of thing?

Not terrible I think, if it's an integration test ;)

Actually, the test itself could be in C#; only the class that declares an indexed property would have to be in VB (IIRC C# 4+ can consume indexed properties, even if it can't declare them)

blairconrad:
I _think_ that's all the possibilities and error messages we need for the basic `A.CallToSet`. Starting on the `To` interaction.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -61,6 +59,123 @@ public IAnyCallConfigurationWithNoReturnTypeSpecified CallTo(object fakeObject)
...
+        private static string GetIndexerName(MethodCallExpression methodCallExpression)
+        {
+            if (methodCallExpression != null &&
+                HasThis(methodCallExpression) &&
+                methodCallExpression.Method.IsSpecialName)
+            {
+                var methodName = methodCallExpression.Method.Name;
+                if (methodName.StartsWith("get_", StringComparison.Ordinal))
+                {
+                    return methodName.Substring(4);
```
thomaslevesque: Shouldn't you check the number of arguments as well? If there's none, then it's not an indexer, it's just a normal property

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -61,6 +59,123 @@ public IAnyCallConfigurationWithNoReturnTypeSpecified CallTo(object fakeObject)
...
+                }
+            }
+
+            return null;
+        }
+
+        private static bool HasThis(MethodCallExpression methodCallExpression)
+        {
+            return (methodCallExpression.Method.CallingConvention & CallingConventions.HasThis) ==
```
thomaslevesque: There's a `HasFlag` method on `Enum` which could make this code clearer (at the cost of a boxing the argument, though).

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -61,6 +59,123 @@ public IAnyCallConfigurationWithNoReturnTypeSpecified CallTo(object fakeObject)
...
+                                            "', which is a field, not a property getter.");
+            }
+
+            var propertySetterInfo = propertyInfo.GetSetMethod(nonPublic: true);
+            if (propertySetterInfo == null)
+            {
+                throw new ArgumentException("The property '" + memberExpression.Member.Name + "' does not have a setter.");
+            }
+
+            var arguments = new[] { BuildArgumentThatMatchesAnything<TValue>() };
```
thomaslevesque: You could also use `Expression.Property(null, typeof(A<>), nameof(A<TValue>.Ignored))`

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -61,6 +59,123 @@ public IAnyCallConfigurationWithNoReturnTypeSpecified CallTo(object fakeObject)
...
+
+        private string GetExpressionDescription<TValue>(Expression<Func<TValue>> expression)
+        {
+            var matcher = new ExpressionCallMatcher(
+                expression,
+                ServiceLocator.Current.Resolve<ExpressionArgumentConstraintFactory>(),
+                ServiceLocator.Current.Resolve<MethodInfoManager>(),
+                this.callExpressionParser);
+
+            var expressionDescription = matcher.DescriptionOfMatchingCall;
```
thomaslevesque: Creating an `ExpressionCallMatcher` just to get its description doesn't seem quite right to me. Perhaps the description generation should be taken out of `ExpressionCallMatcher`? (I haven't looked at the code yet, so I'm not sure if it makes sense)

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
```diff
@@ -0,0 +1,92 @@
...
+                .Invokes(action);
+        }
+
+        public IAfterCallSpecifiedConfiguration CallsBaseMethod()
+        {
+            return this.CreateArgumentValidationConfiguration(this.setterExpression)
+                .CallsBaseMethod();
+        }
+
+        public IAfterCallSpecifiedConfiguration AssignsOutAndRefParametersLazily(
```
thomaslevesque: Does this make sense for a property setter?

src/FakeItEasy/Configuration/IPropertySetterConfiguration.cs
```diff
@@ -0,0 +1,19 @@
...
+        : IVoidArgumentValidationConfiguration
+    {
+        /// <summary>
+        /// Configures the property assignment to be accepted when the value <see cref="object.Equals(object)"/>
+        /// <paramref name="value"/>.
+        /// </summary>
+        /// <param name="value">The value to match.</param>
+        /// <returns>A configuration object.</returns>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "To", Justification = "There's no need for clients to implement the member.")]
+        IVoidArgumentValidationConfiguration To(TValue value);
```
thomaslevesque: Shouldn't there be an overload that takes a lambda? I think we mentioned that at some point in #175.

thomaslevesque:
Great work @blairconrad! Looks like it's almost done.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: Not in 3.5. And adding an extension method to an enum is quite tricky.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: The check on https://github.com/FakeItEasy/FakeItEasy/pull/673/files/e7918307aa92a1d7e91730d61752a74f58f46cee#diff-ab0b145be13922e815c0f04e2bc3b878R143 would have caught that possibility.
Is it too far removed from this code for comfort?

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: All the `Expression.Property`s I see take 2 arguments. I don't doubt you, but I'm confused.
Do you prefer the new version (if so, teach me why!), or was that just for education?

src/FakeItEasy/Configuration/IPropertySetterConfiguration.cs
blairconrad: There now. I didn't push right. :grimacing:

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
blairconrad: It seems not. `out` and `ref` aren't supported even for indexers, it seems?
It's unlikely to be used by anyone. We could throw an invalid operation (we may anyhow).
Or would you rather we made up new interfaces for the fluent API to avoid confusion?

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: I also did not look that deeply. I saw what the other code was using and ripped it off, hackily.
I'm open to other options, but will not be able to spend too much time on it in the next nearly 3 weeks…

blairconrad:
@thomaslevesque, thanks for the great comments. I rebutted. I'm not against making any changes in particular, but did not know how to address some (the alternative way to make `Ignored`).
Work, yardwork, and vacation prep are robbing me of time, so if we want this done before I go, we may have to cut corners. Alternatively, I'm happy to pass the branch off to have anyone else finish (although this has been sort of fun).

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
blairconrad: As an aside, if someone _tries_ to use the method, they're confronted with

```
System.InvalidOperationException

The number of values for out and ref parameters specified does not match the number of out and ref parameters in the call.
```

blairconrad: >  would you rather we made up new interfaces for the fluent API to avoid confusion?

I'm beginning to think that I would rather we did…

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
thomaslevesque: Well, the way you check it works in C#, but maybe not in another language. It depends on the fact that the code `foo.Bar` will be converted to a `MemberExpression`, but a different compiler might generate an expression like `foo.get_Bar()` instead (after all, that's what the C# does for indexed properties).

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
thomaslevesque: Ah, yes, forgot about 3.5. Adding a *generic* extension method for enums is indeed tricky, since you can't use `Enum` as a generic constraint, but you can still create an extension method on the `Enum` type itself. However in this case it's probably not worth the trouble, because the implementation would have to depend on the underlying type.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
thomaslevesque: https://msdn.microsoft.com/en-us/library/dd294147%28v=vs.110%29.aspx

But this overload was added in 4.0, unfortunately...

Your way is fine, it's just that using the compiler to generate an expression from a lambda and extracting its body seemed a bit roundabout to me.

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
thomaslevesque: > out and ref aren't supported even for indexers, it seems?

No, they're not.

> We could throw an invalid operation (we may anyhow).

We could, but it would be better if we could prevent it at compile time.

> Or would you rather we made up new interfaces for the fluent API to avoid confusion?

It's unfortunate, but yes, I think it would be the right way to handle that...

Maybe we could change `IVoidConfiguration` so that it no longer inherits `IOutAndRefParametersConfiguration`, and introduce a new `IVoidWithOutAndRefParametersConfiguration`? (it would be a breaking change, but unlikely to affect many people). It would probably be the easiest route; creating a completely new interface would require an adapter.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: … and for whatever reason, the overloads all showed up for me today. I swear last night I saw 3, and today 5. I wonder what I was really typing…

I ended up needing 

```csharp
Expression.Property(null, typeof(A<TValue>), nameof(A<TValue>.Ignored));
```

Omitting the first `TValue` went poorly.

Thanks for persevering!

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: I was about to add the check, and realized: what if there are no arguments? We'll try to configure a 1-argument "indexed setter", which should be fine. Perhaps a better option would be to just rename the "Indexer" methods?

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: Well this is embarrassing. Remember when I said things went well 11 minutes ago? They don't, and you figured out why:

> But this overload was added in 4.0, unfortunately...

So it's not in 3.5. Shall we be a bit roundabout? Or would you prefer diverging implementations for 3.5 and 4.0+?

thomaslevesque: > … and for whatever reason, the overloads all showed up for me today. I swear last night I saw 3, and today 5. I wonder what I was really typing…

Probably because the current project was FakeItEasy.Net35. You can't use this method, it's not available in 3.5.

thomaslevesque: > So it's not in 3.5.

Yes, that's why I said "your way is fine" :wink:

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
blairconrad: I've removed `IVoidWithOutAndRefParametersConfiguration` from `IPropertySetterConfiguration` with no problems, so it will not appear until _after_ one uses `WhenArgumentsMatch` or similar. Continuing to track down remaining appearances.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
thomaslevesque: Yes, I guess that would work

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
blairconrad: @thomaslevesque, do you absolutely hate an adapter? I wasn't that worried about the breaking change, but was worried about messing with the existing interfaces _too much_ and decided to try out an adapter, to what I think is success.

thomaslevesque: I don't hate it, it's just that I'm always reluctant to add extra layers of indirections when it can be avoided. But if it seems to be the best option, then go ahead :+1:

blairconrad:
I think all functionality is in place, and that I've addressed all comments except for https://github.com/FakeItEasy/FakeItEasy/pull/673#discussion_r60991134, so I'm going to optimistically remove "[WIP]".
I'll try to investigate how better to describe the "bad" expressions. So far, I have no ideas, though.

blairconrad:
Of course, I welcome any and all comments. And would expect to squash before the PR is merged. Probably keeping bfab58c separate, but I'm open to whatever.

adamralph:
Thanks @blairconrad and @thomaslevesque for your efforts on this so far.

I've been keeping one eye on the progress but I've not had time to get involved in the details. I'll try and make a pass over the changes now, however.

blairconrad:
Thanks, @adamralph. It occurs to me that I need to add diagnostic attributes. I'm so excited!

thomaslevesque:
Last minute thought: the name `To` is a little ambiguous. Have we considered another name? I see that `ToValue` was mentioned at some point in #175, and I think it would be better.

adamralph:
I'm not sure that `To` is ambigious. If Intellisense leads me to

```C#
A.CallToSet(question.Answer).To(
```

Then my mind sees "a call to set the question answer to..." and I'd feel compelled to write "42" next.

blairconrad:
The two versions read about the same to me, but lemme ask: @adamralph, would you find `ToValue` confusing? Or worse than `To` in any way other than that it takes you 5 keystrokes closer to death?

If you've no objection, and `To`'s a little unclear to @thomaslevesque, why not change it to be more explicit? Maybe `To` works really well for native speakers of English but will confuse others.
As another benefit, I don't think `ToValue` is a reserved word like `To` is.

blairconrad:
Oh, and a last-minute nit. @adamralph, you'd more likely see

```C#
A.CallToSet(() => question.Answer).To(
```

thomaslevesque:
> If you've no objection, and To's a little unclear to @thomaslevesque, why not change it to be more explicit? Maybe To works really well for native speakers of English but will confuse others.

It's not really unclear to me, but `ToValue` is more specific.

I don't have a strong opinion on this, I just wanted to make sure we considered the alternatives and made a conscious decision.

> As another benefit, I don't think ToValue is a reserved word like To is.

Indeed. It doesn't seem to be an issue in VB.NET, though; I don't know about other languages.

adamralph:
`ToValue` isn't confusing but `Value` just feels redundant. Just as it would be in

```C#
A.CallTo(() => question.Answer).ReturnsValue(42);
```

blairconrad:
I see the redundancy argument. Also,

```C#
A.CallToSet(() => question.Answer).ToValue(42); // reads okay
A.CallToSet(() => question.Answer).ToValue(() => A<int>.Ignored); // I think reads less well than
A.CallToSet(() => question.Answer).To(() => A<int>.Ignored);
```

I'm coming around to keeping `To`.

blairconrad:
I just noticed some typos in the last commits and repushed 3. Like a jerk.

blairconrad:
Regarding moving the method description off `ExpressionCallMatcher`, it seems possible, but would require double work - the `ExpressionCallMatcher` relies on fancy expression parsing, which is also needed to describe the call. I'm essentially duplicating the work in `FakeConfigurationManager.GetExpressionDescription` anyhow, I suppose, so I could take a stab at extracting a call describer, if you're interested.

thomaslevesque:
> Regarding moving the method description off ExpressionCallMatcher, it seems possible, but would require double work - the ExpressionCallMatcher relies on fancy expression parsing, which is also needed to describe the call. I'm essentially duplicating the work in FakeConfigurationManager.GetExpressionDescription anyhow, I suppose, so I could take a stab at extracting a call describer, if you're interested.

It can be done later in another PR. No point in delaying this one :wink:

I have no more objections. Do you have anything else to add before we merge @blairconrad?

blairconrad:
@thomaslevesque, I'm content, although the commits are a bunch of rework. I should probably squash.
Should I just keep bfab58c separate, as it's not necessary for this work, and combine the rest?

Also let me say @thomaslevesque, that I _really_ enjoyed just being able to slap attributes on the new intermediate API steps whose results must be used. Thanks again for the hard work on the analyzer!

blairconrad:
I squashed a lot. Kept the unsquashed bits at https://github.com/blairconrad/FakeItEasy/tree/new-property-setter-syntax-before-squash, in case you hate it.

thomaslevesque:
> I squashed a lot.

Looking good! Merging now.

blairconrad:
Oh, thanks. I've said it before, but I really appreciated your comments on this review (and the others, of course). They really improved things, as well as speeding the changes along. I'm new to the crazy world of Expressions. And even more new to tidbits like "in VB.NET, indexers can have any name".

PR #690, opened 2016-04-29
Fix AssignsOutAndRefParameters for faked delegates

### LINKED ISSUE #677: AssignsOutAndRefParameters on delegate proxies
Inspired by https://twitter.com/th0maswe1ss/status/723799176941699072.

`AssignsOutAndRefParameters` appears not to work for faked delegates.
[`FakeItEasy.Tests.Creation.DelegateProxies.DelegateProxyGeneratorTests.Should_return_proxy_where_out_parameter_can_be_set`](https://github.com/FakeItEasy/FakeItEasy/blob/126d6056334227cda86eeded31df36b60378cf70/tests/FakeItEasy.Tests/Creation/DelegateProxies/DelegateProxyGeneratorTests.cs#L154) has been `Ignore`d since the project came to GitHub.

If there's a way to enable assigning of out and ref parameters for faked delegates, I think we should do it.
If not, we should document the lack and provide an error when it's attempted.

thomaslevesque:
Fixes #677

WIP because I still need to add specs

I took into account the other remarks from https://github.com/thomaslevesque/FakeItEasy/commit/5570756fa1dbefc872804ad7a80bf3ad87aa9f5d

### DIFF
--- a/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
+++ b/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
@@ -74,22 +74,106 @@ namespace FakeItEasy.Creation.DelegateProxies
         }
 
         private static Expression CreateBodyExpression(
-            MethodInfo delegateMethod, DelegateCallInterceptedEventRaiser eventRaiser, IEnumerable<Expression> parameterExpressions)
+            MethodInfo delegateMethod,
+            DelegateCallInterceptedEventRaiser eventRaiser,
+            ParameterExpression[] parameterExpressions)
         {
-            var parameterExpressionsCastToObject =
-                parameterExpressions.Select(x => Expression.Convert(x, typeof(object))).Cast<Expression>().ToArray();
+            /*
+             * Generate a method that:
+             * - wraps its arguments in an object array
+             * - passes this array to eventRaiser.Raise()
+             * - assigns the output values back to the ref/out parameters
+             * - casts and returns the result of eventRaiser.Raise()
+             *
+             * For instance, for a delegate like this:
+             *
+             * delegate int Foo(int x, ref int y, out int z);
+             *
+             * We generate a method like this:
+             *
+             * int ProxyFoo(int x, ref int y, out int z)
+             * {
+             *     var arguments = new[]{ (object)x, (object)y, (object)z };
+             *     var result = (int)eventRaiser.Raise(arguments);
+             *     y = (int)arguments[1];
+             *     z = (int)arguments[2];
+             *     return result;
+             * }
+             *
+             * Or, for a delegate with void return type:
+             *
+             * delegate void Foo(int x, ref int y, out int z);
+             *
+             * void ProxyFoo(int x, ref int y, out int z)
+             * {
+             *     var arguments = new[]{ (object)x, (object)y, (object)z };
+             *     eventRaiser.Raise(arguments);
+             *     y = (int)arguments[1];
+             *     z = (int)arguments[2];
+             * }
+             */
 
-            Expression body = Expression.Call(
+            bool isVoid = delegateMethod.ReturnType == typeof(void);
+
+            // Local variables of the generated method
+            var arguments = Expression.Variable(typeof(object[]), "arguments");
+            var result = isVoid ? null : Expression.Variable(delegateMethod.ReturnType, "result");
+
+            var bodyExpressions = new List<Expression>();
+
+            bodyExpressions.Add(Expression.Assign(arguments, WrapParametersInObjectArray(parameterExpressions)));
+
+            Expression call = Expression.Call(
                 Expression.Constant(eventRaiser),
                 DelegateCallInterceptedEventRaiser.RaiseMethod,
-                new Expression[] { Expression.NewArrayInit(typeof(object), parameterExpressionsCastToObject) });
+                arguments);
 
-            if (!delegateMethod.ReturnType.Equals(typeof(void)))
+            if (!isVoid)
             {
-                body = Expression.Convert(body, delegateMethod.ReturnType);
+                // If the return type is non void, cast the result of eventRaiser.Raise()
+                // to the real return type and assign to the result variable
+                call = Expression.Assign(result, Expression.Convert(call, delegateMethod.ReturnType));
             }
 
+            bodyExpressions.Add(call);
+
+            // After the call, copy the values back to the ref/out parameters
+            for (int index = 0; index < parameterExpressions.Length; index++)
+            {
+                var parameter = parameterExpressions[index];
+                if (parameter.IsByRef)
+                {
+                    var assignment = AssignParameterFromArrayElement(arguments, index, parameter);
+                    bodyExpressions.Add(assignment);
+                }
+            }
+
+            // Return the result if the return type is non-void
+            if (!isVoid)
+            {
+                bodyExpressions.Add(result);
+            }
+
+            var variables = isVoid ? new[] { arguments } : new[] { arguments, result };
+            var body = Expression.Block(variables, bodyExpressions);
             return body;
+        }
+
+        private static BinaryExpression AssignParameterFromArrayElement(ParameterExpression arguments, int index, ParameterExpression parameter)
+        {
+            var conversion =
+                Expression.Convert(
+                    Expression.ArrayAccess(arguments, Expression.Constant(index)),
+                    parameter.Type);
+            var assignment = Expression.Assign(parameter, conversion);
+            return assignment;
+        }
+
+        private static NewArrayExpression WrapParametersInObjectArray(ParameterExpression[] parameterExpressions)
+        {
+            return Expression.NewArrayInit(
+                typeof(object),
+                parameterExpressions.Select(x => Expression.Convert(x, typeof(object))));
         }
 
         private class DelegateCallInterceptedEventRaiser
--- a/tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
+++ b/tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
@@ -6,6 +6,10 @@ namespace FakeItEasy.Specs
     using FluentAssertions;
     using Xbehave;
 
+    public delegate void VoidDelegateWithOutAndRefParameters(int x, ref int y, out int z);
+
+    public delegate int NonVoidDelegateWithOutAndRefParameters(int x, ref int y, out int z);
+
     public interface IHaveAnOut
     {
         [SuppressMessage("Microsoft.Design", "CA1021:AvoidOutParameters", MessageId = "0#", Justification = "Required for testing.")]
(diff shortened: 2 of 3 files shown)


src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -74,21 +74,88 @@ private static Delegate CreateDelegateProxy(Type typeOfProxy, MethodInfo invokeM
...
+             *
+             * int ProxyFoo(int x, ref int y, out int z)
+             * {
+             *     var arguments = new[]{ (object)x, (object)y, (object)z };
+             *     var result = (int)eventRaiser.Raise(arguments);
+             *     y = (int)arguments[1];
+             *     z = (int)arguments[2];
+             *     return result;
+             * }
+             ********************************************************************/
```
blairconrad: A matter of style, I think, but I found the lines of `****` jarring. I think omitting them and using regular single-line comments (`//`) would be (slightly) better.
This is a very minor comment!

However, the content is very clear and really aids my understanding of the following code.

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -74,21 +74,88 @@ private static Delegate CreateDelegateProxy(Type typeOfProxy, MethodInfo invokeM
...
+
+            bool isVoid = delegateMethod.ReturnType == typeof(void);
+
+            // Local variables of the generated method
+            var arguments = Expression.Variable(typeof(object[]), "arguments");
+            var result = isVoid ? null : Expression.Variable(delegateMethod.ReturnType, "result");
+
+            var bodyExpressions = new List<Expression>();
+
+            // arguments = new[]{ (object)x, (object)y, (object)z };
```
blairconrad: I'm not sure this adds much over the comment above. But if we think it's necessary, this is an example of where I'd introduce a method rather than a comment, such as

```c#
var arrayInit = WrapArgumentsInObjectArray(parameterExpressions);
```

blairconrad: Similarly almost every time there's a comment below.

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -74,21 +74,88 @@ private static Delegate CreateDelegateProxy(Type typeOfProxy, MethodInfo invokeM
...
                 Expression.Constant(eventRaiser),
                 DelegateCallInterceptedEventRaiser.RaiseMethod,
-                new Expression[] { Expression.NewArrayInit(typeof(object), parameterExpressionsCastToObject) });
+                arguments);
+
+            if (!isVoid)
+            {
+                // If the return type is non void, cast the result of eventRaiser.Raise()
+                // to the real return type
+                call = Expression.Convert(call, delegateMethod.ReturnType);
```
blairconrad: I would almost consider pulling `call = Expression.Assign(result, call)` from line 133 into this block, then having line 133 just becomes `bodyExpressions.Add(call)`, rather than having the reader re-evaluate the `isVoid`ness.

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
thomaslevesque: I used `/* */` because Stylecop is a bit touchy about single-line comments:

> SA1512 : CSharp.Layout : A single-line comment must not be followed by a blank line. To ignore this error when commenting out a line of code, begin the comment with '////' rather than '//'.

But I can use `////` instead.

blairconrad:
@thomaslevesque, looks good! I made some very minor controversial comments about code style, but really I'm content as things are.

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
thomaslevesque: OK

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
thomaslevesque: Good idea :+1:

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
blairconrad: Oh. But that seems an abuse of `////` as well.
The problems is that you wanted

```c#
//     return result;
// }

bool isVoid = delegateMethod.ReturnType == typeof(void);
```

but StyleCop hated the line before `isVoid` and you don't like

```c#
//     return result;
// }
 bool isVoid = delegateMethod.ReturnType == typeof(void);
```

because it's too squished together?

I'm not wild about the squished up thing either.
I'm happy if you keep what you have, especially since I got around to opening the file alone and see that the block is indented. When I first looked at it, GitHub made me think it was left-aligned, which was probably the real "punch to the eye".

thomaslevesque: OK, I'll keep the `/* */` then

thomaslevesque: And remove the lines of `*` since they bother you ;)

blairconrad: Haha. Don't feel you have to. It was just the initial view I had:

image

that looked so weird. Seeing

image

is not so bad.

thomaslevesque: Too late :wink:

It would be nice to be able to use markdown in C# comments for better formatting...

thomaslevesque:
Made a few changes in the implementation and added specs

Now rewriting existing specs to GWT style

thomaslevesque:
> Now rewriting existing specs to GWT style

It seems to me that the existing specs in `AssignsOutAndRefParametersSpecs` don't make much sense... They all use a condition on the input value of an `out` parameter. But in C#, you can't read the value of an out parameter before you assign it, so it's not possible to get the initial value... It works only because the proxy uses generated code, which doesn't have this restriction. Unless you have a better idea, I'll change the parameter to `ref` so that it doesn't seem as weird.

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
blairconrad: Surely there's an extension somewhere.

thomaslevesque:
All done, I think

thomaslevesque:
I'll squash the first 3 commits together, and leave the last one separate.

thomaslevesque:
Oops, looks like I introduced a ton of code analysis violations... will fix ASAP

tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
```diff
@@ -2,122 +2,169 @@ namespace FakeItEasy.Specs
...
+        }
+
+        [Scenario]
+        public static void AssignOutAndRefParameterForNonVoidDelegate(
+            NonVoidDelegateWithOutAndRefParameters subject,
+            int refValue,
+            int outValue,
+            int result)
+        {
+            "Given a faked delegate with void return type and ref and out parameters"
```
blairconrad: not void

tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
```diff
@@ -2,122 +2,169 @@ namespace FakeItEasy.Specs
...
-                        string value = condition;
-                        subject.MightReturnAKnownValue(out value);
-                        value.Should().BeEquivalentTo(knownOutput);
-                    });
+                {
+                    outValue = Condition;
+                    subject.MightReturnAKnownValue(ref outValue);
+                });
+
+            "Then it should assign the conditional value to the ref field"
```
blairconrad: Based on recent comments in my property setter issue, I think @adamralph would push for something more like "Then is assigns the conditional value to the ref field". But he's not here.

blairconrad:
Looks quite good to me, @thomaslevesque. Did note 1.5 small things up.
And it may not be to everyone's taste, but I did find the new version of the generator to be much more readable. Thanks.

tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
thomaslevesque: Nice catch! Copy/paste error...

tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
thomaslevesque: Since I'm at it, I might as well do it right. Will change it.

tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
```diff
@@ -2,122 +2,175 @@ namespace FakeItEasy.Specs
...
     {
-        [SuppressMessage("Microsoft.Design", "CA1021:AvoidOutParameters", MessageId = "0#", Justification = "Required for testing.")]
-        void MightReturnAKnownValue(out string andThisIsWhoReallyDidIt);
+        [SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "0#", Justification = "Required for testing.")]
+        void MightReturnAKnownValue(ref string andThisIsWhoReallyDidIt);
     }
 
     public static class AssignsOutAndRefParametersSpecs
     {
+        const string Condition = "someone_else";
```
blairconrad: Late-breaking comment. I don't think we have a test where we establish a condition and then don't meet it when exercising the subject. If we care about the ability to decide whether or not to assign the output, then maybe we should have an extra test. 
If we don't, the tests could become simpler...

thomaslevesque: Actually, I don't think the condition really matters. The out parameter *is* set to the `KnownOutput`, which proves that the lambda was executed; I think it's all that matters. And yes, the tests could be made simpler by removing the condition (and then I wouldn't have needed to change `out` to `ref`...)

thomaslevesque:
@blairconrad I made the requested changes (except the one mentioned in https://github.com/FakeItEasy/FakeItEasy/pull/690#discussion_r61631099, but I'm not sure I should). I'll squash when I have your go.

blairconrad:
@thomaslevesque, as we discussed off line, https://github.com/FakeItEasy/FakeItEasy/pull/690#discussion_r61631099 is out of scope and could be done later if we want.
Nice work. Squash it up!

thomaslevesque:
Squashed

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -74,24 +74,108 @@ private static Delegate CreateDelegateProxy(Type typeOfProxy, MethodInfo invokeM
...
+             * delegate void Foo(int x, ref int y, out int z);
+             *
+             * void ProxyFoo(int x, ref int y, out int z)
+             * {
+             *     var arguments = new[]{ (object)x, (object)y, (object)z };
+             *     eventRaiser.Raise(arguments);
+             *     y = (int)arguments[1];
+             *     z = (int)arguments[2];
+             * }
+             */
```
adamralph: I'm late to the game here, but I find this comment style a little odd. The leading `*` on each line seems redundant. I would expect

```C#
            /*
            Generate a method that:
            - wraps its arguments in an object array
            - passes this array to eventRaiser.Raise()
            ...
            */
```

or better still

```C#
            // Generate a method that:
            // - wraps its arguments in an object array
            // - passes this array to eventRaiser.Raise()
            ...
```

This would be consistent with other places where we have multi-line comments (I've never been a fan of `/* ... */` style comments):

* https://github.com/FakeItEasy/FakeItEasy/blob/3724e900b687981898f91d17c87c742568b5fce8/src/FakeItEasy/Core/FakeManager.EventRule.cs#L58-L62
* https://github.com/FakeItEasy/FakeItEasy/blob/3724e900b687981898f91d17c87c742568b5fce8/tests/FakeItEasy.Tests/Core/ArgumentConstraintTrapTests.cs#L68-L76

There are also quite a few 3 line comments and even plenty of 2 line comments, and they all use the `//` style.

adamralph: OK, so if the `//` style is used then StyleCop gives:

> SA1512 : CSharp.Layout : A single-line comment must not be followed by a blank line. To ignore this error when commenting out a line of code, begin the comment with '////' rather than '//'.

Perhaps `/*..*/` style is appropriate here. I do find the leading `*`s quite weird though. Is there any precedent for that style?

thomaslevesque: @adamralph, see the first comment thread (hidden - click "show comments" where it says "blairconrad and 1 other commented on an outdated diff") for the discussion about this.

Basically, if you have `//` comments followed by a blank line, StyleCop complains (because obviously, it knows better than us mere humans...). In the case of `Should_track_constraints_supplied_in_calls_made_from_overlapping_threads`, it works only because there's another comment after the blank line - if you remove it, it gives SA1512. That's why I used `/* */`. The alternatives are:

- use `////` instead of `//` (looks a bit weird)
- don't leave a blank line after the comment block (but it looks like it only applies to the next line)
- put another comment before the next line of code
- suppress the warning
- drop StyleCop :stuck_out_tongue:

thomaslevesque: > Perhaps /*..*/ style is appropriate here. I do find the leading *s quite weird though. Is there any precedent for that style?

Well, it must be pretty standard, because if you write a multiline `/* */` comment, Visual Studio automatically introduces the `*` on each new line (unless it's ReSharper?)

adamralph: @thomaslevesque it looks like our last comments crossed each other transit.

adamralph: It must be Resharper being "helpful". Vanilla VS doesn't insert `*`.

thomaslevesque: I've been using R# for so long that I don't even know what the normal VS behavior is :wink:

thomaslevesque: Now that I think of it, the comment could be outside the method, just above it. This way StyleCop won't complain, even with `//` comments.

adamralph: Ah, of course! And that's probably where the comment does belong...

I did a few other minor refactorings while reading this file (also the older code, not just the latest changes) so I'll send a :lipstick: refactoring PR shortly...

Issue #666, opened 2016-04-14
codify how we're handling the analyzer package


blairconrad:
All of a sudden, thanks to @thomaslevesque, we have an analyzer package. :tada: 

It's built alongside the main package, and currently takes the same version number as the main package.

How are we going to release this?

1. It'll need its own NuGet package.
2. How shall we maintain its release notes? Separately, or will there be one set of notes for both packages?
3. Ditto the documentation.
4. What do we do when we want to 
    1. release a package that has FakeItEasy changes but no Analyzer changes? 
    2. release a package that has Analyzer changes but no FakeItEasy changes? 

I'm sure there are other questions that I just haven't thought of right now.

thomaslevesque:
> How shall we maintain its release notes? Separately, or will there be one set of notes for both packages?

I'd say one set of release notes for both packages. The analyzer is probably not going to change very often anyway.

> Ditto the documentation.

Does it need documentation? I don't think there's much to say about it, besides how to install it. It can probably be included in the main FakeItEasy documentation.

> What do we do when we want to
> 
>     i. release a package that has FakeItEasy changes but no Analyzer changes?
>     ii. release a package that has Analyzer changes but no FakeItEasy changes?

Increment the version for both. This way, you don't have to think about which analyzer version you need for which FakeItEasy version: just pick the same version of both.
Having different version numbers would be a maintenance nightmare IMO, unless we move the analyzer to its own repo.

blairconrad:
> I'd say one set of release notes for both packages

So the analyzer NuGet package will have all kinds of release notes that talk about how we added a 
new property setting syntax or whatever, even though the analyzer doesn't change at all?
Sounds a little confusing.

> Increment the version for both

Works for me, if nobody minds having FakeItEasy bump up its release number with no change in functionality.

> unless we move the analyzer to its own repo.

That's an interesting question. Is there a reason for it to be in the same repo?
We've wondered whether samples and docs should be tied to the same repo. Of any of these, the analyzer seems like the the one the least tied to the FakeItEasy source repo. Not that I want to move it out.

thomaslevesque:
> That's an interesting question. Is there a reason for it to be in the same repo?

Not really. The analyzer tests (not the analyzer itself) references FakeItEasy, but it can do so via NuGet. The only issue is that if we add a new diagnostic for a new FakeItEasy feature, we might need to reference the latest version of FakeItEasy before its released. It could probably be handled by including FakeItEasy as a submodule.

[...]

PR #694, opened 2016-05-01
Remove analyzer

thomaslevesque:
which now lives in its own repo at https://github.com/FakeItEasy/FakeItEasy.Analyzer


thomaslevesque:
I also removed the `MustUseReturnValue` attribute, but I realize now that it should probably be a separate PR.

thomaslevesque:
> I also removed the MustUseReturnValue attribute, but I realize now that it should probably be a separate PR.

Fixed

adamralph:
Let's make sure we have a clear path forward with https://github.com/FakeItEasy/FakeItEasy/issues/666#issuecomment-216068360 before going ahead with this.

PR #702, opened 2016-05-05
Add CollectionOfDummy. Issue #675.

### LINKED ISSUE #675: Helper method for list of dummies

```cs
```
```cs
{
    {
        {
            yield return A.Dummy<T>();
        }
    }
}
```

---

- [x] implement - ~~#702~~
- [ ] update docs - #710

t0mburton:
fixes #675

### DIFF
--- a/src/FakeItEasy/A.cs
+++ b/src/FakeItEasy/A.cs
@@ -76,6 +76,19 @@ namespace FakeItEasy
         }
 
         /// <summary>
+        /// Creates a collection of dummies of the specified type.
+        /// </summary>
+        /// <typeparam name="T">The type of dummies to create.</typeparam>
+        /// <param name="numberOfDummies">The number of dummies in the collection.</param>
+        /// <returns>A collection of dummy objects of the specified type.</returns>
+        /// <exception cref="ArgumentException">Dummies of the specified type can not be created.</exception>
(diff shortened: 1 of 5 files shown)


adamralph:
Thanks @t0mburton! The change set looks spot-on :ok_hand: 

@FakeItEasy/owners this PR LGTM. Any objection to taking this straight to RTM and bypassing 2.0.0-rc? The only public API change is the addition of `A.IList<T> CollectionOfDummy<T>(int numberOfDummies)`.

In fact, it would be interesting to merge #703 first and use this as a pilot for our API approval test.

thomaslevesque:
Looks good to me too :+1:
Thanks @t0mburton!

thomaslevesque:
> @FakeItEasy/owners this PR LGTM. Any objection to taking this straight to RTM and bypassing 2.0.0-rc? The only public API change is the addition of A.IList<T> CollectionOfDummy<T>(int numberOfDummies).

No objection from me

> In fact, it would be interesting to merge #703 first and use this as a pilot for our API approval test.

Good idea :+1:

adamralph:
Thanks again @t0mburton! This will be in the 2.0.0 release. Look out for your name in the release notes! :trophy:

PR #703, opened 2016-05-05
add API approval test

### LINKED ISSUE #701: Add an API approval test
Using https://github.com/approvals/ApprovalTests.Net and https://www.nuget.org/packages/PublicApiGenerator/

This will allow us to see the effect of any change to the public API.

Initially, at least, I don't think we should include this in the default rake tasks, since the management of this test will be a little different to usual and I think it will add unwanted friction for contributors. We can still eyeball PRs as we do know to spot breaking changes, and if unsure, we can pull the changes down locally and run the approval test (`bundle exec rake approve`). We also have the option of setting up a second build config which runs this test and make it a "not required" check in GitHub for merging a PR. I.e. if a contributor changes the public API in a non-breaking way, we can eyeball the results of the failed build and verify that, merge the PR, and take care of updating the test later.

adamralph:
closes #701 

It works like this:

* `tests/FakeItEasy.Tests.Approval/ApiApproval.ApproveApi.approved.txt` contains the state of the public API when the approval test was last run and approved
* when you run `FakeItEasy.Tests.Approval.ApiApproval.ApproveApi` or `bundle exec rake approve`, the current state of the API will be written to `tests/FakeItEasy.Tests.Approval/ApiApproval.ApproveApi.received.txt`
* if the state differs, you will either be presented with a diff (when running from VS) or you will see a test  failure (when running from console)
* if you approve of the change, you update `ApiApproval.ApproveApi.approved.txt` with the contents of `ApiApproval.ApproveApi.received.txt` (if in a diff viewer, right click -> "use this file")
* you commit the changed `ApiApproval.ApproveApi.approved.txt` as the newly approved state of the API

### DIFF
--- a/tests/FakeItEasy.Tests.Approval/ApiApproval.ApproveApi.approved.txt
+++ b/tests/FakeItEasy.Tests.Approval/ApiApproval.ApproveApi.approved.txt
@@ -1,0 +1,600 @@
+﻿[assembly: System.CLSCompliantAttribute(true)]
+[assembly: System.Resources.NeutralResourcesLanguageAttribute("en-US")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute(@"DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
+[assembly: System.Runtime.CompilerServices.InternalsVisibleToAttribute(@"FakeItEasy.IntegrationTests, PublicKey=00240000048000009400000006020000002400005253413100040000010001002f50d82092713be482c885861c984334606da0f54c78738a3dd0862fb2dfc6080f0780132cc65d88f0f0c70af74e8a53430962395bfc1a36fab08b7a2549d387e805c13cc84acd884447ec8c4dcfb6216df720f0998380f9c906b5de8141798d64661f036d47274e6ecb76c9cde5f4cf2b521040601e44b3914fbeb9f39127f9")]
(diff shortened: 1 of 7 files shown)


tests/FakeItEasy.Tests.Approval/ApiApproval.ApproveApi.approved.txt
```diff
@@ -0,0 +1,600 @@
...
+        public static void SetStaticField(this System.Type type, string fieldName, System.Reflection.BindingFlags additionalFlags, object value) { }
+        public static System.Reflection.MemberInfo[] Sort(System.Reflection.MemberInfo[] members) { }
+    }
+}
+namespace FakeItEasy
+{
+    
+    public class static A
+    {
+        [FakeItEasy.Analysis.MustUseReturnValueAttribute("UnusedCallSpecification")]
```
thomaslevesque: Is this really part of the public API? That's weird, the attribute is internal... based on this, there *will* be a public API change when we remove the attribute.

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
```diff
@@ -0,0 +1,19 @@
...
+    using ApprovalTests;
+    using ApprovalTests.Reporters;
+    using NUnit.Framework;
+    using PublicApiGenerator;
+
+    public class ApiApproval
+    {
+        [Test]
+        [UseReporter(typeof(DiffReporter))]
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
thomaslevesque: Any reason for this? Did you run into trouble because of inlining?

thomaslevesque:
Thanks @adamralph!
LGTM, just two minor remarks.

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
thomaslevesque: Ah, I see. It's because the name of the method is used in the `.approved.txt` file...

adamralph: Yes, that's right. At first, it didn't work, but I found the answer somewhere, IIRC some SO question.

tests/FakeItEasy.Tests.Approval/ApiApproval.ApproveApi.approved.txt
adamralph: Indeed. I wonder if this is a bug. @JakeGinnivan any opinion?

adamralph: @xxxxxxxxxxxxxx to clarify, this PR introduces the use of `PublicApiGenerator`, which generates this output.

`FakeItEasy.Analysis.MustUseReturnValueAttribute` is an `internal` type, so we're surprised that `PublicApiGenerator` thinks it is part of the public API.

adamralph:
@thomaslevesque the internal attribute concern is valid, but I'm not sure it should hold up this PR. 

After all, there is always the possibility of bugs being present in PublicApiGenerator but the thing I like about this workflow is that it takes a human to eyeball the changes and, ultimately, approve the output and commit it.

If, indeed, this specific case is a bug in PublicApiGenerator, we can always upgrade when it is patched and update the approved API output to remove those attributes.

thomaslevesque:
> @thomaslevesque the internal attribute concern is valid, but I'm not sure it should hold up this PR.

Agreed. I'll merge now.

(I probably won't be able to do much more in the week to come)

PR #705, opened 2016-05-08
update approved API

### LINKED ISSUE #675: Helper method for list of dummies

adamralph:
approve API change from #702

closes #675


blairconrad:
thanks!

blairconrad:
Hmm. Trying to merge, but the button does nothing in browser on my phone. :(

thomaslevesque:
:+1:

PR #710, opened 2016-05-14
docs: added CollectionOfDummy to dummies.md

### LINKED ISSUE #675: Helper method for list of dummies

adamralph:
closes #675

### DIFF
--- a/docs/dummies.md
+++ b/docs/dummies.md
@@ -34,6 +34,14 @@ library.Checkout(MakeExpiredCard(), A.Dummy<Book>());
 This signals that the actual value of the `Book` is really not
 important. The code is intention-revealing.
 
+You can also create a collection of dummies by writing:
+
+```csharp
+A.CollectionOfDummy<Book>(10);
+```
+
+This will return an `IList` containing 10 dummy `Book` instances.
+
 ## How FakeItEasy uses them
 
 When [creating Fakes](creating-fakes.md) or Dummies of class types,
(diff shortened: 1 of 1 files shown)


docs/dummies.md
```diff
@@ -63,7 +71,7 @@ of approaches in turn, until one succeeds:
   Fake `T`.
 1. If `T` is a value type, the Dummy will be a `T` created via
   `Activator.CreateInstance`.
-1. If nothing above matched, then `T` is a class. Loop over all its constructors in _descending order of argument list length_.  
+1. If nothing above matched, then `T` is a class. Loop over all its constructors in _descending order of argument list length_.
```
adamralph: EditorConfig :wink:

thomaslevesque:
LGTM :+1: 

Unrelated to this PR, but shouldn't the method be named `CollectionOfDummies` instead of `CollectionOfDummy` (and similarly, `CollectionOfFakes` instead of `CollectionOfFake`)? Sure, you could make a collection with only one dummy or fake, but what's the point?

adamralph:
@thomaslevesque I believe the intention behind the singular "fake" in `A.CollectionFake<T>()` is so that, e.g.

```C#
A.CollectionOfFake<Widget>();
```

reads fluently, i.e. "a collection of fake widget". (Of course, it would be better if "widget" was pluralised.)

This tends to read better than

```C#
A.CollectionOfFakes<Widget>();
```

i.e. "a collection of fakes widget".

`A.CollectionOfDummy<T>()` followed this precedent.

thomaslevesque:
You're right, it does read more fluently with the type argument.

PR #713, opened 2016-05-19
Fix code analysis violations

thomaslevesque:


tests/FakeItEasy.Specs/ConfiguringPropertySetterSpecs.cs
```diff
@@ -54,6 +55,7 @@ public interface IHaveInterestingProperties
                 .x(() => exception.Should().BeAnExceptionOfType<FakeConfigurationException>());
         }
 
+        [SuppressMessage("Microsoft.Globalization", "CA1305:SpecifyIFormatProvider", MessageId = "System.String.Format(System.String,System.Object)", Justification = "Not relevant in this case")]
```
thomaslevesque: Alternatively, I could use `FormattableString.Invariant`. Since the project doesn't target .NET 4.6, it would require adding the [StringInterpolationBridge](https://www.nuget.org/packages/StringInterpolationBridge) package.

tests/FakeItEasy.Specs/ConfiguringPropertySetterSpecs.cs
```diff
@@ -110,6 +112,7 @@ public interface IHaveInterestingProperties
                 .x(() => exception.Message.Should().Be("The specified expression is not a method call or property getter."));
         }
 
+        [SuppressMessage("Microsoft.Globalization", "CA1305:SpecifyIFormatProvider", MessageId = "System.String.Format(System.String,System.Object)", Justification = "Not relevant in this case")]
```
thomaslevesque: In this case, using `FormattableString.Invariant` is also possible, but would be a little awkward, because the interpolated string is a step description, so it would read `Invariant($"When I assign the property to {value})"` (assuming `FormattableString` is imported with `using static`).

adamralph: I'm wondering if we shouldn't switch off CA1305 for the test projects?

thomaslevesque: I guess it would make sense. Unless anyone has a good reason to want to keep it enabled, I'll switch it off.

blairconrad: I thought we had, to tell the truth.

thomaslevesque: I disabled CA1305 and amended the commit

adamralph:
https://twitter.com/adamralph/status/733572470440001537

adamralph:
Thanks @thomaslevesque !

PR #717, opened 2016-05-21
refactor: fix CA violations

adamralph:
connects to #336


blairconrad:
For my money, the `Concat` is about as readable as the original (and `+` would've been acceptable as well). Thanks for the change.

PR #721, opened 2016-05-22
update to nuget.exe 3.4.3

adamralph:
To allow building on machines with MSBuild 14 and 15 installed.


thomaslevesque:
Thanks @adamralph!

Issue #436, opened 2015-01-13
Adding additional attributes to a fake of an already faked type fails silently


cmerat:
In this example:

``` csharp
var constructor = typeof(BarAttribute).GetConstructor(new[] { typeof(string) });

var fake1 = A.Fake<Foo>(x => x.WithAdditionalAttributes(new[] { new CustomAttributeBuilder(constructor, new object[] { "1" }) }));
var fake2 = A.Fake<Foo>(x => x.WithAdditionalAttributes(new[] { new CustomAttributeBuilder(constructor, new object[] { "2" }) }));
var fake3 = A.Fake<Foo>(x => x.WithAdditionalAttributes(new[] { new CustomAttributeBuilder(constructor, new object[] { "3" }) }));

Assert.AreEqual("3", fake3.GetType().GetCustomAttributes(typeof(BarAttribute), true).Cast<BarAttribute>().First().Value);
```

This test will fail because the attribute with the value of "1" is returned although I'd expect it to return the attribute with value "3" (since I'm checking fake3). This appears to be due to all fakes inheriting from the same type (`Castle.Proxies.FooProxy, DynamicProxyGenAssembly2, DynamicProxyGenAssembly2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=a621a9e7e5c32e69`).

The solution could be that when custom attributes are specified, another type name is created (FooProxy1, FooProxy2 and so on). I am not sure how easy this is to do using DynamicProxy or if it is a limitation but the current behavior feels like a bug.

adamralph:
@cmerat can you confirm how many instances of `BarAttribute` each fake has after the first 3 lines are run? Do they all have all three instances? This is what I would expect, since all three fakes are of the same type, and any custom attributes are being applied to that type.

As you suggest, the only way of having different attributes for the type of each fake would be for each fake it be of a different type (it's nothing to do with inheritance, it's just the type of the fake). I'm also unsure how feasible this is.

cmerat:
@adamralph no, only a single value is available (I've tried with both AllowMultiple set to true and false -- no difference). It appears that the first Fake to "create" the proxy type gets the attribute assignments and the type then lives for the entire duration of the AppDomain (which means it also impacts other tests).

adamralph:
[...]

PR #729, opened 2016-05-26
upgrade to Castle.Core 3.3.3

adamralph:


thomaslevesque:
LGTM, thanks @adamralph!

adamralph:
First feature ready to go for 2.1.0!

blairconrad:
Do we have an actual issue reported against the lack of 3.3.0?
Or expectation of user-noticeable change in behaviour?

blairconrad:
I'm all for the upgrade, but from Core's [changes.txt](https://github.com/castleproject/Core/blob/5d7e3fce57fd6c76227996e942313be2498063a2/Changes.txt):

```
3.3.3
==================
- fixed #70 - Serilog integration modifies LoggerConfiguration.MinimumLevel
- fixed #69 - Added SourceContext to the Serilog Logger - contributed by @KevivL

3.3.2
==================
- fixed #66 - SerilogLogger implementation bug where exceptions were passed through incorrectly

3.3.1
==================
- implemented #61 - Added support for Serilog - contributed by Russell J Baker (@ruba1987)
```

I'm not sure this is a "feature"…

adamralph:
> Do we have an actual issue reported against the lack of 3.3.0?
> Or expectation of user-noticeable change in behaviour?

Nope, I just saw the upgrade available so went ahead and did it. We've done the same for Castle.Core in the past. Of course, if there were some user issue dependent on it then we'd now put that in the 2.1.0 milestone too.

You're right about the lack of a "feature". In fact, I think 3.3.1, 3.3.2 and 3.3.3 were simply lock-step releases with https://www.nuget.org/packages/Castle.Core-Serilog/ - no changes to the Castle.Core package at all.

I'd be happy to label this as a refactoring or something else. In fact, given this new info, I probably wouldn't have raised the PR at all, but hey ho, it's in now.

I guess it does at least remove some noise from the NuGet "updates" tab, but that's already fairly polluted now with all the code analysis packages which we can't upgrade without forcing our analyzer users on to VS 2015 Update 2. :cry:

blairconrad:
>  if there were some user issue dependent on it then we'd now put that in the 2.1.0 milestone too

I think the only thing we have is #436, which is waiting on a Castle.Core release with castleproject/Core#78 in it. I think the 4.0.0-alpha001 release has this, but I'm not tempted…

adamralph:
No, I think we'll hold off on that one.. :smile:

adamralph:
3.3.0 -> 3.3.3 looks like there's no change to the Castle.Core package itself so the only thing it achieves is to remove one item from the NuGet UI update tab. :feelsgood:

PR #732, opened 2016-05-29
avoid argument null exception in UnusedReturnValueAnalyzer

### LINKED ISSUE #731: AD0001	Analyzer 'FakeItEasy.Analyzer.UnusedReturnValueAnalyzer' threw an exception
I was trying out https://www.nuget.org/packages/FakeItEasy.Analyzer/
and got a bunch of warnings, all in the same style.
The `SyntaxNode` exists, but isn't called in any way by FakeItEasy (`A.`...) code.

Don't know if you guys can make something of this, if not then close as I wont be spending time analyzing this further.

```
Severity	Code	Description	Project	File	Line	Source	Suppression State	Detail Description
Warning	AD0001	Analyzer 'FakeItEasy.Analyzer.UnusedReturnValueAnalyzer' threw an exception of type 'System.ArgumentNullException' with message 'Value cannot be null.
Parameter name: syntax'.	<Project>		1	IntelliSense	Active	Analyzer 'FakeItEasy.Analyzer.UnusedReturnValueAnalyzer' threw the following exception:
'Exception occurred with following context:
Compilation: <Project>
SyntaxTree: C:\Source\<Project>\AnalyzerTests.cs
SyntaxNode: CreateId() [InvocationExpressionSyntax]@[4016..4026) (118,46)-(118,56)

System.ArgumentNullException: Value cannot be null.
Parameter name: syntax
   at Microsoft.CodeAnalysis.CSharp.CSharpSemanticModel.CheckSyntaxNode(CSharpSyntaxNode syntax)
   at Microsoft.CodeAnalysis.CSharp.CSharpSemanticModel.GetSymbolInfo(ExpressionSyntax expression, CancellationToken cancellationToken)
   at Microsoft.CodeAnalysis.CSharp.CSharpExtensions.GetSymbolInfo(SemanticModel semanticModel, ExpressionSyntax expression, CancellationToken cancellationToken)
   at FakeItEasy.Analyzer.UnusedReturnValueAnalyzer.GetCalledMethodSymbol(InvocationExpressionSyntax call, SyntaxNodeAnalysisContext context)
   at FakeItEasy.Analyzer.UnusedReturnValueAnalyzer.AnalyzeCall(SyntaxNodeAnalysisContext context)
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c__DisplayClass42_1`1.<ExecuteSyntaxNodeAction>b__1()
   at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock(DiagnosticAnalyzer analyzer, Action analyze, Nullable`1 info)
-----
'.
```

adamralph:
fixes #731

### DIFF
--- a/src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
+++ b/src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
@@ -58,7 +58,9 @@ namespace FakeItEasy.Analyzer
         private static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax call, SyntaxNodeAnalysisContext context)
         {
             var memberAccess = call?.Expression as MemberAccessExpressionSyntax;
-            return context.SemanticModel.GetSymbolInfo(memberAccess?.Name).Symbol as IMethodSymbol;
+            return memberAccess == null
+                ? null
+                : context.SemanticModel.GetSymbolInfo(memberAccess.Name).Symbol as IMethodSymbol;
         }
 
         private static ImmutableDictionary<string, DiagnosticDescriptor> CreateDiagnosticsMap()


blairconrad:
@adamralph, no test?

adamralph:
@blairconrad the first thing I did was to try and add a test which reproduces the problem by taking one of the current tests and putting the offending code into it, as identified at https://github.com/FakeItEasy/FakeItEasy/issues/731#issuecomment-222370130

However, the "test framework" produced by the analyzer template didn't give me any errors back so I guess it's simply ignoring the AD0001 warning. Side note: this made me realise we are now carrying a load of new code which I don't understand in the slightest (the auto-generated test framework) and I wonder if any of the other @FakeItEasy/owners understand it?

I then considered crafting a test which invokes the analyzer directly, reproducing the conditions to force the code path where `call.Expression` is not of type `MemberAccessExpressionSyntax` but I didn't consider it for very long.

If you generalised this, you'd end up writing tests for our entire code base reproducing every possible permutation of input arguments for every function, with appropriate `null` values, type variation, returning of `null` from method calls, `null` properties, etc.

Ultimately, we were misusing the Elvis operator here. If `call.Expression` is not of type `MemberAccessExpressionSyntax` then we clearly don't want to proceed with `context.SemanticModel.GetSymbolInfo()`, passing `null` as the argument. Instead, the code should branch and return `null`, which is what my fix does. I believe our static analysis of the code is enough here, and really should have been caught when we reviewed the original PR.

src/CommonAssemblyInfo.cs
```diff
@@ -3,9 +3,9 @@
 using System.Resources;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyVersion("2.1.0")]
-[assembly: AssemblyFileVersion("2.1.0")]
-[assembly: AssemblyInformationalVersion("2.1.0")]
+[assembly: AssemblyVersion("2.0.1")]
+[assembly: AssemblyFileVersion("2.0.1")]
+[assembly: AssemblyInformationalVersion("2.0.1")]
```
blairconrad: @adamralph, I'm not so sure about changing the version number here. In general, I'd be inclined to do it in a separate PR (I know, the commit is separate, and that's good, but it still feels a little weird to me. Any reason you didn't use `rake set_version[new_version]` to set the version? Not convenient enough?)

(While we're on the subject, I cannot find the 2.0.1 release. Am I blind? Stupid?
I can find the milestone and release checklist issue. Did `rake create_milestone` you?)

All this may be academic, as I've merged #736, so we have a new feature - Source stepping via GitLink, and we're probably better off releasing 2.1.0 as you suggested.

blairconrad:
I'm content (also sad, but for a different reason) knowing that you tried the test and it didn't work.
I definitely don't understand the test framework to the level required to know why you test fell short. Perhaps @thomaslevesque does.

I'm very happy with the fix you have.

But, I'd be happier if I  understood it. For my education, why is `call.Expression` not a `MemberAccessExpressionSyntax`? Because we're talking about member access from within the class and we have no `thing.` or `this.` at the front? Where do people learn these things?

src/CommonAssemblyInfo.cs
adamralph: Yep, I'll back out this change now that #736 is merged.

TBH I don't think I've ever used `set_version[new_version]` :grin: 

The 2.0.1 release was here #733 - now withdrawn.

adamralph:
@blairconrad I've removed the version change commit.

adamralph:
@blairconrad I'm not sure I understand the expression scenario any more than you. :wink:

src/CommonAssemblyInfo.cs
blairconrad: #733 is the "release checklist issue", not a release. I was looking for something like https://github.com/FakeItEasy/FakeItEasy/releases/tag/untagged-501bff0d4409887cd5ef, only more "2.0.1"-like.

> TBH I don't think I've ever used `set_version[new_version]`

Well, you've not actually initiated many releases since, well, since I showed up. Certainly not since the rake target existed.

Does this also mean that you didn't use `rake create_milestone` to create the milestone and release issue checklist? I'd feel better if I knew you didn't, and the lack of GitHub Release was because you didn't make one, not because the rake target failed without alerting you.

blairconrad:
Changes look good to me now. Merging. If @thomaslevesque shows up and says he understands more about the difficulty with the test and sees a way forward, we can address that as a tech-debt issue or similar.

Nice job, @adamralph.

thomaslevesque:
> However, the "test framework" produced by the analyzer template didn't give me any errors back so I guess it's simply ignoring the AD0001 warning.

That worries me a little... I'll look into it when I can.

> I wonder if any of the other @FakeItEasy/owners understand it?

Not every bit, but it's mostly boilerplate code. As far as I can tell, it just creates a compilation from the code you provide, and runs the analyzer on it.

> But, I'd be happier if I understood it. For my education, why is call.Expression not a MemberAccessExpressionSyntax?

I was wondering about that too. This PR fixes the exception, but I think it might miss some cases. E.g. if you have something like this:

```
using static FakeItEasy.A;

...

CallTo(() => foo.Bar());
```

Is `CallTo(() => foo.Bar())` a `MemberAccessExpressionSyntax`? If not, we won't emit a warning... Of course, it's a corner case, and most people probably won't use FIE that way, but who knows? I'll investigate.

thomaslevesque:
> Is `CallTo(() => foo.Bar())` a `MemberAccessExpressionSyntax`? If not, we won't emit a warning... Of course, it's a corner case, and most people probably won't use FIE that way, but who knows? I'll investigate.

As I suspected, we don't support that case; `CallTo(() => foo.Bar())` is an `IdentifierNameSyntax`. I'll send a PR to fix it.

PR #734, opened 2016-05-29
rename UnusedCallSpecificationAnalyzerTests to UnusedReturnValueAnalyzerTests

adamralph:


blairconrad:
I think this makes sense, but just noticed that our diagnostic messages include the phrase "Unused call specification". They read well enough, so maybe best to leave that alone?

blairconrad:
If nothing else, the change in message could be another issue even if we wanted to pursue it.

blairconrad:
Good eye, @adamralph.

thomaslevesque:
Actually, the name `UnusedCallSpecificationAnalyzerTests ` was intentional. `UnusedReturnValueAnalyzer` is designed to diagnose several issues related to unused return values; "unused call specification" is just one of those issues. Since the tests only covers the "unused call specification" case, I had named it `UnusedCallSpecificationAnalyzerTests `.

blairconrad:
OH! That makes perfect sense. I am chagrined.

thomaslevesque:
It doesn't really matter anyway... we can always split it later if we extend `UnusedReturnValueAnalyzer` to check for other issues.

adamralph:
For me these are more like black box tests (acceptance tests/specs) so ultimately they should be named by feature than by implementation. Given that our "unit" tests for FIE, which are named by implementation, are named FakeItEasy.Tests, it probably follows that FakeItEasy.Analyzer.Tests should also be named by implementation. However, I think we need to overhaul our test project naming generally, but we can address that in https://github.com/FakeItEasy/FakeItEasy/issues/280 (which needs updating).

PR #736, opened 2016-05-29
add source stepping via GitLink

### LINKED ISSUE #735: Source stepping via GitLink
*from https://github.com/FakeItEasy/FakeItEasy/issues/362#issuecomment-207228506*

Replaces #362.

adamralph:
closes #735

### DIFF
--- a/docs/source-stepping.md
+++ b/docs/source-stepping.md
@@ -1,0 +1,9 @@
+# Source Stepping
+
+In most cases, you'll never need to step through the FakeItEasy source code, but occasionally, it may help to see what FakeItEasy is doing under the hood. FakeItEasy uses [GitLink](https://github.com/GitTools/GitLink) to make it a breeze to step into the FakeItEasy source code.
+
+To enable GitLink to work in Visual Studio:
+
+1. Tools → Options → Debugging → General → Enable Just My Code: **Uncheck**
+1. Tools → Options → Debugging → General → Enable source server support: **Check**
+1. Tools → Options → Debugging → Symbols → Cache symbols in this directory: **Path to a suitable folder**
--- a/rakefile.rb
+++ b/rakefile.rb
@@ -5,19 +5,23 @@ if !File.file?(msbuild_command)
   raise "MSBuild not found"
 end
 
-nuget_command  = ".nuget/nuget.exe"
-nunit_command  = "packages/NUnit.Runners.2.6.3/tools/nunit-console.exe"
-xunit_command = "packages/xunit.runner.console.2.0.0/tools/xunit.console.exe"
+nuget_command   = ".nuget/nuget.exe"
+gitlink_command = "packages/gitlink.2.3.0/lib/net45/GitLink.exe"
+nunit_command   = "packages/NUnit.Runners.2.6.3/tools/nunit-console.exe"
+xunit_command   = "packages/xunit.runner.console.2.0.0/tools/xunit.console.exe"
 
 solution        = "FakeItEasy.sln"
 assembly_info   = "src/CommonAssemblyInfo.cs"
 version         = IO.read(assembly_info)[/AssemblyInformationalVersion\("([^"]+)"\)/, 1]
 version_suffix  = ENV["VERSION_SUFFIX"]
+repo_url        = "https://github.com/FakeItEasy/FakeItEasy"
 nuspec          = "src/FakeItEasy/FakeItEasy.nuspec"
 analyzer_nuspec = "src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.nuspec"
 logs            = "artifacts/logs"
 output          = "artifacts/output"
 tests           = "artifacts/tests"
+
+gitlinks        = ["FakeItEasy", "FakeItEasy.Analyzer"]
 
 unit_tests = [
   "tests/FakeItEasy.Tests/bin/Release/FakeItEasy.Tests.dll",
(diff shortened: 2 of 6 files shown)


rakefile.rb
```diff
@@ -5,20 +5,34 @@
...
+  "FakeItEasy.Tests",
+  "FakeItEasy.IntegrationTests",
+  "FakeItEasy.IntegrationTests.VB",
+  "FakeItEasy.Examples.VB",
+  "FakeItEasy.Specs",
+  "FakeItEasy.IntegrationTests.External",
+  "FakeItEasy.Analyzer.Tests",
+  "FakeItEasy.Tests.Approval"
+]
+
```
adamralph: GitLink 2.3.0 introduces `-include`, so when that goes RTM when can upgrade and list the two assemblies we want to include, instead of listing the 9 we want to ignore.

rakefile.rb
```diff
@@ -162,6 +176,12 @@
   run_msbuild solution, "Build", msbuild_command
 end
 
+desc "GitLink PDB's"
+exec :gitlink => [:build] do |cmd|
+  cmd.command = gitlink_command
+  cmd.parameters ". -u #{repo_url} -ignore " + gitlink_ignores.join(",")
```
adamralph: The URL turned out to be the only thing that GitLink coudn't work out for itself, perhaps because I'm using SSH instead of HTTP.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.nuspec
```diff
@@ -16,6 +16,7 @@
   </metadata>
   <files>
     <file src="bin\Release\FakeItEasy.Analyzer.dll" target="analyzers\dotnet\cs" />
+    <file src="bin\Release\FakeItEasy.Analyzer.pdb" target="analyzers\dotnet\cs" />
```
adamralph: It's questionable whether this will be of much use, but I don't think it hurts to include it. I guess there may be a scenario where someone one wants to troubleshoot a problem in the analyzer by invoking it directly and stepping through, e.g. I potentially could have done this when investigating #731.

rakefile.rb
adamralph: BTW - I tried to use [regex](https://github.com/GitTools/GitLink#ignoring-projects-and-explicitly-including-them) here but, for the life of me, I couldn't get it to work, so I resorted to listing the nine projects explicitly. I don't it's worth spending any more time trying to get it to work via regex given that we can switch to `-include` when GitLink 2.3.0 is released.

blairconrad: Intentionally not excluded from `build.cmd vars` output?

docs/source-stepping.md
```diff
@@ -0,0 +1,9 @@
+# Source Stepping
+
+In most cases, you'll never need to step through the FakeItEasy source code, but occasionally, it may help to see what FakeItEasy is doing under the hood. FakeItEasy uses [GitLink](https://github.com/GitTools/GitLink) to make it a breeze to step into the FakeItEasy source code.
+
+To enable GitLink to work in Visual Studio:
+
+1. Tools -> Options -> Debugging -> General -> Enable Just My Code: **Uncheck**
+1. Tools -> Options -> Debugging -> General -> Enable source server support: **Check**
+1. Tools -> Options -> Debugging -> Symbols -> Cash symbols in this directory: **Path to a suitable folder**
```
blairconrad: I'm surprised you're a "->" guy and not a "→" guy.

Also "Cache" instead of "Cash".
Or is there a finder's fee I don't know about?

rakefile.rb
```diff
@@ -5,20 +5,34 @@
...
+nuget_command   = ".nuget/nuget.exe"
+gitlink_command = "packages/gitlink.2.2.0/lib/net45/GitLink.exe"
+nunit_command   = "packages/NUnit.Runners.2.6.3/tools/nunit-console.exe"
+xunit_command   = "packages/xunit.runner.console.2.0.0/tools/xunit.console.exe"
 
 solution        = "FakeItEasy.sln"
 assembly_info   = "src/CommonAssemblyInfo.cs"
 version         = IO.read(assembly_info)[/AssemblyInformationalVersion\("([^"]+)"\)/, 1]
 version_suffix  = ENV["VERSION_SUFFIX"]
+repo_url        = "https://github.com/FakeItEasy/FakeItEasy"
```
blairconrad: We have `repo` downstairs. Is it silly of me to think that we could make these variables friends?

docs/source-stepping.md
adamralph: Will fix.

rakefile.rb
adamralph: Good spot. I was hoping I'd get away with it. :wink: 

I thought about it, but it would mean constructing a URL from the scheme+host and the repo name. Seemed less hassle to duplicate.

rakefile.rb
adamralph: Yep, I didn't see a reason to exclude. We're not excluding things like the various "tests" vectors either.

adamralph: BTW - GitLink 2.3.0 [has just been released](https://github.com/GitTools/GitLink/issues/75#issuecomment-222444188), so I'm going to change this to use `-include` instead.

adamralph:
@blairconrad thanks for the review. All comments addressed/replied to.

rakefile.rb
blairconrad: I was on the fence as well.
I'll allow it, but when we move the repo, _you_ have to make the change in both places.

blairconrad:
Oh! You used the arrows! I'm touched.

thomaslevesque:
:+1:, I'll try it ASAP

Thanks @adamralph!

PR #737, opened 2016-05-29
add build.cmd as a shortcut for bundle exec rake

adamralph:
For convenience, also includes `bundle install`, although I'm in two minds about that. Thoughts?

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,0 +1,3 @@
+@echo off
+
+call bundle exec rake %*
--- a/how_to_build.md
+++ b/how_to_build.md
@@ -41,24 +41,26 @@ At the time of writing the build is only confirmed to work on Windows using the 
 
 Using a command prompt, navigate to your clone root folder and execute:
 
-`bundle exec rake`
+`build.cmd`
 
 This executes the default build tasks. After the build has completed, the build artifacts will be located in `artifacts`.
 
 ## Extras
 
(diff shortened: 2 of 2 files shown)


blairconrad:
> For convenience, also includes bundle install, although I'm in two minds about that. Thoughts?

I'm lazy. How bad is the drag when I don't need to `bundle install`, and what's the behaviour if you don't include and I need to?
If I get everything else wrapped up before you answer, I'll check.

adamralph:
On second thoughts, it's a step too far. I'll dial it back to just wrapping `bundle exec rake` for now, which I think is good in isolation. We can consider adding further candy later.

adamralph:
amended as discussed

blairconrad:
Thanks for the update, @adamralph.

It's nice to have a build.cmd, as we discussed the other day when looking at the `paket` problem, IIRC correctly.

thomaslevesque:
Usually I just run `rake`, so I didn't really feel the need for a shortcut. But maybe it doesn't do exactly what I think?

blairconrad:
I mostly run `rake` too. I'm sure there's a reason to run the `bundle` bit, but I don't know what it is.
I think the big benefit here is that often people will check out a repo and look for the `build.cmd`.
And if you're going to provide one, why not use the "best" version of running rake?

thomaslevesque:
Good point

adamralph:
Running just `rake` doesn't work for me because I have multiple versions of the gems (Albacore, etc.) installed and the latest version (chosen by default) doesn't work with our Rakefile. The build server has the same problem.

PR #739, opened 2016-05-30
New release targets

### LINKED ISSUE #665: streamline production/pre-release handling
The current system, of updating the release notes in Git (or forgetting to) and then copying to NuGet, it tedious and error-prone.
(And the temporary aberration of keeping the "changes in 2.0" in the wiki does not help.)

I think we should do something to reduce the effort.
For example, we could keep the release notes in the source, and they could be updated with PRs, as are the docs.
Then we could see about having the notes automatically copied into the NuGet package when it's packed.
If we couple this with some automated way to attach the NuGet package to the release, we could automatically take a snapshot of the release notes and apply them to the GitHub release at the same time.

A big advantage (to my mind) of making this change is that I think we'd be more likely to update the release notes as PRs are merged, and we'd end up having more eyes on them earlier. Up 'til now, I've been writing most of the release notes, and probably the other owners are taking a peek at them before release, but that's never been explicit in our process.

There are tools to help integrate release notes management into our daily work and then release cycle. Eventually we may want to explore them.

Note that there are some risks with putting the release notes in source, not dissimilar to those of putting the documentation in source.
For example, if a grievous error is found in the release notes, what do we do?

Oh! We could edit the NuGet package and GitHub release notes by hand, and submit a PR to the code. Not great, and the old tagged version of the source would have the wrong info, but that's probably not a huge concern.

Thoughts?

blairconrad:
Connects to #665.

### DIFF
--- a/rakefile.rb
+++ b/rakefile.rb
@@ -38,31 +38,42 @@ specs = "tests/FakeItEasy.Specs/bin/Release/FakeItEasy.Specs.dll"
 approval_tests = "tests/FakeItEasy.Tests.Approval/bin/Release/FakeItEasy.Tests.Approval.dll"
 
 repo = 'FakeItEasy/FakeItEasy'
+ci_server_url = 'http://teamcity.codebetter.com/admin/editBuildParams.html?id=buildType:bt929'
 release_issue_labels = ['P2', 'build', 'documentation']
-release_issue_body = <<-eos
-**Ready** when all other issues on this milestone are **Done** and closed.
+release_issue_common_steps = <<-eos
+**in-progress** when all other issues on this milestone are closed.
 
 - [ ] run code analysis in VS in *Release* mode and address violations (send a regular PR which must be merged before continuing)
-- [ ] if necessary, change `VERSION_SUFFIX` on [CI Server](http://teamcity.codebetter.com/admin/editBuildParams.html?id=buildType:bt929)
+- [ ] if necessary, change `VERSION_SUFFIX` on the [CI Server](#{ci_server_url})
       to appropriate "-beta123" or "" (for non-betas) value and initiate a build
 - [ ] check build
--  edit draft release in [GitHub UI](https://github.com/FakeItEasy/FakeItEasy/releases):
+- edit draft release in [GitHub UI](https://github.com/FakeItEasy/FakeItEasy/releases):
     - [ ] complete release notes, mentioning non-owner contributors, if any (move release notes forward from any pre-releases to the current release)
-    - [ ] attach nupkg(s) - main package and/or analyzer, whichever have new content
+    - [ ] attach main nupkg and/or analyzer nupkg, whichever have content to release
     - [ ] publish the release
-- [ ] push NuGet package
+- [ ] push nupkg(s) to NuGet
 - [ ] de-list pre-release or superseded buggy NuGet packages if present
 - [ ] update website with contributors list (if in place)
 - [ ] tweet, mentioning contributors and post link as comment here for easy retweeting ;-)
 - [ ] post tweet in [Gitter](https://gitter.im/FakeItEasy/FakeItEasy)
-- [ ] post links to the NuGet and GitHub release in each issue in this milestone, with thanks to contributors
+- [ ] post links to the GitHub Release in each issue in this milestone, with thanks to contributors
+eos
+
+next_pre_release_steps = <<-eos
+- if there's to be a new pre-release issue
+    - [ ] run `rake pre_release[version_suffix]` to
+        - create a new draft GitHub Release
+        - new issue (like this one) for the next release, adding it to the current milestone
+    - [ ] change `VERSION_SUFFIX` on the [CI Server](#{ci_server_url})
+eos
+
+next_version_steps = <<-eos
 - [ ] run `rake next_version[new_version]` to
     - create a pull request that changes the version in CommonAssemblyInfo.cs to the expected version (of form _xx.yy.zz_)
+    - create a new draft GitHub Release
     - create a new milestone for the next release
     - create a new issue (like this one) for the next release, adding it to the new milestone
-    - create a new draft GitHub Release
 - [ ] close this milestone
-
 eos
 
 release_body = <<-eos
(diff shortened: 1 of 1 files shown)


blairconrad:
Will probably need a rebase after all the other work that @adamralph did today, but figured I'd get it up there for eyes to see.

rakefile.rb
```diff
@@ -34,32 +34,42 @@
...
+        - new issue (like this one) for the next release, adding it to the current milestone
+    - [ ] change `VERSION_SUFFIX` on the [CI Server](#{ci_server_url})
+eos
+
+next_version_steps = <<-eos
+- [ ] run `rake next_version[new_version]` to
+    - create a pull request that changes the version in CommonAssemblyInfo.cs to the expected version (of form _xx.yy.zz_)
+    - create a new draft GitHub Release
     - create a new milestone for the next release
     - create a new issue (like this one) for the next release, adding it to the new milestone
```
blairconrad: Not super wild about the "next_*_steps" vars, but it avoided duplicating most of the steps and works for now.
If you prefer another approach, I'm open.

rakefile.rb
```diff
@@ -38,32 +38,42 @@
 approval_tests = "tests/FakeItEasy.Tests.Approval/bin/Release/FakeItEasy.Tests.Approval.dll"
 
 repo = 'FakeItEasy/FakeItEasy'
+ci_server_url = 'http://teamcity.codebetter.com/admin/editBuildParams.html?id=buildType:bt929'
```
adamralph: Whilst we're here, shall we press this button

image

to get rid of the opaque number and change it to

image

?

rakefile.rb
```diff
@@ -103,15 +113,17 @@
   run_msbuild solution, "Clean", msbuild_command
 end
 
-desc "Update version number"
-task :set_version, :new_version do |asm, args|
+desc "Update version number and create milestone, release, and release checklist issue"
```
adamralph: > Update version number and _create pull request,_ milestone, release, and release checklist issue

?

rakefile.rb
```diff
@@ -38,32 +38,42 @@
...
+- [ ] post links to the GitHub Release in each issue in this milestone, with thanks to contributors
+eos
+
+next_pre_release_steps = <<-eos
+- if there's to be a new pre-release issue
+    - [ ] run `rake pre_release[version_suffix]` to
+        - create a new draft GitHub Release
+        - new issue (like this one) for the next release, adding it to the current milestone
+    - [ ] change `VERSION_SUFFIX` on the [CI Server](#{ci_server_url})
+eos
```
adamralph: I don't think we need to include this in the pre-release issue. We usually don't know if there's going to be another pre-release until we decide to do one, so I think this will usually be redundant noise.

rakefile.rb
```diff
@@ -38,32 +38,42 @@
 approval_tests = "tests/FakeItEasy.Tests.Approval/bin/Release/FakeItEasy.Tests.Approval.dll"
 
 repo = 'FakeItEasy/FakeItEasy'
+ci_server_url = 'http://teamcity.codebetter.com/admin/editBuildParams.html?id=buildType:bt929'
 release_issue_labels = ['P2', 'build', 'documentation']
-release_issue_body = <<-eos
-**Ready** when all other issues on this milestone are **Done** and closed.
+release_issue_common_steps = <<-eos
+**in-progress** when all other issues on this milestone are closed.
```
adamralph: Not sure I agree. It's **ready** when the other issues are closed, but it won't be **in-progress** until someone picks up the task and does it.

rakefile.rb
blairconrad: Oh? Of the 11 betas and 2 RC for 2.0.0, we knew there was going to be another pre-release 12 times.

blairconrad: Or I thought we did, unless we were always planning going from beta to 2.0.0 and I just didn't know?

rakefile.rb
blairconrad: Sounds good to me. I don't think anbody cares about the current value of the number. But it's catchy!

rakefile.rb
blairconrad: But there has to be (I think) a note about handling new pre-releases in the owner's guide anyhow, for the situations where we didn't plan to have pre-releases, so we _could_ drop the whole block, if you prefer.

adamralph:
@blairconrad I did a pass over the changes and left some minor comments. Overall LGTM.

rakefile.rb
adamralph: Perhaps best done in a later PR, since we'd also have to change the badge in README.md.

rakefile.rb
blairconrad: Ah. I thought the old  **Ready** referred to the label, given its boldness. I can support non-bold "Ready".

rakefile.rb
adamralph: Yep, I think given that we need to have that block separately anyway, in the wiki, then probably best not to repeat it here.

rakefile.rb
adamralph: How about:

> Can be labelled **ready** when all other issues on this milestone are closed.

blairconrad: For some reason I thought we'd replaced **ready** with **in-progress**, when really **ready** survives and I think **working** or something had been renamed. I've been out to lunch this whole time. I would've left the original had I been on the ball. :blush:

I much prefer your new version, though.

rakefile.rb
```diff
@@ -38,32 +38,34 @@
...
+- [ ] push nupkg(s) to NuGet
 - [ ] de-list pre-release or superseded buggy NuGet packages if present
 - [ ] update website with contributors list (if in place)
 - [ ] tweet, mentioning contributors and post link as comment here for easy retweeting ;-)
 - [ ] post tweet in [Gitter](https://gitter.im/FakeItEasy/FakeItEasy)
-- [ ] post links to the NuGet and GitHub release in each issue in this milestone, with thanks to contributors
-- [ ] run `rake set_version[new_version]` to create a pull request that changes the version in
-       CommonAssemblyInfo.cs to the expected version (of form _xx.yy.zz_)
-- [ ] run `rake create_milestone` (whilst on the branch containing the version update) to:
+- [ ] post a link to the GitHub Release in each issue in this milestone, with thanks to contributors
```
blairconrad: I added a bonus change. Just one link, because there's just one GitHub Release.

blairconrad:
I think I've addressed all comments. Please hoot me if there's anything else I can do. Or if you prefer squashes or something.
Thanks for looking!

adamralph:
Thanks @blairconrad!

I *almost* asked for a squash. Perhaps I should have, not sure.

blairconrad:
Thanks for the comments and merge. About the squash, the important thing is that you're tortured over it.

PR #741, opened 2016-05-30
remove nuget downloads badge

adamralph:
shields.io no longer supports it.

We could switch to https://buildstats.info/ but then we'd have to abandon the square badge style and go with rounded corners for all:

[NuGet Badge](https://www.nuget.org/packages/FakeItEasy/)


adamralph:
Any hey presto, as soon as I write that comment, https://buildstats.info/ goes down. Perhaps we won't use it :wink:

adamralph:
Ahhhhh, so nice to have that broken badge gone from our landing page. It was playing havoc with my OCD.

blairconrad:
I'm surprised it wasn't below the fold with those landscape monitors you have there.

PR #742, opened 2016-05-30
Support `using static FakeItEasy.A` in unused call specifications

thomaslevesque:
e.g. `Foo()` instead of `this.Foo()`, or `CallTo(...)` instead of `A.CallTo(...)` (with `using static FakeItEasy.A`)

as discussed in #732

### DIFF
--- a/src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
+++ b/src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
@@ -57,10 +57,12 @@ namespace FakeItEasy.Analyzer
 
         private static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax call, SyntaxNodeAnalysisContext context)
         {
-            var memberAccess = call?.Expression as MemberAccessExpressionSyntax;
-            return memberAccess == null
+            var name = (call?.Expression as MemberAccessExpressionSyntax)?.Name
+                ?? call?.Expression as IdentifierNameSyntax;
+
+            return name == null
                 ? null
-                : context.SemanticModel.GetSymbolInfo(memberAccess.Name).Symbol as IMethodSymbol;
+                : context.SemanticModel.GetSymbolInfo(name).Symbol as IMethodSymbol;
         }
 
         private static ImmutableDictionary<string, DiagnosticDescriptor> CreateDiagnosticsMap()
--- a/tests/FakeItEasy.Analyzer.Tests/UnusedReturnValueAnalyzerTests.cs
+++ b/tests/FakeItEasy.Analyzer.Tests/UnusedReturnValueAnalyzerTests.cs
@@ -110,6 +110,39 @@ namespace TheNamespace
         [Test]
         [SetUICulture("en-US")] // so that the message is in the expected language regardless of the OS language
         [SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "CallTo", Justification = "It's an identifier")]
+        public void Diagnostic_Should_Be_Triggered_When_Call_Specification_Made_In_Global_Scope_Is_Not_Used()
+        {
+            var test = @"using static FakeItEasy.A;
+namespace TheNamespace
+{
+    class TheClass
+    {
+        void Test()
+        {
+            var foo = Fake<IFoo>();
+            CallTo(() => foo.Bar());
+        }
+    }
+
+    interface IFoo { int Bar(); }
+}
+";
+
+            this.VerifyCSharpDiagnostic(
+                test,
+                new DiagnosticResult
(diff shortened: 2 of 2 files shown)


tests/FakeItEasy.Analyzer.Tests/UnusedReturnValueAnalyzerTests.cs
```diff
@@ -110,6 +110,39 @@ interface IFoo { int Bar(); }
         [Test]
         [SetUICulture("en-US")] // so that the message is in the expected language regardless of the OS language
         [SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "CallTo", Justification = "It's an identifier")]
+        public void Diagnostic_Should_Be_Triggered_When_Call_Specification_Is_Not_Used_With_UsingStatic()
```
adamralph: I think the "Not" shouldn't be there:

> Diagnostic_Should_Be_Triggered_When_Call_Specification_*Is*_Used_With_UsingStatic()

?

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
```diff
@@ -57,10 +57,12 @@ private static void AnalyzeCall(SyntaxNodeAnalysisContext context)
 
         private static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax call, SyntaxNodeAnalysisContext context)
         {
-            var memberAccess = call?.Expression as MemberAccessExpressionSyntax;
-            return memberAccess == null
+            var name = (call?.Expression as MemberAccessExpressionSyntax)?.Name
+                ?? call?.Expression as IdentifierNameSyntax;
```
adamralph: How does this work? Is there an implicit cast operator from `IdentifierNameSyntax` to `string`?

adamralph:
@thomaslevesque thanks! I did a pass over the changes. LGTM. One small concern over the test name. The other question is just for my education.

tests/FakeItEasy.Analyzer.Tests/UnusedReturnValueAnalyzerTests.cs
thomaslevesque: Actually, the "Not" should be there, but it applies to the usage of the call specification... I copied and modified `Diagnostic_Should_Be_Triggered_When_Call_Specification_Is_Not_Used` and added `_With_UsingStatic` to the name, so the resulting name sounds confusing. I'll try to think of a better name.

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
thomaslevesque: `MemberAccessExpressionSyntax.Name` is a `SimpleNameSyntax`, and `IdentifierNameSyntax` is a subclass of `SimpleNameSyntax`. So `name` is of type `SimpleNameSyntax`.

tests/FakeItEasy.Analyzer.Tests/UnusedReturnValueAnalyzerTests.cs
blairconrad: Something a little more like

```c#
Diagnostic_Should_Be_Triggered_When_Call_Specification_Made_In_Global_Scope_Is_Not_Used
```

I've read docs that say that the "using static" "[renders static methods available in global scope](https://msdn.microsoft.com/en-us/magazine/dn879355.aspx)", which is why I put "global scope" in there. And I think deduplicating "use" in the name helps a little.

I'm not sure the method name is quite right yet, but I _think_ it's a step in the right directin.

blairconrad: `Diagnostic_Should_Be_Triggered_By_Unused_Call_Specification_In_Global_Scope`?

thomaslevesque: :+1: @blairconrad 

I think I'll pick your first suggestion for consistency with the other tests

thomaslevesque:
I renamed the test

blairconrad:
LGTM, @thomaslevesque!

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
adamralph: Aha! For some reason I assumed `MemberAccessExpressionSyntax.Name` was a `string`. Thanks for clearing that up.

blairconrad:
This issue has been fixed in [release 2.1.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/2.1.0).

PR #745, opened 2016-05-31
remove trailing space from crefs and redundant type

adamralph:


src/FakeItEasy/Core/FakeManagerProvider.cs
```diff
@@ -10,7 +10,7 @@ namespace FakeItEasy.Core
     /// <remarks>
     /// Note that we just need to serialize the <see cref="FakeManager"/> + the lock (an "empty", *new* object will be deserialized)
     /// because <see cref="IFakeCallProcessorProvider"/> doesn't require serializability before the first call of <see cref="Fetch"/> or
-    /// <see cref="EnsureInitialized "/> (see remarks section of <see cref="IFakeCallProcessorProvider"/>).
+    /// <see cref="EnsureInitialized"/> (see remarks section of <see cref="IFakeCallProcessorProvider"/>).
```
adamralph: It turns out the Mono compiler is stricter than Roslyn in this case. It picked this out as an error and failed the build.

blairconrad: OH! "cref"!

blairconrad:
Cool!

Issue #482, opened 2015-04-27
Non-generic Fake and Dummy creation


thomaslevesque:
I just realized that `A.Fake` doesn't have a non-generic overload accepting a `Type` parameter, which would be very useful in some scenarios. I realize that it's not something most people would need on a regular basis, but there doesn't seem to be a good workaround. The only way I found was to call `A.Fake<T>()` via reflection, which is far from ideal.

Am I missing something?

Perhaps adding a non-generic overload to `A.Fake` wouldn't be the best option, because would pollute the API with something that is not usually useful, but it would be nice to have an easy workaround.

blairconrad:
Hi, @thomaslevesque. Is there a particular use case you envision? Have you encountered a need for this, or is it just speculation? About a year ago, #288 was raised, which is nearly the same, but for `A.Dummy`. It was ultimately closed "wontfix" after we discussed it with the potential contributor and it turns out that even he considered it a niche feature.

thomaslevesque:
Hi @blairconrad,

Yes, I have a specific use case in mind: http://stackoverflow.com/a/29905525/98713
As you can see I had to use `GetMethod` + `MakeGenericMethod` to achieve what I wanted.

I realize it's not a common use case, but it's not so far-fetched either, and currently there is no clean way to do it, as far as I can see. It's probably you don't really need it when you use FakeItEasy directly, but as soon as you build some tooling around it, it becomes almost indispensable.

thomaslevesque:
Interestingly, it's kind of the opposite of #470, where I wanted to add a generic version of a method that accepted a `Type` parameter ;)
It's common practice to offer both generic and non-generic forms of an API that accepts a type, because depending on the use case, you might need one of the other.

thomaslevesque:
The only issue that I see is that this non-generic overload couldn't accept an `Action<IFakeOptionsBuilder<T>>`, since there would be no `T`... (in my use case, it's an acceptable limitation)

blairconrad:
Ah, yes. Auto-faking. I added some comments in the StackOverflow question.
Generally in these sorts of situations, I'd recommend AutoFixture/AutoFixture, especially with its AutoFakeItEasy extension.
It's _built_ for this sort of thing, and I'm sure @ploeh and contributors have added all kinds of nice features. There are even extensions for other containers, such as [the one for Autofac](http://docs.autofac.org/en/latest/integration/fakeiteasy.html).

[...]

PR #748, opened 2016-06-05
Fix specs after 461

### LINKED ISSUE #582: Clean up specs touched by #574
Rework all the Specs, according to a comment made by @adamralph: https://github.com/FakeItEasy/FakeItEasy/pull/576#discussion_r49584676

- [x] CreationOptionsSpecs.cs
- [x] CreationSpecs.cs
- [x] FakeOptionsBuilderSpecs.cs
- [x] FakeScopeSpecs.cs

In addition to changes inspired by that comment,  @thomaslevesque and I have had some additional conversation in #574, where I ultimately ended up proposing replacing steps like

```c#
"Given a type that has an implicit options builder defined"
    .x(() => { }); // see RobotActivatedEvent

"And the options builder updates the options to configure the fake"
    .x(() => { }); // see DomainEventFakeOptionsBuilder.BuildOptions
```

With 

```c#
"Given a type that has an implicit options builder defined"
    .See<RobotActivatedEvent>();

"And the options builder updates the options to configure the fake"
    .See<DomainEventFakeOptionsBuilder>();
```

Which would be defined by this extension method

```c#
public static class SpecExtensions
{
    public static IStepBuilder See<T>(this string text)
    {
        return text.x(() => { });
    }
}
```

blairconrad:
Fixes #582.

I incorporated the `See` extension method we talked about, except where I had to talk about a protected method. Then I fell back to a textual note.

### DIFF
--- a/tests/FakeItEasy.Specs/CreationOptionsSpecs.cs
+++ b/tests/FakeItEasy.Specs/CreationOptionsSpecs.cs
@@ -6,10 +6,10 @@ namespace FakeItEasy.Specs
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
     using System.Reflection.Emit;
-    using Creation;
+    using FakeItEasy.Creation;
+    using FakeItEasy.Tests;
     using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
-    using Tests;
     using Xbehave;
 
     public static class CreationOptionsSpecs
@@ -725,7 +725,7 @@ namespace FakeItEasy.Specs
                     var customAttributeBuilders1 = new List<CustomAttributeBuilder> { attribute1 };
 
                     var constructor2 = typeof(ExampleAttribute).GetConstructor(new[] { typeof(object[]) });
-                    var attribute2 = new CustomAttributeBuilder(constructor2, new[] { new object[] { 1, null } });
+                    var attribute2 = new CustomAttributeBuilder(constructor2, new object[] { new object[] { 1, null } });
                     var constructor3 = typeof(DebuggerStepThroughAttribute).GetConstructor(new Type[0]);
                     var attribute3 = new CustomAttributeBuilder(constructor3, new object[0]);
                     var customAttributeBuilders2 = new List<CustomAttributeBuilder> { attribute2, attribute3 };
--- a/tests/FakeItEasy.Specs/CreationOptionsSpecs.cs
+++ b/tests/FakeItEasy.Specs/CreationOptionsSpecs.cs
@@ -6,12 +6,12 @@ namespace FakeItEasy.Specs
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
     using System.Reflection.Emit;
+    using Creation;
     using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
     using Tests;
     using Xbehave;
 
-    [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling", Justification = "Required for testing.")]
     public static class CreationOptionsSpecs
     {
         [SuppressMessage("Microsoft.Design", "CA1040:AvoidEmptyInterfaces", Justification = "It's just used for testing.")]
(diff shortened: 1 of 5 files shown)


blairconrad:
At https://github.com/FakeItEasy/FakeItEasy/pull/748/files#diff-1274aa9800e421ca6f1b68f96f370649R388,

ReSharper thought to change 

```c#
if (options != null)
{
    options.Wrapping(null);
}
```

to

```c#
options?.Wrapping(null);
```

But the latter doesn't (yet) jump off the page to me as "_maybe_ do something", so I resisted. Do either of you care?

tests/FakeItEasy.Specs/CreationOptionsSpecs.cs
```diff
@@ -6,12 +6,12 @@ namespace FakeItEasy.Specs
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
     using System.Reflection.Emit;
+    using Creation;
```
adamralph: I see we already have mix here, but I think we usually prefer fully qualified namespaces.

tests/FakeItEasy.Specs/XBehaveExtensions.cs
```diff
@@ -0,0 +1,70 @@
+namespace FakeItEasy.Specs
+{
+    using System;
+    using System.Diagnostics.CodeAnalysis;
+    using Creation;
+    using Xbehave;
+    using Xbehave.Sdk;
+
+    public static class XBehaveExtensions
```
adamralph: I think (correct me if I'm wrong) we usually prefer `{Type}Extensions`, which would make this `StringExtensions`.

thomaslevesque: Normally I would agree with @adamralph, but in this case I think we can make an exception to the rule, because this method only makes sense in the context of xBehave

adamralph: It makes sense in the context of our specs, where we want these extension
methods on strings.

blairconrad: My initial reasoning had been along the same lines as @thomaslevesque, but in some sense, that's the case for other extensions that we write: they're really for _our_ use, regardless of the type they "extend". E.g., `TypeExtensions.GetDefaultValue`.
Unless @thomaslevesque objects, I am inclined to rename the class `StringExtensions`, following the other examples. The system is easy to remember (my example today notwithstanding) and apply, with no soul-searching required.

If nothing else, switching to a more sophisticated system as @thomaslevesque hints at could be done as part of a separate change…

blairconrad:
Thanks for the comments, gents. I'm happy to update the usings and to rename the extensions class (pending @thomaslevesque's' approval).

thomaslevesque:
No objection from me. I didn't have time to do a full review, but if @adamralph is happy with the changes, I'm OK to merge.

blairconrad:
Thanks, @thomaslevesque. I'm not sure if @adamralph has finished his review or not, but I doubt it. He usually comments.

In the meantime, I amended my commits, incorporating all suggested changes up 'til now.
(I figured they changes weren't that extensive and an amendment wouldn't confuse people.)

blairconrad:
Take your time, sir. I'm in no rush.

tests/FakeItEasy.Specs/CreationOptionsSpecs.cs
```diff
@@ -725,7 +725,7 @@ public interface IInterfaceThatWeWillAddAttributesTo3
                     var customAttributeBuilders1 = new List<CustomAttributeBuilder> { attribute1 };
 
                     var constructor2 = typeof(ExampleAttribute).GetConstructor(new[] { typeof(object[]) });
-                    var attribute2 = new CustomAttributeBuilder(constructor2, new[] { new object[] { 1, null } });
+                    var attribute2 = new CustomAttributeBuilder(constructor2, new object[] { new object[] { 1, null } });
```
adamralph: Why was this required?

adamralph:
LGTM! I left one question about the explicit array type declaration.

adamralph:
BTW

> Fixes #574.

#574 is a pull request. (And this can't fix #461 because that's already fixed.) :smile: 

Shall we label this as an independent refactoring?

tests/FakeItEasy.Specs/CreationOptionsSpecs.cs
thomaslevesque: ReSharper warning I think:

> Covariant array conversion from `object[][]` to `object[]` can cause runtime exception on write operation

adamralph: Resharper at it's best! :trollface:

thomaslevesque: Well, it wasn't really an issue in that case, since the array is never written to after it's created, but it's a valid warning in general...

thomaslevesque:
> Shall we label this as an independent refactoring?

Done.

Anything else to do before we can merge? I need to make changes to the same specs for #482, so I'd rather use the latest version to avoid conflicts...

blairconrad:
@adamralph, "Fixes #574" was a typo. Supposed to be #582. I've made the correction. No idea what I did there.

blairconrad:
And @thomaslevesque was correct about the explicit cast. Does anyone mind if we leave as is?

Thanks for the comments, folks!

thomaslevesque:
> Does anyone mind if we leave as is?

I don't. The code is "more correct" this way, regardless of where the suggestion came from ;)

thomaslevesque:
Merged. Thanks @blairconrad!

blairconrad:
Thanks again!

tests/FakeItEasy.Specs/StringExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        /// <summary>
+        /// Creates a step builder that can be used to refer to a type's method.
+        /// Useful when we want to establish a condition with actually performing an action.
+        /// </summary>
+        /// <typeparam name="T">The type to look at.</typeparam>
+        /// <param name="text">A description of the method's relevant quality.</param>
+        /// <param name="func">The method to look at. Will not be executed.</param>
+        /// <returns>A step builder.</returns>
+        [SuppressMessage("Microsoft.Usage", "CA1801:ReviewUnusedParameters", MessageId = "func", Justification = "Used to indicate the method to see.")]
+        public static IStepBuilder See<T>(this string text, Func<T, Action<Type, IFakeOptions>> func)
```
thomaslevesque: A bit late to notice it, but... was this intentional? I don't think there should be an overload as specific as that.

thomaslevesque: It should probably be something like `public static IStepBuilder See<T>(this string text, Func<T, object> func)` instead

adamralph: True, but I think we can invoke YAGNI for now and widen the parameter type when we find we need to.

In fact, we could probably drop all the constraints and widen it all the way up to:

```C#
See(this string text, params object[] args)
```

thomaslevesque: Actually, it' s not that simple... Currently it's used like this:


    .See<DomainEventFakeOptionsBuilder>(_ => _.BuildOptions);

where `BuildOptions` is a method. For this to work, there must be an overload of `See` with the right signature, i.e. with a delegate parameter whose return type is a delegate that matches the method group. And we can't really make it generic (`See<T, T1, T2>(this string text, Func<T, Action<T1, T2>> func)`), because then, we would need to specify all type parameters. And anyway, we would need to make as many overloads as there are possible signatures...

So, even though it's nice to be able to use the method group like that, maybe we should do something like this instead:

    .See(nameof(DomainEventFakeOptionsBuilder.BuildOptions));

blairconrad: Yeah, the only nice-looking (from the call site) option is to have a `See` with the fully-specified delegate.
An alternative would be (depending on how we end up using `See` in the future) to just declare those extensions nearer usage, e.g. in `FakeOptionsBuilderSpecs.cs`.
However, I'm content with using `nameof`. It doesn't help when methods have overloads, but I can live with that.

thomaslevesque: > It doesn't help when methods have overloads, but I can live with that.

It doesn't matter for `nameof` if the method has overloads, since we don't need to know which overload it refers to. Not sure what happens if one of the overloads is renamed, though.

blairconrad: I'd just meant that by using `nameof`, we don't identify a particular overload.
Your point is a good one, though.

I don't really think it's going to matter. `nameof` will probably suffice.

(Then again, I'm not that bothered by the very specific `See` that we have now either.)

PR #753, opened 2016-06-10
Call out use of prefixes for property method names

### LINKED ISSUE #752: more clearly point out use of set_ and get_ to specify properties' method names
It's not super obvious what's going on, as a client recently [pointed out to me](https://gitter.im/FakeItEasy/FakeItEasy?at=57586cde2eaa837d71e794ae):

> I saw that, but it was not clear if I need to specify 'set_', or that was what the developer chose for a name, but after you point it out, I (get) see it now

The section to update would be [Specifying a call to any method or property](http://fakeiteasy.readthedocs.io/en/stable/specifying-a-call-to-configure/#specifying-a-call-to-any-method-or-property)

blairconrad:
Fixes #752


adamralph:
Thanks @blairconrad !

PR #756, opened 2016-06-12
Add non-generic overloads for A.Fake

### LINKED ISSUE #482: Non-generic Fake and Dummy creation

thomaslevesque:
and for `A.Dummy`, `A.CollectionOfFake` and `A.CollectionOfDummy`

Fixes #482

### DIFF
--- a/docs/creating-fakes.md
+++ b/docs/creating-fakes.md
@@ -8,6 +8,17 @@ var foo = A.Fake<IFoo>();
 ```
 This will return a faked object that is an actual instance of the type specified (`IFoo` in this case).
 
+You can also create a collection of fakes by writing:
+```csharp
+var foos = A.CollectionOfFake<Foo>(10);
+```
+
+For cases where the type to fake isn't statically known, non-generic methods are also available:
+```csharp
+object foo = A.Fake(typeof(Foo));
+IList<object> foos = A.CollectionOfFake(typeof(Foo), 10);
+```
+
 ##Explicit Creation Options
 When creating fakes you can, through a fluent interface, specify options for how the fake should be created:
 
@@ -31,14 +42,14 @@ var foo = A.Fake<FooClass>(x => x.WithArgumentsForConstructor(() => new FooClass
 var foo = A.Fake<FooClass>(x => x.WithArgumentsForConstructor(new object[] { "foo", "bar" }));
 
 // Specifying additional interfaces to be implemented. Among other uses,
-// this can help when a fake skips members because they have been 
+// this can help when a fake skips members because they have been
 // explicitly implemented on the class being faked.
 var foo = A.Fake<FooClass>(x => x.Implements(typeof(IFoo)));
-// or 
+// or
 var foo = A.Fake<FooClass>(x => x.Implements<IFoo>());
 
 // Assigning custom attributes to the faked class.
-// Get a parameterless constructor for our attribute and create a builder 
+// Get a parameterless constructor for our attribute and create a builder
 var constructor = typeof(FooAttribute).GetConstructor(new Type[0]);
 var builder = new CustomAttributeBuilder(constructor, new object[0]);
 var builders = new List<CustomAttributeBuilder>() { builder };
--- a/src/FakeItEasy/A.cs
+++ b/src/FakeItEasy/A.cs
@@ -113,6 +113,7 @@ namespace FakeItEasy
         /// <param name="typeOfDummy">The type of dummy to return.</param>
         /// <returns>A dummy object of the specified type.</returns>
         /// <exception cref="ArgumentException">Dummies of the specified type can not be created.</exception>
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
         public static object Dummy(Type typeOfDummy)
         {
             return FakeCreator.CreateDummy(typeOfDummy);
@@ -126,6 +127,7 @@ namespace FakeItEasy
         /// <returns>A collection of dummy objects of the specified type.</returns>
         /// <exception cref="ArgumentException">Dummies of the specified type can not be created.</exception>
         [SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter", Justification = "Used to specify the type of dummy.")]
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
         public static IList<T> CollectionOfDummy<T>(int numberOfDummies)
         {
             return FakeCreator.CollectionOfDummy<T>(numberOfDummies);
@@ -138,6 +140,7 @@ namespace FakeItEasy
         /// <param name="numberOfDummies">The number of dummies in the collection.</param>
         /// <returns>A collection of dummy objects of the specified type.</returns>
         /// <exception cref="ArgumentException">Dummies of the specified type can not be created.</exception>
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
         public static IList<object> CollectionOfDummy(Type typeOfDummy, int numberOfDummies)
         {
             return FakeCreator.CollectionOfDummy(typeOfDummy, numberOfDummies);
--- a/src/FakeItEasy/A.cs
+++ b/src/FakeItEasy/A.cs
@@ -49,6 +49,27 @@ namespace FakeItEasy
         }
 
         /// <summary>
+        /// Creates a fake object of the specified type.
+        /// </summary>
+        /// <param name="typeOfFake">The type of fake object to create.</param>
+        /// <returns>A fake object.</returns>
+        public static object Fake(Type typeOfFake)
+        {
+            return FakeCreator.CreateFake(typeOfFake, x => { });
+        }
+
+        /// <summary>
+        /// Creates a fake object of the specified type.
+        /// </summary>
+        /// <param name="typeOfFake">The type of fake object to create.</param>
+        /// <param name="optionsBuilder">A lambda where options for the built fake object can be specified.</param>
+        /// <returns>A fake object.</returns>
+        public static object Fake(Type typeOfFake, Action<IFakeOptions> optionsBuilder)
+        {
+            return FakeCreator.CreateFake(typeOfFake, optionsBuilder);
+        }
+
+        /// <summary>
         /// Creates a collection of fakes of the specified type.
         /// </summary>
         /// <typeparam name="T">The type of fakes to create.</typeparam>
@@ -58,6 +79,17 @@ namespace FakeItEasy
         public static IList<T> CollectionOfFake<T>(int numberOfFakes)
         {
             return FakeCreator.CollectionOfFake<T>(numberOfFakes);
+        }
+
+        /// <summary>
+        /// Creates a collection of fakes of the specified type.
+        /// </summary>
+        /// <param name="typeOfFake">The type of fakes to create.</param>
+        /// <param name="numberOfFakes">The number of fakes in the collection.</param>
+        /// <returns>A collection of fake objects of the specified type.</returns>
+        public static IList<object> CollectionOfFake(Type typeOfFake, int numberOfFakes)
+        {
+            return FakeCreator.CollectionOfFake(typeOfFake, numberOfFakes);
         }
 
         /// <summary>
@@ -75,6 +107,18 @@ namespace FakeItEasy
         }
 
         /// <summary>
+        /// Gets a dummy object of the specified type. The value of a dummy object
+        /// should be irrelevant. Dummy objects should not be configured.
+        /// </summary>
+        /// <param name="typeOfDummy">The type of dummy to return.</param>
+        /// <returns>A dummy object of the specified type.</returns>
+        /// <exception cref="ArgumentException">Dummies of the specified type can not be created.</exception>
+        public static object Dummy(Type typeOfDummy)
+        {
+            return FakeCreator.CreateDummy(typeOfDummy);
+        }
+
+        /// <summary>
         /// Creates a collection of dummies of the specified type.
         /// </summary>
         /// <typeparam name="T">The type of dummies to create.</typeparam>
@@ -85,6 +129,18 @@ namespace FakeItEasy
         public static IList<T> CollectionOfDummy<T>(int numberOfDummies)
         {
             return FakeCreator.CollectionOfDummy<T>(numberOfDummies);
+        }
+
+        /// <summary>
+        /// Creates a collection of dummies of the specified type.
+        /// </summary>
+        /// <param name="typeOfDummy">The type of dummy to return.</param>
+        /// <param name="numberOfDummies">The number of dummies in the collection.</param>
+        /// <returns>A collection of dummy objects of the specified type.</returns>
+        /// <exception cref="ArgumentException">Dummies of the specified type can not be created.</exception>
+        public static IList<object> CollectionOfDummy(Type typeOfDummy, int numberOfDummies)
+        {
+            return FakeCreator.CollectionOfDummy(typeOfDummy, numberOfDummies);
         }
 
         /// <summary>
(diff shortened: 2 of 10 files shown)


tests/FakeItEasy.Specs/NonGenericCreationOptionsSpecs.cs
```diff
@@ -0,0 +1,845 @@
...
+    using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
+    using System.Reflection.Emit;
+    using FakeItEasy.Creation;
+    using FakeItEasy.Tests;
+    using FakeItEasy.Tests.TestHelpers;
+    using FluentAssertions;
+    using Xbehave;
+
+    public static class NonGenericCreationOptionsSpecs
```
thomaslevesque: Note: this is an exact copy of `CreationOptionsSpecs`, except that it tests the non-generic `A.Fake`

src/FakeItEasy/A.cs
```diff
@@ -75,6 +107,18 @@ public static T Dummy<T>()
         }
 
         /// <summary>
+        /// Gets a dummy object of the specified type. The value of a dummy object
+        /// should be irrelevant. Dummy objects should not be configured.
+        /// </summary>
+        /// <param name="typeOfDummy">The type of dummy to return.</param>
+        /// <returns>A dummy object of the specified type.</returns>
+        /// <exception cref="ArgumentException">Dummies of the specified type can not be created.</exception>
```
blairconrad: It's odd that this method seems to be more accessible than `Dummy<T>` (`EditorBrowsable`). I'd think they should be the same.

src/FakeItEasy/A.cs
```diff
@@ -49,6 +49,27 @@ public static T Fake<T>(Action<IFakeOptions<T>> optionsBuilder)
...
+        {
+            return FakeCreator.CreateFake(typeOfFake, x => { });
+        }
+
+        /// <summary>
+        /// Creates a fake object of the specified type.
+        /// </summary>
+        /// <param name="typeOfFake">The type of fake object to create.</param>
+        /// <param name="optionsBuilder">A lambda where options for the built fake object can be specified.</param>
+        /// <returns></returns>
```
blairconrad: Empty `returns`

tests/FakeItEasy.Specs/NonGenericCreationOptionsSpecs.cs
blairconrad: This tempts me to suggest parameterized specs, or a common base, but I wonder if it's worth it.

src/FakeItEasy/A.cs
thomaslevesque: Oops!

src/FakeItEasy/A.cs
thomaslevesque: Ah, I missed the `EditorBrowsable` attribute on `Dummy<T>`... I'll add it

tests/FakeItEasy.Specs/NonGenericCreationOptionsSpecs.cs
thomaslevesque: It might be worth it (would avoid a lot of duplication); I'm not sure it could be done, but I'll give it a try.

src/FakeItEasy/A.cs
adamralph: Shouldn't the same apply to `CollectionOfDummy`?

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
```diff
@@ -35,6 +35,14 @@ public T CreateFake<T>(Action<IFakeOptions<T>> optionsBuilder)
             return (T)this.fakeAndDummyManager.CreateFake(typeof(T), options => optionsBuilder((IFakeOptions<T>)options));
         }
 
+        public object CreateFake(Type typeOfFake, Action<IFakeOptions> optionsBuilder)
+        {
+            Guard.AgainstNull(typeOfFake, nameof(typeOfFake));
+            Guard.AgainstNull(optionsBuilder, nameof(optionsBuilder));
```
adamralph: This method does not de-reference either argument. These guards should only be added if it's not possible for a method further down the call stack to give the same quality of exception message. Has that been verified?

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
```diff
@@ -55,6 +63,20 @@ public IList<T> CollectionOfFake<T>(int numberOfFakes)
             return result;
         }
 
+        public IList<object> CollectionOfFake(Type typeOfFake, int numberOfFakes)
+        {
+            Guard.AgainstNull(typeOfFake, nameof(typeOfFake));
```
adamralph: Same question again about the guard.

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
```diff
@@ -68,6 +90,13 @@ public T CreateDummy<T>()
             return (T)this.fakeAndDummyManager.CreateDummy(typeof(T));
         }
 
+        public object CreateDummy(Type typeOfDummy)
+        {
+            Guard.AgainstNull(typeOfDummy, nameof(typeOfDummy));
```
adamralph: And again.

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
```diff
@@ -87,5 +116,19 @@ public IList<T> CollectionOfDummy<T>(int numberOfDummies)
 
             return result;
         }
+
+        public IList<object> CollectionOfDummy(Type typeOfDummy, int numberOfDummies)
+        {
+            Guard.AgainstNull(typeOfDummy, nameof(typeOfDummy));
```
adamralph: And again.

src/FakeItEasy/A.cs
blairconrad: If we're adding `EditorBrowsable`s, absolutely

adamralph: I'd be happy with the missing ones being added in a separate commit in this PR.

adamralph:
LGTM. My comments are minor concerns.

blairconrad:
Me too.

Oh! Docs.

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
thomaslevesque: Ah, I wasn't aware of this rule. I saw a `Guard.AgainstNull` in the generic `CreateFake<T>`, so I added it here as well.

thomaslevesque:
> Oh! Docs.

Ah, yes. Thanks for the reminder!

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
adamralph: No problem, fully understandable. I believe we're using null guards in a lot of redundant places. It may be worth doing a cross cutting sweep of all of them at some point and we should consider our approach then.

For this PR, I think it's fine as-is.

docs/creating-fakes.md
```diff
@@ -8,6 +8,17 @@ var foo = A.Fake<IFoo>();
 ```
 This will return a faked object that is an actual instance of the type specified (`IFoo` in this case).
 
+You can also create a collection of fakes by writing:
+```csharp
+var foos = A.CollectionOfFake<Foo>(10);
+```
```
thomaslevesque: Not really related to this issue, but I noticed `CollectionOfFake<T>` wasn't introduced. It would have been weird to mention the non-generic one without mentioning the generic one, so I added it as well.

docs/creating-fakes.md
```diff
@@ -31,14 +42,14 @@ var foo = A.Fake<FooClass>(x => x.WithArgumentsForConstructor(() => new FooClass
 var foo = A.Fake<FooClass>(x => x.WithArgumentsForConstructor(new object[] { "foo", "bar" }));
 
 // Specifying additional interfaces to be implemented. Among other uses,
-// this can help when a fake skips members because they have been
```
thomaslevesque: editorconfig

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
```diff
@@ -35,6 +35,14 @@ public T CreateFake<T>(Action<IFakeOptions<T>> optionsBuilder)
             return (T)this.fakeAndDummyManager.CreateFake(typeof(T), options => optionsBuilder((IFakeOptions<T>)options));
         }
 
+        public object CreateFake(Type typeOfFake, Action<IFakeOptions> optionsBuilder)
```
thomaslevesque: I realized I didn't copy the documentation from the interface. It's an internal class, so I'm not sure it really matters, but if you feel it's important, I'll add it.

src/FakeItEasy/Creation/IFakeCreatorFacade.cs
```diff
@@ -18,8 +18,17 @@ internal interface IFakeCreatorFacade
...
+        /// <param name="typeOfFake">The type of fake to create.</param>
+        /// <param name="optionsBuilder">Action that builds options for the created fake object.</param>
+        /// <returns>The created fake object.</returns>
+        /// <exception cref="FakeItEasy.Core.FakeCreationException">Was unable to generate the fake in the current configuration.</exception>
+        object CreateFake(Type typeOfFake, Action<IFakeOptions> optionsBuilder);
+
+        /// <summary>
         /// Creates a dummy object, this can be a fake object or an object resolved
-        /// from the current IFakeObjectContainer.
+        /// from a dummy factory.
```
thomaslevesque: There seems to be a few places where `IFakeObjectContainer` is still mentioned... we should probably make a sweep to fix it everywhere.

tests/FakeItEasy.Specs/CreationOptionsSpecs.cs
```diff
@@ -835,12 +835,30 @@ public interface IInterfaceThatWeWillAddAttributesTo3
...
             "When I create a fake using the options builder"
-                .x(() => exception = Record.Exception(() => A.Fake(optionsBuilder)));
+                .x(() => exception = Record.Exception(() => this.CreateFake(optionsBuilder)));
 
             "Then fake creation fails"
                 .x(() => exception.Should().BeAnExceptionOfType<ArgumentException>()
                 .WithMessage("*Only expression of the type ExpressionType.New (constructor calls) are accepted.*"));
         }
+
+        protected abstract T CreateFake<T>(Action<IFakeOptions<T>> optionsBuilder);
```
thomaslevesque: This is the simplest way I found to test both the generic and non-generic overloads. Do you see any issue with it?

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
```diff
@@ -0,0 +1,129 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using FakeItEasy.Tests.TestHelpers;
+    using FluentAssertions;
+    using Xbehave;
+
+    public abstract class DummyCreationSpecsBase
```
thomaslevesque: There were no specs for dummy creation, so I added some. Probably partially redundant with `DummyTests`

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
adamralph: I don't think we need it.

src/FakeItEasy/Creation/IFakeCreatorFacade.cs
adamralph: Raised https://github.com/FakeItEasy/FakeItEasy/issues/757

src/FakeItEasy/A.cs
```diff
@@ -113,6 +113,7 @@ public static T Dummy<T>()
         /// <param name="typeOfDummy">The type of dummy to return.</param>
         /// <returns>A dummy object of the specified type.</returns>
         /// <exception cref="ArgumentException">Dummies of the specified type can not be created.</exception>
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
```
adamralph: This should really have been added in the commit which added this new overload, but I'm happy to proceed as-is.

(Same applies to `CollectionOfDummy(Type typeOfDummy, int numberOfDummies)`.)

adamralph:
LGTM - I'm :+1: to merge.

src/FakeItEasy/A.cs
thomaslevesque: At some point you said you wanted all `EditorBrowsable` attributes to be added in a separate commit; maybe I misunderstood. Anyway, I don't mind squashing this commit if you prefer.

adamralph: Sorry, I wasn't clear. I meant only the ones that are currently missing in master (there's only one).

Anyway, I'm happy as-is. No need to squash.

src/FakeItEasy/Creation/IFakeCreatorFacade.cs
```diff
@@ -28,6 +37,16 @@ internal interface IFakeCreatorFacade
         T CreateDummy<T>();
 
         /// <summary>
+        /// Creates a dummy object, this can be a fake object or an object resolved
+        /// from a dummy factory.
+        /// </summary>
+        /// <param name="typeOfDummy">The type of dummy to create.</param>
+        /// <returns>The created dummy.</returns>
+        /// <exception cref="FakeItEasy.Core.FakeCreationException">Was unable to generate the fake in the current configuration and
+        /// no dummy was registered in the container for the specified type..</exception>
```
blairconrad: The double "." is odd here, but the docs were just copied, so not your fault. At least it's consistent.

tests/FakeItEasy.Specs/CreationOptionsSpecs.cs
blairconrad: This is about what I imagined happening. An alternative would be to split the two concrete classes into their own files, but everything's so tightly related that I don't mind it as is.

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
blairconrad: Commendable! I see no need to root out the redundancies.

blairconrad: … but the day the redundant unit tests inconvenience me: :skull:

src/FakeItEasy/Creation/IFakeCreatorFacade.cs
blairconrad: We could leave and clean up in another PR.

thomaslevesque: Fixed

blairconrad:
As usual, sterling work, @thomaslevesque. I did make a minor note about (mostly existing) XML docs, but ~~I think we can wait~~ you corrected it while I was typing this note!

thomaslevesque:
Thanks!

adamralph:
:+1: great work all round!

BTW - don't forget to add the closed issue to the next milestone. I've done it for #482

thomaslevesque:
> BTW - don't forget to add the closed issue to the next milestone.

will do

blairconrad:
> BTW - don't forget to add the closed issue to the next milestone.

As merger, I should've done this. :blush:

PR #761, opened 2016-06-14
Options for fake collection

### LINKED ISSUE #755: Add overload with optionsBuilder parameter for CollectionOfFake
Unlike `A.Fake`, `A.CollectionOfFake` doesn't have an overload with an `optionsBuilder` parameter. It's probably an oversight, which could easily be remedied.

thomaslevesque:
fixes #755

### DIFF
--- a/src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
+++ b/src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
@@ -53,11 +53,16 @@ namespace FakeItEasy.Creation
         /// </returns>
         public IList<T> CollectionOfFake<T>(int numberOfFakes)
         {
+            return this.CollectionOfFake<T>(numberOfFakes, x => { });
+        }
+
+        public IList<T> CollectionOfFake<T>(int numberOfFakes, Action<IFakeOptions<T>> optionsBuilder)
+        {
             var result = new List<T>();
 
             for (var i = 0; i < numberOfFakes; i++)
             {
-                result.Add(this.CreateFake<T>(x => { }));
+                result.Add(this.CreateFake(optionsBuilder));
             }
 
             return result;
@@ -65,13 +70,16 @@ namespace FakeItEasy.Creation
 
         public IList<object> CollectionOfFake(Type typeOfFake, int numberOfFakes)
         {
-            Guard.AgainstNull(typeOfFake, nameof(typeOfFake));
+            return this.CollectionOfFake(typeOfFake, numberOfFakes, x => { });
+        }
 
+        public IList<object> CollectionOfFake(Type typeOfFake, int numberOfFakes, Action<IFakeOptions> optionsBuilder)
+        {
             var result = new List<object>();
 
             for (var i = 0; i < numberOfFakes; i++)
             {
-                result.Add(this.CreateFake(typeOfFake, x => { }));
+                result.Add(this.CreateFake(typeOfFake, optionsBuilder));
             }
 
             return result;
--- a/tests/FakeItEasy.Specs/CreationSpecs.cs
+++ b/tests/FakeItEasy.Specs/CreationSpecs.cs
@@ -94,11 +94,33 @@ namespace FakeItEasy.Specs
                 .x(() => fakes.Should().OnlyContain(item => Fake.GetFakeManager(item) != null));
         }
 
+        [Scenario]
+        [Example(2)]
+        [Example(10)]
+        public void CollectionOfFakeWithOptionBuilder(
+            int count,
+            IList<ICollectionItem> fakes)
+        {
+            "When I create a collection of {0} fakes that also implement another interface"
+                .x(() => fakes = this.CreateCollectionOfFake<ICollectionItem>(count, options => options.Implements<IDisposable>()));
+
+            "Then {0} items are created"
+                .x(() => fakes.Should().HaveCount(count));
+
+            "And all items extend the specified type and the extra interface"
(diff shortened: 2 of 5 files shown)


src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
```diff
@@ -53,25 +53,33 @@ public object CreateFake(Type typeOfFake, Action<IFakeOptions> optionsBuilder)
...
+                result.Add(this.CreateFake(optionsBuilder));
             }
 
             return result;
         }
 
-        public IList<object> CollectionOfFake(Type typeOfFake, int numberOfFakes)
+        public IList<object> CollectionOfFake(Type typeOfFakes, int numberOfFakes)
         {
-            Guard.AgainstNull(typeOfFake, nameof(typeOfFake));
```
thomaslevesque: This guard clause was redundant, because we end up calling `CreateFake`, which already has the guard clause.

tests/FakeItEasy.Specs/CreationSpecs.cs
```diff
@@ -177,5 +204,10 @@ protected override IList<T> CreateCollectionOfFake<T>(int numberOfFakes)
         {
             return A.CollectionOfFake(typeof(T), numberOfFakes).Cast<T>().ToList();
         }
+
+        protected override IList<T> CreateCollectionOfFake<T>(int numberOfFakes, Action<IFakeOptions<T>> optionsBuilder)
+        {
+            return A.CollectionOfFake(typeof(T), numberOfFakes, options => optionsBuilder((IFakeOptions<T>)options)).Cast<T>().ToList();
```
adamralph: Why the wrapped call here? Can't we just pass `optionsBuilder` directly?

thomaslevesque: The non-generic overload accepts an `Action<IFakeOptions>`, not `Action<IFakeOptions<T>>` There's no inheritance relation between the interfaces, so contravariance doesn't work here.

thomaslevesque: [This code](https://gist.github.com/thomaslevesque/a4ad1fc87dd03edfb3b8dfc589146156) works, but if you remove the inheritance between the interfaces, it doesn't.

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
```diff
@@ -53,25 +53,33 @@ public object CreateFake(Type typeOfFake, Action<IFakeOptions> optionsBuilder)
...
             {
-                result.Add(this.CreateFake<T>(x => { }));
+                result.Add(this.CreateFake(optionsBuilder));
             }
 
             return result;
         }
 
-        public IList<object> CollectionOfFake(Type typeOfFake, int numberOfFakes)
+        public IList<object> CollectionOfFake(Type typeOfFakes, int numberOfFakes)
```
adamralph: The parameter in `A` is named `typeOfFake` so I think we should be consistent.

If we'd rather pluralise it in `A` then I think that's OK, but only because we haven't released those overloads yet, as this could be called a breaking change (I'm not sure what our breaking/non-breaking policy is on parameter name changes).

Whatever the decision, I'd prefer this commit to be consistent with `A` and remain singular. This would avoid having this unrelated change here. If we decide to pluralise both here and in `A`, that should be a separate commit.

thomaslevesque: > The parameter in A is named typeOfFake so I think we should be consistent.

Ah, yes... I changed it so that it would be the same as on the interface. I'll make it singular everywhere.

adamralph: I see that the existing parameter in `IFakeCreatorFacade` is plural so there is already inconsistency. I think the cleanest way is to remove this change from this commit, since it's unrelated to adding the new overloads, and add another commit which makes everything consistent, whether singular or plural. I have no preference which.

adamralph: My last comment overlapped with yours. Seems we're in agreement.

tests/FakeItEasy.Specs/CreationSpecs.cs
adamralph: Ah OK, makes sense.

src/FakeItEasy/Creation/DefaultFakeCreatorFacade.cs
thomaslevesque: Do you still want a separate commit? If we make it singular everywhere, the only place on `master` where it needs to be fixed is `IFakeCreationFacade.CollectionOfFake(Type typeOfFakes, int numberOfFakes)`

adamralph: For a such small change, no, same commit is fine.

thomaslevesque:
All done, I think.

adamralph:
Thanks!

thomaslevesque:
Thanks for the comments :smile:

blairconrad:
Nice. Thank you.
Is it worthwhile adding to the docs? At some point, I'm of the mind that "given the docs for the simpler examples, these things are figureoutable"…

thomaslevesque:
> Is it worthwhile adding to the docs?

I considered it, and thought it wasn't. We have examples of `A.Fake` with options and `A.FakeCollection` without, so people will probably figure out they can pass options when creating a collection of fakes, too. And the Intellisense helps, too.

adamralph:
Agreed. No need to cover every single point in the overload matrix in the docs.

PR #674, opened 2016-04-22
[WIP] Add WithNonVoidReturnType function

balazsbonis:
DO NOT MERGE.

I started working on the https://github.com/FakeItEasy/FakeItEasy/issues/648 issue, just to see if I can manage to get it done, and I got to weird exception here.

First of all, I don't know if I am taking the right approach, so feel free to tell me where should I look at.

Secondly, I added a unit test to see how my code works so far (Should_return_null_from_trapper_when_set_to_return_null in ExpressionArgumentConstraintFactoryTests.cs), and it blows up saying this:

`Expected object to be <null>, but found <NULL>.`

which I cannot wrap my head around...

Can I ask one of the FakeItEasy gurus to help me out with this? :smile:

PR #762, opened 2016-06-14
Add WithNonVoidReturnType

### LINKED ISSUE #648: Common default return value of functions with generics

```c#
```

thomaslevesque:
Supersedes #674 

I left @balazsbonis's commit intact for now so you can see what I changed; I'll squash the first two commits before merge.

Fixes #648


tests/FakeItEasy.Analyzer.Tests/Helpers/CodeFixVerifier.cs
```diff
@@ -46,7 +46,6 @@ protected virtual CodeFixProvider GetBasicCodeFixProvider()
         /// <param name="newSource">A class in the form of a string after the code fix was applied to it.</param>
         /// <param name="codeFixIndex">Index determining which code fix to apply if there are multiple.</param>
         /// <param name="allowNewCompilerDiagnostics">A boolean controlling whether or not the test will fail if the code fix introduces other warnings after being applied.</param>
-        [SuppressMessage("Microsoft.Design", "CA1026:DefaultParametersShouldNotBeUsed", Justification = "It's not intended to be a public API, so it doesn't matter")]
```
thomaslevesque: I disabled this warning in the ruleset for tests, since it only matters for public APIs

tests/FakeItEasy.Tests.ruleset
```diff
@@ -21,4 +22,4 @@
     <Rule Id="CA2000" Action="None" />
     <Rule Id="CA2210" Action="None" />
   </Rules>
-</RuleSet>
\ No newline at end of file
+</RuleSet>
```
adamralph: Are our editorconfig settings battling against VS again?

tests/FakeItEasy.Specs/AnyCallConfigurationSpecs.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            "And the configured method returns null when called with generic argument IFoo"
+                .x(() => fake.Bar<IFoo>().Should().BeNull());
+        }
+
+        [Scenario]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "IFoo", Justification = "It's an identifier")]
+        public static void WithReturnType(
+            IFoo fake,
+            string returnValue = "hello world")
```
adamralph: Are you sure this works? It would come as a surprise to me. Default parameter values are baked into call sites at compile time. I don't think xbehave could even see this value if it wanted to. I would assign it inline:

```C#
.x(() => A.CallTo(fake).Where(call => call.Method.Name == "Bar").WithReturnType<string>().Returns(returnValue = "hello world"));
```

thomaslevesque: Oh, well spotted! I'll fix it.

tests/FakeItEasy.Tests.ruleset
thomaslevesque: Looks like it... I assume that's because of the ruleset designer; editorconfig only works when you edit the source (which is what I did)

tests/FakeItEasy.Specs/AnyCallConfigurationSpecs.cs
thomaslevesque: If I change this, I no longer need to suppress the CA warning. Do you want me to rollback the changes to the ruleset and suppressions?

thomaslevesque: Anyway, I think there are quite a few warnings that we should disable for tests; this could be done in its own PR.

adamralph: :+1: to rollback and tackle it all separately.

adamralph:
If you squash the first two commits, I presume it will keep @balazsbonis as the author?

thomaslevesque:
> If you squash the first two commits, I presume it will keep @balazsbonis as the author?

I'm almost certain it will, but I'll check before pushing

adamralph:
It's academic in this case, since he's not GPG signing his commits, but if he was, that would remove the `Verified` from this commit. I wonder how we'd handle that case... I'd guess we'd just leave his commit as-is. But then, we wouldn't necessarily be able to rebase it as is, because of conflicts, and editing those conflicts during rebase would remove the GPP signature too...

thomaslevesque:
I don't think it's possible to rebase a GPG signed commit, even if there's no conflict, since the SHA1 would be different

thomaslevesque:
(I mean someone else's commit, of course; you can still rebase your own commit)

thomaslevesque:
I made all the changes, including the squash. Thanks for the review!

tests/FakeItEasy.Tests/Configuration/AnyCallConfigurationTests.cs
```diff
@@ -50,6 +50,18 @@ public void WithReturnType_should_set_the_type_to_the_configured_rule()
...
         [Test]
+        public void WithNonVoidReturnType_should_set_the_flag_to_apply_to_all_return_types()
+        {
+            // Arrange
+
+            // Act
+            this.configuration.WithNonVoidReturnType();
+
+            // Assert
+            Assert.That(this.callRule.ApplicableToAllNonVoidReturnTypes, Is.EqualTo(true));
```
blairconrad: Is this the only reason the property has a getter? I'd be inclined to drop it (and this test) in that case. Of course, that leaves  a set-only property, but we could switch to a method.
On the other hand, the property as is _does_ follow the pattern established by `ApplicableToMembersWithReturnType`, so if it stays as is, I'm okay.

blairconrad:
Generally looks good to me. Thanks, @thomaslevesque!

tests/FakeItEasy.Tests/Configuration/AnyCallConfigurationTests.cs
thomaslevesque: Well, it's also read in `AnyCallCallRule`... but yes, it's the only place outside the class where it's read. It doesn't really bother me, though; IMO, if we can set it, we should be able to get it too.

adamralph: I'm fine with it remaining. I'd care more if the type was public, but since it's internal.. meh.

adamralph:
Thanks @thomaslevesque !

blairconrad:
I remembered right after I went up to bed last night. Docs?

thomaslevesque:
Argh, I keep forgetting...

thomaslevesque:
I'll send a new PR soon

PR #763, opened 2016-06-15
Add documentation example for WithNonVoidReturnType

### LINKED ISSUE #648: Common default return value of functions with generics

thomaslevesque:
Fixes #648

### DIFF
--- a/docs/specifying-a-call-to-configure.md
+++ b/docs/specifying-a-call-to-configure.md
@@ -55,6 +55,10 @@ A.CallTo(fakeShop).Throws(new Exception());
 // Or limit the calls by return type
 A.CallTo(fakeShop).WithReturnType<string>().Returns("sugar tastes good");
 
+// Or limit the calls to methods that return a value. Note that it will throw at runtime
+// if the configured return value doesn't match the called method's return type.
+A.CallTo(fakeShop).WithNonVoidReturnType().Returns("sugar tastes good");
+
 // Or create a sophisticated test with a predicate that acts on an IFakeObjectCall
 A.CallTo(fakeShop).Where(call => call.Arguments.Count > 4)
                   .Throws(new Exception("too many arguments is bad");


docs/specifying-a-call-to-configure.md
```diff
@@ -55,6 +55,9 @@ A.CallTo(fakeShop).Throws(new Exception());
 // Or limit the calls by return type
 A.CallTo(fakeShop).WithReturnType<string>().Returns("sugar tastes good");
 
+// Or limit the calls to methods that return a value
+A.CallTo(fakeShop).WithNonVoidReturnType().Returns("sugar tastes good");
```
blairconrad: I generally like it. Should we mention that it's not typesafe? I'd like to steer people toward `WithReturnType` if all possible.

docs/specifying-a-call-to-configure.md
```diff
@@ -55,6 +55,9 @@ A.CallTo(fakeShop).Throws(new Exception());
 // Or limit the calls by return type
 A.CallTo(fakeShop).WithReturnType<string>().Returns("sugar tastes good");
 
+// Or limit the calls to methods that return a value
+A.CallTo(fakeShop).WithNonVoidReturnType().Returns("sugar tastes good");
+
```
adamralph: I think this example should go above the `WithReturnType` example, since it is less restrictive. Each example seems to be more restrictive than the one above.

blairconrad: Interesting idea. Me, I like to put the scary powerful things at the bottom, in the  hopes that people stop reading before they found it, because they found the safe thing that's probably what the really want.

`ReturnsLazily` is less restrictive than `Returns`, but I'd rather talk about `Returns` first.

docs/specifying-a-call-to-configure.md
adamralph: Agreed. We should mention that an exception will be thrown at runtime if the argument doesn't match the return type of a method that is called.

docs/specifying-a-call-to-configure.md
blairconrad: So my "typesafe" comment above should've suggested indicating that `WithNonVoidReturnType` is good to use _when you don't know the return type at compile time_.

adamralph: Yep, fair enough. The methods which ought to be favoured should come first.

docs/specifying-a-call-to-configure.md
thomaslevesque: I added a comment mentioning it

adamralph:
:+1: LGTM. @blairconrad?

blairconrad:
Thanks, @thomaslevesque!

PR #764, opened 2016-06-16
Update analyzer for calls to WithNonVoidReturnType

### LINKED ISSUE #648: Common default return value of functions with generics

thomaslevesque:
closes #648


adamralph:
Should we add a test?

adamralph:
Would it be possible to parameterise a single test to do it for all the methods?

thomaslevesque:
> Would it be possible to parameterise a single test to do it for all the methods?

Not really, since we have to write actual valid code that uses the method... Of course, we could pass that code as a parameter, but it would be quite ugly.

adamralph:
I think we ought to write a test. It feels a bit icky pushing this out without any proof that it works.

We have tech debt in that some of other methods aren't covered, but we can pay that off later.

thomaslevesque:
OK. I amended the commit to add a test.

adamralph:
That was quick! Thanks!

blairconrad:
:+1: Thanks!

thomaslevesque:
> That was quick!

The magic of copy-paste-edit ;)

PR #767, opened 2016-06-17
Replace NUnit with xUnit.net

### LINKED ISSUE #545: Switch from NUnit to xUnit.net

blairconrad:
Fixes #545.


blairconrad:
As far as I know, this PR contains all of the necessary updates for us to support xUnit.net. We'll be functional, running all the tests and recording the results.
I would like to add a cleanup commit or two to the end, but I figured I'd let you guys get a head start on the existing commits.
Sorry the third one is so big, but once I pulled the "xUnit.net switch" many things _had_ to be changed, so it was very awkward to break up.

blairconrad:
The CI server reported 1298 tests, vs. 1328 in master, but I had checked the lists locally. I may have made an error.

thomaslevesque:
Thanks @blairconrad ! That's a big one, reviewing it is going to take a while :wink: 

I'll try to make a first pass this week end

blairconrad:
Thanks, @thomaslevesque. It is big. And much more painful to read than to make the changes, I'm sure. Although the were some moments…

blairconrad:
Oh, and I wouldn't want this to delay 2.1.0!

adamralph:
It probably makes sense for us to get 2.1.0 out and then tackle this.

tests/FakeItEasy.Tests/TestHelpers/MethodDataAttribute.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// <summary>
+    /// Provides a data source for a data theory, with the data coming from a public static method.
+    /// The target method must return the test data in something something compatible with IEnumerable&lt;object[]&gt;.
+    /// Caution: the method's result is completely enumerated before any test is run.
+    /// </summary>
+    [SuppressMessage("Microsoft.Design", "CA1019:DefineAccessorsForAttributeArguments",
+        Justification = "No need to access arguments at runtime.")]
+    [DataDiscoverer("Xunit.Sdk.MemberDataDiscoverer", "xunit.core")]
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+    public sealed class MethodDataAttribute : DataAttribute
```
thomaslevesque: If I understand correctly, the difference between this and `MemberDataAttribute` is that it looks for the method on the `ReflectedType` rather than the `DeclaringType`? Maybe the name should reflect that (not sure how, though)

blairconrad: A good idea. I am also not immediately sure how. (There is a secondary difference - this attribute only looks for methods, not properties or fields, but I think I captured that in the name.)

adamralph: `ReflectedTypeMethodData`?

tests/FakeItEasy.IntegrationTests/Assertions/ExceptionMessagesTests.cs
```diff
@@ -4,13 +4,13 @@ namespace FakeItEasy.IntegrationTests.Assertions
     using FakeItEasy.Tests;
     using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
-    using NUnit.Framework;
+    using Xunit;
+    using Record = FakeItEasy.Tests.TestHelpers.Record;
```
adamralph: Shall we kill `FakeItEasy.Tests.TestHelpers.Record` in this changeset? Presumably it's now a redundant duplication?

adamralph:
LGTM. The only thing I think we should also do in this same changeset is remove our `Record` since it's not required anymore.

blairconrad:
Ah. You guys have put more effort into this than I was anticipating at this point. I have cleanup commits coming, but figured there'd be more comments to address first and it would be less confusing if I held off on the cleanup. I'll change `MethodDataAttribute`'s name and push that and the other commits in a bit.

blairconrad:
(Removing `Record` was indeed one of the cleanup steps.)

blairconrad:
Renamed MethodDataAttribute (in commit cf2230cb3dba9df6031cdcfdd0e6e63c23fe5bdb) and added cleanup.
I'm very happy to squash the rename into the previous commit, after you've seen it. And of course any other squashes or changes. Thanks for your comments!

adamralph:
Apologies @blairconrad - in https://github.com/FakeItEasy/FakeItEasy/pull/767#issuecomment-226730149 I thought you meant that all that would come after this PR.

I'll give the latest changes a once-over...

tests/FakeItEasy.Specs/FakeOptionsBuilderSpecs.cs
```diff
@@ -118,6 +118,7 @@ public class FakeOptionsBuilderSpecs
                 .x(() => exception.Should().BeAnExceptionOfType<ExpectationException>());
         }
 
+        [Scenario]
```
adamralph: Oops! :stuck_out_tongue:

blairconrad:
No apologies necessary, @adamralph. And if you prefer, they could come after this PR as well. I would've put the other commits in sooner had 2.1.0 not been looming. Also I had people over yesterday, so that took up some time.

tests/FakeItEasy.Specs/FakeOptionsBuilderSpecs.cs
blairconrad: I know! I accidentally killed it a few PRs ago. I think during the big language cleanup. I can find the reference, if you want.

tests/FakeItEasy.Tests/ExceptionThrowerConfigurationExtensionsTests.cs
```diff
@@ -524,17 +524,7 @@ public void Should_configure_fake_to_throw_exceptions_returned_by_the_factory()
...
-                    var result = x.Invoke(null);
-
-                    if (result == null)
-                    {
-                        return false;
-                    }
-
-                    return typeof(T).IsAssignableFrom(result.GetType());
-                },
+                x => x.Invoke(null) is T,
```
adamralph: :laughing:

tests/FakeItEasy.Tests/Expressions/ArgumentConstraints/AggregateArgumentConstraintTests.cs
```diff
@@ -14,7 +14,7 @@ public AggregateArgumentConstraintTests()
             this.ConstraintField = new AggregateArgumentConstraint(new[] { new EqualityArgumentConstraint("foo"), new EqualityArgumentConstraint("bar") });
         }
 
-        public interface ITypeWithPaldkf
```
adamralph: WTF?

tests/FakeItEasy.Specs/FakeOptionsBuilderSpecs.cs
adamralph: It's OK, I believe you. And forgive you :wink:

adamralph:
LGTM!

I would vote to squash the 3rd, 4th and 5th commits ("Replace NUnit with xUnit", "Rename MethodDataAttribute to ReflectedMethodDataAttribute" and "Use xUnit's record for catching exceptions") into one as they seem to me to be the entirety of "Replace NUnit with xUnit". The other commits make sense as before/after refactorings.

tests/FakeItEasy.Tests/ExceptionThrowerConfigurationExtensionsTests.cs
blairconrad: This was perhaps my favourite part of the whole exercise.

tests/FakeItEasy.Tests/Expressions/ArgumentConstraints/AggregateArgumentConstraintTests.cs
blairconrad: I can put it back, if you prefer.

But if you're asking where this came from, 8fce4e74354ab120eaa20667fa207b5556d4c447.

adamralph: No, the new version is better. I was just surprised at the old version.

tests/FakeItEasy.Tests/TestHelpers/MethodDataAttribute.cs
thomaslevesque: > `ReflectedTypeMethodData`?

:+1:
Not great if it was for a public API, but since it's for our own internal use, I think it's explicit enough.

thomaslevesque:
Looks good to me too :+1: 

Thanks for the good work @blairconrad, it wasn't exactly a fun task :wink: 

Just a small worry: when running all the tests with the ReSharper runner, some of them fail. If I run them again separately, they pass... I haven't looked into it yet, but there's probably some shared state that doesn't play well with parallelization.

adamralph:
In that case we should switch off parallelisation for all the converted projects until we can make them threadsafe - https://github.com/FakeItEasy/FakeItEasy/blob/794fe3c74bdef06647a585d6c83202a8be64a2a1/tests/FakeItEasy.Specs/Properties/XunitAssemblyInfo.cs#L1

adamralph:
In fact, although the method we've used works, this is the proper way to do it - https://github.com/NancyFx/Nancy/blob/cac60523e7d5e3663d378875355cccf608ef02eb/test/Nancy.ViewEngines.DotLiquid.Tests/Properties/AssemblyInfo.cs#L3

blairconrad:
I can switch XUnitAssemblyInfo to use `[assembly: CollectionBehavior(DisableTestParallelization = true)]` and add a similar file to all tests.
Do we expect this means we can stop setting `-maxthreads` to `1` in the rakefile?

adamralph:
Yeah, we should take that out. It's up to each project whether it's thread safe or not.

blairconrad:
Assuming you like the last commit, I'll eventually squash with the other non-cleanup commits and rebase on top of master (probably once 2.1.0 is out).

adamralph:
@blairconrad LGTM :+1: 

Squash and rebase away...

blairconrad:
Thanks! Done. The only conflict was in `ATests.cs`, where Git was confused by my having removed `OnSetup` and @adamralph having removed some test case that I forget what it was.

adamralph:
Great, @blairconrad, thanks! I think this is ready to :ship: 

@thomaslevesque - any outstanding concerns?

thomaslevesque:
> @thomaslevesque - any outstanding concerns?

Nope, looks good to me too.

adamralph:
Thanks @blairconrad. What an effort!!! :clap:

thomaslevesque:
:tada:

blairconrad:
Thanks folks. I'll make that outstanding issue now.

PR #768, opened 2016-06-19
move non-generic methods from A to Sdk.Create

### LINKED ISSUE #482: Non-generic Fake and Dummy creation

adamralph:
closes #482

### DIFF
--- a/src/FakeItEasy/A.cs
+++ b/src/FakeItEasy/A.cs
@@ -4,6 +4,7 @@ namespace FakeItEasy
     using System.Collections.Generic;
     using System.ComponentModel;
     using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
     using System.Linq.Expressions;
     using FakeItEasy.Configuration;
     using FakeItEasy.Creation;
@@ -57,7 +58,7 @@ namespace FakeItEasy
         [SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter", Justification = "Used to specify the type of fake.")]
         public static IList<T> CollectionOfFake<T>(int numberOfFakes)
         {
-            return FakeCreator.CollectionOfFake<T>(numberOfFakes);
+            return Enumerable.Range(0, numberOfFakes).Select(_ => Fake<T>()).ToList();
         }
 
         /// <summary>
@@ -70,7 +71,7 @@ namespace FakeItEasy
         [SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "This is by design when using the Expression-, Action- and Func-types.")]
         public static IList<T> CollectionOfFake<T>(int numberOfFakes, Action<IFakeOptions<T>> optionsBuilder)
         {
-            return FakeCreator.CollectionOfFake(numberOfFakes, optionsBuilder);
+            return Enumerable.Range(0, numberOfFakes).Select(_ => Fake<T>(optionsBuilder)).ToList();
         }
 
         /// <summary>
@@ -98,7 +99,7 @@ namespace FakeItEasy
         [EditorBrowsable(EditorBrowsableState.Advanced)]
         public static IList<T> CollectionOfDummy<T>(int numberOfDummies)
         {
-            return FakeCreator.CollectionOfDummy<T>(numberOfDummies);
+            return Enumerable.Range(0, numberOfDummies).Select(_ => Dummy<T>()).ToList();
         }
 
         /// <summary>
--- a/src/FakeItEasy/A.cs
+++ b/src/FakeItEasy/A.cs
@@ -49,27 +49,6 @@ namespace FakeItEasy
         }
 
         /// <summary>
-        /// Creates a fake object of the specified type.
-        /// </summary>
-        /// <param name="typeOfFake">The type of fake object to create.</param>
-        /// <returns>A fake object.</returns>
-        public static object Fake(Type typeOfFake)
-        {
-            return FakeCreator.CreateFake(typeOfFake, x => { });
-        }
-
-        /// <summary>
-        /// Creates a fake object of the specified type.
-        /// </summary>
-        /// <param name="typeOfFake">The type of fake object to create.</param>
-        /// <param name="optionsBuilder">A lambda where options for the built fake object can be specified.</param>
-        /// <returns>A fake object.</returns>
-        public static object Fake(Type typeOfFake, Action<IFakeOptions> optionsBuilder)
-        {
-            return FakeCreator.CreateFake(typeOfFake, optionsBuilder);
-        }
-
-        /// <summary>
         /// Creates a collection of fakes of the specified type.
         /// </summary>
         /// <typeparam name="T">The type of fakes to create.</typeparam>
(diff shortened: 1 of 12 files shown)


src/FakeItEasy/A.cs
```diff
@@ -14,14 +14,14 @@ namespace FakeItEasy
     [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "A", Justification = "Is spelled correctly.")]
     public static class A
     {
-        private static IFakeConfigurationManager ConfigurationManager
+        internal static IFakeCreatorFacade FakeCreator
```
thomaslevesque: Could it be injected into `Create` instead?

adamralph: It could, but then I didn't want knowledge of how the service locator works (are they singletons or not?) to have to propagate to the new type. I thought it safest just to reuse.

Incidentally, I wonder if it would probably cleaner to refactor `A` to forward calls to `Sdk.Create`? That would remove the need for the generic members on `IFakeCreatorFacade`.

thomaslevesque: > It could, but then I didn't want knowledge of how the service locator works (are they singletons or not?) to have to propagate to the new type. I thought it safest just to reuse.

Works for me

> Incidentally, I wonder if it would probably cleaner to refactor A to forward calls to Sdk.Create? That would remove the need for the generic members on IFakeCreatorFacade.

I'm not sure I like it... it's yet another indirection level, in a library that already has so many.
Anyway, if you feel it's worth it, don't let me stop you :wink:

adamralph: > It could, but then I didn't want knowledge of how the service locator works (are they singletons or not?) to have to propagate to the new type. I thought it safest just to reuse.

Scratch that. It's a property which calls the service locator every time, so no reason `Sdk.Create` can't have the same property. I'll amend.

adamralph:
I amended the commit so that `Sdk.Create` gets the `IFakeCreatorFacade` itself instead of reaching into `A` (and making `A`'s `IFakeCreatorFacade`internal).

adamralph:
I added a refactoring commit which simplifies things quite a lot.

src/FakeItEasy/A.cs
```diff
@@ -78,7 +58,7 @@ public static object Fake(Type typeOfFake, Action<IFakeOptions> optionsBuilder)
         [SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter", Justification = "Used to specify the type of fake.")]
         public static IList<T> CollectionOfFake<T>(int numberOfFakes)
         {
-            return FakeCreator.CollectionOfFake<T>(numberOfFakes);
+            return Enumerable.Range(0, numberOfFakes).Select(_ => Fake<T>()).ToList();
```
thomaslevesque: :sunglasses:

adamralph:
Whilst I'm here, shall I remove the XML docs on `IFakeCreatorFacade` and `DefaultFakeCreatorFacade`? They're adding very little indeed.

thomaslevesque:
> Whilst I'm here, shall I remove the XML docs on IFakeCreatorFacade and DefaultFakeCreatorFacade? They're adding very little indeed.

By all means! Actually, I'm wondering about the value of `IFakeCreatorFacade` itself... Why can't we use `IFakeAndDummyManager` directly?

adamralph:
OK, I'll add another commit which removes the docs.

> I'm wondering about the value of IFakeCreatorFacade itself... Why can't we use IFakeAndDummyManager directly?

My exact thoughts! I've tried removing them, and everything seems fine, but for some bizarre reason the unit test count drops from over 1,000 to 90. I think there's something really weird going on in the unit tests. More generally, I'd love to remove most of the unit tests and rely on the specs. Most of the unit tests are probably not required. I'd like to continue killing `IFakeCreatorFacade` but I think it makes sense to leave that for a later PR.

src/FakeItEasy/Sdk/Create.cs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// <summary>
+    /// Provides methods for generating fake objects.
+    /// </summary>
+    [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = "A", Justification = "Is spelled correctly.")]
+    public static class Create
+    {
+        private static IFakeCreatorFacade FakeCreator
+        {
+            get { return ServiceLocator.Current.Resolve<IFakeCreatorFacade>(); }
```
thomaslevesque: I just realized that `IFakeCreatorFacade` is not registered as a singleton... and most of its dependencies are not singletons either. This means that everytime we call `A.Something`, we create new instances of all these things. I wonder if this could be improved...

thomaslevesque: Making it a singleton doesn't seem to break anything. The same is true for `IFakeAndDummyManager` (it just breaks `ServiceLocatorTests`, which tests specifically that we get a new instance every time)

adamralph: Yes, this is probably worth pursuing too.

adamralph:
Docs removed.

src/FakeItEasy/Sdk/Create.cs
thomaslevesque: Probably better to treat it as a separate issue, though

adamralph: Yep, absolutely.

thomaslevesque:
LGTM :+1:

But I think we should merge #767 first, because it will be harder to rebase...

adamralph:
The only conflict should be in `ATest.cs`, so it should be a trivial rebase either way, but sure, let's get #767 in first. I'd rather not add any extra burden on @blairconrad on top of the already Herculean effort he's put in there :wink:

blairconrad:
Pfah, by all means let's have #767 wait. This is the PR that adds actual value. I don't mind the rebase, and I may not get the "don't run parallel" stuff ready 'til my morning.

blairconrad:
Generally looks good. Thanks for your work!
I have two things, though.

1. docs?
2. Do we want to move the specs that were affected? I realize they deal with creation options or Dummy creation, but they focus on a whole separate section of the API. I don't feel particularly strongly, so if you want to keep as is, I am content.

adamralph:
1. Of course. One day I might learn to remember this myself. I'll add some docs and push another commit.
1. The specs are named by feature rather than by implementation, so I'm not sure I can think of a much better structure than we currently have. I.e. we have generic and non-generic creation, so we have `GenericCreationSpecs` and `NonGenericCreationSpecs`. The former satisfied by `A` and the latter satisfied by `Sdk.Create` but this seems to be coincidental to our naming of our specs. It would probably make sense to separate dummy and fake specs but that's outside the scope of this PR.

adamralph:
I've inserted a commit which changes the docs. If you're happy with the changeset, I propose squashing that new docs commit into the previous commit.

thomaslevesque:
@adamralph didn't you want to document these methods in a separate "SDK" doc page?

adamralph:
Given that the SDK is so small, I think it's OK to sit where it is for now. If the SDK gets bigger then we might want to consider creating an SDK section.

adamralph:
Or not. If we feel that it's better to have an SDK section now I don't mind creating one.

thomaslevesque:
I don't mind if they're in the same section; it makes them more discoverable. The description makes it clear that it's not the "normal" way of creating fakes, so I don't think there's any risk of confusion for the user.

blairconrad:
I like the docs as they are. Having them live with their functional siblings and indicating that they are not the normal way to work would've been my preference.
As far as I'm concerned, squash away and we can merge.

adamralph:
Squashed.

src/FakeItEasy/Sdk/Create.cs
blairconrad: Good observation, @thomaslevesque. I created #769.

thomaslevesque:
Since @blairconrad doesn't seem to mind, I'll merge this now.

thomaslevesque:
Thanks @adamralph !

blairconrad:
Indeed, thanks!

PR #770, opened 2016-06-20
Fix sparse SDK namespace warning

blairconrad:


adamralph:
Good spot. I'd probably prefer to switch this rule off globally. IMHO it's
utter nonsense.

blairconrad:
> I'd probably prefer to switch this rule off globally.

No objection. Next time it comes up? ;)

blairconrad:
> Good spot.

Actually, no credit for me. It's all about the checklist.

thomaslevesque:
> I'd probably prefer to switch this rule off globally.

:+1:

blairconrad:
Okay. Working on it.

blairconrad:
Thanks for the suggestion for the global switch. And merge.

samples/FakeItEasy.Examples.ruleset
```diff
@@ -15,4 +14,4 @@
     <Rule Id="CA2201" Action="None" />
     <Rule Id="CA2210" Action="None" />
   </Rules>
-</RuleSet>
\ No newline at end of file
+</RuleSet>
```
adamralph: I presume VS is removing the EOL here. Do we need a `.editorconfig` tweak?

blairconrad: I think this is the same effect that @thomaslevesque mentioned recently somewhere. The editorconfig only applies when VS is used as an editor. I was working in a funny GUI.

You're suggesting touching the file to say that it's okay to have EOLs in rulesets?

adamralph: AFAIK editorconfig only applies when you're manipulating the raw text file itself (in VS or any other editor which supports editorconfig). If you use the "ruleset GUI" in VS then it bypasses editorconfig since, I'm guessing, the editorconfig extension doesn't get the required (saving a text file) events. It's probably easier to alter our editorconfig settings to stop trying to add EOL's at the end of file types that battle against the various GUI's in VS for various file types. I believe we've already done this for csproj and vbproj.

blairconrad: Whoops. I misread your comment (and the diff). I thought you were upset that the CRLF went away, not that they were added. Sorry.
Yes, adding an exception seems like a good idea.
Although it might not have made a difference in this case, as I can't promise I used VS as an editor. May have been emacs. I honestly do not remember.

adamralph: https://github.com/editorconfig/editorconfig-emacs#readme :wink:

blairconrad: Yeah, I saw it.

thomaslevesque: > May have been emacs.

What?! You're not using vim?
(I didn't think it was possible for Windows users to have this debate, but apparently it is  :laughing:)

blairconrad: Indeed. A treasured Day Job colleague and I have had this tussle in the past. Some day he'll see the light.

PR #771, opened 2016-06-20
set version to 2.2.0 for next release

blairconrad:
preparing for 2.2.0

### DIFF
--- a/src/CommonAssemblyInfo.cs
+++ b/src/CommonAssemblyInfo.cs
@@ -3,9 +3,9 @@ using System.Reflection;
 using System.Resources;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyVersion("2.1.0")]
-[assembly: AssemblyFileVersion("2.1.0")]
-[assembly: AssemblyInformationalVersion("2.1.0")]
+[assembly: AssemblyVersion("2.2.0")]
+[assembly: AssemblyFileVersion("2.2.0")]
+[assembly: AssemblyInformationalVersion("2.2.0")]
 
 [assembly: AssemblyCompany("Patrik Hägne")]
 [assembly: AssemblyCopyright("Copyright (c) FakeItEasy contributors. (fakeiteasyfx@gmail.com)")]


src/CommonAssemblyInfo.cs
```diff
@@ -3,9 +3,9 @@
 using System.Resources;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyVersion("2.1.0")]
-[assembly: AssemblyFileVersion("2.1.0")]
-[assembly: AssemblyInformationalVersion("2.1.0")]
+[assembly: AssemblyVersion("2.2.0")]
+[assembly: AssemblyFileVersion("2.2.0")]
+[assembly: AssemblyInformationalVersion("2.2.0")]
```
thomaslevesque: How would you feel about introducing a `VersionInfo` class with a const `Version` field, to have the version in one place instead of 3?

blairconrad: I don't mind so much. Just to make the PR review easier?

thomaslevesque: Mostly to make it less error-prone; if the version is only in one place, you can't change it somewhere and forget to change it somewhere else...

thomaslevesque: This is how I usually do it: https://github.com/thomaslevesque/Linq.Extras/blob/master/Linq.Extras/Properties/AssemblyInfo.cs#L16-L25

blairconrad: The technique is sound, but if we're only ever updating the version via 
`rake next_version[new_version]`, I don't think we'll run into problems.
And I think we'd lose the ability to leverage albacore for this (which we may eventually anyhow, if/when we drop rake).
Still, if you're keen on it, I see no other downside.
I'd prefer to make it its own issue, though, rather than doing as part of the last step of closing off 2.1.0.

thomaslevesque: > The technique is sound, but if we're only ever updating the version via
> rake next_version[new_version], I don't think we'll run into problems.

True

> And I think we'd lose the ability to leverage albacore for this (which we may eventually anyhow, if/when we drop rake).

Good point, forgot about that. In large solutions, when I need to read/update the version as part of the build process, I do this:
- create a `version.txt` file that contains nothing but the version (to make it easier to manipulate without having to parse or generate C# code)
- create a T4 template that generates `VersionInfo.cs` from `version.tt`
- include `VersionInfo.cs` in every project that must share the same version number

> I'd prefer to make it its own issue, though, rather than doing as part of the last step of closing off 2.1.0.

Sure

thomaslevesque:
Thanks @blairconrad 

As discussed, let's leave the AssemblyInfo/version issue out of this for now.

blairconrad:
Thanks for the merge!

src/CommonAssemblyInfo.cs
blairconrad: Your "generate a VersionInfo.cs file" technique is another one I'm familiar with. We use it at the Day Job.

I'm indifferent to the mechanism we use for setting the version to tell the truth. I _do_ enjoy running `rake next_version[new_version]` to do all the half a dozen or so little steps we'd otherwise have to do as humans, but I don't really care how it accomplishes things under the hood (capot/bonnet). If you want a change, go for it

thomaslevesque: > If you want a change, go for it!

No, it's OK... I was worried about the fact that the version was specified in multiple places, but since it's generated by the build script, it doesn't really matter

PR #775, opened 2016-06-20
Fix concurrency issue in CastleDynamicProxyGenerator

### LINKED ISSUE #773: FakeItEasy.Core.FakeCreationException while creating (many) fakes
We are using FakeItEasy 2.0.0 and are creating thousands of fake objects during our unit-tests. Unfortunately we get a `FakeItEasy.Core.FakeCreationException` occasionally in different test-classes for different fake-objects on our CI:

```
MESSAGE:
SetUp : FakeItEasy.Core.FakeCreationException : 
  Failed to create fake of type "IConditionElementCollector".

  Below is a list of reasons for failure per attempted constructor:
    No constructor arguments failed:
      No usable default constructor was found on the type IConditionElementCollector.
      An exception of type System.InvalidOperationException was caught during this call. Its message was:
      Collection was modified; enumeration operation may not execute.
         bei System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
         bei System.Collections.Generic.List`1.Enumerator.MoveNextRare()
         bei System.Collections.Generic.List`1.Enumerator.MoveNext()
         bei Castle.DynamicProxy.Generators.Emitters.AbstractTypeEmitter.AddCustomAttributes(ProxyGenerationOptions proxyGenerationOptions)
         bei Castle.DynamicProxy.Generators.BaseProxyGenerator.CreateTypeAttributes(ClassEmitter emitter)
         bei Castle.DynamicProxy.Generators.ClassProxyGenerator.GenerateType(String name, Type[] interfaces, INamingScope namingScope)
         bei Castle.DynamicProxy.Generators.ClassProxyGenerator.<>c__DisplayClass1.<GenerateCode>b__0(String n, INamingScope s)
         bei Castle.DynamicProxy.Generators.BaseProxyGenerator.ObtainProxyType(CacheKey cacheKey, Func`3 factory)
         bei Castle.DynamicProxy.Generators.ClassProxyGenerator.GenerateCode(Type[] interfaces, ProxyGenerationOptions options)
         bei Castle.DynamicProxy.DefaultProxyBuilder.CreateClassProxyType(Type classToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options)
         bei Castle.DynamicProxy.ProxyGenerator.CreateClassProxy(Type classToProxy, Type[] additionalInterfacesToProxy, ProxyGenerationOptions options, Object[] constructorArguments, IInterceptor[] interceptors)
         bei FakeItEasy.Creation.CastleDynamicProxy.CastleDynamicProxyGenerator.GenerateClassProxy(Type typeOfProxy, IEnumerable`1 argumentsForConstructor, IInterceptor interceptor, IEnumerable`1 allInterfacesToImplement)
         bei FakeItEasy.Creation.CastleDynamicProxy.CastleDynamicProxyGenerator.DoGenerateProxy(Type typeOfProxy, IEnumerable`1 additionalInterfacesToImplement, IEnumerable`1 argumentsForConstructor, IInterceptor interceptor)
         bei FakeItEasy.Creation.CastleDynamicProxy.CastleDynamicProxyGenerator.CreateProxyGeneratorResult(Type typeOfProxy, IEnumerable`1 additionalInterfacesToImplement, IEnumerable`1 argumentsForConstructor, IFakeCallProcessorProvider fakeCallProcessorProvider)


+++++++++++++++++++
STACK TRACE:
bei FakeItEasy.Core.DefaultExceptionThrower.ThrowFailedToGenerateProxyWithResolvedConstructors(Type typeOfFake, String reasonForFailureOfUnspecifiedConstructor, IEnumerable`1 resolvedConstructors)
bei FakeItEasy.Creation.FakeObjectCreator.TryCreateFakeWithDummyArgumentsForConstructor(Type typeOfFake, IProxyOptions proxyOptions, IDummyValueCreationSession session, String failReasonForDefaultConstructor, Boolean throwOnFailure)
bei FakeItEasy.Creation.FakeObjectCreator.CreateFake(Type typeOfFake, IProxyOptions proxyOptions, IDummyValueCreationSession session, Boolean throwOnFailure)
bei FakeItEasy.Creation.DefaultFakeAndDummyManager.CreateFake(Type typeOfFake, Action`1 optionsBuilder)
bei FakeItEasy.Creation.DefaultFakeCreatorFacade.CreateFake[T](Action`1 optionsBuilder)
bei FakeItEasy.A.Fake[T]()
bei VisibilityPropertyHandlerTest.Setup()
```

Any idea, how to fix this Heisenbug? The tests are running sequentially on the CI (no tests are running concurrently). It happens about once every week at different locations, usually where a lot of fakes are created rapidly: in the unit-tests.

thomaslevesque:
Fixes #773


blairconrad:
I don't mean to quibble, but we don't _know_ that this fixes the issue that @apacha sees. I believe it might, and it's a good step, but maybe we should instead claim that we're no longer using a static ProxyGenerationOptions?

blairconrad:
Also, bigger change than I anticipated. I won't get to for several hours. Sorry.

thomaslevesque:
> maybe we should instead claim that we're no longer using a static ProxyGenerationOptions?

True. I'll change the description.

> Also, bigger change than I anticipated. I won't get to for several hours. Sorry.

No worry, it can wait until then.

It's not as big as it seems; I'll put a summary in the PR description.

blairconrad:
>  It's not as big as it seems; I'll put a summary in the PR description.

Oh, I'll never turn down a summary, but I'm not super scared by it. It's just that there's the "2 minute review while compiling" and then there's the "wait until you're not at work review".

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
```diff
@@ -67,14 +89,11 @@ public ProxyGeneratorResult GenerateProxy(Type typeOfProxy, IEnumerable<Type> ad
                 return new ProxyGeneratorResult(DynamicProxyResources.ProxyIsSealedTypeMessage.FormatInvariant(typeOfProxy));
             }
 
+            Guard.AgainstNull(additionalInterfacesToImplement, "additionalInterfacesToImplement");
             GuardAgainstConstructorArgumentsForInterfaceType(typeOfProxy, argumentsForConstructor);
+            Guard.AgainstNull(fakeCallProcessorProvider, "fakeCallProcessorProvider");
```
blairconrad: The order of the Guard/GuardAgainstConstructorArgumentsForInterfaceType/Guard is a little peculiar.
Also, since we're not dereferencing additionalInterfacesToImplement or fakeCallProcessorProvider, I'd've expected to see the Guards deeper in. 
Functionally, none of this makes a difference

thomaslevesque: It's the same order as the parameters, but I can change it. Anyway, seeing it again, I find it a bit strange too... I'll make another pass at it, and try to move the guard clauses to the method where the parameters are dereferenced.

adamralph:
> maybe we should instead claim that we're no longer using a static ProxyGenerationOptions?

:+1: Given the non-deterministic nature of concurrency bugs, perhaps we should re-title this "Improve thread safety by removing shared ProxyGenerationOptions" and label it as an independent `enhancement`. After release, and given some time, we can see if #773 is still an issue, and if not, close it as `invalid`, i.e. "not a bug". It's weird, but then so is the non-deterministic nature of concurrency bugs.

The alternative is to say that this PR "fixes #773" if we are reasonably confident it does, and then, after some time, if there are still problems, we create a new bug issue to be fixed and attempt it to fix it again.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
adamralph: Having guard clauses for arguments half way through methods is also a massive code smell.

blairconrad:
> The alternative is to say that this PR "fixes #773" if we are reasonably confident it does, and then, after some time, if there are still problems, we create a new bug issue to be fixed and attempt it to fix it again.

I did not think that this would be my preference. Only once I read it, it is. I think @thomaslevesque nailed the analysis, and this is the cause of the problem. @apacha's [comment](https://github.com/FakeItEasy/FakeItEasy/issues/773#issuecomment-227170076) that some of his tests spawn background tasks only increases my confidence.

Sorry to flip-flop on this, but I do believe that this will resolve the problem.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
blairconrad: It looked weird to me too, but I felt better when I noticed that we were short-circuiting above. I guess pushing the guards into `CreateProxyGeneratorResult` will help with this as well.

adamralph:
OK, I'm also convinced.

I've taken the liberty to change the link to #773 in the description to "Fixes #773" and moved #773 into progress.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
thomaslevesque: > Having guard clauses for arguments half way through methods is also a massive code smell.

The `GuardAgainstConstructorArgumentsForInterfaceType` was already in the middle, it's why I put the others there. I think it was an optimization to bail early if the type is a value type or is sealed. But it's not a very useful optimization, so I'll just put everything at the start of the method.

thomaslevesque:
I restored the guard clauses to their previous state (except a redundant one that I removed). Not sure what I had in mind when I changed them...
The only other change is that I inlined `GetConstructorArgumentsArray`, using the null propagation operator.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
```diff
@@ -40,18 +40,42 @@ public CastleDynamicProxyGenerator(CastleDynamicProxyInterceptionValidator inter
...
-        public ProxyGeneratorResult GenerateProxy(Type typeOfProxy, IEnumerable<Type> additionalInterfacesToImplement, IEnumerable<object> argumentsForConstructor, IFakeCallProcessorProvider fakeCallProcessorProvider)
+        public bool MethodCanBeInterceptedOnInstance(MethodInfo method, object callTarget, out string failReason)
+        {
+            return this.interceptionValidator.MethodCanBeInterceptedOnInstance(method, callTarget, out failReason);
+        }
+
+        private static ProxyGenerationOptions CreateProxyGenerationOptions()
+        {
+            return new ProxyGenerationOptions(ProxyGenerationHook);
+        }
```
adamralph: Can we inline this?

thomaslevesque: We could, but why? I created a method because it was used in two places; if, for some reason or another, we need to make changes to the `ProxyGenerationOptions`, there will be only one place to change.

adamralph: YAGNI though. Do we really need a private method around a call to a constructor?

adamralph: Anyway, it's a minor point. I won't insist on it.

adamralph:
Thanks @thomaslevesque!

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
thomaslevesque: Well, unless it really bothers you, I'd rather keep it. Looks cleaner to me. And it will be inlined at runtime anyway ;)

thomaslevesque:
Thanks for the merge!

Issue #571, opened 2016-01-09
Remove all write-only properties


blairconrad:
Per @thomaslevesque's remark: https://github.com/FakeItEasy/FakeItEasy/pull/529#discussion_r49126608.

While we're at it, I think it has some friends that are in a similar state. See the [FakeManager constructor](https://github.com/FakeItEasy/FakeItEasy/blob/bd57a05c0de5ff37609230fd1b8bc4ec45bc9d4d/Source/FakeItEasy/Core/FakeManager.cs#L43).

adamralph:
I think that we should sweep up all instances of this anti-pattern in one go. Shall we change the title to

> Remove all write-only properties

?

blairconrad:
Some may be harder than the family that we've identified, but I'm all for taking a crack at it.

blairconrad:
Good thinking.

adamralph:
If any of them turn out to be public, I think we should remove the `refactoring` label and add the `feature` and `breaking` labels, and probably get it into 2.0.

blairconrad:
Cool. So... how does one _find_ write-only properties? Anything easier than writing code to reflect over all the members of all the types and examine access levels?
Regardless, we should do all the libraries, not just the .Net 4.0 like I did when looking for unused types…

blairconrad:
Adding to 2.0 for now, since we'll need to do the analysis before deciding whether it's in 2.0…

thomaslevesque:
> If any of them turn out to be public, I think we should remove the refactoring label and add the feature and breaking labels, and probably get it into 2.0.

If any of them are public, they should probably be treated in priority. Private usages of this pattern are not a really big issue, they can be fixed at any time

> how does one find write-only properties

if they're auto-properties, it's pretty easy: just search for `private get`

adamralph:
> Adding to 2.0 for now, since we'll need to do the analysis before deciding whether it's in 2.0…

:+1: good call

blairconrad:
Because I love this sort of thing, I've been poking. I used ReSharper to find property bits that could be made private, as a first step. Because why purge the write-only properties and then find out there are more?I found some gets that could be made private while I was at it. Would you rather I left them for a separate issue?
(Also some properties that can just be removed altogether.)

thomaslevesque:
> Would you rather I left them for a separate issue?

I don't think it needs to be done in a separate issue (perhaps just a separate commit)

adamralph:
[...]

PR #779, opened 2016-06-21
remove write only properties

adamralph:
connects to #571

### DIFF
--- a/src/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
@@ -10,7 +10,12 @@ namespace FakeItEasy.Core
         private class AutoFakePropertyRule
             : IFakeObjectCallRule
         {
-            public FakeManager FakeManager { private get; set; }
+            private readonly FakeManager fakeManager;
+
+            public AutoFakePropertyRule(FakeManager fakeManager)
+            {
+                this.fakeManager = fakeManager;
+            }
 
             public int? NumberOfTimesToCall
             {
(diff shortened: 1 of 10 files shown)


adamralph:
I'm not sure if I've captured all the write only properties. I just searched for "private get".

There may well be more which don't have a getter defined at all.

src/FakeItEasy/Core/FakeManager.AutoFakePropertyRule.cs
```diff
@@ -30,15 +35,15 @@ public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
 
                 var newRule = new CallRuleMetadata
                                   {
-                                      Rule = new PropertyBehaviorRule(fakeObjectCall.Method, FakeManager)
```
blairconrad: I'm amazed there was no `this.` on here.

src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
```diff
@@ -10,7 +10,12 @@ public partial class FakeManager
         private class PropertySetterRule
             : IFakeObjectCallRule
         {
-            public FakeManager FakeManager { get; set; }
+            private readonly FakeManager fakeManager;
```
blairconrad: This appears to have been a read/write property. I don't it going away, though.

blairconrad:
> I'd had half a plan to reflect through the assembly at some point, but this is a great start (and maybe finish), @adamralph. Thanks.

src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
adamralph: Yes, I think in this case it was public readable by accident, since the getter was never used outside the class. The class was written in the same pattern as the other Rule classes.

PR #617, opened 2016-02-25
[WIP] Porting to .Net Core

jeremymeng:
This is for #531.  

----

**Edited: 2016-7-20 by @jeremymeng**

In current code the default AssemblyLoadContext is used instead of custom one.

----

**Update: 2016-7-18 (UTC) by @blairconrad**

We've for some time wanted to merge the coreclr branch to master and disable the netcore bits via a feature toggle, would be the inclusion/exclusion of that build from the nupkg. 

I'm stealing the top of this "description" to list what I see as next steps that will achieve that goal as quickly as posssble. Note that these steps will not result in a releasable FakeItEasy with netcore support, but will make it so we're back to one active branch (master).

The primary goal is to preserve the current beaviour of the .NET 4.0 library while allowing use to finish our netcore-related work without interfering with .NET 4.0-targeted improvements.

Here's what I think should do before merging:

- [x] revert the 4.0 build to rely on Castle.Core 3.3.3 (I've pushed aa9f77e9187232851ca10d11d0f42412a644bb08 to my coreclr branch already - if it looks good, and @jeremymeng is willing, it can be cherry-picked)
- [x] update the Self-Initialized Fakes behaviour:
  - revert the 4.0 project's implementation to match the 2.2.0 release and
  - use `#if`s and updated `.csproj` and `projecxt.json` files to exclude the Self-Initialized Fakes from the netcore project 
- split netcore and "regular" nuspecs
  - [x] remove the 4.5 targetFramework from the nuspec(pushed c95ba3ad187a17cf1fabaa6b9efb47c57eff729d to my coreclr branch)
  - [x] remove the netcore build from the official nupkg  and provide a second nuspec so we can continue to publish pre-release netcore packages from @jeremymeng's AppVeyor. ;-) (pushed ab528fd8e0c2ea5a5776fee9511d0d903f10a97b to my coreclr branch)

- [ ] review usage of ~~custom~~ AssemblyLoadContext (I don't entirely understand this - I stole it from below)
- [ ] General review - there are some typos and grammar that need fixing in comments, at least. I made comments in the PR already. Probably should've just pushed a commit. I may do that
- [x] Fix `approve` test - we have an `[assembly: InternalsVisibleTo("FakeItEasy.Tests.dotnet, …]` that is messing up the public API. One option is to adjust `ApiApproval.ApproveApi.approved.txt`, but I'd prefer to find a way to call the test assembly `FakeItEasy.Tests`.

[...]

PR #780, opened 2016-06-21
Refactor argument constraint tests

thomaslevesque:
to avoid the use of custom ReflectedMethodData attribute, because it can't be ported to .NET Core (as discussed in https://github.com/FakeItEasy/FakeItEasy/pull/617#issuecomment-227571784)


blairconrad:
Things look good, @thomaslevesque. I wouldn't mind a little time to investigate another approach, though. If you don't mind.

thomaslevesque:
> I wouldn't mind a little time to investigate another approach, though.

I expected as much. It works, but it's certainly not an ideal solution...

> If you don't mind.

Not at all :wink:

blairconrad:
Hmm… my initial idea would work pretty easily (I think) were it not for the subtypes declaring additional methods. I'd thought to provide instances of the subtypes (although they wouldn't have to be subtypes) that would act as test contexts. Here's a sample. I just converted 2 of the leaf classes to give an idea. Probably your method is better. I made #781 just to give an idea. It's by no means complete.

adamralph:
If the other methods we're considering prove too cumbersome, I'd be happy quite happy with this approach. It's verbose, but dead simple. KISS.

blairconrad:
> If the other methods we're considering prove too cumbersome, I'd be quite happy with this approach. It's verbose, but dead simple. KISS.

I feel the same way. There are a few things to remember when creating a new case (because the base tests aren't abstract, we aren't reminded to override them), but probably someone will just copy an existing class anyhow, so it'll be okay. I'd just wanted to try a different tack.

thomaslevesque:
I find @blairconrad's solution more elegant, and probably less brittle. What worries me a little is how the tests will appear: instead of each test class appearing to have its own tests, only the base class will have tests with many test cases.

adamralph:
I find this solution simple, and the other one complex.

I'm not sure why this solution would be more brittle. Ultimately, we're
trying to save ourselves the trouble of writing each test explicitly, so I
feel like an explicit declaration of each test, internally delegating the
heavy lifting to some thing common is more natural and easier to grok.

thomaslevesque:
> I'm not sure why this solution would be more brittle.

Well, not very brittle... it's just that we could forget to override the test methods and/or apply the attributes in the derived classes.

blairconrad:
I've no objection to merging @thomaslevesque's solution. I just thought it would be a fun exercise to see if approaching the problem from another angle would yield something suitable and maybe with a little less ceremony in the tests, since  @adamralph's [initial comment](https://github.com/FakeItEasy/FakeItEasy/pull/617#issuecomment-227572383) mentioned the verbosity.
I had fun, and appreciate you guys holding off on the merge while I played.

adamralph:
> we could forget to override the test methods and/or apply the attributes in the derived classes

True, but we could equally forget to write tests for certain things, or forget attributes on any other test method. These tests are really no different to other tests in that respect, it's just that we have shortcuts under the hood for implementing them.

adamralph:
Thanks @thomaslevesque! For the changes, and for letting us persuade you that your approach is best :-D

PR #781, opened 2016-06-22
[DO NOT MERGE] an alternative way to build a test hierarchy - test contexts

blairconrad:


tests/FakeItEasy.Tests/ArgumentConstraintTestBase.cs
```diff
@@ -1,43 +1,48 @@
...
-        internal IArgumentConstraint ConstraintField { get; set; }
-
-        protected abstract string ExpectedDescription { get; }
-
-        private IArgumentConstraint Constraint => this.ConstraintField;
+        public static Type[][] TestContextTypes =
+        {
+            new[] { typeof(CollectionContainsTests)},
+            new[] { typeof(CollectionIsEmptyTests)},
+        };
```
blairconrad: If we have reflection, we could instead find subclasses of `ArgumentConstraintTestBase<IEnumerable<object>>` or whatever base test context we want to provide.

tests/FakeItEasy.Tests/FakeItEasy.Tests.csproj
```diff
@@ -79,12 +79,6 @@
     <Compile Include="ActionOverWriterValueFormatter.cs" />
     <Compile Include="ArgumentConstraintManagerExtensions\CollectionContainsTests.cs" />
     <Compile Include="ArgumentConstraintManagerExtensions\CollectionIsEmptyTests.cs" />
-    <Compile Include="ArgumentConstraintManagerExtensions\DerivedTypeArgumentTests.cs" />
-    <Compile Include="ArgumentConstraintManagerExtensions\SameAsConstraintTests.cs" />
-    <Compile Include="ArgumentConstraintManagerExtensions\EqualToConstraintTests.cs" />
-    <Compile Include="ArgumentConstraintManagerExtensions\IsSameSequenceAsTests.cs" />
-    <Compile Include="ArgumentConstraintManagerExtensions\NullCheckedMatchesConstraint.cs" />
-    <Compile Include="ArgumentConstraintManagerExtensions\StringEndsWithTests.cs" />
```
blairconrad: I just didn't want to take the time to convert everything.

tests/FakeItEasy.Tests/ArgumentConstraintTestBase.of.T.cs
```diff
@@ -14,5 +18,7 @@ protected ArgumentConstraintTestBase()
         }
 
         protected abstract void CreateConstraint(IArgumentConstraintManager<T> scope);
+        public abstract IEnumerable<object[]> ValidValues();
+        public abstract IEnumerable<object[]> InvalidValues();
     }
 }
```
blairconrad: I used ` ArgumentConstraintTestBase<T>` as the test context base, because it was convenient. If this were a serious proposal, we'd probably pick something better.

blairconrad:
You can see the test results near http://teamcity.codebetter.com/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=219880#_focus=1541

thomaslevesque:
Interesting approach @blairconrad. I guess it could work. However I wonder if it couldn't be a little simpler: the test case source could return instances of the "test contexts", rather than their types. This way you could just use `MemberData`, instead of a custom attribute.

adamralph:
That is what I was thinking. I single test class, and then `MemberData` used to switch between two variants of *something*. This could be some type that wraps around the two SUT's, or even just a flag to pass to some static helper which alternates between the two SUT's based on the flag.

blairconrad:
@thomaslevesque, that's a good idea. I was stuck on the idea of having to work with a type at first because I was thinking of needing to feed the types into an attribute constructor, which I originally did. I ended up amending the approach to use the method to generate the test case sources, but didn't go further.

@adamralph, I don't really understand the "two variants" or "two SUTs" comments. I have 2 classes in my example just because I was rushing and didn't have the time to convert all the subclasses. Is that what you're talking about?

blairconrad:
I've pushed a slightly more complete version of the spike. The downside is that `Constraint_should_provide_correct_description`, which had been overridden in `AggregateArgumentConstraintTests` and `EqualityArgumentConstraintTests`, fail on `CommonArgumentConstraintTests`, because the common implementation of the test is also run.

blairconrad:
I've worked around the downside with a hack: don't supply a set of test parameters if the test case overrides the common test method implementation.

So, the code needs a horrible polish, but pending a successful build, I think it's actually functional. But perhaps too convoluted.

adamralph:
Closing in favour of #780

PR #783, opened 2016-06-26
Better message non fake

### LINKED ISSUE #682: Better exception message when configuring/asserting a non-fake object
*Extracted from #681*

Currently:

```
    - Sealed methods can not be intercepted.
```

better would be something like

```
    - The object is not a fake.
```

blairconrad:
Fixes #682.


blairconrad:
Adding a complaint when the configured object isn't a fake. The
straightforward fix would've resulted in extra tmes parsing the call
configuration expression, so I added a bit of refactoring first.

* 2c5bd2a9cea0f486d30803b1946ca55b30a4e60e - Remove IExpressionParser

  `GetFakeManagerCallIsMadeOn` was only used in the one spot, and the
  extra type added confusion, I thought.

* 686f8d1c2eaf021e2cb3c9ebb8ca34f398188e7a - Evaluate call target expression at most once

   Why evaluate an expression more than once?

* 8b707330a20bfc33705799697dacc1998aabaf90 - Remove ParsedCallExpression.Arguments

  Nobody used it.

* 87beb3e88a90942bcd4a8289a4313637dc1b7ce5 - Add specs to reinforce too-granular unit tests

  The next refactoring will make some (to my mind) overly restrictive
  tests awkward, so let's cover the funcitonality first.

* 692797761ba95a6821418f726f4e01a9c9b0c2be - Parse expressions earlier and only once

  The major refactoring here. We were always parsing the expression at
  some point, so let's do it up front. Then we can operate on the
  ParsedCallExpression without a need to reparse.

  The PropertySetterConfiguration class still takes a getter expression

* 39fc03929c2e1ae9cf83cfadc82b1384850bd8c5 - Make some methods static

  We weren't using `this` anymore after the previous commit, but the
  StyleCop rules make us reorder when methods become static, and it
  really messed up the diff, so I kept the change separate. Can always
  squash (this and anything else) later.

* e5d87bc38a4862497df18e800e1a12b489467e8d - Build property setter from parsed call expression

  It still creates a setter expression and parses it. I've half a mind
  to revisit that later, to see if we can't make a new
  ParsedCallExpression from the one that descibes the getter, but I
  think it can wait 'til later.
  
  This is the last commit in the "refactoring" phase. If you'd like, I
  can make all the ones up 'til now their own PR.
  
* 263a009b632db8abd0873f542a5c577bca9cf294 - Reword specs as Given/When/Then
 
  I'll be adding to these specs in a second, so they might as well use our preferred language.

* 8234ae7af9064e93f52bc864b8cf94f54cd79f60 - Add specs showing bad message when configuring unconfigurable methods

  Some of these fail at this point.
  
* 916fd41594d3dd1cb9bdd4c5164c42df5134e124 - Complain early when configuring a non-fake

  The actual fix.
   
* 3b164f97b641986a80968512b797f7c303c83110 - Clean up after complaining early when configuring a non-fake

  My common post-fix cleanup on touched files. Aside from the usual
  (Re#er warnings, mostly), I spotted a couple of set-only
  properties. And updated some `Guard` calls to use `nameof` to get
  the parameter name.

thomaslevesque:
Thanks @blairconrad! So many changes for such a small issue :wink:

> Why evaluate an expression more than once?

Because the result might change between the calls? This prints 0, then 42:

    int x = 0;
    Expression<Func<int>> expr = () => x;
    var func = expr.Compile();
    Console.WriteLine(func());
    x = 42;
    Console.WriteLine(func());

I'm not sure it matters in our case, though... If a call spec refers to a variable, and the value of the variable changes, we already don't take it into account. In other words, changing the value of `x` in this example has no effect:

    var foo = A.Fake<IFoo>();
    int x = 0;
    A.CallTo(() => foo.Bar(x)).Returns(42);
    Console.WriteLine(foo.Bar(0)); // 42
    Console.WriteLine(foo.Bar(1)); // 0
    x = 1;
    Console.WriteLine(foo.Bar(0)); // still 42
    Console.WriteLine(foo.Bar(1)); // still 0

I'll try to review the changes tonight, no time right now.

blairconrad:
> So many changes for such a small issue

Yeah, I felt kind of bad about that. I am certainly happy to squash. I figure it's easier to combine than to re-extract.

>> Why evaluate an expression more than once?
> Because the result might change between the calls? This prints 0, then 42:
> …
> I'm not sure it matters in our case, though... If a call spec refers to a variable, and the value of the variable changes, we already don't take it into account.

Yes, exactly. I should've been more explicit: "why evaluate one of these expressions more than once?".

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -61,11 +62,14 @@ public IAnyCallConfigurationWithNoReturnTypeSpecified CallTo(object fakeObject)
...
+            var parsedCallExpression = this.callExpressionParser.Parse(propertySpecification);
+            this.AssertThatMemberCanBeIntercepted(parsedCallExpression);
 
-            var setterExpression = BuildSetterFromGetter(propertySpecification);
+            var setterExpression = BuildSetterFromGetter(parsedCallExpression, propertySpecification);
 
-            return new PropertySetterConfiguration<TValue>(setterExpression, this.CreateVoidArgumentValidationConfiguration);
+            return new PropertySetterConfiguration<TValue>(
+                setterExpression,
+                lambda => this.CreateVoidArgumentValidationConfiguration(this.callExpressionParser.Parse(lambda)));
```
thomaslevesque: It took me a moment to understand this change... Sometimes I wish C# had function composition like F#... then we could do `this.callExpressionParser.Parse >> this.CreateVoidArgumentValidationConfiguration` instead of having to use a lambda.

tests/FakeItEasy.Tests/CompletedFakeObjectCallExtensionsTests.cs
```diff
@@ -1,62 +0,0 @@
...
-    using System.Collections.Generic;
-    using System.Linq;
-    using System.Linq.Expressions;
-    using FakeItEasy.Core;
-    using FakeItEasy.Expressions;
-    using FakeItEasy.Tests.TestHelpers;
-    using FluentAssertions;
-    using Xunit;
-
-    public class CompletedFakeObjectCallExtensionsTests : ConfigurableServiceLocatorTestBase
```
thomaslevesque: Why remove these tests? Couldn't they be updated to reflect the new signature of `CreateCallMatcher`?

thomaslevesque: (actually, forget it... it's covered by the specs added in the previous commit)

tests/FakeItEasy.Tests/Configuration/StartConfigurationTests.cs
```diff
@@ -1,223 +0,0 @@
...
-    using System.Linq.Expressions;
-    using FakeItEasy.Configuration;
-    using FakeItEasy.Core;
-    using FakeItEasy.Expressions;
-    using FakeItEasy.Tests;
-    using FakeItEasy.Tests.TestHelpers;
-    using FluentAssertions;
-    using Xunit;
-
-    public class StartConfigurationTests
```
thomaslevesque: I assume you removed this because it's covered everywhere else?

thomaslevesque: Ah, I see now; it's covered in `UnnaturalFakeSpecs`. I didn't know about `StartConfiguration` and didn't realized it was used in unnatural fakes.

tests/FakeItEasy.Specs/ConfigurationSpecs.cs
```diff
@@ -184,20 +184,255 @@ public interface IFoo
...
+        {
+            "Given an object that is not a fake"
+                .x(() => notAFake = new BaseClass());
+
+            "When I start to configure the object"
+                .x(() => exception = Record.Exception(() => A.CallTo(notAFake)));
+
+            "Then it throws an argument exception"
+               .x(() => exception.Should().BeAnExceptionOfType<ArgumentException>()
+                   .And.Message.Should().Contain("The specified object is not recognized as a fake object."));
```
thomaslevesque: Out of scope for this PR, but I think error messages like this one should be extracted to resources. Right now they're repeated in many places (especially in tests/specs), so if we want to change them, we'll have to do it in multiple places.

src/FakeItEasy/CompletedFakeObjectCallExtensions.cs
```diff
@@ -27,7 +27,7 @@ public static IEnumerable<ICompletedFakeObjectCall> Matching<TFake>(this IEnumer
         {
             var factory = ServiceLocator.Current.Resolve<IExpressionCallMatcherFactory>();
             var callExpressionParser = ServiceLocator.Current.Resolve<ICallExpressionParser>();
-            var matcher = factory.CreateCallMathcer(callExpressionParser.Parse(callSpecification));
+            var matcher = factory.CreateCallMatcher(callExpressionParser.Parse(callSpecification));
```
thomaslevesque: Ah, this one had been bothering me during the whole review :+1:

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -97,10 +93,10 @@ public virtual IAfterCallSpecifiedConfiguration AssignsOutAndRefParametersLazily
 
         public UnorderedCallAssertion MustHaveHappened(Repeated repeatConstraint)
         {
-            Guard.AgainstNull(repeatConstraint, "repeatConstraint");
+            Guard.AgainstNull(repeatConstraint, nameof(repeatConstraint));
 
             this.manager.RemoveRule(this.RuleBeingBuilt);
-            var asserter = this.asserterFactory.Invoke(this.Calls.Cast<IFakeObjectCall>());
+            var asserter = this.asserterFactory.Invoke(this.Calls);
```
thomaslevesque: IIRC, the cast was necessary before C# 4. But since we dropped .NET 3.5 support, it wouldn't make sense to compile with C# 3 anyway, so :+1:

src/FakeItEasy/Expressions/ExpressionCallMatcher.cs
```diff
@@ -87,15 +84,9 @@ public virtual void UsePredicateToValidateArguments(Func<ArgumentCollection, boo
...
-            {
-                var valueProvidingConstraint = constraint as IArgumentValueProvider;
-                if (valueProvidingConstraint != null)
-                {
-                    values.Add(valueProvidingConstraint.Value);
-                }
-            }
+            var values = this.argumentConstraints.OfType<IArgumentValueProvider>()
+                .Select(valueProvidingConstraint => valueProvidingConstraint.Value)
+                .ToList();
```
thomaslevesque: ReSharper suggestion? :wink:

thomaslevesque:
Great work @blairconrad. The big refactoring to use parsed expressions is a big improvement. It seems to make things a bit more complex in some places, but I think it's still worth it. I only have minor comments (and they're just comments, not things that need to be changed), LGTM :+1:

thomaslevesque:
> I've half a mind
to revisit that later, to see if we can't make a new
ParsedCallExpression from the one that descibes the getter, but I
think it can wait 'til later.

Maybe I'm missing something, but I think it might actually be easier that way... Anyway, as you say it can be done later.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: It's not the most obvious, it's true. An alternative would be to pass the parser into the `PropertySetterConfiguration`.
**also**, I had not intended to call that value `lambda`. :blush:. Something like `propertySetterExpression` would be better, I think.

src/FakeItEasy/Expressions/ExpressionCallMatcher.cs
blairconrad: Yeah.

tests/FakeItEasy.Specs/ConfigurationSpecs.cs
blairconrad: Not a bad idea. Many of the the exception messages are already in resources. Although I'm funny... I don't mind them being repeated as literals so much in the tests (I wouldn't fight if you wanted to use the resource in the test as well though). It's when I see them repeated in the production code that I'm less comfortable.

tests/FakeItEasy.Tests/CompletedFakeObjectCallExtensionsTests.cs
blairconrad: Yeah. I actually worked for a while trying to convert existing tests. In some places, like for `FakeConfigurationManagerTests`, I succeeded. Elsewhere, not so much. I cannot remember the difficulty here. Maybe I just thought the tests were "too thin" to bother with. For `StartConfigurationTests`, I tried for a while, and failed. Truth be told, I couldn't even understand them.

tests/FakeItEasy.Tests/Configuration/StartConfigurationTests.cs
blairconrad: I'd seen the class before, when trying to remove unused classes pre-2.0. But I didn't really understand what it did. I may not still.

blairconrad:
> The big refactoring to use parsed expressions is a big improvement. It seems to make things a bit more complex in some places

Oh? There's [bit with the the lambda lambda](https://github.com/FakeItEasy/FakeItEasy/pull/783/files#diff-ab0b145be13922e815c0f04e2bc3b878R74) (I'll send another commit with the rename soon, which we can squash in later). Anything else? I don't want to make things worse.


> > I've half a mind
to revisit that later, to see if we can't make a new
ParsedCallExpression from the one that describes the getter, but I
think it can wait 'til later.

> Maybe I'm missing something, but I think it might actually be easier that way... Anyway, as you say it can be done later.

I think it might be in some respects. Well, really I don't think it'll be tricky at all, but the (internal) interface for `ParsedCallExpression` may need to change to accommodate, and I just felt like that might require a bit of thinking.

Thanks for the review. I enjoyed the comments!

src/FakeItEasy/CompletedFakeObjectCallExtensions.cs
blairconrad: It's been bothering me for 2 years or so, but I never took the time before...

blairconrad:
> I'll send another commit with the rename soon, which we can squash in later

Oh, I might just make us build the new setter from the `ParsedCallExpression` instead. Looking at the code around the rename just made me feel silly.

thomaslevesque:
> Oh? There's bit with the the lambda lambda (I'll send another commit with the rename soon, which we can squash in later). Anything else? I don't want to make things worse.

It was just a general impression, but looking at it again, I can't find what exactly had bothered me... maybe I was just tired.

adamralph:
I haven't had time to review yet, but as a general comment, I love the granular refactoring commits. Each has a specific purpose and seems to be reviewable in isolation. I'm also happy to keep them all in this PR since they are ultimately driven by the desire ease the eventual change.

blairconrad:
I'm building property setters directly from ParsedCallExpressions, with no need to go through an Expression! I appended a commit for the sake of @thomaslevesque, but it could be merged into e5d87bc38a4862497df18e800e1a12b489467e8d.

blairconrad:
> they are ultimately driven by the desire ease the eventual change.

That was my thought for not sending a separate PR. I believe in https://twitter.com/kentbeck/status/250733358307500032

adamralph:
:+1: me too!

thomaslevesque:
> I'm building property setters directly from ParsedCallExpressions, with no need to go through an Expression!

Perfect :+1:

tests/FakeItEasy.Specs/ConfigurationSpecs.cs
adamralph: I'm fine with them being in resources for the production code, but they should be spelled out in the tests (or centralised in a separate resource for the tests) otherwise you are just testing whatever the production code is written to do.

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -100,12 +102,14 @@ private static Expression BuildArgumentThatMatchesAnything<TValue>()
 
         private static FakeManager GetFakeManagerCallIsMadeOn(ParsedCallExpression parsedCallExpression)
         {
-            if (parsedCallExpression.CallTarget == null)
-            {
-                throw new ArgumentException("The specified call is not made on a fake object.");
-            }
+            return parsedCallExpression.CallTarget == null
+                ? null
```
adamralph: I'm detecting a code smell here. `CallTo<T>(Expression<Func<T>> callSpecification)` calls this method and then de-reference the return value. Now, that method also calls `AssertThatMemberCanBeIntercepted()` before de-referencing the return value from this method, and I'm guessing that there's no way that method can not throw an exception if the return value from this method is null, hence an NRE is never thrown. However, none of this is obvious to the reader of the code.

Also, it's not obvious that `EnsureCallIsMadeOnAFake` will never be called in a situation where `parsedCallExpression.CallTarget != null`.

I think we should get rid of `GetFakeManagerCallIsMadeOn` and inline the usage of `Fake.GetFakeManager(parsedCallExpression.CallTarget)`.

adamralph:
@blairconrad I did a full pass of the changes and left just one comment - https://github.com/FakeItEasy/FakeItEasy/pull/783#discussion_r68766899

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
blairconrad: Well spotted. I think I like it. You're okay relying on `FakeFacadeManager.GetFakeManager` complaining that the faked object is null, if ever it comes up, not that I think it will…

About inlining `Fake.GetFakeManager`, do you mean everywhere? If we remove `EnsureCallIsMadeOnAFake`, `CallTo(Expression<Action>)` becomes

```c#
public IVoidArgumentValidationConfiguration CallTo(Expression<Action> callSpecification)
{
    Guard.AgainstNull(callSpecification, nameof(callSpecification));

    var parsedCallExpression = this.callExpressionParser.Parse(callSpecification);
    Fake.GetFakeManager(parsedCallExpression.CallTarget);
    this.AssertThatMemberCanBeIntercepted(parsedCallExpression);

    return this.CreateVoidArgumentValidationConfiguration(parsedCallExpression);
}
```

In this case, there's an unused return value from `Fake.GetManager` which looks funny. (There's _always_ an unused return value, but if it's hidden in `EnsureCallIsMadeOnAFake`, then it's not so weird looking.)

adamralph: I think we can keep `EnsureCallIsMadeOnAFake`. It's only `GetFakeManagerCallIsMadeOn` that I think we should inline.

adamralph: Or rather, remove `GetFakeManagerCallIsMadeOn`, so we'd be left with:

```C#
public static void EnsureCallIsMadeOnAFake(ParsedCallExpression parsedCallExpression) =>
    Fake.GetFakeManager(parsedCallExpression.CallTarget);
```

and

```C#
public IReturnValueArgumentValidationConfiguration<T> CallTo<T>(Expression<Func<T>> callSpecification)
{
    Guard.AgainstNull(callSpecification, "callSpecification");

    var parsedCallExpression = this.callExpressionParser.Parse(callSpecification);
    var fake = Fake.GetFakeManager(parsedCallExpression.CallTarget);
    this.AssertThatMemberCanBeIntercepted(parsedCallExpression);

    var rule = this.ruleFactory.Invoke(parsedCallExpression);

    fake.AddRuleFirst(rule);

    return this.configurationFactory.CreateConfiguration<T>(fake, rule);
}
```

and

```C#
private IVoidArgumentValidationConfiguration CreateVoidArgumentValidationConfiguration(ParsedCallExpression parsedCallExpression)
{
    var fake = Fake.GetFakeManager(parsedCallExpression.CallTarget);
    var rule = this.ruleFactory.Invoke(parsedCallExpression);
    fake.AddRuleFirst(rule);

    return this.configurationFactory.CreateConfiguration(fake, rule);
}
```

adamralph: Note that `CallTo(Expression<Action>)` still needs to `fake.AddRuleFirst(rule);` so we still need to assign the return value.

blairconrad: I generally like the shape of it, but `Fake.GetFakeManager` will ultimately call `FakeFacade.GetFakeManager`, which will throw an `ArgumentNullException` when the call is static. The current behaviour is to indicate that the method is static. Unfortunately, that error is coming from the `CastleDynamicProxyInterceptionValidator`.
The existing design puts as much of the decision of whether a call can be configured into the hands of the proxy, by returning a null `FakeManager` when one can't be found. As you hint at, this is a bit of a problem, because we could then fail when we dereference the object. (Currently we won't, because there's no way have `AssertThatMemberCanBeIntercepted` pass when we can't find a `FakeManager`.)

In theory if we ever had a proxy that could support static method calls (e.g. some CodeCop proxy), we'd want to preserve this behaviour. It would mean that the way we'd find a FakeManager  would change: perhaps by looking up by target object, as we do now, for instance methods, but looking up by type if the method were static.

In practice, we're unlikely ever to have such a situation, so perhaps acting like we'll always have a faked object is reasonable.

That still leaves us with the problem of how to structure the code so it gives users the best experience and is still understandable. I think the current user experience is good, but as you point out, we're still not giving a good reading experience—why should I expect a passing `AssertThatMemberCanBeIntercepted` to mean that there's a `FakeManager` to hang rules on?

I'm wondering if we shouldn't keep all the code as is, except to explicitly check the manager for nullness before we add a rule to it. An alternative is to embrace our lack of static (sealed, non-virtual) method handling and hoist some of the checks out of the proxy interception validators, but we'd have to be careful to put them some place where they'd be hit if we tried to abuse an unnatural fake.

adamralph: > I'm wondering if we shouldn't keep all the code as is, except to explicitly check the manager for nullness before we add a rule to it.

Sure, that would work.

I have another question though. When `parsedCallExpression.CallTarget == null` we are returning `null` from `GetFakeManagerCallIsMadeOn` which means no exception is thrown from `EnsureCallIsMadeOnAFake`. That implies that `parsedCallExpression.CallTarget == null` means "the calls are made on a fake". Is that correct?

More generally, all calls to `GetFakeManagerCallIsMadeOn` will now receive a null instead of an exception being thrown when `parsedCallExpression.CallTarget == null`. Is that what we want?

blairconrad: > I have another question though. When `parsedCallExpression.CallTarget == null` we are returning `null` from `GetFakeManagerCallIsMadeOn` which means no exception is thrown from `EnsureCallIsMadeOnAFake`. That implies that `parsedCallExpression.CallTarget == null` means "the calls are made on a fake". Is that correct?

Oh, bother. It is. That's not good, if only from a naming perspective. If the method were `EnsureCallIsNotMadeOnANonFake` it would take the lie away. I don't _love_ it.

> More generally, all calls to `GetFakeManagerCallIsMadeOn` will now receive a null instead of an exception being thrown when `parsedCallExpression.CallTarget == null`. Is that what we want?

I don't mind it so much. Sometime's there's just no fake manager that the call is made on. I think it's safe for the usages we have right now.

I _may_ be able to get rid of `EnsureCallIsMadeOnAFake` and make things look a little bit better. I'll try it out tonight and get back to you.

blairconrad: Okay. I added two commits. There's a little redundancy, in that we'll attempt to get the fake manager twice in many cases, but it's all I could think of to make the code readable and not introduce too much temporal coupling and hidden assumptions. Tell me what you think.

thomaslevesque: > Okay. I added two commits. There's a little redundancy, in that we'll attempt to get the fake manager twice in many cases, but it's all I could think of to make the code readable and not introduce too much temporal coupling and hidden assumptions. Tell me what you think.

Much better, I think. I'm not too worried about the duplicate calls to `Fake.GetFakeManager`; it's not like it's very expensive.

thomaslevesque:
@adamralph, is there anything still bothering you? As far as I'm concerned, it's ready to merge.

adamralph:
Thanks @blairconrad!

I still feel quite uncomfortable reading `FakeConfigurationManager` but I don't think the final result is any worse than what we previously had in `master`. If I feel strongly enough about it, I may propose further refactorings, but I think it's good to go in it's current state.

blairconrad:
I'm not in love with the new form either, but I think it's better than the middle bit I had. Ultimately, I think we may always have something that looks a little unsavoury if we want "that object isn't a fake" to be among the first error messages we issue while at the same time sort of leaving the window open for proxy types that have fewer restrictions than we currently do.

If we right off the bat said: no statics (maybe with an extra check for extension methods), no sealeds, no non-virtual, I think the code could be cleaned up quite a bit.
We'd just need to decide to take that step.

PR #784, opened 2016-06-30
Clean up properties

blairconrad:
Taking advantage of a little C# 6 to remove private setters from auto-props. While I was here, I noticed some getters that were effectively private. There are more, but on public classes. :disappointed:

### DIFF
--- a/src/FakeItEasy/Configuration/BuildableCallRule.cs
+++ b/src/FakeItEasy/Configuration/BuildableCallRule.cs
@@ -53,7 +53,7 @@ namespace FakeItEasy.Configuration
         /// Gets a collection of actions that should be invoked when the configured
         /// call is made.
         /// </summary>
-        public virtual ICollection<Action<IFakeObjectCall>> Actions { get; private set; }
+        public virtual ICollection<Action<IFakeObjectCall>> Actions { get; }
 
         /// <summary>
         /// Gets or sets a function that provides values to apply to output and reference variables.
@@ -214,9 +214,9 @@ namespace FakeItEasy.Configuration
                 this.DescriptionWriter = descriptionWriter;
             }
 
-            public Func<IFakeObjectCall, bool> Predicate { get; private set; }
+            public Func<IFakeObjectCall, bool> Predicate { get; }
 
-            public Action<IOutputWriter> DescriptionWriter { get; private set; }
+            public Action<IOutputWriter> DescriptionWriter { get; }
         }
     }
 }
--- a/src/FakeItEasy/Core/ArgumentValueFormatter.cs
+++ b/src/FakeItEasy/Core/ArgumentValueFormatter.cs
@@ -88,7 +88,7 @@ namespace FakeItEasy.Core
                 this.distanceFromKnownType = distanceFromKnownType;
             }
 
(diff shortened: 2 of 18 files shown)


blairconrad:
Huh. That's interesting. I got a NullReferenceException when running the approval tests on my box, so didn't trust the results, but thought I'd be safe anyhow. I see this change from approved to received API:

```diff
@@ -473,7 +473,7 @@ namespace FakeItEasy.Core                                                                           
     }                                                                                                                  
     public class FakeManager : FakeItEasy.Core.IFakeCallProcessor                                                      
     {                                                                                                                  
-        public System.Type FakeObjectType { get; }                                                                     
+        public virtual System.Type FakeObjectType { get; }
         public virtual object Object { get; }                                                                          
         public virtual System.Collections.Generic.IEnumerable<FakeItEasy.Core.IFakeObjectCallRule> Rules { get; }      
```

But here's my change to that file:

```diff
index ed0f561..0b0bcb7 100644                                                            
--- a/src/FakeItEasy/Core/FakeManager.cs                                                 
+++ b/src/FakeItEasy/Core/FakeManager.cs                                                 
@@ -76,7 +76,7 @@ namespace FakeItEasy.Core                                              
         /// <summary>                                                                   
         /// Gets the faked type.                                                        
         /// </summary>                                                                  
-        public virtual Type FakeObjectType { get; private set; }                        
+        public virtual Type FakeObjectType { get; }
                                                                                         
         /// <summary>                                                                   
         /// Gets the interceptions that are currently registered with the fake object.  
```

Smart guys, any ideas why the approved API text doesn't have a `virtual`?

blairconrad:
For what it's worth, in a FakeItEasy.dll built from master, JustDecompile reports `FakeManager.FakeObjectType` as non-virtual. Although when I try to extend the class in VS, the complaint I get on overriding the property is not that it isn't virtual, but that the constructor is internal.  Which it is. So FakeManager has a number of virtual members and no inheritors. I think the members are virtual so they can be faked in our tests. :sob: 

Confirmed.

Anyhow, i'm not sure that the new `virtual` is a problem as such. Thoughts?

adamralph:
> I think the members are virtual so they can be faked in our tests.

:confounded:

I don't think there is an API change. If anything the new version is correct and the old version incorrect, so I think you can go ahead and approve the change.

I guess there is a bug in PublicApiGenerator. @JakeGinnivan does this look familiar?

blairconrad:
Thanks, @adamralph. That was my feeling as well. I'll amend the first commit.
The odd thing is that JustDecompile agrees with PublicApiGenerator. Two bugs? Seems unlikely.

adamralph:
Here's the CIL diff (file attached):

image

The get method is identical (and virtual) but the set method no longer exists. The set method was non-virtual since it was private, which means the entire property could not be considered to be virtual. This is a really odd effect of combining the get and set into a single false member. Once again, the argument against the existence of properties is bolstered!

Anyhow, I agree, this change won't hurt anyone and AFAIK is not breaking, since it's a relaxation of constraints rather than tightening.

blairconrad:
Thanks for the merge, @adamralph. And the bonus check.

PR #785, opened 2016-07-01
Use nameof to refer to parameters

blairconrad:
A little :lipstick: so we don't have to update these piecemeal.

### DIFF
--- a/src/FakeItEasy/A.cs
+++ b/src/FakeItEasy/A.cs
@@ -46,7 +46,7 @@ namespace FakeItEasy
         [SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter", Justification = "Used to specify the type of fake.")]
         public static T Fake<T>(Action<IFakeOptions<T>> optionsBuilder)
         {
-            Guard.AgainstNull(optionsBuilder, "optionsBuilder");
+            Guard.AgainstNull(optionsBuilder, nameof(optionsBuilder));
 
             return (T)FakeAndDummyManager.CreateFake(typeof(T), options => optionsBuilder((IFakeOptions<T>)options));
         }
--- a/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
+++ b/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
@@ -20,7 +20,7 @@ namespace FakeItEasy
         /// <returns>A dummy argument value.</returns>
         public static T IsNull<T>(this IArgumentConstraintManager<T> manager) where T : class
         {
-            Guard.AgainstNull(manager, "manager");
+            Guard.AgainstNull(manager, nameof(manager));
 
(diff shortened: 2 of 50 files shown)


adamralph:
:+1: thanks @blairconrad. Tedious, highly appreciated.

I wonder if there's an analyzer which can tell us when we're not using `nameof()` and should be?

thomaslevesque:
Yes, thanks @blairconrad!

> I wonder if there's an analyzer which can tell us when we're not using nameof() and should be?

Yes, ReSharper :wink: (actually, I suspect that's how @blairconrad did it).
I think CodeCracker does it too.

adamralph:
There's also [C# Essentials](https://visualstudiogallery.msdn.microsoft.com/a4445ad0-f97c-41f9-a148-eae225dcc8a5).

I'd prefer an analyser that lived in the project though. Relying on everyone to install appropriate tooling is a poor solution.

thomaslevesque:
> I'd prefer an analyser that lived in the project though. Relying on everyone to install appropriate tooling is a poor solution.

CodeCracker is a Roslyn analyzer too; it can be installed either as a VSIX or as a NuGet package

adamralph:
Ah OK, I knew it was a Roslyn Analyser, but didn't realise it also came in NuGet form. (C# Essentials is also a Roslyn analyser but only comes in VSIX form.)

https://twitter.com/adamralph/status/748874069894836224

blairconrad:
> Yes, ReSharper :wink: (actually, I suspect that's how @blairconrad did it).

For the non-`Guard.AgainstNull` changes, yes. But ReSharper knew nothing about `Guard.AgainstNull`. That was me and a good ol' regex search-and-replace.

blairconrad:
Oh, and thanks for the merge. And the CodeCracker/C# Essentials learning. I am curious, but should probably learn more about .NET Core instead.

thomaslevesque:
> But ReSharper knew nothing about Guard.AgainstNull. That was me and a good ol' regex search-and-replace.

You could have (temporarily) applied the [`[InvokerParameterName]`](https://www.jetbrains.com/help/resharper/2016.1/Reference__Code_Annotation_Attributes.html#InvokerParameterNameAttribute) attribute to `Guard.AgainstNull`'s parameter, then R# would have picked it up :wink:

blairconrad:
> You could have (temporarily) applied the [`[InvokerParameterName]`](https://www.jetbrains.com/help/resharper/2016.1/Reference__Code_Annotation_Attributes.html#InvokerParameterNameAttribute) attribute to `Guard.AgainstNull`'s parameter, then R# would have picked it up :wink:

Oh, good point! But regexing was just as easy. I'll keep that trick in mind for the future, though.

adamralph:
I don't think there's much to learn about CodeCracker/C# Essentials. AFAIK they're just extensions that perform static analysis and give you warnings and suggest refactorings.

blairconrad:
Sure. But this is what I wanted to learn:

image

PR #787, opened 2016-07-01
Use fully-qualified namespaces in usings

blairconrad:


thomaslevesque:
LGTM :+1: 

Shouldn't we have a "cleanup" tag? It's not really a refactoring...

blairconrad:
Thanks, Quick-Draw.

> Shouldn't we have a "cleanup" tag? It's not really a refactoring...

I could've gone for `tech-debt`, I suppose. Better? Better but not close enough? I'm afraid that I'm a little lax with the labels.

thomaslevesque:
`tech-debt` is slightly better, but isn't a really good match either... Anyway, I'm not convinced there's much value in labeling pull requests. Issues, yes, but PRs? They're (usually) short-lived, and we rarely go back to them after they've been merged.

adamralph:
I think it makes sense to label PRs when they're not connected to an issue. It's good to know what kind of changes are being made, i.e. feature, bug, refactoring, tech debt, docs, etc. When the PR is connected to an issue, that issue has the relevant label. When it's an independent PR, the work still falls into one of those categories.

blairconrad:
And I'd noticed @adamralph's recent tendency to label the PRs, assumed that was why, and so I labelled this one.

adamralph:
I only label them when they're independent, since they're effectively a PR
and issue in one.

blairconrad:
Sorry, yes. "the independent PRs" I should've said.

thomaslevesque:
OK, it makes sense

adamralph:
For more context (I realise I've said this before), this is my view of our world: https://waffle.io/FakeItEasy/FakeItEasy

Sometimes you might see me add "connects to #xxx" in an issue description. This is to make PR's which serve parent issues fold into that issue on the board. Any unconnected PR appears as it's own card, so labelling them means that every card on the board has an appropriate "type" label (feature, bug, refactoring, etc.)

adamralph:
Sometimes, I'll add "connects to..." in an HTML comment, as I did for https://github.com/FakeItEasy/FakeItEasy/pull/617 - you can see it if you hit the "edit" button on the PR description.

thomaslevesque:
> in an HTML comment

That's sneaky :wink:
IMO it doesn't need to be hidden in a comment; if it's visible, anyone can see the connection.

adamralph:
@thomaslevesque yeah, true. It didn't feel quite right altering a contributor's description, but perhaps it's better. I have no qualms altering your or @blairconrad's descriptions :stuck_out_tongue:

PR #788, opened 2016-07-02
Remove SmellyAttribute

blairconrad:
If we see a code smell, we should probably make a **tech-debt** issue.


thomaslevesque:
I'm surprised the `SmellyAttribute` class didn't have the `[Smelly]` attribute :wink:

Good riddance, thanks @blairconrad!

blairconrad:
Ha. Thanks guys. Of course, if it did have the `[Smelly]` attribute, I _might_ not've been smart enough to realize that it wans't used

PR #792, opened 2016-07-04
Use nameof for some literals

blairconrad:


src/FakeItEasy/Core/DelegateRaiser.cs
```diff
@@ -68,7 +68,7 @@ private static void AssertThatValuesSatisfyCallSignature(MethodInfo callMethod,
 
             throw new FakeConfigurationException(
                 "The event has the signature ({0}), but the provided arguments have types ({1})."
-                    .FormatInvariant(fakeSignature, actionSignature, "Raise"));
+                    .FormatInvariant(fakeSignature, actionSignature, nameof(Raise)));
```
blairconrad: In retrospect, this one may be an error. It refers to the class `Raise`. I will restore.

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -163,7 +163,7 @@ private static NewArrayExpression WrapParametersInObjectArray(ParameterExpressio
 
         private class DelegateCallInterceptedEventRaiser
         {
-            public static readonly MethodInfo RaiseMethod = typeof(DelegateCallInterceptedEventRaiser).GetMethod("Raise");
+            public static readonly MethodInfo RaiseMethod = typeof(DelegateCallInterceptedEventRaiser).GetMethod(nameof(Raise));
```
blairconrad: Mostly I wanted the change so I'd stop getting complaints that `Raise` wasn't used.

blairconrad:
Oh, thanks for the label! So forgetful. Sigh.

src/FakeItEasy/Core/DelegateRaiser.cs
thomaslevesque: Actually, I think `"Raise"` could be removed... there are only 2 placeholders in the format strings, not 3

src/FakeItEasy/Core/DelegateRaiser.cs
```diff
@@ -5,7 +5,7 @@ namespace FakeItEasy.Core
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
     using System.Reflection;
-
+    using Castle.DynamicProxy.Generators;
```
thomaslevesque: Is this intentional?

blairconrad: Ah. No. I have no explanation.

blairconrad: Ah. Not true. I think it came from the change to `nameof(Raise)` that I incompletely reverted.

src/FakeItEasy/Core/DelegateRaiser.cs
blairconrad: YES!

blairconrad:
Thanks for the keen reviews, @thomaslevesque.

thomaslevesque:
LGTM :+1:

I'll merge when the build completes

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
adamralph: Does this still refer to the "Raise" type rather than the method?

adamralph:
Please see https://github.com/FakeItEasy/FakeItEasy/pull/792#discussion_r69491028 before merging.

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
thomaslevesque: No, it refers to the method, which is closer in scope (just checked). If we want to be extra careful, we could fully qualify it (`DelegateCallInterceptedEventRaiser.Raise`).

adamralph: OK, I think it's fine as is.

PR #795, opened 2016-07-05
Remove IFakeAndDummyManager.TryResolveDummyValue

blairconrad:
It's not used. Probably was missed from removal when we changed unconfigured fakes to always return Dummies. But that's just a guess.


adamralph:
Thanks!

PR #804, opened 2016-07-21
Introduce Then syntax to configure successive calls to the same member

### LINKED ISSUE #803: Provide a fluent syntax to configure successive calls to the same member.
Currently, configuring successive calls to the same member is a bit counter-intuitive, because the configuration order is the reverse of the call order. For instance, if I do this:

```
A.CallTo(() => foo.Bar()).Returns(1);
A.CallTo(() => foo.Bar()).Throws<InvalidOperationException>().Once();
A.CallTo(() => foo.Bar()).Returns(0).Twice();
```

It will return 0 on the first two calls, then throw an exception, then return 1 forever. Even though it makes sense from an implementation perspective, this behavior is very surprising and confusing to people using FakeItEasy for the first time, and there's no way you can guess it works like this without looking at the [documentation](http://fakeiteasy.readthedocs.io/en/stable/limited-call-specifications/#changing-behavior-between-calls).

I'm not suggesting we change the current behavior, because too much existing code relies on it, but we could provide a more intuitive syntax:

```
A.CallTo(() => foo.Bar()).Returns(0).Twice()
    .Then.Throws<InvalidOperationException>().Once()
    .Then.Returns(1);
```

I know that having two ways to do the same thing is something we usually try to avoid, but in this case I think it's worth it, because:
- it makes the code less surprising and easier to understand, especially for newcomers
- it's more discoverable
- as a side benefit, the call specification no longer needs to be repeated.

The return type of `Once`/`Twice`/`NumberOfTimes` would have to change from `void` to something else; this would be a breaking change at the binary level, but not at the source level, except in the (unlikely) case where one of these methods is assigned to a delegate.

`Then` would only be available after a repeat configuration, since it doesn't make sense for a rule that is applied forever.

thomaslevesque:
Fixes #803 

- The first commit refactors the configuration interfaces so that the relevant type (*) is propagated
- The second commit introduces the new API

(*) by "relevant type", I mean the configuration interface we get after the call is fully specified (i.e. after `Where`/`WhenArgumentsMatch`/`WithReturnType` etc):
- for `A.CallTo(Expression<Action>)` and `A.CallTo(object)`, this is `IVoidConfiguration`
- for `A.CallTo(Expression<Func<T>>)` and `A.CallTo(object).WithReturnType<T>()`, this is `IReturnValueConfiguration<T>`
- for `A.CallToSet(Expression<Func<TValue>>)`, this is `IPropertySetterConfiguration`


thomaslevesque:
Most serious issues so far:

- `OutAndRefParametersConfigurationExtensions`

  Many of the methods already have generic parameters; adding a new one (the configuration interface) means that users who explicitly stated the types will have to also specify the configuration interface... which can be pretty ugly. However, it can still work if the parameter types are specified in the lambda (which is what I usually do):
    ```
...
.AssignsOutAndRefParametersLazily((string s) => new object[] { s.Length })
    ```
- `ExceptionThrowerConfigurationExtensions`

  Same issue as above, plus another (probably the worst):  `Throws<T>()` can no longer work if we add a generic parameter for the interface type; since it takes no parameter, type inference can't work. And I don't think people will enjoy writing things like `.Throws<IAnyCallConfigurationWithReturnTypeSpecified<Foo>, ArgumentException>()`...

src/FakeItEasy/ExceptionThrowerConfigurationExtensions.cs
```diff
@@ -130,6 +134,11 @@ public static IAfterCallSpecifiedConfiguration Throws<T1>(this IExceptionThrower
                 });
         }
 
+#if DISABLED
+
+        // This method already has a generic parameter, and it can't be inferred; if we add another, we make it unusable,
+        // especially since the new one is something unintelligible like IAnyCallConfigurationWithReturnTypeSpecificied<Foo>...
+
```
blairconrad: I've always seen these extension methods as coding convenience, not necessity. Could we make this not be an extension method? e.g., on `RuleBuilder` (and other leaf classes) have a method like

```c#
public IAfterCallSpecifiedConfiguration<IVoidConfiguration> Throws<T>() where T : Exception, new()
{
    return this.Throws(() => new T());
}
```

If I add this, and tart up `IVoidArgumentValidationConfiguration` (possibly not the best place, but it was expedient), then this passes, where `Bar` returns `void`, if I can say that:

```c#
[Scenario]
public static void Throws(
    IFoo fake,
    Exception exception)
{
    "Given a fake"

    "And I configure a method to throw"
        .x(() => A.CallTo(() => fake.Bar()).Throws<ArgumentOutOfRangeException>());

    "When I call the method"
        .x(() => exception = Record.Exception(() => fake.Bar()));

    "Then it throws"
        .x(() => exception.Should().BeAnExceptionOfType<ArgumentOutOfRangeException>());
}
```

blairconrad:
@thomaslevesque, may have a way to, um... patch and reinflate your bubble (this is almost certainly not a real phrase in English. Inspired by [burst your bubble](http://www.knowyourphrase.com/phrase-meanings/Burst-Your-Bubble.html)). [Check it out](https://github.com/FakeItEasy/FakeItEasy/pull/804/files#r71817665). Apologies if my idea is rubbish.

blairconrad:
Also, I should've said. I remain excited about the API and love the separation of the first commit.

src/FakeItEasy/ExceptionThrowerConfigurationExtensions.cs
thomaslevesque: > I've always seen these extension methods as coding convenience, not necessity.

Well, they *are* a convenience, but without them, the API becomes quite inconvenient...

> Could we make this not be an extension method? e.g., on RuleBuilder (and other leaf classes) have a method like

Yes, but we don't expose the classes directly, only the interfaces... so the method wouldn't be available (unless we make it part of the interface, but that would make it harder to implement)

However, your comment made me realize that it should be possible to write more specific extension methods that wouldn't require the extra type parameter; looking into it now.

blairconrad: > > I've always seen these extension methods as coding convenience, not necessity.

> Well, they are a convenience, but without them, the API becomes quite inconvenient...

What I meant was that any API endpoint that we provide via extension method we could also have provided using traditional, non-extension methods. Implementing via extension method is just something that reduces a little of the FakeItEasy code.

> Yes, but we don't expose the classes directly, only the interfaces... so the method wouldn't be available (unless we make it part of the interface, but that would make it harder to implement)

Absolutely.

> But in other cases I still have to add a type param

Ah, so I just got lucky by using this particular case as a test. Too bad.

thomaslevesque: > Ah, so I just got lucky by using this particular case as a test. Too bad.

No, it works fine if the method is in the interface. I guess we could do it like this in this particular case.

In fact, migrating all the extension methods to the interface they extend would solve all the usability problems, but it feels wrong somehow... Interfaces should be as small as possible; the current approach of having only general purpose methods in the interface and extension methods for specific use cases is clean and elegant.

blairconrad: > In fact, migrating all the extension methods to the interface they extend would solve all the usability problems,

Oh, good. That's what I originally thought. I thought you'd seen something that prevented it.

>  Interfaces should be as small as possible; the current approach of having only general purpose methods in the interface and extension methods for specific use cases is clean and elegant.

That's never really been my opinion. I'm all for focused interfaces, but using extension methods to provide core functionality has always felt a little off to me. There's the advantage that you don't have to provide implementations on as many concrete classes, but it seems a little "polyfill" to me. Especially in FIE, where we don't expect users to implement these interfaces. To contrast with the `IEnumerable` extension methods, where everyone's encouraged to provide their own minimal implementation of an interface and the LINQ methods extend it to a rich ecosystem.

If the interfaces were abstract classes would you feel better about it?

thomaslevesque: > If the interfaces were abstract classes would you feel better about it?

I guess so; especially since only one of the methods would need to be abstract. But it's not really an option for FIE, because RuleBuilder et al. implement multiple configuration interfaces, and .NET only supports one base class.

Anyway, having the extension methods declared in the interfaces is probably the best option here. What bothers me the most is the repetition. For instance, 3 classes implement `IOutAndRefParametersConfiguration.AssignsOutAndRefParametersLazily`, and there are 5 related helper extension methods; these 5 methods will turn into 15 when implemented in each of the configuration classes, with almost exactly the same code. Maybe we could keep the extension methods, but make them internal, and use them to implement the interface methods...

blairconrad: > Maybe we could keep the extension methods, but make them internal, and use them to implement the interface methods...

An interesting approach.

blairconrad: >  3 classes implement `IOutAndRefParametersConfiguration.AssignsOutAndRefParametersLazily`

A quibble, but I only see two: `AnyCallConfiguration` and `RuleBuilder`.
Even if I'm right, that still leaves 10 methods.
The real reason I bring it up is that I'm doubting my counting. What have I done wrong?

thomaslevesque: > A quibble, but I only see two: `AnyCallConfiguration` and `RuleBuilder`.

And `RuleBuilder.ReturnValueConfiguration<TMember>`

I used the R# "Derived symbols" command (<kbd>Alt</kbd>+<kbd>End</kbd>)

blairconrad: Oh, and based on your original comment, is there a _need_ to hoist the `AssignsOutAndRefParameters…` methods up to the interfaces? I thought `Throws` was the only one that absolutely required it, and that the `Assigns…` methods would only cause difficulties if users specified the typesparams. I'm just making sure I understand. I'm not arguing for making extra difficulties for the users, just wondering if pushing the methods up to the interfaces is technically necessary in all cases.

blairconrad: > And `RuleBuilder.ReturnValueConfiguration<TMember>`

So it does!

> I used the R# "Derived symbols" command (<kbd>Alt</kbd>+<kbd>End</kbd>)

So did I. And I **swear** that `RuleBuilder.ReturnValueConfiguration<TMember>` didn't show up. And now it does. User error, I'm sure.

thomaslevesque: > Oh, and based on your original comment, is there a need to hoist the AssignsOutAndRefParameters… methods up to the interfaces? I thought Throws was the only one that absolutely required it, and that the Assigns… methods would only cause difficulties if users specified the typeparams.

Exactly. But I suspect many users do specify the typeparams explicitly... I think it's better to keep both options usable.

thomaslevesque: I made the changes we discussed (two new commits).

thomaslevesque:
I added a commit that implement the actual feature. The StyleCop warnings are still there until we decide whether or not to go ahead with this. Also, I didn't write any test yet, I just tested the feature manually (and not very thoroughly), so it's still a work in progress.

Initially, having to add a generic type param to the configuration interfaces annoyed me, but in hindsight, I think it's actually a good thing in the long run; it means we stop losing the information about what kind of call it is (void, return value, property set). It might make it easier to implement other new features in the future.

thomaslevesque:
I added some specs, and fixed the StyleCop and FxCop warnings.

src/FakeItEasy/Configuration/IExceptionThrowerConfiguration.cs
```diff
@@ -1,20 +1,77 @@
...
+        IAfterCallSpecifiedConfiguration<TInterface> Throws<T1, T2, T3, T4>(Func<T1, T2, T3, T4, Exception> exceptionFactory);
+
+        /// <summary>
+        /// Throws the specified exception when the currently configured
+        /// call gets called.
+        /// </summary>
+        /// <typeparam name="T">The type of exception to throw.</typeparam>
+        /// <returns>Configuration object.</returns>
+        [SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter", Justification = "By design.")]
+        IAfterCallSpecifiedConfiguration<TInterface> Throws<T>() where T : Exception, new();
```
thomaslevesque: These methods were initially extension methods, but had to be moved to the interface because of the new generic type param (see the discussion [here](https://github.com/FakeItEasy/FakeItEasy/pull/804#discussion_r71857913))

src/FakeItEasy/Configuration/IOutAndRefParametersConfiguration.cs
```diff
@@ -19,6 +20,68 @@ public interface IOutAndRefParametersConfiguration
...
+        /// <typeparam name="T1">Type of the first argument of the faked method call.</typeparam>
+        /// <typeparam name="T2">Type of the second argument of the faked method call.</typeparam>
+        /// <typeparam name="T3">Type of the third argument of the faked method call.</typeparam>
+        /// <typeparam name="T4">Type of the fourth argument of the faked method call.</typeparam>
+        /// <param name="valueProducer">A function that produces the output values.</param>
+        /// <returns>A configuration object.</returns>
+        /// <exception cref="FakeConfigurationException">
+        /// The signatures of the faked method and the <paramref name="valueProducer"/> do not match.
+        /// </exception>
+        IAfterCallSpecifiedConfiguration<TInterface> AssignsOutAndRefParametersLazily<T1, T2, T3, T4>(Func<T1, T2, T3, T4, object[]> valueProducer);
```
thomaslevesque: These methods were initially extension methods, but had to be moved to the interface because of the new generic type param (see the discussion [here](https://github.com/FakeItEasy/FakeItEasy/pull/804#discussion_r71857913))

src/FakeItEasy/Configuration/IThenConfiguration.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// <summary>
+    /// Configures the next behavior for this call.
+    /// </summary>
+    /// <typeparam name="TInterface">The type of configuration interface to return.</typeparam>
+    public interface IThenConfiguration<out TInterface>
+    {
+        /// <summary>
+        /// Gets an object that lets you specify how the call will behave after the previous configuration has been consumed.
+        /// </summary>
+        [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = nameof(Then), Justification = "Part of the fluent syntax.")]
```
thomaslevesque: I considered looking for another name, but the only language I can think of that uses `Then` (with this casing) as a keyword is VB.NET, and it seems to handle it just fine.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -103,6 +101,22 @@ public virtual void AddRuleLast(IFakeObjectCallRule rule)
...
+        public virtual void AddRuleAfter(IFakeObjectCallRule previousRule, IFakeObjectCallRule newRule)
+        {
+            var previousNode = this.AllUserRules.Nodes().FirstOrDefault(n => n.Value.Rule == previousRule);
+            if (previousNode == null)
+            {
+                throw new InvalidOperationException("The rule after which to add the new rule was not found in the list.");
+            }
+
+            this.AllUserRules.AddAfter(previousNode, new CallRuleMetadata { Rule = newRule });
+        }
```
thomaslevesque: I wonder if I should make this internal... there's probably no harm in exposing it publicly, but if it's public, there should probably be an `AddRuleBefore` method as well. Let me know what you think.

src/FakeItEasy/ExceptionThrowerConfigurationExtensions.cs
```diff
@@ -67,11 +70,12 @@ public static IAfterCallSpecifiedConfiguration Throws<T1>(this IExceptionThrower
         /// </summary>
         /// <param name="configuration">The configuration to use.</param>
         /// <param name="exceptionFactory">A function that returns the exception to throw when invoked.</param>
+        /// <typeparam name="TInterface">The type of configuration interface to return.</typeparam>
         /// <typeparam name="T1">Type of the first argument of the faked method call.</typeparam>
         /// <typeparam name="T2">Type of the second argument of the faked method call.</typeparam>
         /// <returns>Configuration object.</returns>
         /// <exception cref="FakeConfigurationException">The signatures of the faked method and the <paramref name="exceptionFactory"/> do not match.</exception>
-        public static IAfterCallSpecifiedConfiguration Throws<T1, T2>(this IExceptionThrowerConfiguration configuration, Func<T1, T2, Exception> exceptionFactory)
+        internal static IAfterCallSpecifiedConfiguration<TInterface> Throws<TInterface, T1, T2>(this IExceptionThrowerConfiguration<TInterface> configuration, Func<T1, T2, Exception> exceptionFactory)
```
thomaslevesque: The addition of the `TInterface` type param made this method, and the following ones, too impractical to use, so they had to be moved to the `IExceptionThrowerConfiguration<TInterface>`. The extension methods are still used by the classes that implement the interface, to avoid code duplication.

src/FakeItEasy/LinkedListExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        public static IEnumerable<LinkedListNode<T>> Nodes<T>(this LinkedList<T> linkedList)
+        {
+            var node = linkedList.First;
+            while (node != null)
+            {
+                yield return node;
+                node = node.Next;
+            }
+        }
```
thomaslevesque: Convenience method used in `FakeManager.AddRuleAfter`

src/FakeItEasy/OutAndRefParametersConfigurationExtensions.cs
```diff
@@ -34,15 +35,16 @@ public static class OutAndRefParametersConfigurationExtensions
...
         /// <param name="configuration">The configuration to extend.</param>
         /// <param name="valueProducer">A function that produces the output values.</param>
         /// <returns>A configuration object.</returns>
         /// <exception cref="FakeConfigurationException">
         /// The signatures of the faked method and the <paramref name="valueProducer"/> do not match.
         /// </exception>
-        public static IAfterCallSpecifiedConfiguration AssignsOutAndRefParametersLazily<T1>(
-            this IOutAndRefParametersConfiguration configuration, Func<T1, object[]> valueProducer)
+        internal static IAfterCallSpecifiedConfiguration<TInterface> AssignsOutAndRefParametersLazily<TInterface, T1>(
+            this IOutAndRefParametersConfiguration<TInterface> configuration, Func<T1, object[]> valueProducer)
```
thomaslevesque: The addition of the `TInterface` type param made this method, and the following ones, too impractical to use, so they had to be moved to the `IOutAndRefParametersConfiguration<TInterface>` interface. The extension methods are still used by the classes that implement the interface, to avoid code duplication.

thomaslevesque:
I made a few fixes and squashed some commits.

I haven't yet approved the API changes, so the approval tests are failing. I'll add a commit to approve the changes after we agree on them.

src/FakeItEasy/Configuration/IExceptionThrowerConfiguration.cs
```diff
@@ -1,20 +1,77 @@
...
+        IAfterCallSpecifiedConfiguration<TInterface> Throws(Func<IFakeObjectCall, Exception> exceptionFactory);
+
+        /// <summary>
+        /// Throws the specified exception when the currently configured
+        /// call gets called.
+        /// </summary>
+        /// <param name="exceptionFactory">A function that returns the exception to throw when invoked.</param>
+        /// <typeparam name="T1">Type of the first argument of the faked method call.</typeparam>
+        /// <returns>Configuration object.</returns>
+        /// <exception cref="FakeConfigurationException">The signatures of the faked method and the <paramref name="exceptionFactory"/> do not match.</exception>
```
blairconrad: Bonus docs!

src/FakeItEasy/Configuration/BuildableCallRule.cs
```diff
@@ -158,6 +158,24 @@ public virtual void ApplyWherePredicate(Func<IFakeObjectCall, bool> predicate, A
...
+        /// <returns>The cloned rule.</returns>
+        public BuildableCallRule CloneCallSpecification()
+        {
+            var clone = this.CloneCallSpecificationCore();
+            clone.wherePredicates.AddRange(this.wherePredicates);
+            return clone;
+        }
+
+        /// <summary>
+        /// When overridden in a derived class, returns an new instance of the same type and takes care of copying the
```
blairconrad: nit: "a new"

And possibly "takes care of copying" → "copies". Not that there's anything _wrong_ with what you have. I leave it to you.

src/FakeItEasy/Configuration/IExceptionThrowerConfiguration.cs
```diff
@@ -1,20 +1,77 @@
 namespace FakeItEasy.Configuration
 {
     using System;
+    using System.Diagnostics.CodeAnalysis;
```
blairconrad: For the life of me, I can't see where this is used. I could checkout and build and remove it, but so lazy!

src/FakeItEasy/Configuration/IRepeatConfiguration.cs
```diff
@@ -3,12 +3,14 @@ namespace FakeItEasy.Configuration
...
-    public interface IRepeatConfiguration : IHideObjectMembers
+    /// <typeparam name="TInterface">The type of configuration interface to return.</typeparam>
+    public interface IRepeatConfiguration<out TInterface> : IHideObjectMembers
     {
         /// <summary>
         /// Specifies the number of times for the configured event.
         /// </summary>
         /// <param name="numberOfTimesToRepeat">The number of times to repeat.</param>
-        void NumberOfTimes(int numberOfTimesToRepeat);
+        /// <returns>A configuration object that lets you define the following behavior.</returns>
```
blairconrad: It's probably just me, but when I read "the following behavior", I expected to see the behavior on the next line or so. I wonder if maybe we shouldn't say "…define subsequent behavior for the method call"?

src/FakeItEasy/Configuration/IThenConfiguration.cs
blairconrad: We had a similar concern the last time we introduced `Then`, and it all worked out. I think it's fine.

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -107,7 +164,10 @@ public UnorderedCallAssertion MustHaveHappened(Repeated repeatConstraint)
         }
 
         public class ReturnValueConfiguration<TMember>
-            : IAnyCallConfigurationWithReturnTypeSpecified<TMember>, ICallCollectionAndCallMatcherAccessor
+            : IAnyCallConfigurationWithReturnTypeSpecified<TMember>,
+              IAfterCallSpecifiedWithOutAndRefParametersConfiguration<IReturnValueConfiguration<TMember>>,
```
blairconrad: This is a bit of a surprise. To me, at least.

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -118,19 +178,54 @@ public ReturnValueConfiguration(RuleBuilder parentConfiguration)
...
+                return this.Throws<IReturnValueConfiguration<TMember>, T>();
+            }
+
+            public IAfterCallSpecifiedWithOutAndRefParametersConfiguration<IReturnValueConfiguration<TMember>> ReturnsLazily(Func<IFakeObjectCall, TMember> valueProducer)
             {
                 Guard.AgainstNull(valueProducer, nameof(valueProducer));
 
                 this.ParentConfiguration.RuleBeingBuilt.Applicator = x => x.SetReturnValue(valueProducer(x));
-                return this.ParentConfiguration;
+                return this;
```
blairconrad: Aha! There's the other part of the surprise from line 168.

src/FakeItEasy/Core/FakeManager.cs
blairconrad: I'd hide it, if at all possible. It's always easier to expose it later if you change your mind.

tests/FakeItEasy.Specs/ThenSpecs.cs
```diff
@@ -0,0 +1,163 @@
...
+                .x(() => A.CallTo(() => action1()).MustHaveHappened(Repeated.Exactly.Once));
+
+            "And the second delegate is called twice"
+                .x(() => A.CallTo(() => action2()).MustHaveHappened(Repeated.Exactly.Twice));
+
+            "And the fourth call throws an exception"
+                .x(() => exception.Should().BeAnExceptionOfType<InvalidOperationException>());
+        }
+    }
+}
```
blairconrad: I wonder if there shouldn't be some specs to show the interaction between the `Then`-style specifications and the old-style "define behaviour, then defined overriding behaviour", just to make sure it keeps working.
We won't encourage this kind of thing but, it will come up via implicit rules and/or `IFakeOptionBuilder`s.

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
```diff
@@ -52,7 +52,7 @@ public void Returns_called_with_value_returns_parent_configuration()
 
             var result = returnConfig.Returns(10);
 
-            result.Should().Be(this.builder);
+            result.Should().BeSameAs(returnConfig);
```
blairconrad: Are we still returning the "parent configuration"? I think that was `this.builder`.

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
```diff
@@ -78,7 +78,7 @@ public void Returns_with_call_function_should_return_delegate()
 
             var returned = config.ReturnsLazily(x => x.Arguments.Get<int>(0));
 
-            returned.Should().Be(config.ParentConfiguration);
+            returned.Should().BeSameAs(config);
```
blairconrad: Now we're returning the configuration. The method name could stand an update here too, I think.

blairconrad:
Good work, @thomaslevesque. I've taken a pass. The cloning wouldn't have occurred to me. I think it's a good approach.
Also I wouldn't have thought of `AddAfter`. I probably would've just added the rule to the end, which should be the same effect if the methods are used fluently like we expect, but if someone saves away call configuration and then `Then`s it later, your approach will do the right thing.

I've made some quite minor notes. I'm keen to see the API and docs updates.

src/FakeItEasy/Configuration/IAfterCallSpecifiedConfiguration.cs
```diff
@@ -3,8 +3,9 @@ namespace FakeItEasy.Configuration
     /// <summary>
     /// Lets you set up expectations and configure repeat for the configured call.
     /// </summary>
-    public interface IAfterCallSpecifiedConfiguration
-        : IRepeatConfiguration
```
thomaslevesque: Currently, it doesn't seem very useful, but I think it makes sense to have it. If we removed it and returned `IRepeatConfiguration` directly, we couldn't add anything else after a call configuration without making a breaking change to the API. Currently, we just need to add another base interface if we want to do that (which, technically, is also a breaking change, but the users aren't supposed to implement that interface anyway)

That being said, I think the name is confusing; it should probably be called `IAfterCallConfiguredConfiguration` (which doesn't sound very good, but is more accurate).

src/FakeItEasy/Configuration/IRepeatConfiguration.cs
thomaslevesque: Yes, subsequent is better.

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: Can you elaborate? What do you find surprising?

src/FakeItEasy/Configuration/IAfterCallSpecifiedConfiguration.cs
blairconrad: I'm good with keeping. Just wondering if we'd made it skinnier over the years or something. I should've looked in the gits.

Good point about the name! We could add that to our future "breaking changes" list.

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: (I know, the interface is weird... it took me a while to wrap my head around it)

thomaslevesque: Basically, `TInterface` in the generic configuration interfaces represents the kind of call that's being configured (the types I mentioned in the PR description)

blairconrad: I was surprised to see a new interface show up in the list of bases here.
But it's because of [returning this](https://github.com/FakeItEasy/FakeItEasy/pull/804/files/5a163207653185e905712251ee56b0ba7c1732c8#diff-6d9562889008c1f9bfad045e9078aba0R228), no?

tests/FakeItEasy.Specs/ThenSpecs.cs
thomaslevesque: Sure, I can add some specs.

blairconrad: Thanks. You're good folk.
I'm not saying go nuts or anything, but it'd be nice to have at least one demonstrating the interaction.

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
thomaslevesque: No, we're returning the `ReturnValueConfiguration<T>` now (cf. the "other part of the surprise" in your comment above). Anyway, I think this test falls in the Gulliver's Travels Test category... from an external point of view, it shouldn't matter which object we return, as long as it has the correct interface.

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
thomaslevesque: Yes, probably... but I'd rather drop this test anyway, for the same reason as above (GTT)

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: Yes, probably... it's been a while since I did this, I don't remember exactly

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
blairconrad: Oh, yeah. I'd've convinced myself that I had good specs in place to cover whatever we're trying to cover here, and dropped these like hot potatoes.

src/FakeItEasy/Configuration/RuleBuilder.cs
blairconrad: Oh! A dig about the delay in the review. Nasty, M. Levesque.

(I'm just teasing.)

thomaslevesque: :smile:

thomaslevesque:
I made the changes (each in its own commit to make it easier to squash later). I also rebased and added a commit to approve the API.

blairconrad:
> docs/limited-call-specifications.md → docs/changing-behavior-between-calls.md

@thomaslevesque, love the new topic name. I wonder if we shouldn't leave the old file (with a redirect) so as not to break existing links. If that's a thing. Then again, I don't know what RTD does if you hit the old URL.

blairconrad:
… or a redirect could be added via https://readthedocs.org/dashboard/fakeiteasy/redirects/
A little more clicky-clicky and less discoverable/maintainable by us (especially as versions change? not sure how the redirect works across versions). Thoughts?

thomaslevesque:
Yes, the redirect is a good idea

> A little more clicky-clicky and less discoverable/maintainable by us

Looks like there's a way to do it in the docs themselves: http://read-the-docs.readthedocs.io/en/latest/user-defined-redirects.html#page-redirects
But I don't understand where we're supposed to put that configuration...

thomaslevesque:
> Looks like there's a way to do it in the docs themselves: http://read-the-docs.readthedocs.io/en/latest/user-defined-redirects.html#page-redirects
> But I don't understand where we're supposed to put that configuration...

Scratch that, I misunderstood what the docs were saying... looks like it has to be done in the dashboard, which is a shame.

blairconrad:
> looks like it has to be done in the dashboard, which is a shame.

Yeah. "Clicky clicky". :-( Since nothing is available in a config file or whatever, I prefer manual redirect, I think.

(Sorry for lack of comments on the review. Chores and Olympics beckon.)

blairconrad:
Also, 

* `calling-base-methods.md` and
* `doing-nothing.md`

Now have broken links to `limited-call-specifications.md`

docs/changing-behavior-between-calls.md
```diff
@@ -17,17 +19,40 @@ A.CallTo(() => fakeShop.Address).Returns("123 Fake Street").NumberOfTimes(17);
...
+
+```csharp
+// Configure the method to throw an exception once, then succeed forever
+A.CallTo(() => fakeService.DoSomething())
+    .Throws<Exception>().Once()
+    .Then
+    .Returns("SUCCESS");
+```
+
+Note that you can only use `Then` after specifying a number of repetitions for
```
blairconrad: This is consistent with our current usage of "repeats" and "repetitions", but in light of #808, I wonder if we shouldn't, while we're introducing documentation, look for less controversial terms.
Perhaps something like

> Note that you can only use `Then` after specifying the number of times the behavior should occur.

docs/changing-behavior-between-calls.md
```diff
@@ -17,17 +19,40 @@ A.CallTo(() => fakeShop.Address).Returns("123 Fake Street").NumberOfTimes(17);
...
 Fake and then popped off once the number of repetitions defined for a
-call have been exhausted. Thus, it's possible to have a call to a Fake
-act one way, and then another. In order to test the System Under
-Test's retry logic, a Fake service could be configured to fail once
-and then function properly thereafter:
+call have been exhausted.
+
+This can be useful when you have a fake with a default behavior configured
+by a `FakeOptionsBuilder`, and you need to override this behavior for a
+specific test.
```
blairconrad: Perhaps stress a little more why you'd use this and not the other. It's clear to me, but I worry that new readers might miss the subtlety.

Consider something like

> This can be useful when you are unable to use `Then` to speicfy
> different behaviour for successive calls.  For example, when you have
> a fake with a default behavior configured by a
> [`FakeOptionsBuilder`](implicit-creation-options.md), and you need to
> override this behavior for a specific test.

(Note the link added to `implicit-creation-options`.)

As an alternative to referring to FakeOptionsBuilders (or perhaps just a precursor), we could use the example of specifying common behaviour in a `Setup` method (I know, different words for each test framework, but I think it would be clear). I think it would be accessible to a broader section of readers.

docs/changing-behavior-between-calls.md
```diff
@@ -17,17 +19,40 @@ A.CallTo(() => fakeShop.Address).Returns("123 Fake Street").NumberOfTimes(17);
...
+
+Thus, it's possible to have a call to a Fake act one way, and then another. For
+instance, the same effect as the previous sample can be achieved by overriding
+the behavior of the fake:
```
blairconrad: Love the reimplemented example. Very useful to confused people.

(In general, I like the all the docs changes. I think they'd be acceptable as is (except for the broken links I mentioned in [the conversation](https://github.com/FakeItEasy/FakeItEasy/pull/804)), but I'm trying to be super picky just in case we can alleviate confusion for even one person.

thomaslevesque:
> I prefer manual redirect, I think.

You mean just putting a link to the new page?

docs/changing-behavior-between-calls.md
thomaslevesque: OK, I'll reword it

docs/changing-behavior-between-calls.md
thomaslevesque: Good suggestion, thanks

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -164,6 +260,38 @@ public IAnyCallConfigurationWithReturnTypeSpecified<TMember> Where(Func<IFakeObj
                 this.ParentConfiguration.RuleBeingBuilt.ApplyWherePredicate(predicate, descriptionWriter);
                 return this;
             }
+
+            public virtual IAfterCallSpecifiedConfiguration<IReturnValueConfiguration<TMember>> AssignsOutAndRefParametersLazily(Func<IFakeObjectCall, ICollection<object>> valueProducer)
```
blairconrad: I do not think this needs to be virtual.

blairconrad:
> You mean just putting a link to the new page?

Yessir. In light of this confusion, perhaps best to ignore all my ideas for "making docs clearer". :stuck_out_tongue_closed_eyes:

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: Good point. I must have copied it from the enclosing class, which was virtual. I'll remove it.

blairconrad: (Likewise none of the `RuleBuilder` methods need to be virtual, but that was the case even before this PR showed up, so maybe I digress.)

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -118,19 +178,54 @@ public ReturnValueConfiguration(RuleBuilder parentConfiguration)
```
blairconrad: `ParentConfiguration` can now be private, thanks to your test changes.

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -118,19 +178,54 @@ public ReturnValueConfiguration(RuleBuilder parentConfiguration)
 
             public ICallMatcher Matcher => this.ParentConfiguration.Matcher;
 
+            public IReturnValueConfiguration<TMember> Then
+            {
+                get
+                {
+                    var newConfiguration = (RuleBuilder)this.ParentConfiguration.Then;
```
blairconrad: A perfectly valid cast in this case, but these downcasts always make me nervous. What's your opinion on removing the cast and making `RuleBuilder.Then` look more like 

```c#
IVoidConfiguration IThenConfiguration<IVoidConfiguration>.Then => this.Then;

private RuleBuilder Then
{
    get
    {
        var newRule = this.RuleBeingBuilt.CloneCallSpecification();
        this.manager.AddRuleAfter(this.RuleBeingBuilt, newRule);
        return new RuleBuilder(newRule, this.manager, this.asserterFactory);
    }
}
```
? Am I just being silly?

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: > (Likewise none of the RuleBuilder methods need to be virtual, but that was the case even before this PR showed up, so maybe I digress.)

I can change it, but maybe it would be better to do it in a separate PR, since it's unrelated to the changes in this PR.

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: Sure. I didn't like this cast either, and your workaround is quite elegant.

thomaslevesque: (btw, "Sure" wasn't referring to "Am I just being silly", in case that wasn't clear :wink:)

docs/changing-behavior-between-calls.md
thomaslevesque: I also moved this paragraph after the code example, because it was getting difficult to follow (it wasn't clear what the "Thus" in the following paragraph was referring to)

thomaslevesque:
Thanks for the review @blairconrad. I made the requested changes (again, in separate commits; there will be quite a bit of squashing to do before we merge)

tests/FakeItEasy.Specs/FakeItEasy.Specs.ruleset
```diff
@@ -5,4 +5,4 @@
     <Rule Id="CA1003" Action="None" />
     <Rule Id="CA1009" Action="None" />
   </Rules>
-</RuleSet>
\ No newline at end of file
+</RuleSet>
```
blairconrad: Surprising, given no other changes to the file. I'm not that worked up either way. I'm sure you either opened the text view and editorconfig changed things, or you _didn't_ open the text view and regular VS changed things.

tests/FakeItEasy.Specs/ThenSpecs.cs
```diff
@@ -0,0 +1,233 @@
...
+
+            "And the second call returns 3"
+                .x(() => result2.Should().Be(3));
+
+            "And the third call returns 1"
+                .x(() => result3.Should().Be(1));
+        }
+
+        [Scenario]
+        public void NonVoidMethodWithThenAndOverride2(
```
blairconrad: An alternative to the 1/2 naming would be to distinguish based on the ordering of the configs - first with then, or second with then. Not a big deal to me, as the step descriptions make all clear.

tests/FakeItEasy.Specs/FakeItEasy.Specs.ruleset
thomaslevesque: Yes, I must have opened it in text view. Anyway, having the empty line at the end is correct according to the .editorconfig, right?

tests/FakeItEasy.Specs/ThenSpecs.cs
thomaslevesque: You're right, I'll rename them.

thomaslevesque:
I amended some recent commits where I had introduced StyleCop warnings.
And I renamed the specs as discussed.

tests/FakeItEasy.Specs/FakeItEasy.Specs.ruleset
blairconrad: It is. It just looked odd as a change when nothing else happened here. Not a complaint. Just an observation. It's all goo.d

blairconrad:
@thomaslevesque, thanks for indulging me. I think everything looks good. I'd say you might as well squash, since as far as I can tell, @adamralph hasn't been available to be confused by the commits up to now.
I guess the next question is when we want to merge this, since that'll start the "3.0 clock". :smile:

thomaslevesque:
I squashed some of the commits, but since I had quite a lot of conflicts, I'm not sure each commits builds properly... The final result is identical to before the squash, though.

blairconrad:
For what it's worth, I'm not that worked up over every commit compiling. Although it's a nice bonus.
(And it's not that hard to check, with a `git rebase -x .\\build.cmd -i master` (and they don't all compile, if we're keeping score (I'm not trying to)).)

Were I to squash commits, my goal would be more to have a nicer story, with each commit having a well-defined purpose in a logical chain leading to fixing the issue. I do not achieve this goal as often as I'd like. And really, it's not entirely clear that the effort expended is worth it. Sometimes yes, sometimes no. Mostly the changes will just be read during review, and I _think_ reviewers appreciate the effort, but again, there's a cost/benefit to weigh.

blairconrad:
Oh. I realized that my comment above lacked resolution. I don't want any changes.

thomaslevesque:
Thanks @blairconrad.

Let's wait a bit before merging in case @adamralph wants to have a look (and anyway, I think we didn't even decide whether or not we wanted the feature).

blairconrad:
Yeah, I like waiting.
I think we decided we wanted the feature (that's how I understand the [on-hold](https://github.com/FakeItEasy/FakeItEasy/labels/on-hold) label (which could use [an explanation](https://github.com/FakeItEasy/FakeItEasy/wiki/Issue-labels)).
The only thing that would delay a merge (for me) is that it's breaking, so once we merge, we have to decide what other [breaking changes](https://github.com/FakeItEasy/FakeItEasy/labels/breaking) we'd want in 3.0. It's that or leave 3.0 open forever again, which I'd like to avoid.

Perhaps reviewing the public API and/or our positions on the identified breaking issues would be a good idea.

adamralph:
I'm on the road right now so can't contribute much, but I do think we
should leave this PR unmerged until we decide we have a critical mass of
breaking change issues in at least a **ready** state, which together
warrant the major version bump. As mentioned, that is indeed the purpose of
the **on-hold** label.

On 8 Aug 2016 6:55 p.m., "Blair Conrad" <notifications@github.com> wrote:

Yeah, I like waiting.
I think we decided we wanted the feature (that's how I understand the
on-hold <https://github.com/FakeItEasy/FakeItEasy/labels/on-hold> label
(which could use an explanation
<https://github.com/FakeItEasy/FakeItEasy/wiki/Issue-labels>).
The only thing that would delay a merge (for me) is that it's breaking, so
once we merge, we have to decide what other breaking changes
<https://github.com/FakeItEasy/FakeItEasy/labels/breaking> we'd want in
3.0. It's that or leave 3.0 open forever again, which I'd like to avoid.

Perhaps reviewing the public API and/or our positions on the identified
breaking issues would be a good idea.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/FakeItEasy/FakeItEasy/pull/804#issuecomment-238299452>,
or mute the thread

adamralph:
Marked as WIP since the associated issue is `on-hold`.

thomaslevesque:
I think this PR is ready to merge (has been for a while :wink:). I just rebased it onto master.

adamralph:
Sorry, @thomaslevesque looks like it needs another rebase 😟

thomaslevesque:
> Sorry @thomaslevesque, looks like it needs another rebase 😟

Ah, yes, there were conflicts in the ApiApproval files. I did the rebase.

src/FakeItEasy/Configuration/IThenConfiguration.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace FakeItEasy.Configuration
+{
+    using System.Diagnostics.CodeAnalysis;
+
+    /// <summary>
+    /// Configures the next behavior for this call.
+    /// </summary>
+    /// <typeparam name="TInterface">The type of configuration interface to return.</typeparam>
+    public interface IThenConfiguration<out TInterface>
```
blairconrad: I hate to add a late-breaking comment, but shouldn't this hide object members?
So could many other interfaces in the review, and probably others. I only highlight this one because it's completely new.
On the other hand, we could just survey the exposed types and hide their object members later.

tests/FakeItEasy.Tests/Configuration/AnyCallConfigurationTests.cs
```diff
@@ -32,18 +32,18 @@ public class AnyCallConfigurationTests : AutoInitializedFixture
...
-            var result = this.configuration.WithReturnType<int>();
+    // Act
+    var result = this.configuration.WithReturnType<int>();
 
-            // Assert
-            returnConfig.Should().BeSameAs(result);
-        }
+    // Assert
+    returnConfig.Should().BeSameAs(result);
+}
```
blairconrad: formatting: there appears to be a loss of indentation

src/FakeItEasy/Configuration/IThenConfiguration.cs
thomaslevesque: Good point; I'll update it.

thomaslevesque: Some other configuration interfaces that lack `IHideObjectMembers`:

- `IDoNothingConfiguration<TInterface>`
- `IOutAndRefParametersConfiguration<TInterface>`
- `IWhereConfiguration<out T>` (which probably needs to be updated anyway, because the doc comment about `T` is incorrect)

Should I add it to them as well, or do it in a separate PR?

thomaslevesque:
I pushed a few fixup commits.

While I was at it, I also updated `IWhereConfiguration`, which I had missed somehow (just a non-functional change, I only renamed the generic type parameter and updated the incorrect doc comment).

src/FakeItEasy/Configuration/IThenConfiguration.cs
blairconrad: I think put them in a separate PR. The change isn't really related to this issue, and it might make the change (not that anyone will notice) easier to spot if it's a separate issue and line in the release notes.
Also, it gets this PR in faster!

blairconrad:
Thanks, @thomaslevesque. The changes look great. Leaving here in case @adamralph wants to eyeball it. I'm not sure whether his comment 

> Sorry @thomaslevesque, looks like it needs another rebase 😟

Means he was about to merge or not.

But as far as I'm concerned, you're ready for a squash and merge.

Thank you!

thomaslevesque:
@adamralph did you have a chance to review the fixup commits? Please let me know when I can squash

adamralph:
Sorry @thomaslevesque I actually haven't had a chance to review this PR at all yet. I'll try and find some time to do so tomorrow, but in the meantime, you may as well go ahead and squash, since having the commits separate won't benefit me.

Also, if you're both confident in the changes, you don't have to wait for me to review before merging. It's your call.

thomaslevesque:
> Sorry @thomaslevesque I actually haven't had a chance to review this PR at all yet. I'll try and find some time to do so tomorrow, but in the meantime, you may as well go ahead and squash, since having the commits separate won't benefit me.

OK, I squashed

> Also, if you're both confident in the changes, you don't have to wait for me to review before merging. It's your call.

I'm reasonably confident, but it's a pretty big change, so I wouldn't mind another pair of eyes, if you can manage it.

adamralph:
@thomaslevesque which are the fix up commits? In fact, it would be *easier* for me to review this if they were already squashed in wherever they need to be.

thomaslevesque:
> @thomaslevesque which are the fix up commits? In fact, it would be easier for me to review this if they were already squashed in wherever they need to be.

I already squashed the fixup commits; the remaining ones were intended to stay, since they represent logical sets of changes. I could perhaps squash some of them, but it would become difficult to maintain a logical separation.

adamralph:
Ah, apologies, I thought there was still some squashing to do.

OK, I'll review ASAP. Sorry for the delay, I hope you're in no huge rush with this.

thomaslevesque:
I can squash commits 1, 3 and 4, if you prefer. Would it help?

> OK, I'll review ASAP. Sorry for the delay, I hope you're in no huge rush with this.

No, not in a rush. This PR is almost 4 months old, so a day or two won't make a huge difference :wink:

adamralph:
> I can squash commits 1, 3 and 4, if you prefer. Would it help?

If each commit has independent value, i.e. it could potentially have been in a separate PR on it's own, then I'd leave it separate. Otherwise, I'd squash.

---

(The following is academic/observational, and somewhat of a brain dump. I'll leave it to you to judge how much, if any of it, you want to apply.)

The commit pattern I usually aim for, and find the easiest to review, is:

*{refactor<sub>1</sub>, ..., refactor<sub>x</sub>,}* **red**, **green**, *{refactor<sub>x+1</sub>, ..., refactor<sub>n</sub>}*

Pre- and post- refactorings exist only when they are relevant and beneficial. Sometimes it can be better to combine the red and green commits, specially if the red commit alone would fail to compile.

Based purely on my reading of the commit messages:

- 1, 3, 4 and 8 look like refactorings
- 2, 5 and 10 look like they could be combined to become the green commit.
- 6 looks like the red commit.
- 7 looks like a candidate for squashing into the commit which caused the warnings to appear, but if that was more than commit, I'd leave 7 separate. The more pedantic approach in that case would be to tease 7 apart into separate commits, and squash each of those into the commit which it fixes up, but I think that's overkill
- 9 could arguably be part of the green commit but I think it's OK (perhaps better, even) to leave it separate

adamralph:
BTW another way to form a separate red commit is to write failing tests and empty/exception-throwing  implementations, with the green commit filling in the implementations. In this case I guess that could be achieved by combining commits 2 and 6.

I'm not a religious advocate of TDD though (and specially not at the "unit" testing level). It's often not practical and/or contrived to form separate red and green commits just to tick that box.

thomaslevesque:
Good advice, thanks!

However, it has become pretty hard to rearrange the commits. Any interactive rebase that does more than squashing successive commits produces hard-to-fix conflicts. I've already been through a lot of pain to make the commits you see now, so I'd rather not do it all over again... I'll try to follow your advice next time :wink:

adamralph:
Sure, no problem.

src/FakeItEasy/Configuration/IAfterCallSpecifiedConfiguration.cs
adamralph: Is the breakage worth considering for 3.0?

blairconrad: While we're breaking things, why not?
I assume very few people refer to these types by name.

src/FakeItEasy/Configuration/BuildableCallRule.cs
```diff
@@ -160,6 +160,24 @@ public virtual void ApplyWherePredicate(Func<IFakeObjectCall, bool> predicate, A
...
+            clone.wherePredicates.AddRange(this.wherePredicates);
+            return clone;
+        }
+
+        /// <summary>
+        /// When overridden in a derived class, returns a new instance of the same type and copies the call
+        /// specification members for that type.
+        /// </summary>
+        /// <returns>The cloned rule.</returns>
+        protected abstract BuildableCallRule CloneCallSpecificationCore();
```
adamralph: I can't stop myself from detecting a code smell here:

- Only the docs are telling an implementer that they should return the same type as the implementer. Should there be a generic type param enforcing this instead?
- The "core" suffix. Is there a name that could better express the intent of this method and distinguish it from `CloneCallSpecification`?

That said, the type is internal so it's not of utmost importance to address these concerns.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -182,6 +180,22 @@ internal virtual void ClearUserRules()
...
         }
 
+        /// <summary>
+        /// Adds a call rule to the fake object after the specified rule.
+        /// </summary>
+        /// <param name="previousRule">The rule after which to add a rule.</param>
+        /// <param name="newRule">The rule to add.</param>
+        internal void AddRuleAfter(IFakeObjectCallRule previousRule, IFakeObjectCallRule newRule)
+        {
+            var previousNode = this.AllUserRules.Nodes().FirstOrDefault(n => n.Value.Rule == previousRule);
```
adamralph: `Nodes` isn't built in!? 😱

tests/FakeItEasy.Specs/ThenSpecs.cs
```diff
@@ -0,0 +1,233 @@
...
+                    foo.Bar();
+                    foo.Bar();
+                    foo.Bar();
+                });
+
+            "Then the first delegate is invoked once"
+                .x(() => A.CallTo(() => action1()).MustHaveHappened(Repeated.Exactly.Once));
+
+            "And the second delegate is invoked twice"
+                .x(() => A.CallTo(() => action2()).MustHaveHappened(Repeated.Exactly.Twice));
```
adamralph: Is it worth ensuring the order of calls in this spec?

docs/limited-call-specifications.md
```diff
@@ -1,38 +1 @@
...
-and then function properly thereafter:
-
-```csharp
-// set up an action that can run forever, unless superseded
-A.CallTo(() => fakeService.DoSomething()).Returns("SUCCESS");
-
-// set up a one-time exception which will be used for the first call
-A.CallTo(() => fakeService.DoSomething()).Throws<Exception>().Once();
-```
+This page has moved, see [Changing behavior between calls](changing-behavior-between-calls.md).
```
adamralph: 👍 nice touch

src/FakeItEasy/Configuration/IAfterCallSpecifiedConfiguration.cs
adamralph: Raised https://github.com/FakeItEasy/FakeItEasy/issues/910

thomaslevesque:
> Out of interest, how is it that Then can be a property whereas other things such as Once need to be methods?

I don't know, it just felt more natural...

> Or could Once, etc. actually be properties too?

Maybe they could, but I don't think they should, because they're not side-effect free. `Then,` on the other hand, doesn't mutate the current rule, it just creates a new one with the same call specification.

src/FakeItEasy/Configuration/IPropertySetterConfiguration.cs
```diff
@@ -4,11 +4,11 @@ namespace FakeItEasy.Configuration
...
     /// </summary>
     public interface IPropertySetterConfiguration :
-        IExceptionThrowerConfiguration,
+        IExceptionThrowerConfiguration<IPropertySetterConfiguration>,
         ICallbackConfiguration<IPropertySetterConfiguration>,
-        ICallBaseConfiguration,
+        ICallBaseConfiguration<IPropertySetterConfiguration>,
         IAssertConfiguration,
-        IDoNothingConfiguration
+        IDoNothingConfiguration<IPropertySetterConfiguration>
```
thomaslevesque: Not in this case; we're not declaring `IDoNothingConfiguration<out TInterface>` here, just stating that `IPropertySetterConfiguration` inherits from it. (and `IDoNothingConfiguration<out TInterface>`  does have the `out`).

Anyway, I'm not sure how useful it is to declare the fluent interfaces as covariant. We do it because R# says we can, but I don't think we ever take advantage of it.

src/FakeItEasy/Configuration/BuildableCallRule.cs
thomaslevesque: - Generic type param: how ? `BuildableCallRule` itself would need to become generic, using the [CRTP](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern). Is this what you had in mind?

        internal abstract class BuildableCallRule<T>
            : IFakeObjectCallRule
              where T : BuildableCallRule<T>
        {
            ...
            protected abstract T CloneCallSpecificationCore();
            ...
        }

        internal class AnyCallCallRule
            : BuildableCallRule<AnyCallRule>
        {
            ...
            protected override AnyCallRule CloneCallSpecificationCore()
            {
                return new AnyCallCallRule
                {
                    argumentsPredicate = this.argumentsPredicate
                };
            }
        }

    I think it's a little overkill...

- "core" suffix: it's a common naming convention when using the Template Method pattern (very frequent in the BCL). But if you have a better idea, I'm open to suggestions.

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: Nope... but pretty easy to implement

tests/FakeItEasy.Specs/ThenSpecs.cs
thomaslevesque: I guess it is. I'll update the specs.

src/FakeItEasy/Configuration/BuildableCallRule.cs
adamralph: Yeah, I guess CRTP is what I had in mind. Let's leave it as-is. It's internal, so it's only contributors which have to follow that specific doc, and I'm sure we can manage, given our current tempo of creating new types that inherit from `BuildableCallRule`.

adamralph:
Thanks @thomaslevesque !

adamralph:
Doh! I merged without giving you a chance to squash the fixup! commit...

thomaslevesque:
Thanks for the review!

> Doh! I merged without giving you a chance to squash the fixup! commit...

Ah, well... no big deal!

Issue #805, opened 2016-07-25
Request for accessing some internals


morgul77:
Hi

We are doing lots of integration testing using Autofac and Fakeiteasy. However, there's sometimes a hole in what we test, because we A: have to fake it or B: wait for the side effect. 

I really like the Fakeiteasy syntax, and coupled with Autofac's support for dynamic proxy, I have now managed to get a prototype of a concept that lets our tests know if a particular piece of code has been called, and being able to inspect the return value.

Registration in autofac (Dump() is from linqpad):
```c#
var builder = new ContainerBuilder();
	builder.RegisterType<CA>().AsImplementedInterfaces();	
	builder.RegisterType<B>().AsImplementedInterfaces().WhenCompletedCallMatches(builder, x => x.Do(),() => "no return value".Dump())
													  .WhenCompletedCallMatches(builder, x => x.Do(A<string>.That.Matches(y => y.Equals("this"))), s => ("if this is " + s).Dump())
													  .WhenCompletedCallMatches(builder, x => x.Do(A<string>.That.Matches(y => y.Equals("that"))), s => ("if that is " + s).Dump()); ;
	using (var container = builder.Build())
	{
		var ia = container.Resolve<IA>();
		ia.Do();
		ia.Do("this");
		ia.Do("that");
		ia.Do("something else");
	}
```

The actual code, is perhaps misusing Fakeiteasy, but I think it helps, because I can use the ExpressionCallMatcher, the parser, and other useful things. Code as follows:
```c#
public class Interceptor : IInterceptor
{
	public virtual void Intercept(IInvocation invocation)
	{
		invocation.Proceed();
	}
}
// Define other methods and classes here
public static class ContainerBuilderExtensions
{
	public static IRegistrationBuilder<T, ConcreteReflectionActivatorData, SingleRegistrationStyle> WhenCompletedCallMatches<T>(this IRegistrationBuilder<T, ConcreteReflectionActivatorData, SingleRegistrationStyle> registration, ContainerBuilder builder, Expression<Action<T>> callSpecification, Action actionToInvoke)
	{
		var interceptorName = Guid.NewGuid().ToString();
		EnableInterceptionOnce(registration);
		registration.InterceptedBy(interceptorName);
		var fakeItEasyInterception = A.Fake<IInterceptionListener>(d => d.ConfigureFake(
		t =>
		{
[...]

PR #807, opened 2016-07-26
Allow callmatching to happen from external interpection

morgul77:
Open ArgumentCollection constructor for external usage ,and added CompletedFakeObjectCallExtensions.Matches method for callspecifications with return values.

See issue #805

Not used to doing pull requests, so if I missed something, please let me know.

### DIFF
--- a/src/FakeItEasy/CompletedFakeObjectCallExtensions.cs
+++ b/src/FakeItEasy/CompletedFakeObjectCallExtensions.cs
@@ -25,7 +25,7 @@ namespace FakeItEasy
         [SuppressMessage("Microsoft.Design", "CA1011:ConsiderPassingBaseTypesAsParameters", Justification = "The compiler would not be able to figure out the type.")]
         public static IEnumerable<ICompletedFakeObjectCall> Matching<TFake>(this IEnumerable<ICompletedFakeObjectCall> calls, Expression<Action<TFake>> callSpecification)
         {
-            return Matching(calls, callSpecification);
+            return InternalMatching(calls, callSpecification);
         }
 
         /// <summary>
@@ -40,7 +40,7 @@ namespace FakeItEasy
         [SuppressMessage("Microsoft.Design", "CA1011:ConsiderPassingBaseTypesAsParameters", Justification = "The compiler would not be able to figure out the type.")]
         public static IEnumerable<ICompletedFakeObjectCall> Matching<TFake, TResult>(this IEnumerable<ICompletedFakeObjectCall> calls, Expression<Func<TFake, TResult>> callSpecification)
         {
-            return Matching(calls, callSpecification);
+            return InternalMatching(calls, callSpecification);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/FakeItEasy/Configuration/ArgumentCollection.cs
```diff
@@ -25,6 +25,17 @@ public class ArgumentCollection
         ///   Initializes a new instance of the <see cref = "ArgumentCollection" /> class.
         /// </summary>
         /// <param name = "arguments">The arguments.</param>
+        /// <param name = "method">The method.</param>
+        [DebuggerStepThrough]
+        public ArgumentCollection(object[] arguments, MethodInfo method)
+            : this(arguments, GetArgumentNames(method))
+        {
+        }
```
thomaslevesque: Why does this constructor need to be public? How do you plan to use it?

thomaslevesque:
Thanks for the PR @morgul77!

At first glance, it seems good. I'm just wondering why you need to make the `ArgumentCollection` constructor public (not that I would be strongly opposed to expose it publicly, but I'm curious).

> Not used to doing pull requests, so if I missed something, please let me know.

You did a pretty good job! Not even a StyleCop or FxCop warning :wink:

src/FakeItEasy/CompletedFakeObjectCallExtensions.cs
```diff
@@ -25,14 +25,7 @@ public static class CompletedFakeObjectCallExtensions
...
         {
-            var factory = ServiceLocator.Current.Resolve<IExpressionCallMatcherFactory>();
-            var callExpressionParser = ServiceLocator.Current.Resolve<ICallExpressionParser>();
-            var matcher = factory.CreateCallMatcher(callExpressionParser.Parse(callSpecification));
-
-            return
-                from call in calls
-                where matcher.Matches(call)
-                select call;
+            return Matching(calls, callSpecification);
```
thomaslevesque: Are you sure this doesn't result in infinite recursion? I think it calls the same method... You probably need a cast to `Expression`.

morgul77:
1. Yes, exactly the same. I could not find something that would accept ```Expression<Func<TFake, TResult>> ``` and  ```Expression<Action<TFake>>``` at the same time as public method. Way better have one private method. Doing it.

2. The argument constructor is needed for the CastleInvocationCallAdapterExternal to create arguments. See attached linqpad script. Cleaned up a bit, not using fakeiteasy interceptors anymore. And I cluttered it with CastleInvocationCallAdapterExternal.
[autofac testing2.zip](https://github.com/FakeItEasy/FakeItEasy/files/384973/autofac.testing2.zip)

src/FakeItEasy/CompletedFakeObjectCallExtensions.cs
morgul77: That was quick. I did think it went to my private repository first :)

morgul77: Not sure if casting is preferable to another methodname?

thomaslevesque: > That was quick. I did think it went to my private repository first :)

It does! The code is still only in your repo. But since you have a PR open on that branch, Github immediately updates the PR.



> Not sure if casting is preferable to another methodname?

Both ways are fine by me :+1:

src/FakeItEasy/Configuration/ArgumentCollection.cs
blairconrad: Probably to create a custom implementation of `IInterceptedFakeObjectCall`. Essentially an external `CastleInvocationCallAdapter`.

blairconrad: Ah. I've read the note in the issue. It's so.

blairconrad: That being said, I'm resistant to exposing the constructor. A need for this sort of thing just feels like it's outside FakeItEasy's scope.

blairconrad:
Closing, as (I think is) suggested by https://github.com/FakeItEasy/FakeItEasy/issues/805#issuecomment-235556124.
If I misinterpreted, we can always reopen.
Or if the new approach doesn't pan out, we can always revisit this or another approach.
Thanks for your interest, @morgul77!

Issue #531, opened 2015-08-26
Support .NET Platform Standard 1.6 (inc. .NET Core 1.0)


khellang:
I figured it's time to start this discussion :smile:

Here are some threads for reference:

https://github.com/castleproject/Core/issues/90
https://github.com/nsubstitute/NSubstitute/issues/192
https://github.com/Moq/moq4/issues/168

There's a PR pending for .NET Core support in Castle.Core at https://github.com/castleproject/Core/pull/92

adamralph:
As soon as Castle.Core support the dotnet TFN I'm happy to do this.

jeremymeng:
Would it be okay to start the work based on https://github.com/castleproject/Core/tree/netcore? Most of the big changes are already in. I am willing to give it a first stab.

adamralph:
@jeremymeng I think would be fine to grab some custom binary to reference for now, so long as your are prepared to rebase/squash your commits back to something clean (or have someone else do that) when Castle for netcore (or whatever other name is goes by) is on NuGet.

jchannon:
Where are we at with this?

thecodejunkie:
We have a green build against rc2 now (including tests, i.e all tests that do not require FakeItEasy) https://github.com/NancyFx/Nancy/pull/2418 and would be happy to help transfer knowledge on rc2 changes to help this move along

blairconrad:
Hey, @jchannon and @thecodejunkie. As you can see, @jeremymeng had kindly been doing some work on this for us, but the core team had all been focusing on FIE 2.0, which I think we're nearly ready to release the second (and final, one hopes) RC for.

We'd be grateful for any insight you have to offer, but I'm not going to be available for the next two weeks, so can't personally do anything. I'll leave it up to @thomaslevesque and @adamralph to coordinate with you.

jeremymeng:
@jchannon @thecodejunkie 

There are a couple of work items now
- [ ] Replacement of binary serialization used in `FileStorage`.  So far I haven't had success in using a Json or Xml serializer.
- [ ] Seemly incorrect usage of my custom AssemblyLoadContext.  No tests are hitting the code path where I made changes.  I am not sure how to test it yet.
- [ ] Move to RC2
- [ ] integration into build system, etc.

Any helps are welcome :smile:

thomaslevesque:
@jeremymeng, thanks for all your work on this task!

> Any helps are welcome :smile:

[...]

PR #814, opened 2016-07-31
[WIP] add dotnet targets to rakefile

blairconrad:
Add a target (of the form *_dotnet) for each existing target that we have a .NET Standard analogue for.
Also add a "dotnet" target to run all .NET 4.0-related targets as well as the .NET Standard targets.

One downside is that we end up building the .NET 4.0 projects twice.

Connects to #531

### DIFF
--- a/rakefile.rb
+++ b/rakefile.rb
@@ -10,11 +10,13 @@ gitlink_command = "packages/gitlink.2.3.0/lib/net45/GitLink.exe"
 xunit_command   = "packages/xunit.runner.console.2.0.0/tools/xunit.console.exe"
 
 solution        = "FakeItEasy.sln"
+dotnet_solution = "FakeItEasy-NetCore.sln"
 assembly_info   = "src/CommonAssemblyInfo.cs"
 version         = IO.read(assembly_info)[/AssemblyInformationalVersion\("([^"]+)"\)/, 1]
 version_suffix  = ENV["VERSION_SUFFIX"]
 repo_url        = "https://github.com/FakeItEasy/FakeItEasy"
 nuspec          = "src/FakeItEasy/FakeItEasy.nuspec"
+dotnet_nuspec   = "src/FakeItEasy.dotnet/FakeItEasy.nuspec"
 analyzer_nuspec = "src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.nuspec"
 logs            = "artifacts/logs"
 output          = "artifacts/output"
@@ -26,6 +28,10 @@ gitlinks        = ["FakeItEasy", "FakeItEasy.Analyzer"]
 unit_tests = [
   "tests/FakeItEasy.Tests/bin/Release/FakeItEasy.Tests.dll",
   "tests/FakeItEasy.Analyzer.Tests/bin/Release/FakeItEasy.Analyzer.Tests.dll"
+]
+
+dotnet_unit_test_directories = [
+  "tests/FakeItEasy.Tests.dotnet"
 ]
 
 integration_tests = [
@@ -95,6 +101,9 @@ end
 desc "Execute default tasks"
 task :default => [ :vars, :gitlink, :unit, :integ, :spec, :pack ]
 
+desc "Execute default tasks for .NET Core solution"
+task :dotnet => [ :default, :unit_dotnet, :pack_dotnet]
+
 desc "Print all variables"
 task :vars do
   print_vars(local_variables.sort.map { |name| [name.to_s, (eval name.to_s)] })
@@ -111,6 +120,11 @@ directory logs
 desc "Clean solution"
 task :clean => [logs] do
   run_msbuild solution, "Clean", msbuild_command
+end
+
+desc "Clean .NET Core solution"
+task :clean_dotnet => [logs] do
+  run_msbuild dotnet_solution, "Clean", msbuild_command
 end
 
 desc "Update version number and create pull request, milestone, release, and release checklist issue"
@@ -222,6 +236,10 @@ task :build => [:clean, :restore, logs] do
   run_msbuild solution, "Build", msbuild_command, packages
 end
 
+task :build_dotnet => [:clean_dotnet, :restore, logs] do
+  run_msbuild dotnet_solution, "Build", msbuild_command, packages
+end
+
 desc "GitLink PDB's"
 exec :gitlink => [:build] do |cmd|
   cmd.command = gitlink_command
@@ -233,6 +251,11 @@ directory tests
 desc "Execute unit tests"
 task :unit => [:build, tests] do
     run_tests(unit_tests, xunit_command, tests)
+end
+
+desc "Execute .NET Core unit tests"
+task :unit_dotnet => [:build_dotnet, tests] do
+    run_dotnet_tests(dotnet_unit_test_directories, tests)
 end
 
 desc "Execute integration tests"
@@ -256,6 +279,12 @@ desc "create the nuget package"
 exec :pack => [:build, output] do |cmd|
   cmd.command = nuget_command
   cmd.parameters "pack #{nuspec} -Version #{version}#{version_suffix} -OutputDirectory #{output}"
+end
+
+desc "create the nuget package for .NET Core"
+exec :pack_dotnet => [:build_dotnet, output] do |cmd|
+  cmd.command = nuget_command
+  cmd.parameters "pack #{dotnet_nuspec} -Version #{version}-dotnet#{version_suffix} -OutputDirectory #{output}"
 end
 
 desc "create the analyzer nuget package"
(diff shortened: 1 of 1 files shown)


rakefile.rb
```diff
@@ -344,6 +373,18 @@ def run_tests(test_assemblies, command, result_dir)
...
 end
 
+def run_dotnet_tests(test_directories, result_dir)
+  test_directories.each do |test_directory|
+    result_file = File.expand_path(File.join(result_dir, File.basename(test_directory)) + '.TestResults.xml')
+
+    cmd = Exec.new
+    cmd.command = 'dotnet'
+    cmd.working_directory = test_directory
+    cmd.parameters "test -c Release --no-build -xml #{result_file} -nologo -notrait \"explicit=yes\""
```
blairconrad: `dotnet test` hated the `-noshadow` option.

blairconrad:
There's a lot of duplication. I'm not sure it's worthwhile trying to avoid it.
Also, we may end up changing `dotnet`. But that's okay.

rakefile.rb
```diff
@@ -10,11 +10,13 @@
 xunit_command   = "packages/xunit.runner.console.2.0.0/tools/xunit.console.exe"
 
 solution        = "FakeItEasy.sln"
+dotnet_solution        = "FakeItEasy-NetCore.sln"
```
thomaslevesque: Is the different indentation intentional?

rakefile.rb
```diff
@@ -10,11 +10,13 @@
 xunit_command   = "packages/xunit.runner.console.2.0.0/tools/xunit.console.exe"
 
 solution        = "FakeItEasy.sln"
+dotnet_solution        = "FakeItEasy-NetCore.sln"
 assembly_info   = "src/CommonAssemblyInfo.cs"
 version         = IO.read(assembly_info)[/AssemblyInformationalVersion\("([^"]+)"\)/, 1]
 version_suffix  = ENV["VERSION_SUFFIX"]
 repo_url        = "https://github.com/FakeItEasy/FakeItEasy"
 nuspec          = "src/FakeItEasy/FakeItEasy.nuspec"
+dotnet_nuspec          = "src/FakeItEasy.dotnet/FakeItEasy.nuspec"
```
thomaslevesque: same here

rakefile.rb
```diff
@@ -95,6 +101,9 @@
 desc "Execute default tasks"
 task :default => [ :vars, :gitlink, :unit, :integ, :spec, :pack ]
 
+desc "Execute default tasks for .NET Core solution"
+task :dotnet => [ :default, :unit_dotnet, :pack_dotnet]
```
thomaslevesque: This will also execute non-dotnet tasks... Shouldn't it be `[ :vars, :gitlink, :unit_dotnet, :pack_dotnet]`?

thomaslevesque: or maybe not `:gitlink`, I'm not sure...

rakefile.rb
thomaslevesque: Does it matter? I don't see why shadow-copying would be an issue anyway.

thomaslevesque:
Thanks @blairconrad. I made a few comments on the diff.

rakefile.rb
blairconrad: @adamralph [suggested](https://github.com/FakeItEasy/FakeItEasy/issues/531#issuecomment-236426591) having `dotnet` run the whole shebang. I saw no harm.

rakefile.rb
blairconrad: No. I just use a proportional font and don't pay attention to `=` alignment. Will amend. Thanks!

rakefile.rb
blairconrad: I just wanted to highlight the difference. It's probably fine.
Shadow-copying can make running of non-netstandard FIE-using tests slower. But that's not a big deal either.

blairconrad:
Thanks for the notes, @thomaslevesque!

rakefile.rb
blairconrad: I should update the description. Unless you'd rather this target _not_ run the regular stuff.

thomaslevesque: I don't mind either way, but the description should probably make it clear.

blairconrad:
Whoops. Looks like this relies on the dotnet packages being restored. It looked to me like the MSBuild had covered that, but it seems not. Or not for Castle.Core, at least. Will add an update for that.

adamralph:
It seems that the `dotnet` moniker has been used in 3 different ways:

* prefix, e.g. `dotnet_solution`
* suffix, e.g. `clean_dotnet`
* in the middle, e.g. `run_dotnet_tests`

Shall we use one convention throughout?

rakefile.rb
```diff
@@ -95,6 +101,9 @@
 desc "Execute default tasks"
 task :default => [ :vars, :gitlink, :unit, :integ, :spec, :pack ]
 
+desc "Execute default tasks plus .NET Core tasks"
```
adamralph: It's not really ".NET Core", it's ".NET Platform Standard".

".NET Core" is just one of the platforms that our ".NET Platform Standard" DLL supports.

rakefile.rb
```diff
@@ -95,6 +101,9 @@
 desc "Execute default tasks"
 task :default => [ :vars, :gitlink, :unit, :integ, :spec, :pack ]
 
+desc "Execute default tasks plus .NET Core tasks"
+task :dotnet => [ :default, :unit_dotnet, :pack_dotnet]
```
adamralph: I don't think this is the right approach. We don't want to build the net40 solution and then the dotnet solution. We want to build the dotnet solution only. It should really be the `dotnet` analog of the `default` task, i.e.

```ruby
+task :dotnet => [ :vars, :gitlink_dotnet, :unit_dotnet, :integ_dotnet, :spec_dotnet, :pack_dotnet ]
```

adamralph:
I'd like to propose a different approach. If we didn't mind taking a dep on the half baked netstd tooling, we wouldn't preserve the current solution and our rakefile would treat the net40 and netstd assemblies as equal first class citizens. How about we build that rakefile as rakefile-dotnet and leave the current rakefile as is? Equally, we can have a build-dotnet.cmd which wraps up the call to it. (rake takes a param for the rakefile it should run). Yes, it's duplication, but no more than the two solutions. When we take a dep on the tooling we can just delete one rakefile. We also have the option of pulling out the stuff that is common to both rakefiles into a common file.

rakefile.rb
blairconrad: And all the `_dotnet` packages will build/test/pack both the .NET Platform Standard and the .NET 4.0 versions? That would seem to follow along from your suggestion that we be able to complete the whole build by calling the single `dotnet` target.

adamralph: Indeed. I think the `dotnet` task should be what `default` would be if we were taking a dep on the half baked tooling.

blairconrad:
> It seems that the `dotnet` moniker has been used in 3 different ways

I'd opted for Englishy usage.
In `dotnet_solution` and `run_dotnet_tests`, its an adjective. In `clean_dotnet` we've more of a verb-noun vibe.

Although if we take your "2 rakefiles" approach, the problem probably goes away.

blairconrad:
> If we didn't mind taking a dep on the half baked netstd tooling, we wouldn't preserve the current solution

So replace FakeItEasy.sln with FakeItEasy-NetCore.sln?

I'm not terribly excited about the state of the tooling. Especially when I've been trying since yesterday afternoon to get the `restore/build` sequence working.

We need nuget-3.5.0-beta2 in order for `nuget restore` to work on `FakeItEasy-NetCore.sln`, and even when that succeeds, we still need a `dotnet restore` in FakeItEasy.Tests.dotnet.
Oh, and `MSBuild` is chokcing on `FakeItEasy-NetCore.sln` when we start from a clean build. Even @jeremymeng's AppVeyor build just pops in and builds the `FakeItEasy.dotnet.csproj` directly.
Which works, but then the pack was failing because the two `clean` targets were run in a bad order, so the non-dotnet FakeItEasy DLL is gone by the time we pack.

However, using the two rakefiles would probably clean a lot of that up (one `clean` target, in particular).

But I don't understand why the "2 rakefile" plan means we have to drop the current sln.

rakefile.rb
blairconrad: Sorry, I'd written "`_dotnet` packages". I meant "`_dotnet` targets". I think you understood that anyhow.

rakefile.rb
blairconrad: Right. Darn it, that probably wouldn't've been a problem had we settled on a moniker! :ghost:

thomaslevesque:
:+1: for a separate rakefile

adamralph:
> So replace FakeItEasy.sln with FakeItEasy-NetCore.sln?

If we were to enforce use of the new tooling then yes, we would have a single solution.

adamralph:
> But I don't understand why the "2 rakefile" plan means we have to drop the current sln.

It doesn't. I think you must have misunderstood.

The proposal is to keep the two current solution files, and build a new rakefile-dotnet(rakefile-NetCore? - oops we have yet another moniker) on top of FakeItEasy-NetCore.sln.

blairconrad:
Sure. I just don't understand why we couldn't defer enforcing the new tooling by keeping two slns.
Or if we were going to enforce the new tooling why we wouldn't just have one rakefile.

adamralph:
I guess our last comments crossed in the air :wink:

blairconrad:
> we wouldn't preserve the current solution

You didn't mean `.sln`, did you? You meant the "approach in the PR".

PR #815, opened 2016-08-02
[WIP] Add a dotnet rakefile

blairconrad:
This is attempt 2!

for discussion only

Connects to #531


rakefile-dotnet.rb
```diff
@@ -0,0 +1,373 @@
...
+msbuild_command = "C:/Program Files (x86)/MSBuild/14.0/Bin/MSBuild.exe"
+if !File.file?(msbuild_command)
+  raise "MSBuild not found"
+end
+
+nuget_command   = ".nuget/nuget.exe"
+gitlink_command = "packages/gitlink.2.3.0/lib/net45/GitLink.exe"
+xunit_command   = "packages/xunit.runner.console.2.0.0/tools/xunit.console.exe"
+
+solution        = "FakeItEasy-NetCore.sln"
```
blairconrad: @jeremymeng (or anyone), do you know why executing 

```C:/Program Files (x86)/MSBuild/14.0/Bin/MSBuild.exe" FakeItEasy-NetCore.sln /target:Build /p:configuration=Release /nr:false /verbosity:minimal /nologo /fl /flp:LogFile=artifacts/logs/Build.log;Verbosity=Detailed;PerformanceSummary /p:NuGetPackagesDirectory=C:/PortableApps/Documents/Source/GitHub/FakeItEasy/packages```

here results in 

```
  Project FakeItEasy.Tests (.NETCoreApp,Version=v1.0) will be compiled because expected inputs are missing
  Compiling FakeItEasy.Tests for .NETCoreApp,Version=v1.0
C:\PortableApps\Documents\Source\GitHub\FakeItEasy\tests\FakeItEasy.Tests.dotnet\project.json(44,35): error NU1001:
The dependency FluentAssertions >= 4.6.3 could not be resolved. [C:\PortableApps\Documents\Source\GitHub\FakeItEasy\tests\FakeItEasy.Tests.dotnet\FakeItEasy.Tests.dotnet.xproj]
C:\PortableApps\Documents\Source\GitHub\FakeItEasy\tests\FakeItEasy.Tests.dotnet\project.json(45,34): error NU1001:
The dependency Microsoft.CodeAnalysis.Analyzers >= 1.1.0 could not be resolved. [C:\PortableApps\Documents\Source\GitHub\FakeItEasy\tests\FakeItEasy.Tests.dotnet\FakeItEasy.Tests.dotnet.xproj]
C:\PortableApps\Documents\Source\GitHub\FakeItEasy\tests\FakeItEasy.Tests.dotnet\project.json(59,55): error NU1001:
The dependency Microsoft.CodeAnalysis.Analyzers >= 1.1.0 could not be resolved. ```

Can one note build ` .xproj` from within a .`sln` using MSBuild? The rakefile runs the `nuget restore` earlier, so I'd think that wasn't it, but 
```
dotnet restore
dotnet build -c Release
``` 

does pass. 

Is there any way forward but to drop the project from the `.sln`? Which I'm loathe to do.

Thank for any help you can bring.

adamralph:
:+1: I much prefer this approach.

Although there will be now 2 files to maintain (for some time at least), it should be fairly easy to keep the relevant bits in sync using a diff viewer.

I guess it's very much a WIP, since the two rakefiles look identical apart from the `solution` variable. The `dotnet` one will also have to run FakeItEasy.Tests.dotnet and target a different nuspec, etc.

BTW - why do we not have specs and integration tests for `dotnet`? (Added to [list of questions](https://github.com/FakeItEasy/FakeItEasy/issues/531#issuecomment-236366247).)

blairconrad:
@adamralph, sorry, I wasn't clear enough with my "for discussion only". As you say, I've done nothing to the new rakefile except change the `solution`, which brings me to the [compilation problem I mentioned earlier](https://github.com/FakeItEasy/FakeItEasy/pull/815#discussion-diff-73132820). Really I only started the PR to be able to talk about the problem.

We don't have specs and integration tests because we haven't got 'round to them yet. @jeremymeng has [a branch](https://github.com/jeremymeng/FakeItEasy/tree/integration-tests) that begins to include integration tests, with (very few) failing ones right now, mostly because the paths are different for the projects (and the working dir for `dotnet test` is not the same as the .NET 4.0 working dir). 

Mr. Meng and I [had been chatting](https://github.com/blairconrad/FakeItEasy/tree/integration) about the changes. Actually, I'd forgotten. You probably want [my branch instead](https://github.com/blairconrad/FakeItEasy/tree/integration), since I've cleaned some things up, and have fewer failures.

adamralph:
Thanks @blairconrad I've updated the checklist items so they can be checked off as each project is added.

blairconrad:
Regarding my compilation error, Nancy and [StructureMap](https://github.com/structuremap/structuremap/blob/0a520954ec35821c3c448cb3ac26d39adad13224/rakefile.rb#L70) seem to have adopted a "test xproj as endpoints of the builds" approach and are using `dotnet restore` and `dotnet build` run on those projects to drive the entire build.
Given our duplicated projects for 4.0 and .NET Standard, I don't see how that's an option…

rakefile-dotnet.rb
jeremymeng: > Can one not build .xproj from within a .sln using MSBuild?

Maybe because of a bug in dotnet where it doesn't work well with the custom nuget restore package directory.

blairconrad: Oh, maybe! Will try to try without custom dir tonight.

blairconrad: Removing the custom dir did not help. :-(

blairconrad: But calling `dotnet restore` after cleaning but before using MSBuild to build did!

blairconrad:
Progress! The new rakefile will now build, (unit) test, and pack both .NET Standard and old-style projects.
GitLinks is still run on the original DLL only, and I've not shifted the nuspec locations around or anything, but I'm pleased with the night's progress. I'm off to read and sleep. Consecutively.

adamralph:
Not important now, but just a reminder that as part of this, `rakefile-dotnet.rb` and `build-dotnet.cmd` should be added to `Solution Items` in `FakeItEasy-NetCore.sln` (and `rakefile.rb` should be removed).

adamralph:
Also, `how_to_build.md` will need updating.

rakefile-dotnet.rb
```diff
@@ -258,6 +268,12 @@
   cmd.parameters "pack #{nuspec} -Version #{version}#{version_suffix} -OutputDirectory #{output}"
 end
 
+desc "create the nuget package with the .NET Standard"
+exec :pack => [:build, output] do |cmd|
+  cmd.command = nuget_command
+  cmd.parameters "pack #{dotnet_nuspec} -Version #{version}-dotnet#{version_suffix} -OutputDirectory #{output}"
+end
+
```
adamralph: This should replace the "create the nuget package" partial task definition above.

I.e. we should only be building one NuGet package from this rakefile (plus the analyzer).

blairconrad: This is a note for the far future, yes? Until we can build a production version of the package that includes our .NET lib (which we can't do until we have a production Castle.Core), we'll need to produce both nupkgs.

blairconrad:
> Not important now, but just a reminder that as part of this, rakefile-dotnet.rb and build-dotnet.cmd should be added to Solution Items in FakeItEasy-NetCore.sln (and rakefile.rb should be removed).
> Also, how_to_build.md will need updating.

True. I'll do that in the near term. Better than forgetting.

rakefile-dotnet.rb
adamralph: I was thinking that we would continue to use the other rakefile to produce the net40-only package, and this rakefile to produce the net40+netstd package.

adamralph: We'll need a separate env var on the CI server to control the pre-release moniker for this rakefile though, perhaps `VERSION_SUFFIX_DOTNET`.

adamralph: More generally, the I don't think the new rakefile should care at all about the net40-only situation. It should only be concerned with doing everything required for producing a net40+netstandard1.6 package.

blairconrad: > We'll need a separate env var on the CI server to control the pre-release moniker for this rakefile though, perhaps VERSION_SUFFIX_DOTNET.

not `_NETSTD`?

But seriously, why? We have the ability to hardcode such a thing into the version number right now (as I do on line 274 above). We won't want to remove the special dotnet/netstd token until we promote this rakefile to be _the_ rakefile, so why not just leave it in here?

> More generally, I don't think the new rakefile should care at all about the net40-only situation. It should only be concerned with doing everything required for producing a net40+netstandard1.6 package.

Okay.

adamralph: I'm more than happy to switch to `netstd` as the moniker everywhere, but we seem to be delaying that and going with `dotnet` everywhere for now, which is why I suggested `_DOTNET`. I think we should do one or the other, but not mix.

The problem is, we'll need to be able to control the version suffix for the net40 package independently of the dotnet package. E.g. when we are pushing out an RC of the net40 package, I don't think we'll want to use RC in the version suffix of the net40+netstd package. And likewise, when we push out net40 RTM, we'll want the net40+netstd to still have a pre-release suffix. I don't think that we want `dotnet` in the version name there, but perhaps there's an idea there which I don't understand?

blairconrad: > I'm more than happy to switch to `netstd` as the moniker everywhere, but we seem to be delaying that and going with `dotnet` everywhere for now, which is why I suggested `_DOTNET`. I think we should do one or the other, but not mix.

I'm happy with netstd. I didn't realize we were delaying as such.
I haven't converted my PR yet, but am happy to do so the second all the project names and paths and whatnot switch, but it's a separate task from "make a rakefile".
I could take a crack at the conversion to netstd, if nobody else wants to, but otherwise, I'm happy to continue on with this task.

> The problem is, we'll need to be able to control the version suffix for the net40 package independently of the net40+netstd package. E.g. when we are pushing out an RC of the net40 package, I don't think we'll want to use RC in the version suffix of the net40+netstd package.

I don't really see any harm. We could either not publish the RC package or publish and then never publish the production release with the same version number. 

> And likewise, when we push out net40 RTM, we'll want the net40+netstd to still have a pre-release suffix.

Sure! "-netstd". It satisfies [point 9 of SemVer](Satisfies http://semver.org/#spec-item-9)'s requirements for being a 'pre-release' version. And is more intention-revealing than "-beta".

adamralph: The problem with that versioning scheme is that `netstd` sorts after beta, so if we're on 2.3.0 net40 and 2.4.0-netstd net40+netstd then we can't push out 2.4.0-beta net40 when we get closer to release because because it's already trumped by 2.4.0-netstd.

blairconrad: But if you want an extra variable, that's fine by me.
I probably don't understand how you expect the naming to work.
In the end, I'll probably go along with just about anything, but am afraid that I won't be able to manage the version-setting for the release.

blairconrad: Whoops, we crossed again.

So how would that work with the two environment variables?

Alternatively, we could push out 2.5.0-beta.

blairconrad: Back on the `netstd` naming, I can jam some commits at the front of this PR, since we might as well use the changes I have so far to make sure things keep building and working.

blairconrad:
Closing in favour of #821, which (I think) addresses all of @adamralph's concerned except for the one about having an environment variable to specify pre-release versions of the .NET Standard build.
We can continue that there.

PR #816, opened 2016-08-02
address nitpicks from netstd changes

adamralph:
connects to #531

PR #817, opened 2016-08-03
add missing bits to FakeItEasy.dotnet.csproj

adamralph:
Addressing a few more items on [The List](https://github.com/FakeItEasy/FakeItEasy/issues/531#issuecomment-236366247).

connects to #531

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,6 +26,14 @@ indent_size = 2
 insert_final_newline = false
 charset = utf-8-bom
 
+[project.json]
+insert_final_newline = false
+charset = utf-8-bom
+
+[*.nuget.targets]
+insert_final_newline = false
+charset = utf-8-bom
+
 [*.json]
 # 2-column space indentation
 indent_size = 2
--- a/src/FakeItEasy.dotnet/FakeItEasy.dotnet.csproj
+++ b/src/FakeItEasy.dotnet/FakeItEasy.dotnet.csproj
@@ -27,8 +27,12 @@
     <DefineConstants>TRACE;DEBUG;FEATURE_NETCORE_REFLECTION</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <CodeAnalysisRuleSet>..\..\FakeItEasy.ruleset</CodeAnalysisRuleSet>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
+    <StyleCopTreatErrorsAsWarnings>true</StyleCopTreatErrorsAsWarnings>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <DebugSymbols>true</DebugSymbols>
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
     <OutputPath>bin\Release\</OutputPath>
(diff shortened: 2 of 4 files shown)


.editorconfig
```diff
@@ -26,6 +26,14 @@ indent_size = 2
...
 charset = utf-8-bom
 
+[project.json]
+insert_final_newline = false
+charset = utf-8-bom
+
+[*.nuget.targets]
+insert_final_newline = false
+charset = utf-8-bom
+
```
adamralph: This is what NuGet does to these files, so better to embrace it, as we do with `*.{csproj,vbproj}`.

Note that I couldn't define these settings in one section, as the EditorConfig docs state you can do, e.g.

```yaml
[*.{csproj,vbproj},project.json,*.nuget.targets]
```

For some reason it doesn't work. I guess this is a bug in the VS EditorConfig extension.

src/FakeItEasy.dotnet/FakeItEasy.dotnet.csproj
```diff
@@ -604,6 +611,9 @@
     </Compile>
   </ItemGroup>
   <ItemGroup>
+    <None Include="FakeItEasy.dotnet.nuget.targets">
+      <SubType>Designer</SubType>
```
adamralph: This element has never tasted good, but it's what VS did when I added the file, so `¯\_(ツ)_/¯`.

src/FakeItEasy.dotnet/FakeItEasy.dotnet.nuget.targets
```diff
@@ -0,0 +1,9 @@
+﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup Condition="'$(NuGetPackageRoot)' == ''">
+    <NuGetPackageRoot>$(UserProfile)\.nuget\packages\</NuGetPackageRoot>
+  </PropertyGroup>
+  <ImportGroup>
+    <Import Project="$(NuGetPackageRoot)\StyleCop.MSBuild\4.7.54\build\StyleCop.MSBuild.Targets" Condition="Exists('$(NuGetPackageRoot)\StyleCop.MSBuild\4.7.54\build\StyleCop.MSBuild.Targets')" />
+  </ImportGroup>
+</Project>
```
adamralph: Interesting that NuGet does this when the project has a `project.json`, even though there is also a `csproj` where it could have been added.

Also interesting that we initially had these files ignored in the initial .NET PR (fixed in #816). I don't see how we can ignore these files.

src/FakeItEasy.dotnet/project.json
```diff
@@ -15,8 +19,7 @@
     "System.Reflection.TypeExtensions": "4.1.0",
     "System.Runtime.Loader": "4.0.0"
   },
-
   "frameworks": {
-    "netstandard1.6": { }
+    "netstandard1.6": {}
   }
-}
+}
```
adamralph: NuGet did the extra tidy up in this diff.

thomaslevesque:
LGTM :+1:
Thanks!

.editorconfig
adamralph: Looks like there hasn't been a release since October 2014... https://github.com/editorconfig/editorconfig-visualstudio/releases

src/FakeItEasy.dotnet/FakeItEasy.dotnet.nuget.targets
blairconrad: The dotnet build removes this file, so maybe "add and don't ignore" isn't the straightforward solution we'd hoped.

Specifically, the `.nuget\nuget.exe restore .\FakeItEasy-NetCore.sln -PackagesDirectory .\packages` whacks it.

adamralph: image

blairconrad: I don't make 'em up. I just report em.

PR #819, opened 2016-08-03
add nuget.exe to solutions

adamralph:
Believe it or not, this isn't completely useless.

When you "open" the exe from the solution explorer, you get this:

image

If you then double-click the leaf under `Version`, you get:

image

I.e. you can check which version of nuget.exe we are running from inside VS!

### DIFF
--- a/FakeItEasy-NetCore.sln
+++ b/FakeItEasy-NetCore.sln
@@ -32,6 +32,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".nuget", ".nuget", "{4B581366-50F5-42E9-87A5-BA7DC4F6E330}"
 	ProjectSection(SolutionItems) = preProject
 		.nuget\NuGet.Config = .nuget\NuGet.Config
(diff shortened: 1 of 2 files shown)


thomaslevesque:
Thanks @adamralph!

I noticed you started signing your commits. Are there implications we should be aware of?

adamralph:
> I noticed you started signing your commits. Are there implications we should be aware of?

Not really, it just means that any signed commit is one that is guaranteed to have been created by me (at least, as long as my GPG key hasn't been compromised). Any other commits could have theoretically been crafted by anyone, although in practice, any that are in one of my PR's are also guaranteed to be me as long as my GitHub account hasn't been compromised.

thomaslevesque:
But that means we can't rebase/squash/amend your commits, right? Or if we do, we lose the signature?

adamralph:
That's correct, and it makes sense since in any of those scenarios, it's a new commit, so effectively it wasn't *me* who created it, even though it carries me as the author. It doesn't matter though. We shouldn't be afraid to perform actions which lose signatures.

PR #821, opened 2016-08-03
Create a rakefile to build a package with .NET Standard support

blairconrad:
Add a rakefile (and cmd) that will build a NuGet package that includes both .NET 4.0 and .NET Standard libraries. Intended to eventually replace the current rakefile.

Connects to #531


blairconrad:
Supersedes of #815. I think I've addressed all of @adamralph's concerned except for the one about having an environment variable to specify pre-release versions of the .NET Standard build.
We can continue that here.

blairconrad:
As always (and potentially after the pre-release version question is resolved), I'm happy to amend, reorder, squash, what have you.

blairconrad:
This PR is getting a little full, but since the appveyor.yml isn't useful until we have the netstd build, I figured, why not add the commit?
Also, I removed the `-netstd` portion of the nupkg's version. My thinking is that since for now we aren't building the net40-only nupkg on AppVeyor, there's no need for a separate "pre-release" environment var.

blairconrad:
I see that TeamCity claims to be getting/have got .NET Core support.
Does this turn us away from our "move to AppVeyor" approach? I'm content to continue on with the shift, if you all are.

In unrelated news, is there anything more (or less, or different) you'd like to from this PR?

thomaslevesque:
> I see that TeamCity claims to be getting/have got .NET Core support.

Yes, but when will the CodeBetter servers be updated?

> Does this turn us away from our "move to AppVeyor" approach? I'm content to continue on with the shift, if you all are.

Well, as I said before, I don't really like TeamCity, and I love AppVeyor... so as far as I'm concerned, yes, we should go ahead with the move to AppVeyor.

> In unrelated news, is there anything more (or less, or different) you'd like to from this PR?

To be honest, I haven't looked at it yet... (no, it's not a petty revenge for taking a long time to review my PR :wink:)

blairconrad:
> To be honest, I haven't looked at it yet.

Fair enough. I just worried that my trickling in of changes made it look like I wasn't done.
Of course, we're never _done_, but I'm content with the content of this PR, and would prefer to continue other .NET Standard (and other other) work in subsequent PRs.

adamralph:
TeamCity itself doesn't have to "support" .net core, the necessary bits
just need to be on the machines where the agents are running. You can used
TeamCity to execute "anything". That said, I'm more than happy for us to
switch to AppVeyor.

blairconrad:
> Yes, but when will the CodeBetter servers be updated?

I thought one had: 

https://twitter.com/codebetterCI/status/762998588427292672

I think I add confusion by being sloppy with the TeamCity/CodeBetter monikers again.

rakefile.netstd.rb
```diff
@@ -0,0 +1,400 @@
...
+  }
+  asm.input_file = assembly_info
+  asm.output_file = assembly_info
+end
+
+desc "Build solution"
+task :build => [:clean, :restore, logs] do
+  netstd_unit_test_directories.each do | test_directory |
+    restore_dependencies_for_project test_directory
+  end
```
adamralph: Should this block be moved to the `restore` task?

adamralph: Also, will just `dotnet restore` not work?

appveyor.yml
```diff
@@ -0,0 +1,15 @@
...
+    set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
+
+    gem install bundler
+
+    bundle install
+
+    build.netstd.cmd
+test: off
+on_finish:
+    - ps: $root = Resolve-Path .\artifacts; [IO.Directory]::GetFiles($root.Path, '*.*', 'AllDirectories') | % { Push-AppveyorArtifact $_ -FileName $_.Substring($root.Path.Length + 1) -DeploymentName to-publish }
```
adamralph: What does this line do?

If I "Export YAML" from the AppVeyor UI, I get:

```yaml
version: '{build}'
environment:
  VERSION_SUFFIX: -netstd
test: off
artifacts:
- path: artifacts/output/*
```

Is the `artifacts` setting not sufficient?

adamralph:
@blairconrad I did a pass over the changes. Apart from my two comments, this looks good to go.

appveyor.yml
adamralph: Oh, one more thing. The build script should probably be split up into `init` and `install` sections, e.g.

```yaml
version: '{build}'
init:
- cmd: >-
    set RUBY_VERSION=200
    set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
environment:
  VERSION_SUFFIX: -netstd
install:
- cmd: >-
    gem install bundler
    bundle install
build_script:
- cmd: build.netstd.cmd
test: off
...
```

I guess `init` could also be moved to `environment`, e.b.

```yaml
version: '{build}'
environment:
  RUBY_VERSION: 200
  PATH: C:\Ruby%RUBY_VERSION%\bin;%PATH%
  VERSION_SUFFIX: -netstd
install:
- cmd: >-
    gem install bundler
    bundle install
build_script:
- cmd: build.netstd.cmd
test: off
...
```

(I'm not sure if setting the `PATH` variable that way will work.)

I guess it's debatable whether `gem install bundler` should be part of `init` or `install`. On that note, I wonder if the AppVeyor environment comes pre-installed with bundler? It seems Ruby is pre-installed, so it would probably make sense to include bundler by default.

blairconrad: The `on_finish` ensures that the entire artifacts directory is published even on a failed build. Without this, we get no logs for failed builds.

I can do the install/init split and check the environment vars. I don't have an opinion on where `gem install bundler` should go, but will do learning and try to make one.
I believe things didn't work well without the `gem install bundler`. In fact, I think there was a message in the build telling me to run it before I had that line, but I will try again without it to see what happens.

rakefile.netstd.rb
blairconrad: IIRC, `dotnet restore` restored the `csproj` projects, but no the `xproj` projects. Will reconfirm.

appveyor.yml
thomaslevesque: > The on_finish ensures that the entire artifacts directory is published even on a failed build. Without this, we get no logs for failed builds.

Which logs are you talking about? AppVeyor always saves the full console output

adamralph: IIRC we instruct MSBuild to log at a highly summarised level to the
console, and at a detailed level to the file system.

rakefile.netstd.rb
blairconrad: `dotnet restore` is not sufficient.

Initial attempt to move this block to the `restore` task failed miserably.
But it was because I had a malformed (although still technically valid?) task definition. That's probably what happened to me the first time. Turns out I can move it. I'm glad you said something.

appveyor.yml
blairconrad: Yes, exactly as @adamralph says. Clean.log and Build.log in artifacts/logs.
I actually had a mysterious failure on the CI server and had insufficient hints about what went wrong until I added this step.

rakefile.netstd.rb
blairconrad: Although now (and maybe I just see it because I'm on a different machine), the build is mucking about with 
`src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets b/src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets`, leaving it in a modified state.

Given the churn in there, and how I see it removed by the `restore` and created (sometimes differently) by the `build`, I'm not sure what to do with this file. It's not currently impeding the build, but it troubles me.

I see that xunit and others ignore it, but they don't care about StyleCop. I think this is something to address separately, though, since there's bad behaviour before adding the new rakefile.

blairconrad:
Gents, thanks for the comments.

I was able to remove the `gem install bundler`.
The `RUBY_VERSION`/`PATH`environment variable pasting just wouldn't work for me, so I cheated and inlined `RUBY_VERSION`.

Oh, and there's a new nuget exe, so I picked it up while I was there.

I did add some new commits, which I would use to fixup their obvious counterparts and rebase, if y'all liked the new version. Just say the word.

rakefile.netstd.rb
```diff
@@ -106,9 +106,17 @@
...
-exec :restore do |cmd|
-  cmd.command = nuget_command
-  cmd.parameters "restore #{solution} -PackagesDirectory #{packages}"
+task :restore do
+  restore_cmd = Exec.new
+  restore_cmd.command = nuget_command
+  restore_cmd.parameters "restore #{solution} -PackagesDirectory #{packages}"
+  restore_cmd.execute
+
+  # performing restore on the solution doesn't restore
```
adamralph: Does anyone else find this behaviour somewhat bizarre?

rakefile.netstd.rb
```diff
@@ -0,0 +1,405 @@
...
+    test_cmd.execute
+  end
+end
+
+def restore_dependencies_for_project(project_dir)
+  restore_cmd = Exec.new
+  restore_cmd.command = "dotnet"
+  restore_cmd.parameters "restore #{project_dir}"
+  restore_cmd.execute
+end
```
adamralph: This is a nitpick, but this method is only used once. Would it be better to inline it in the `restore` task?

rakefile.netstd.rb
```diff
@@ -0,0 +1,405 @@
...
+
+    xunit = XUnitTestRunner.new
+    xunit.command = command
+    xunit.assembly = test_assembly
+    xunit.options '-noshadow', '-nologo', '-notrait', '"explicit=yes"', '-xml', result_file
+    xunit.execute
+  end
+end
+
+def run_netstd_tests(test_directories, result_dir)
```
adamralph: A nitpick, but this isn't really a method for running netstd tests. The test target happens to be netstd, but the test project is netcoreapp, which is why it uses the `dotnet` CLI. Perhaps this would be better named as `run_dotnet_tests`? Or `run_tests_dotnet`?

appveyor.yml
```diff
@@ -1,15 +1,16 @@
...
-    set RUBY_VERSION=200
 
-    set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
+environment:
+    PATH:           'C:\Ruby200\bin;%PATH%'
+    VERSION_SUFFIX: -netstd
 
-    gem install bundler
+install:
+    - bundle install
```
adamralph: I think should have `cmd:` in front of it. Perhaps that is the default, and that is why it works, but we've specified it explicitly elsewhere.

appveyor.yml
```diff
@@ -1,15 +1,16 @@
 version: '{build}'
-build_script:
-- cmd: >-
-    set RUBY_VERSION=200
 
-    set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
+environment:
+    PATH:           'C:\Ruby200\bin;%PATH%'
+    VERSION_SUFFIX: -netstd
```
adamralph: Why do we have this?

appveyor.yml
```diff
@@ -0,0 +1,16 @@
+version: '{build}'
+
+environment:
+    PATH:           'C:\Ruby200\bin;%PATH%'
+    VERSION_SUFFIX: -netstd
```
adamralph: Why do we have this?

appveyor.yml
```diff
@@ -0,0 +1,16 @@
+version: '{build}'
+
+environment:
+    PATH:           'C:\Ruby200\bin;%PATH%'
+    VERSION_SUFFIX: -netstd
+
+install:
+    - bundle install
```
adamralph: I think should have `cmd:` in front of it. Perhaps that is the default, and that is why it works, but we've specified it explicitly elsewhere.

rakefile.netstd.rb
```diff
@@ -0,0 +1,405 @@
...
+packages        = File.absolute_path("packages")
+
+gitlinks        = ["FakeItEasy", "FakeItEasy.Analyzer", "FakeItEasy.netstd"]
+
+unit_tests = [
+  "tests/FakeItEasy.Tests/bin/Release/FakeItEasy.Tests.dll",
+  "tests/FakeItEasy.Analyzer.Tests/bin/Release/FakeItEasy.Analyzer.Tests.dll"
+]
+
+netstd_unit_test_directories = [
```
adamralph: Similar to my other comment, these aren't really netstd tests. They are tests that target any platform which requires them to be run using the `dotnet` CLI. (In our case, the test project is `netcoreapp`.). Perhaps this would be better named `dotnet_unit_test_directories`.

rakefile.netstd.rb
```diff
@@ -0,0 +1,405 @@
...
+  "tests/FakeItEasy.Tests.netstd"
```
adamralph: You're going to hate me for this comment, but the test project is misnamed, for similar reasons. I think it should be `FakeItEasy.Tests.netcoreapp`.

adamralph: That said, project.json (and whatever csproj variant comes along to replace it) are designed for multi-targetting, so perhaps the original `dotnet` name is better for the test project (and any other project.json project). (I guess you'll hate me even more after this comment.)

appveyor.yml
blairconrad: the -netstd? You [asked for it](https://github.com/FakeItEasy/FakeItEasy/pull/821#discussion_r75471409). It had been specified in the clickly-clicky.
Which makes more sense if it's going to be an environment variable.
I'm happy to put it back.

appveyor.yml
blairconrad: Yes. I missed it.

rakefile.netstd.rb
blairconrad: I do

appveyor.yml
adamralph: Sorry, I didn't mean to ask for it, it just came along for the ride when I exported the YAML for the UI. I don't think we should ever set `VERSION_SUFFIX` in the YAML. It should be a clicky-clicky config item only.

rakefile.netstd.rb
blairconrad: but I'm inclined to go with it. It may get better when the tooling's done. Or we'll learn stuff later.

adamralph: Yep, if it needs to be done then it needs to be done. Definitely bizarre behaviour though.

rakefile.netstd.rb
adamralph: Or even more generally, given that the multi-targettting (project.json+xproj and future csproj) is "the future", those projects should probably have no suffix at all, and it should be the specific .NET Framework projects which should have a suffix, e.g. `net40`. I don't think we need to solve all this in this PR though. We can clear up the suffixes later.

adamralph:
Looking back at my comments, I think the only thing we need to do before merging this (aside from, perhaps, some squashing) is to remove `VERSION_SUFFIX` from `appveyor.yml`. The rest is either nitpicky or relates to the wider issue of naming the new projects correctly, and I think can be done later.

rakefile.netstd.rb
blairconrad: Yes. I super love that comment.

appveyor.yml
blairconrad: > I don't think we should ever set VERSION_SUFFIX in the YAML

Agreed.

> It should be a clicky-clicky config item only.

Sure. For now. I have a secret long-term goal of having us commit all the version changes like  we used to, but I know that's contentious and definitely not a concern for today, so I'm not even going to mention it.

blairconrad:
Thanks, @adamralph. I'll remove `VERSION_SUFFIX` and fix up the `cmd:` for the `install`.
Then let everyone have a look before squashing.

rakefile.netstd.rb
thomaslevesque: Personally I prefer having it as a separate method, even if it's used only once. It makes the intent clearer and keeps the methods small.

rakefile.netstd.rb
thomaslevesque: What exactly do you find bizarre? The fact that we have to specify the solution *and* the packages directory?

rakefile.netstd.rb
blairconrad: Whoops. I'd forgotten this comment after @adamralph's overall "just fix the VERSION_SUFFIX" comment.
Let's keep it as it is.

thomaslevesque: Ruby noob question: is `restore_cmd` a local variable, or is it shared across methods? The fact that it's not declared explicitly bothers me a little... I guess it's been too long since I used any scripting language.

blairconrad: I vote for "local".

As supported by this [article about variable scoping](http://www.techotopia.com/index.php/Ruby_Variable_Scope).

rakefile.netstd.rb
blairconrad: The fact that `restore FakeItEasy.netstd.sln` doesn't restore the dependencies for the `.xproj`s that are part of the solution, but it does for the `.csproj`s.

thomaslevesque:
Don't know if it's related to merging the other #831 (unlikely IMO), but this PR no longer builds... I also can't build locally with `build.netstd.cmd`, it gives me this error:

> A stable release of a package should not have a prerelease dependency. Either modify the version spec of dependency "Castle.Core [4.0.0-beta001, )" or update the version field in the nuspec.

Also, after attempting to build, the file `FakeItEasy.netstd.nuget.targets` is modified:

```diff
diff --git a/src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets b/src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targe
ts
index c89c2f7..ecef765 100644
--- a/src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets
+++ b/src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets
@@ -1,9 +1,9 @@
 <U+FEFF><?xml version="1.0" encoding="utf-8" standalone="no"?>
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup Condition="'$(NuGetPackageRoot)' == ''">
-    <NuGetPackageRoot>$(UserProfile)\.nuget\packages\</NuGetPackageRoot>
+    <NuGetPackageRoot>D:\Docs\GitHub\FakeItEasy\packages</NuGetPackageRoot>^M
   </PropertyGroup>
   <ImportGroup>
-    <Import Project="$(NuGetPackageRoot)\StyleCop.MSBuild\4.7.54\build\StyleCop.MSBuild.Targets" Condition="Exists('$(NuGetPackageRoot)\StyleCop.MSBuild\4.7.54\build\StyleCop.MSBuild.Targets')" />
+    <Import Project="$(NuGetPackageRoot)\stylecop.msbuild\4.7.54\build\StyleCop.MSBuild.Targets" Condition="Exists('$(NuGetPackageRoot)\stylecop.msbuild\4.7.54\build\StyleCop.MSBuild.Targets')" />^M
   </ImportGroup>
 </Project>
\ No newline at end of file
```

thomaslevesque:
> I also can't build locally with build.netstd.cmd, it gives me this error:
> 
>     A stable release of a package should not have a prerelease dependency. Either modify the version spec of dependency "Castle.Core [4.0.0-beta001, )" or update the version field in the nuspec.

Ah, I see. I had to set the VERSION_SUFFIX environment variable. But it still modifies `FakeItEasy.netstd.nuget.targets`...

thomaslevesque:
Strange... I just ran the TC build again without changing anything, and this time it passed

blairconrad:
@thomaslevesque, I'm not sure why this wouldn't build on TC. TC doesn't hit the `netstd` cmd/rakefile at all.
Appveyor has the `VERSION_SUFFIX` set, as you discovered is necessary.

We've talked a little bit about the `FakeItEasy.netstd.nuget.targets` (here and in #531, I think). Yeah, the restore removes it (IIRC) and then it's put back later in the build, with different paths, as you saw.
I wonder if the `NuGetPackageRoot` change is related to us passing the packages directory into `nuget restore`. That was apparently due to disk space considerations, said @jeremymeng. (Some CI servers only allocate so much space in `$(UserProfile)`).

I can try modifying the rakefile a bit to see if it works around it.

(Although I notice that many projects seem to ignore the `nuget.targets` files…)

thomaslevesque:
> @thomaslevesque, I'm not sure why this wouldn't build on TC. TC doesn't hit the netstd cmd/rakefile at all.

I think it was a random failure, unrelated to this PR

> We've talked a little bit about the FakeItEasy.netstd.nuget.targets (here and in #531, I think)

I must have missed it...

This is annoying, as it expands the variable into an absolute path, which might be different between machines. We'll have to be careful not to commit it accidentally.

thomaslevesque:
According to https://github.com/FakeItEasy/FakeItEasy/issues/531#issuecomment-237015154, it's redundant and could be removed.

blairconrad:
I was confused by that comment, as we're not using onovotny/ReferenceGenerator, as far as I know. Something else seems to be making it. Then again, it seems to be making it in a way very similar to the way we commit it. I'm not sure why we couldn't remove and ignore. [Other](https://github.com/xunit/xunit/blob/master/.gitignore#L9) [projects](https://github.com/structuremap/structuremap/blob/master/.gitignore#L35) ignore it.

For fun I found that if I stop passing the package directory to nuget, the difference in the file changes to

```diff
diff --git a/src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets b/src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets
index c89c2f7..d434d86 100644
--- a/src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets
+++ b/src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets
@@ -4,6 +4,6 @@
     <NuGetPackageRoot>$(UserProfile)\.nuget\packages\</NuGetPackageRoot>
   </PropertyGroup>
   <ImportGroup>
-    <Import Project="$(NuGetPackageRoot)\StyleCop.MSBuild\4.7.54\build\StyleCop.MSBuild.Targets" Condition="Exists('$(NuGetPackageRoot)\StyleCop.MSBuild\4.7.54\build\StyleCop.MSBuild.Targets')" />
+    <Import Project="$(NuGetPackageRoot)stylecop.msbuild\4.7.54\build\StyleCop.MSBuild.Targets" Condition="Exists('$(NuGetPackageRoot)stylecop.msbuild\4.7.54\build\StyleCop.MSBuild.Targets')" />
   </ImportGroup>
 </Project>
```

which looks easy enough to fix up, if that's what we prefer. But I don't see the point. Removal seems okay to me, but @adamralph  [had suggested](https://github.com/FakeItEasy/FakeItEasy/pull/817/files#r73284144) that keeping it would be important. I assume because it appears to enrol StyleCop in the build process. But removing the file makes no difference - StyleCop warns about problems in the FakeItEasy.netstd build anyhow.

After all that, I lean toward remove + ignore.

thomaslevesque:
> After all that, I lean toward remove + ignore.

Sounds good to me :+1:

.gitignore
```diff
@@ -17,3 +17,4 @@ packages
 *.[Dd]ot[Ss]ettings.*
 Thumbs.db
 *.received.txt
+*.nuget.targets
```
adamralph: nitpick: this should have a line ending

src/FakeItEasy.netstd/FakeItEasy.netstd.nuget.targets
```diff
@@ -1,9 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
-<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <PropertyGroup Condition="'$(NuGetPackageRoot)' == ''">
-    <NuGetPackageRoot>$(UserProfile)\.nuget\packages\</NuGetPackageRoot>
-  </PropertyGroup>
-  <ImportGroup>
-    <Import Project="$(NuGetPackageRoot)\StyleCop.MSBuild\4.7.54\build\StyleCop.MSBuild.Targets" Condition="Exists('$(NuGetPackageRoot)\StyleCop.MSBuild\4.7.54\build\StyleCop.MSBuild.Targets')" />
-  </ImportGroup>
-</Project>
```
adamralph: There's also an entry in the csproj for this file which will have to be removed:

https://github.com/FakeItEasy/FakeItEasy/blob/3756a07a38b2c02347b72359fdecae70ea2377a7/src/FakeItEasy.dotnet/FakeItEasy.dotnet.csproj#L617-L619

rakefile.netstd.rb
adamralph: It's local, and this is both declaration and assignment:

```ruby
restore_cmd = Exec.new
```

adamralph:
Given that `FakeItEasy.netstd.nuget.targets` is recreated during restore, it can indeed be removed and ignored.

IMHO there are two things which need doing to make this squashable+mergeable:

1. Remove `FakeItEasy.netstd.nuget.targets` from `FakeItEasy.dotnet.csproj`
2. Allow the new rakefile to work without having to set the VERSION_SUFFIX env var. Until we take a dep on Castle.Core 4.0.0 (RTM), we should probably munge a null or empty `version_suffix` variable to something like `"adhoc"` to allow the build to continue. I think this can be done like so:

```ruby
version_suffix = version_suffix.to_s.empty? ? "adhoc" : version_suffix
```

.gitignore
blairconrad: Bah. Yes. I really should jump on that ".editorconfig for emacs". Thanks!

adamralph:
LGTM! I think we're ready to squash a few of the commits and merge.

blairconrad:
… and did you see how I created the penultimate commit?

adamralph:
I did :smile:

blairconrad:
Okay, all squashed up.

Great comments from both of you. In addition to the silly mistakes I made, I'm very happy with the improved "no more *nuget.targets" and "no more scary build failures due to missing VERSION_SUFFIX" (but I'll be more happier when we don't need that. :wink:).

Thanks!

adamralph:
Thanks @blairconrad. Sterling effort!

blairconrad:
Thanks for the merge. And as with many of our issues, sterling **group** effort!

PR #825, opened 2016-08-13
Update how_to_build.md

medmondson:
I tripped up at this point, so thought it worth pointing out this needs to be executed in the project's root (whereas the previous steps are not dependent on this).

### DIFF
--- a/src/FakeItEasy.dotnet/FakeItEasy.dotnet.csproj
+++ b/src/FakeItEasy.dotnet/FakeItEasy.dotnet.csproj
@@ -582,6 +582,9 @@
     <Compile Include="..\FakeItEasy\RootModule.cs">
       <Link>RootModule.cs</Link>
     </Compile>
+    <Compile Include="..\FakeItEasy\Sdk\Create.cs">
+      <Link>Sdk\Create.cs</Link>
+    </Compile>
     <Compile Include="..\FakeItEasy\ServiceLocator.cs">
       <Link>ServiceLocator.cs</Link>
     </Compile>


blairconrad:
Thanks for noticing this, @medmondson. I took the liberty of changing the text a little to match some of the words we'd had elsewhere in the file, and created #828.
If it gets merged, you still get credit for your contribution!

blairconrad:
@adamralph has indicated in #828 that he prefers this approach to the one in #828. I don't mind this style.

how_to_build.md
```diff
@@ -33,7 +33,7 @@ At the time of writing the build is only confirmed to work on Windows using the
 
     `gem install bundler`
 
-1. Install gems:
+1. Install gems  (ensuring you're in the project root):
```
adamralph: @medmondson to keep the language consistent with other parts of the document, could we change this to

> Install gems (after navigating to your clone root folder):

If you're happy to make that change, can you please amend the current commit rather than adding another one, and force push to your branch.

blairconrad:
Thanks, @medmondson. Looks good to me.

adamralph:
Another commit was added rather than an amend. Are we OK with that or do we want a squash?

medmondson:
@adamralph Sorry about that guys. I wasn't sure how git/github dealt with new commits after a PR is open.  I'll happily squash both to keep commits inline.

blairconrad:
Good eye, @adamralph. I'd focused on the content. :blush:

medmondson:
Hope this is okay, I didn't squash but brought the the commits inline.

medmondson:
Sorry for breaking the build! I followed the instructions and somehow have brought in more files. Any pointers?

blairconrad:
@medmondson, no need to apologize. For starters, it's only the build of your PR that's broken. Nobody else is affected. Even if they were, you have good intentions, so don't feel bad.

It looks to me like you've... I'm not sure. Rebased on top of your master, which you've merged from the FakeItEasy master. I guess. In general, you shouldn't be merging into your master at all, just pulling from upstream.

I think we need to do 2 things. 

1. fix your master, and
2. fix this PR, and
3. probably fix the other PR (yes, that's 3 things)

I'm not super sophisticated, so I tend to use a sledgehammer, but if it were me, and I were trying to fix my master, I'd 

1. checkout master
2. fetch upstream (assuming you call the FakeItEasy/FakeItEasy repo upstream)
3. reset your local master to the upstream one
4. force push your local master to origin

Something like

```cmd
git checkout master
git fetch upstream master
git reset --hard upstream/master
git push origin master --force
```

That should make all your masters look just like FakeItEasy.

Then there's this PR.

Again, inelegant, but I'd probably just checkout your local branch, reset it to master and add your change (because it's so simple).

```cmd
git checkout patch-1
git reset --hard master
# edit the how-to-build.md
git commit -m "Your message here"
git push --force origin HEAD
```

That should clear up that branch.

If you're willing, and we've got those branches straightened out, we can talk about your NonVirtualAnalzyerSetup branch elsewhere, if you want.

medmondson:
Firstly, thank you very much for those clear instructions.  For some reason that closed my PR! My master appears fixed now so I may as well create a new PR for this change.

blairconrad:
Whoops. I'm sorry about it closing the PR. As I say I'm not sophisticated! :blush:

PR #826, opened 2016-08-13
Non virtual analyzer setup

### LINKED ISSUE #723: Live code analyzer to highlight dangerous usage
As we know faking an object containing non-virtual members will cause the actual code contained within those members to run.

With the advent of [Roslyn](https://github.com/dotnet/roslyn) it's now possible to provide 'live' code analysis as we type.  We could use this to warn a user of a potentially dangerous fake.

This would hold huge advantage to both new and advanced users.

I have experience in this area, therefore if you ratify my suggestion I'll be happy to begin working on this.

medmondson:
Fixes #723. Analyzer to mark a setup of a non-virtual member


thomaslevesque:
Thanks @medmondson!
The PR looks pretty good, and I think this analyzer is a great addition. I'm unable to do a full review right now, though, so I'll get back to you in a few days.

src/FakeItEasy.Analyzer/Resources.resx
```diff
@@ -117,6 +117,18 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="NonVirtualSetupDescription" xml:space="preserve">
+    <value>Non virtual members can not be intercepted.</value>
+    <comment>An optional longer localizable description of the diagnostic.</comment>
```
blairconrad: I'm not an expert, but should this message be here? I'd've assumed either a longer description, or the lack of the `comment` tag.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System.Linq;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
```
blairconrad: This is a nitpick, but FakeItEasy style is to put the usings inside the namespace, sorted alphabetically (with System* in a group at the top), like on line 6 and below.

This can be seen in the compilation warnings in the IDE (or in the [CI build](http://teamcity.codebetter.com/viewLog.html?buildId=225343&buildTypeId=bt929&tab=buildLog&state=59%2C243#_state=59,243&focus=258)), along with a number of other warnings that have been introduced.

I point this out not to force you to do more work, @medmondson, but because if I were a contributor, I'd want to know about the target project's conventions. Don't worry too much about the warnings @medmondson. We won't reject the PR based on them. But if you _wanted_ to clean them up, either as part of other rework or not, the effort would be welcome and appreciated.

tests/FakeItEasy.Analyzer.Tests/NonVirtualSetupAnalyzerTests.cs
```diff
@@ -0,0 +1,200 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace FakeItEasy.Analyzer.Tests
+{
+    using Helpers;
```
blairconrad: An additional nitpick, which I'm not sure our StyleCop catches, but the preference is to use the full namespace in usings. Again, don't go out of your way to make a change, but if you happen to be in the area later…

tests/FakeItEasy.Analyzer.Tests/NonVirtualSetupAnalyzerTests.cs
```diff
@@ -0,0 +1,200 @@
...
+
+
+        protected override DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
+        {
+            return new NonVirtualSetupAnalyzer();
+        }
+
+    }
+
+}
```
blairconrad: I'm enjoying the tests, @medmondson. Do you think there's value in adding tests (and possibly specific message, etc.) for static members? This can certainly be considered a "bonus", from my point of view.

blairconrad:
Thanks for the work, @medmondson. I've taken a very quick pass and made some minor comments, mostly to do with the code structure. I've not taken time to investigate the actual working of the analyzer yet (I need to start earlier in my day to understand those kinds of things), but intend to. So far, I'm left with a good impression, though.

medmondson:
Many thanks for taking an initial look at this.

I must admit I noticed the namespace convention but it looks as though resharper slipped them in unnoticed.

Agree with the additional tests. I'll do the work on both when I find time.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,77 @@
...
+                context.ReportDiagnostic(diagnostic);
+            }
+        }
+
+        private static bool NotSetupCall(InvocationExpressionSyntax invocationExpression)
+        {
+            var isSetupInvocation = invocationExpression
+                .Ancestors()
+                .OfType<InvocationExpressionSyntax>()
+                .Any(x => x.Expression.ToString() == "A.CallTo");
```
blairconrad: This doesn't handle calls made via a `using static FakeItEasy.A`. That is:
```c#
using static FakeItEasy.A;

…
        var foo = Fake<Foo>();
        CallTo(() => foo.Bar());
```

Will not trigger the diagnostic, even when `Bar` is not virtual. Perhaps following an approach closer to `UnusedReturnValueAnalyzer.AnalyzeCall` and `UnusedReturnValueAnalyzer.GetCalledMethodSymbol` would help.
This may help resolve other deficiencies as well. In particular, I think the whitelist of fully-qualified method names to examine (see `UnusedReturnValueAnalyzer.CreateDiagnosticsMap`) is quite helpful, in that it provides an obvious place to to reference to see what's being examined.

For example, even looking at that list now, I am reminded of ``A.CallToSet`1``, and ``FakeItEasy.Fake`.CallsTo`1``, which would also benefit from your diagnostic.

And the fully-qualifiedness of the names would help avoid false positives in the (extremely unlikely) event that the code being analyzed has another `A.CallTo`.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,77 @@
...
+            if (!symbolInfo.Symbol.IsVirtual)
+            {
+                var location = invocationExpression.GetLocation();
+
+                Diagnostic diagnostic = Diagnostic.Create(DiagnosticDefinitions.NonVirtualSetup, location, symbolInfo.Symbol.Name);
+                context.ReportDiagnostic(diagnostic);
+            }
+        }
+
+        private static bool NotSetupCall(InvocationExpressionSyntax invocationExpression)
```
blairconrad: A nit, but if you're in the area again, our preference is to prefix Boolean properties and methods with "Is/Can/Has", as you've done with `IsInterface`. 

(And my personal preference is to lean toward affirmative phrases, such as `IsSetupCall`, but in this case, given the limited usage, I think your choice of the negative makes a better call site: `if (IsNotSetupCall` vs. `if (!IsSetupCall`. Alternative would be to invert the `if` so that you had `if ( IsSetupCall… ) { … }`, but that introduces nesting, so it's tough. I'm not asking for any action in this digression; it's just food for thought.)

src/FakeItEasy.Analyzer/Resources.resx
```diff
@@ -129,4 +141,4 @@
     <value>Unused call specification</value>
     <comment>The title of the diagnostic.</comment>
   </data>
-</root>
+</root>
```
blairconrad: I'm a little sad to see all the distracting whitespace changes here.
For my curiosity, did you edit the file using the visual editor in the IDE? I suspect that's what introduced the changes.

medmondson:
Thanks @blairconrad. Don't worry if it's a nit or not I'm taking it all on board :-)

I actually toyed with the method naming for a while before settling with the negative phrase for the reason you noted.

Thanks for pointing out the analysis scenario I missed, I'll write the appropriate tests/fixes.

blairconrad:
I mentioned some scenarios that are missed, but worse, we have false positives. Consider:

```c#
namespace TheNamespace
{
    using FakeItEasy;
    class TheClass
    {
        void Test()
        {
            var aClass = new AClass();
            var foo = A.Fake<IFoo>();
            A.CallTo(() => foo.Bar(aClass.Method())).Returns(42);
        }
    }

    class AClass { public int Method() { return 3; } }
    interface IFoo { int Bar(int i); }
}
```

This is a valid usage, but the analyzer sees that `aClass.Method()` is not virtual and complains, even though it need not be virtual. The analyzer makes no distinction between the `foo.Bar` and the `aClass.Method` when searching for non-virtual methods.

blairconrad:
@medmondson, I've looked at the analyzer itself a little and made some additional comments. The "false positives" one is a show-stopper for me, but I think we're on the right track!

medmondson:
Modified as per comments above, _except_ for the static stuff which I'll pick up when I find the time.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            var isSetupInvocation = invocationExpression
+                                    .Ancestors()
+                                    .OfType<InvocationExpressionSyntax>()
+                                    .Take(1)
+                                    .Single();
```
blairconrad: `.Take(1).Single()` is probably better written as `First()`, but in this case, I'd be inclined to go with `FirstOrDefault()` and return `true` if the result is `null` - as written, we're throwing a lot of unnecessary exceptions whenever there is no `InvocationExpressionSyntax` in `Ancestors`, and extra exceptions are at least wasteful, and can also interfere with usage of the analyzer, as seen in #731.

medmondson: Good point, think I was suffering from lack of sleep. I'll pick it up when I sort the extra static work.

adamralph:
Note that the PR branch is not a straight line from master.

@medmondson please see https://github.com/FakeItEasy/FakeItEasy/blob/master/CONTRIBUTING.md#handling-updates-from-upstreammaster for details about how the PR branch should be formed.

medmondson:
@adamralph No problem, I assumed I was that quick there were no upstream changes!

blairconrad:
Hey, @medmondson, I'd promised to chat about how to "fix" your branch.

I think there's an easy fix to apply, working on top of an up-to-date master. (which you have right now, I think, but I'll include what I normally do to "have an up-to-date master". I do this every time before submitting a pull request. It's a little verbose, but I like to be explicit and one can always look for shortcuts when one's experienced.)

To make sure you have an up-to-date master, follow the first 3 steps that @adamralph  pointed you to in https://github.com/FakeItEasy/FakeItEasy/blob/master/CONTRIBUTING.md#handling-updates-from-upstreammaster:

1. clean up your working directory by stashing changes (if necessary)
2. `git checkout master`
3. `git pull upstream master`

Now your master is exactly like FakeItEasy's.

Then you want to do something to your PR branch, so check it out:

1. `git checkout NonVirtualAnalyzerSetup`

Of course, what you want to do to your branch will vary on a case-by-case basis. This time, we want to retain your desirable changes while untangling what you got from the merge earlier. It looks like
you've only picked up one rogue commit as a result of your merge from master: ba45ef98032f9ecaf30da560ab91513f54848660.

So I recommend just rebasing your branch on top of master:

1. `git rebase -i master`

This will bring up an editor with a bunch of commits. You can reorder them or modify them or do all kinds of things. You should see something like this:

```
pick 8ee09ce Putting structure in place to begin work on analyzer.
pick ba45ef9 Correct name for tests folder
pick bb4f6ce Began working on analyzer
pick 32a4de3 Begin analyzer code
pick b1abc12 Further work on analyzer and diagnostics - currently failing test on interfaces, need to exclude these
pick 938bdbb All tests passing, now to check for other scenarios
pick 5e12b6d Added more tests
pick ba8dfa3 Added test and resolution for false positives on nested non-virtual members
pick fe89404 Removed unnecessary white space changes
pick d9fe8e8 Organised usings as per project standards

# Rebase 33d5fa3..d9fe8e8 onto 33d5fa3 (10 command(s))
#
# Commands:
# p, pick = use commit
# r, reword = use commit, but edit the commit message
# e, edit = use commit, but stop for amending
# s, squash = use commit, but meld into previous commit
# f, fixup = like "squash", but discard this commit's log message
# x, exec = run command (the rest of the line) using shell
# d, drop = remove commit
#
# These lines can be re-ordered; they are executed from top to bottom.
#
# If you remove a line here THAT COMMIT WILL BE LOST.
#
# However, if you remove everything, the rebase will be aborted.
#
# Note that empty commits are commented out
```

To remove the extra commit, just delete this line

> `pick ba45ef9 Correct name for tests folder`

and save the file and exit. Git will replay all the other commits on top of master, and then you can force push your branch as you've done before.

(As an aside, most of your commits are sort of partial work that don't make a lot of sense on their own, and the whole PR might benefit from having them squashed all into one. You could do this by changing all but the first "pick" to "squash". This will change all the commits into one, and you'll get the opportunity to edit the commit message. If you're not comfortable with this at this point, feel free to leave all the "pick"s as picks for now. You can always get rid of the "Correct name for tests folder" commit as I described above, and then later on try the `git rebase -i master` again to squash things. Or we can look into that much later when the PR is nearly done.)

There are two important things to remember:

1. Don't panic. If the `git rebase -i` seems to be going horribly you can recover at any time. In the editor, just delete ALL the lines, and save, and it'll be like you never ran the command. Or if you are at the command prompt and git is asking you confusing questions (I don't think it will), you can always `git rebase --abort`. In either case, you're back where you started.
2. Don't panic. Even after the git rebase, if you don't like how your working copy or your history look, you can always throw away your local branch and recreate it from the one in your origin, just like you did using `git reset` with your master branch earlier, except pointing at "origin/NonVirtualAnalyzerSetup" instead of "upstream/master". Then you can try again.

blairconrad:
Oh, and @medmondson, if you're in the middle of things and want to chat, you can always try https://gitter.im/FakeItEasy/FakeItEasy. We're sometimes there.

medmondson:
@blairconrad WOW thank you so much for taking the time to provide this information, it's very much appreciated. I've not used git extensively yet so those pointers above are invaluable, thanks very much.

I'll learn for your comments and add the extra static tests (and fixes) when I find time. Seriously thanks :-)

blairconrad:
@medmondson, I'm glad the info seems useful. I'm not a git expert and was _completely_ unschooled when I first showed up in FakeItEasyLand. I'm sure @adamralph has horror stories he could tell.
To this day, I just have a few bags of tricks, but they _generally_ work out. If anything in my comment is confusing, do come back and I'll see what I can do. Or we can enlist help!

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
thomaslevesque: I also prefer the affirmative form; `if (!IsSetupCall(invocationExpression))` would be better IMO.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            const string setupCall = "A.CallTo";
+
+            return isSetupInvocation.Expression.ToString() != setupCall;
```
thomaslevesque: This won't cover all cases. Even though it's unlikely, `A.CallTo` _could_ refer to something completely unrelated to FakeItEasy. And as mentioned by @blairconrad, it doesn't handle the case (admittedly not very common) of `using static FakeItEasy.A`. You can get the full member name (including the namespace) [like this](https://github.com/FakeItEasy/FakeItEasy/blob/master/src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs#L42-L43)

thomaslevesque:
Thanks again for the PR @medmondson, and sorry for not being able to review earlier. I only have a few comments, which are mostly the same as @blairconrad's.

medmondson:
Okay, I've done the extra work and attempted to fix (and squash) my commits.  Everything seems good this end. Please let me know what you think.

As a side I've really enjoyed contributing to this project. Hope I get the chance to help with anything else.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,99 @@
...
+            var invocationExpression = (InvocationExpressionSyntax)context.Node;
+
+            if (IsInterface(context, invocationExpression))
+                return;
+
+            var invocationParent = invocationExpression
+                                    .Ancestors()
+                                    .OfType<InvocationExpressionSyntax>()
+                                    .Take(1)
+                                    .ToList();
```
thomaslevesque: Why are you making a list if it will never have more than one element? You could use `FirstOrDefault` instead

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            if (!IsSetupInvocation(invocationParent))
+                return;
+
+            if (!IsFakeItEasy(context, invocationParent.Single()))
+                return;
```
thomaslevesque: I don't really understand why `IsSetupInvocation` and `IsFakeItEasy` have to be separate methods... as far as I can tell, you don't need `IsSetupInvocation` at all, it's redundant. I think you should
- remove `IsSetupInvocation`
- rename `IsFakeItEasy` to `IsSetupInvocation`

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            var symbolInfo = context.SemanticModel.GetSymbolInfo(context.Node);
+
+            if (!symbolInfo.Symbol.IsVirtual)
```
thomaslevesque: :warning: `symbolInfo.Symbol` could be null (probably not in this case, but you should probably check, just to make sure)

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            var methodFullName =
+                string.Concat(methodSymbol.ContainingType.GetFullName(), ".", methodSymbol.GetDecoratedName());
+
+            return methodFullName == fakeItEasyMethod;
+        }
+
+        private static bool IsInterface(SyntaxNodeAnalysisContext context, InvocationExpressionSyntax invocationExpression)
+        {
+            var typeKind = context.SemanticModel.GetSymbolInfo(invocationExpression).Symbol.ContainingType.TypeKind;
```
thomaslevesque: :warning: `Symbol` could be null here too

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        private static ImmutableArray<DiagnosticDescriptor> CreateSupportedDiagnostics()
+        {
+            var supportedDiagnostics = new[]
+            {
+                DiagnosticDefinitions.NonVirtualSetup
+            };
+
+            return supportedDiagnostics.ToImmutableArray();
+        }
```
thomaslevesque: This method could be reduced to `ImmutableArray.Create(DiagnosticDefinitions.NonVirtualSetup)` and inlined without loss of readability.

tests/FakeItEasy.Analyzer.Tests/NonVirtualSetupAnalyzerTests.cs
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    internal class Foo
+    {
+        internal virtual void Bar()
+        {
+            throw new NotImplementedException();
+        }
+    }
```
thomaslevesque: Nitpick: although the non-virtual warning should indeed not be triggered here, this code wouldn't work, because the class and method are internal and can't be faked. It would make the test less confusing if they were public.

thomaslevesque:
@medmondson I have a few extra remarks, but I think we're getting close to something we can merge :smile: 

> As a side I've really enjoyed contributing to this project. Hope I get the chance to help with anything else.

And we enjoy your contribution, thanks !

src/FakeItEasy.Analyzer/SymbolHelpers.cs
```diff
@@ -0,0 +1,26 @@
...
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using Microsoft.CodeAnalysis.Diagnostics;
+
+    internal static class SymbolHelpers
+    {
+        /// <summary>
+        /// Returns the called method symbol for the invocation expression
+        /// </summary>
+        /// <param name="call"></param>
```
blairconrad: I tell people at work that empty param and return elements just clutter the code. Could you provide the documentation, please?

tests/FakeItEasy.Analyzer.Tests/NonVirtualSetupAnalyzerTests.cs
blairconrad: Excellent point, @thomaslevesque. In fact, I'd upgrade this from "nitpick". The "test code" should be an otherwise-valid use of FakeItEasy.

blairconrad:
And I hate to bring this up again, @medmondson, but your branch is still in a wonky state. Here's what I see when I look at the repos. (I used [GitKraken](https://www.gitkraken.com/) to make the pictures. I am a big fan of using the command line, but I do think that a graphical tool can sometimes give a nice overview when things get confusing.)

image

You can see your branch (and a little mergey symbol to represent the PR) at the very top. This is commit 501ea2c8e36d372d9fab51d692a977410cc5d000.
The next node down is _your_ master at commit ba58dc42c6498d467fe5810c48444c623cf54a79.
You can see that it has two parents.
One is FakeItEasy's master, at commit 0d825433325f971a5319768278c0332f3a6367b2, and one is an illegal (by our contribution rules) commit c2e80687a8c6766da85d8e8beb05e034e703f4ff, which exists only on your master. It was created by merging FakeItEasy master into your master.

We can fix this. As before, the first step is to make sure your master matches FakeItEasy's. I do this by:

```
git fetch upstream
git checkout master
git reset --hard upstream/master
```

The `git fetch upstream` just makes sure your computer has the up-to-date information about the upstream (FakeItEasy) repo, so when you `reset` below, you're resetting to right commit.

After this, my picture looks like this:

image


You can see that my master is now at 0d825433325f971a5319768278c0332f3a6367b2 ("Merge pull request #829 from medmondson/patch-1").

That leaves NonVirtualAnalyzerSetup in a bit of an odd state where it still includes the extra merge commit. However, this is even easier to resolve. We wish we'd never merged and just done our work on the "proper" master, so let's tell git that:

```
git checkout NonVirtualAnalyzerSetup
git rebase master
```

And as easy as that we have

image

NonVirtualAnalyzerSetup is now a straight line of 2 commits from the FakeItEasy master.

And everything will stay clean and proper if we never merge into master (or our topic branches) ever again and just pull the master from upstream (and push it to origin) and then rebase our topic branches on top of master.

blairconrad:
Hey, @medmondson, the git structure looks much better for this branch. Thanks!

One thing that you should do to save yourself additional trouble in the future is push your local master up to origin, since I see that the latter still has an additional couple of merges on it:

image

I'd just

```
git checkout master
git push --force origin master
```

And you should be all set up.

And to tell the truth, I've lost track of the status of the contents of the PR. I'll have to review again.

blairconrad:
I don't know what I was looking at. The PR branch still comes off @medmondson's master, which has a merge from FakeItEasyMaster. You should probably ignore my previous message.

tests/FakeItEasy.Analyzer.Tests/NonVirtualSetupAnalyzerTests.cs
blairconrad: Also, is this another opportunity for a diagnostic? Might be tricky, as one would have to navigate any `InternalsVisibleTo` that were around.

blairconrad:
I've taken another pass at the code. I think @thomaslevesque's comments are vital, and that the remaining ones of mine that are still not addressed can be either done now or can be addressed after the merge. To summarize, since they're now all hidden in "outdated diff"s above:

- only `A.CallTo` is checked, not the other call-specifying methods such as `A.CallToSet` and `Fake<T>.Callsto` or `Fake<T>.AnyCall`
- no tests for static methods
- no tests for specifiers accessed via `using static`
- StyleCop violations (and as a bonus, I'd update the "class strings" in the test cases to conform as well, but I'm like that...)

Thanks for your efforts, @medmondson. Even though it may not seem like it, things are moving along!

medmondson:
Apologies for the broken build there. I'd managed to get into a tangle by committing to another (unintended branch) and the merge brought it into scope of this PR.  I hope it doesn't spam you with notifications.

I've not squashed or ensured inline commits from 'main' (yet).

On the plus side I only have the following left to work on:

- only A.CallTo is checked, not the other call-specifying methods such as A.CallToSet and Fake<T>.Callsto or Fake<T>.AnyCall

- no tests for static methods

Getting it to report property setups (for Callsto) might take a bit of extra work as I couldn't seem to get them to appear for an invocation expression, however my time was short and I intend to look into this further.

No need to comment, unless you feel compelled to.

blairconrad:
Hi, @medmondson. I enjoy commenting, so here I am.

Don't worry about squashing the commits at this point. Your new commits continue to be on a nice straight line from the FakeItEasy master (as it was a few days ago, but it's a moving target, so don't worry about that), so it's very easy for us to follow, and bodes well for the eventual squash and rebase, which we'll ask for once everyone's happy with the content.

And don't feel bad about the broken build. It's only broken for your branch, so there's no negative impact on anyone else, and we don't get unduly spammed.

Thanks for the updates. They're not necessary or anything, but it is nice to know that we haven't yet scared you off!

blairconrad:
> Getting it to report property setups (for Callsto) might take a bit of extra work as I couldn't seem to get them to appear for an invocation expression

I don't really understand anything, but the [UnusedReturnValueAnalyzer](https://github.com/FakeItEasy/FakeItEasy/blob/f65e61a2ea63a3851a6fb1d8b27c606f711d3bbf/src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs) successfully deals with `A.CallToSet` and `Fake<T>.CallsTo`, so maybe there's a hint there.

tests/FakeItEasy.Analyzer.Tests/NonVirtualSetupAnalyzerTests.cs
```diff
@@ -171,6 +171,47 @@ public void Bar()
...
+            const string test = @"using System;
+
+//Temporary class to get non-virtual test correct
+namespace FakeItEasy.Analyzer.Tests
+{
+    class TheClass
+    {
+        void TheTest()
+        {
+            var foo = A.Fake<Foo>();
```
blairconrad: `foo` is never used.

medmondson:
Okay, we're now good for property setups (with a specific test written for A.CallToSet()).  However because property accesses are treated as MemberAccess expressions rather than InvocationExpressions I have had to make some changes - which I fear have affected the readability.  I've done what makes sense to me but I'm very welcome of your views.

Other than that I think I've covered off all the remaining tasks.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            var methodFullName =
+                string.Concat(methodSymbol.ContainingType.GetFullName(), ".", methodSymbol.GetDecoratedName());
+
+            return DiagnosticsMap.ContainsKey(methodFullName);
+        }
+
+        private static ImmutableDictionary<string, DiagnosticDescriptor> CreateDiagnosticsMap()
+        {
+            var memberNames = DiagnosticSubjects.CallSpecMemberNames();
```
blairconrad: I appreciate the desire to reuse code, but many of the items in `CallSpecMemberNames` are not applicable here.
There's no need for this analyzer to examine usages of `WithAnyArguments`, for example, and inclusion of the extra items in the list causes problems:

```c#

File: src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs [hunk 19]
```diff
@@ -0,0 +1,108 @@
...
+            //Method so exclude non-methods and interfaces
+
+            AnalyzeCall<InvocationExpressionSyntax>(context, x => NotMethod(x) || Interface(x));
+        }
+
+        public static void AnalyzeProperty(SyntaxNodeAnalysisContext context)
+        {
+            //Propery so exclude methods
+
+            AnalyzeCall<MemberAccessExpressionSyntax>(context, x => Method(x));
```
blairconrad: I don't understand why we wouldn't also exclude calls to interface members here as we did in `AnalyzeMethod`.

blairconrad:
```c#
namespace TheNamespace
{
    using FakeItEasy;

    public class TheClass
    {
        public void Test()
        {
            var foo = A.Fake<IFoo>();
            A.CallTo(() => foo.Bar);
        }
    }

    public interface IFoo
    {
        int Bar{get;}
    }
}
```

is detected as an invalid use:

`Test0.cs(11,28): warning FakeItEasy0002: Non virtual member 'Bar' cannot be intercepted.`

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        private static void AnalyzeCall<T>(SyntaxNodeAnalysisContext context, Func<SymbolInfo, bool> exclusions) where T : ExpressionSyntax
```
blairconrad: Rather than excluding items to examine, could we not just include items?
That way, rather than saying the property analyzer excludes methods, and the method analyzer excludes non-methods, we could just say that the property analyzer applies to properties and the method analyzer applies to methods?

The `Interface` portion of the exclusion could be placed right into this method, where it would be run whatever we're analyzing (which would eliminate the problem I mention with `AnalyzeProperty` on line 43.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+            return DiagnosticsMap.ContainsKey(methodFullName);
+        }
+
+        private static ImmutableDictionary<string, DiagnosticDescriptor> CreateDiagnosticsMap()
+        {
+            var memberNames = DiagnosticSubjects.CallSpecMemberNames();
+
+            return memberNames.ToImmutableDictionary(name => name, name => DiagnosticDefinitions.NonVirtualSetup);
+        }
+
```
blairconrad: The new methods below have all deviated from the preferred `IsSomething` naming convention.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            return x.Symbol?.Kind != SymbolKind.Method;
+        }
+
+        private static bool Method(SymbolInfo x)
+        {
+            return x.Symbol?.Kind == SymbolKind.Method;
+        }
+
+        private static bool Interface(SymbolInfo x)
```
blairconrad: I find the naming of this method troubling. Unlike `NotMethod` and `Method` above, this method does not judge `x` (which is not a great parameter name—strive for something more meaningful)'s Interfaceness. Rather it checks to see if `x` were declared on an interface.
This really confused me when I was originally reading line 36, which I thought was saying to exclude items that were not methods or were interfaces (which aren't methods, and I couldn't understand why we'd be examining a symbol the could either be a method or an interface).

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            context.RegisterSyntaxNodeAction(AnalyzeMethod,
+                SyntaxKind.InvocationExpression);
+
+            context.RegisterSyntaxNodeAction(AnalyzeProperty,
+                SyntaxKind.SimpleMemberAccessExpression);
+        }
+
+        public static void AnalyzeMethod(SyntaxNodeAnalysisContext context)
```
blairconrad: This method and `AnalyzeProperty` can be made private.

blairconrad:
> Okay, we're now good for property setups (with a specific test written for A.CallToSet()).

Thanks for adding that, @medmondson! In retrospect, I should've made a separate issue for this to give the option of spreading the work out over time. It's good of you to take it as part of this PR.

> However because property accesses are treated as MemberAccess expressions rather than InvocationExpressions I have had to make some changes - which I fear have affected the readability.
> I've done what makes sense to me but I'm very welcome of your views.

For me, I found some minor readability issues as you predicted. I've pointed them out for your consideration. In addition, you introduced a few false positives, which I've also noted.

> Other than that I think I've covered off all the remaining tasks.

Indeed. I think we're heading down the home stretch.

medmondson:
Thanks once again @blairconrad I agree with all your comments, thanks for being the second pair of eyes around those areas I knew weren't quite right.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
thomaslevesque: Agreed, the members to analyze are specific to each analyzer and shouldn't be extracted to a common class.

blairconrad:
@medmondson, also note that the analyzer currently doesn't consider abstract methods to be interceptable, so this test fails:

```c#
        [Fact]
        public void Diagnostic_Should_Not_Be_Triggered_For_Abstract_Method()
        {
            const string test = @"
namespace TheNamespace
{
    using FakeItEasy;

    public class TheClass
    {
        public void Test()
        {
            var foo = A.Fake<Foo>();
            A.CallTo(() => foo.Bar()).Returns(42);
        }
    }

    public abstract class Foo { public abstract int Bar(); }
}
";

            VerifyCSharpDiagnostic(test);
        }
```

blairconrad:
The analyzer also seems to be confused by chained member access. At least where the intermediate "call" is field access:

```c#
namespace TheNamespace
{
    using FakeItEasy;

    public class Bar
    {
        public virtual int TheInt => 7;
    }

    public class Foo
    {
        public Bar TheBar;
    }

    public class TheClass
    {
        public void Test()
        {
            Foo foo = new Foo();
            foo.TheBar = A.Fake<Bar>();
            A.CallTo(() => foo.TheBar.TheInt).Returns(5);
        }
    }
}
```

gives

```
// Test0.cs(22,28): warning FakeItEasy0002: Non virtual member 'TheBar' cannot be intercepted.
GetCSharpResultAt(22, 28, NonVirtualSetupAnalyzer.FakeItEasy0002)
```

blairconrad:
It may be useful, once you'd updated the analyzer, to point it at a large corpus of FakeItEasy-using tests. I found the two most recent problems i mentioned above by using the Analyzer with the FakeItEasy tests themselves. Of course, those tests do things that are intended to fail from time to time, so one has to be picky with the violations that are detected.

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -85,24 +84,30 @@ private static bool IsSetupInvocation(SyntaxNodeAnalysisContext context, Invocat
...
 
-        private static bool Method(SymbolInfo x)
+        private static bool IsMethod(SymbolInfo symbolInfo)
         {
-            return x.Symbol?.Kind == SymbolKind.Method;
+            return symbolInfo.Symbol?.Kind == SymbolKind.Method;
         }
 
-        private static bool Interface(SymbolInfo x)
+        private static bool IsInterface(SymbolInfo symbolInfo)
```
thomaslevesque: Perhaps this should be renamed to `IsInterfaceMember`, since it doesn't actually check if `symbolInfo` is an interface, but rather if its containing type is an interface.

src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
```diff
@@ -57,10 +57,23 @@ private static void AnalyzeCall(SyntaxNodeAnalysisContext context)
 
         private static ImmutableDictionary<string, DiagnosticDescriptor> CreateDiagnosticsMap()
         {
-            var memberNames = DiagnosticSubjects.CallSpecMemberNames();
-
-            return memberNames.ToImmutableDictionary(name => name, name => DiagnosticDefinitions.UnusedCallSpecification);
+            var callSpecMemberNames = new[]
+                        {
```
thomaslevesque: wrong indentation

src/FakeItEasy.Analyzer/NonVirtualSetupAnalyzer.cs
```diff
@@ -0,0 +1,113 @@
...
+    using Microsoft.CodeAnalysis.CSharp;
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+    using Microsoft.CodeAnalysis.Diagnostics;
+
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    class NonVirtualSetupAnalyzer : DiagnosticAnalyzer
+    {
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } = ImmutableArray.Create(DiagnosticDefinitions.NonVirtualSetup);
+
+        private static readonly ImmutableDictionary<string, DiagnosticDescriptor> DiagnosticsMap = CreateDiagnosticsMap();
```
thomaslevesque: You don't actually need this to be a dictionary, since you only check if the key exists. This could be a `HashSet<string>` instead.

thomaslevesque: Or rather an `ImmutableHashSet<string>`

thomaslevesque:
@medmondson the latest changes look good, thanks! I made a few minor comments on the diff.

Of the two scenarios mentioned by @blairconrad above, the "chained call on a field" now seems to work, but the "abstract method" still doesn't. It's probably just a matter of considering abstract methods as if they were virtual (which they are, technically, but not according to the Roslyn API...)

thomaslevesque:
@medmondson, thanks for the corrections. All that remains now is the case of the abstract method, which is trivial (just test `IsAbstract` in addition to `IsVirtual`, and add the test from @blairconrad's [comment](https://github.com/FakeItEasy/FakeItEasy/pull/826#issuecomment-243117906) above), and it should be good to go :+1:

thomaslevesque:
> Of the two scenarios mentioned by @blairconrad above, the "chained call on a field" now seems to work

Looks like I was mistaken. [This code]:

    A.CallTo(() => this.callRule.UsePredicateToValidateArguments(predicate))
        .MustHaveHappened();

warns that `UsePredicateToValidateArguments` is not virtual, but it is.

thomaslevesque:
OK, here's a test that fails:

        [Fact]
        public void Diagnostic_Should_Not_Be_Triggered_For_Non_Sealed_Override_Method()
        {
            const string test = @"
namespace TheNamespace
{
    using FakeItEasy;

    public class TheClass
    {
        public void Test()
        {
            var foo = A.Fake<Foo2>();
            A.CallTo(() => foo.Bar()).Returns(42);
        }
    }

    public class Foo { public virtual int Bar() => 0; }
    public class Foo2 : Foo { public override int Bar() => 1; }
}
";

            VerifyCSharpDiagnostic(test);
        }

It's not related to the field, but to the fact that the method is an override. Apparently an override has `IsVirtual == false`... not what I expected.

Here's a fix for this issue:

            var member = symbolInfo.Symbol;
            bool isVirtual = member.IsVirtual
                             //|| member.IsOverride && !member.IsSealed
                             || member.IsAbstract;
            if (!isVirtual)
            {
                var location = invocationExpression.GetLocation();

                Diagnostic diagnostic = Diagnostic.Create(DiagnosticDefinitions.NonVirtualSetup, location, symbolInfo.Symbol.Name);
                context.ReportDiagnostic(diagnostic);
            }

src/FakeItEasy.Analyzer/Resources.resx
```diff
@@ -117,6 +117,19 @@
...
   </resheader>
+  <data name="NonVirtualSetupDescription" xml:space="preserve">
+    <value>Non virtual members can not be intercepted.</value>
+    <comment>An optional longer localizable description of the diagnostic.</comment>
+  </data>
+  <data name="NonVirtualSetupMessageFormat" xml:space="preserve">
+    <value>Non virtual member '{0}' cannot be intercepted.</value>
+    <comment>The format-able message the diagnostic displays.</comment>
+  </data>
+  <data name="NonVirtualSetupSpecificationTitle" xml:space="preserve">
```
blairconrad: This name is incorrect. The diagnostic doesn't contain "Specification", so we don't see the name in the Solution Explorer:

image

The fix is to rename to `NonVirtualSetupTitle`.

(kudos to @thomaslevesque for noticing the problem)

medmondson:
Thanks for pointing out those issues.

I've taken a quick(ish) look at the analyzer title issue and made some changes to bring it inline with the unused call naming, although it appears I can't get changes to reflect on the analyzer, for example if I change the description.

Currently I've been testing it through adding my freshly built .dll to the FakeItEasy project.

untitled


Is this how you do it?

blairconrad:
Thanks for the continued updates, @medmondson. I didn't know about adding an analyzer via the DLL. 
I'm not fancy, so I've been building the package (`build.cmd pack` from the command line) and then adding adding the nupkg to my project (using a local NuGet feed):

image

I'm a little short on time right now, but will take a look at your updates (especially the title change) by my morning (I'm in UTC+4 right now).

blairconrad:
> I've taken a quick(ish) look at the analyzer title issue and made some changes to bring it inline with the unused call naming, although it appears I can't get changes to reflect on the analyzer, for example if I change the description.

My mistake. I thought you'd pushed this to your branch. I guess I won't be looking at the title now!

I will look at other things.

medmondson:
Okay, I think we're there. As I suspected the missing analyzer title was resolved with a restart of the project / visual studio / laptop (not sure which).

Other than that I've tidied up the loose ends around project specific compiler warnings and given the analyzer a good run through the existing fakeiteasy code.

The last thing that comes to mind, are we happy with the analyzer message? 

> Non virtual member '{0}' cannot be intercepted.

seems a bit blunt, whereas

> Unused call specification '{0}'; did you forget to configure or assert the call?

is a friendly hint in the right direction.

I guess sometimes there's usually nothing that can be done about a non-virtual member in an attempted mock, but perhaps something like:

Non virtual member '{0}' cannot be intercepted. Members of concrete objects need to be overridden.

*may* just prevent a bit of confusion? Could we provide a link to the docs perhaps?

blairconrad:
HI, @medmondson. The title looks good, thanks for updating that.

> I've tidied up the loose ends around project specific compiler warnings

I'm not sure what a project specific compiler warning is, but I see this output when I build the Analyzer in the IDE in Debug mode:

```
1>------ Rebuild All started: Project: FakeItEasy.Analyzer, Configuration: Debug Any CPU ------
1>CSC : warning AD0001: Analyzer 'FakeItEasy.Analyzer.NonVirtualSetupAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
1>CSC : warning AD0001: Analyzer 'FakeItEasy.Analyzer.NonVirtualSetupAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
1>CSC : warning AD0001: Analyzer 'FakeItEasy.Analyzer.NonVirtualSetupAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
1>CSC : warning AD0001: Analyzer 'FakeItEasy.Analyzer.NonVirtualSetupAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
1>  FakeItEasy.Analyzer -> C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\bin\Debug\FakeItEasy.Analyzer.dll
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(47,1): warning : SA1503 : CSharp.Layout : The body of the if statement must be wrapped in opening and closing curly brackets.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(49,1): warning : SA1503 : CSharp.Layout : The body of the if statement must be wrapped in opening and closing curly brackets.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(59,1): warning : SA1503 : CSharp.Layout : The body of the if statement must be wrapped in opening and closing curly brackets.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(12,1): warning : SA1400 : CSharp.Maintainability : The class must have an access modifier.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(74,1): warning : SA1408 : CSharp.Maintainability : Insert parenthesis within the conditional AND and OR expressions to declare the operator precedence.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(14,1): warning : SA1201 : CSharp.Ordering : All properties must be placed after all fields.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(25,1): warning : SA1116 : CSharp.Readability : If the method parameters are on separate lines, the first parameter must begin on the line beneath the name of the method.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(28,1): warning : SA1116 : CSharp.Readability : If the method parameters are on separate lines, the first parameter must begin on the line beneath the name of the method.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(46,13,46,14): warning : SA1000 : CSharp.Spacing : The spacing around the keyword 'if' is invalid.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\SymbolHelpers.cs(16,1): warning : SA1629 : CSharp.Documentation : The documentation text within the summary tag must end with a period.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\SymbolHelpers.cs(16,1): warning : SA1629 : CSharp.Documentation : The documentation text within the param tag must end with a period.
1>C:\PortableApps\Documents\Source\GitHub\FakeItEasy\src\FakeItEasy.Analyzer\SymbolHelpers.cs(16,1): warning : SA1629 : CSharp.Documentation : The documentation text within the return tag must end with a period.
========== Rebuild All: 1 succeeded, 0 failed, 0 skipped ==========
```

I'm not that concerned about the SA* warnings. It would be nice to have them cleaned up, but I'm happy to do that in a compensating or superseding PR.

The four `AD0001` warnings do trouble me, though. They come from the Analyzer itself (which I'd added to each project in the FakeItEasy solution, just to see what it would detect). There are even more that arise when I build the FakeEasy and FakeItEasy.Specs projects. So far I've not been able to figure out what in the Analyzer is throwing the exception, nor why it doesn't appear when I build from the command line. Have you any insight? I'm not had any experience debugging Analyzers running in the IDE. Note that if warnings are treated as errors, these would halt the compilation, so it's quite a concern.

Other than that, the main code of the analyzer looks good to me. The test project has a few CSharp.Spacing problems that can be cleared up later just like the earlier ones, if need be, and I've a few minor comments to add to the "test strings" in the tests. Again, those could be cleaned up later.

The main thing that's troubling me is the `NullReferenceExceptions`. Could you look into that? I'll also take a look as time allows.

medmondson:
@blairconrad - thanks for showing me your build details. Unfortunately my build is identical to yours, but without the exceptions.

It might be worth doing a force fetch.

I'm running:

````
Microsoft Visual Studio Enterprise 2015
Version 14.0.25425.01 Update 3
Microsoft .NET Framework
Version 4.6.01038
````

Is anyone else on the project seeing these exceptions?

My build for reference:

```
1>------ Rebuild All started: Project: FakeItEasy.Analyzer, Configuration: Debug Any CPU ------
1>  FakeItEasy.Analyzer -> C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\bin\Debug\FakeItEasy.Analyzer.dll
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(47,1): warning : SA1503 : CSharp.Layout : The body of the if statement must be wrapped in opening and closing curly brackets.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(49,1): warning : SA1503 : CSharp.Layout : The body of the if statement must be wrapped in opening and closing curly brackets.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(59,1): warning : SA1503 : CSharp.Layout : The body of the if statement must be wrapped in opening and closing curly brackets.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(12,1): warning : SA1400 : CSharp.Maintainability : The class must have an access modifier.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(74,1): warning : SA1408 : CSharp.Maintainability : Insert parenthesis within the conditional AND and OR expressions to declare the operator precedence.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(14,1): warning : SA1201 : CSharp.Ordering : All properties must be placed after all fields.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(25,1): warning : SA1116 : CSharp.Readability : If the method parameters are on separate lines, the first parameter must begin on the line beneath the name of the method.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(28,1): warning : SA1116 : CSharp.Readability : If the method parameters are on separate lines, the first parameter must begin on the line beneath the name of the method.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\NonVirtualSetupAnalyzer.cs(46,13,46,14): warning : SA1000 : CSharp.Spacing : The spacing around the keyword 'if' is invalid.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\SymbolHelpers.cs(16,1): warning : SA1629 : CSharp.Documentation : The documentation text within the summary tag must end with a period.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\SymbolHelpers.cs(16,1): warning : SA1629 : CSharp.Documentation : The documentation text within the param tag must end with a period.
1>C:\Users\matt\FakeItEasy\src\FakeItEasy.Analyzer\SymbolHelpers.cs(16,1): warning : SA1629 : CSharp.Documentation : The documentation text within the return tag must end with a period.
========== Rebuild All: 1 succeeded, 0 failed, 0 skipped ==========
```

**Comment by blairconrad** [comment]
@medmondson, that's a build of FakeItEasy.Analyzer _using FakeItEasy.Analyzer as an analyzer_?

@thomaslevesque saw the same thing. I'll check my versions in a bit.

No rush on the SAs or anything. We're a picky bunch at FakeItEasy Enterprises, but we're patient. ;-) 

If your lack of Internets is due to vacation time, has fun!

**Comment by medmondson** [comment]
I managed to track it down to a null reference within the `IsInterfaceMember` method, although I considered the symbol may be null I hadn't considered `ContainingType` could also be.

Yes may lack of internet will be due to a much needed vacation :-)

**Comment by blairconrad** [comment]
> I managed to track it down to a null reference within the `IsInterfaceMember` method, although I considered the symbol may be null I hadn't considered `ContainingType` could also be.

Excellent!

Also, bah. I had an eye on that at one point and hadn't gotten around to trying it out. Very nice.

**Comment by blairconrad** [comment]
Enjoying the tidied SA warnings. Very clean. Thanks, @medmondson.

What was your technique for finding the offending NullReferenceException? I had no idea how to debug an analyzer error that only showed up in the IDE.
As a followup to that question, do you know what code was triggering the error?

**Comment by blairconrad** [comment]
Scratch that last question. 

I haven't looked at all the places where the NRE was arising, but they seem to be related to the access of types within namespaces. Here's the list:

```
src\FakeItEasy\Creation\CastleDynamicProxy\CastleDynamicProxyInterceptionValidator.cs(58,18,58,37): 'DynamicProxy'
src\FakeItEasy\Creation\CastleDynamicProxy\CastleDynamicProxyInterceptionValidator.cs(58,18,58,46): 'Internal'
src\FakeItEasy\Creation\CastleDynamicProxy\CastleDynamicProxyInterceptionValidator.cs(58,18,58,60): 'InternalsUtil'
src\FakeItEasy\Creation\CastleDynamicProxy\CastleDynamicProxyInterceptionValidator.cs(62,21,62,40): 'DynamicProxy'
src\FakeItEasy\Creation\CastleDynamicProxy\CastleDynamicProxyInterceptionValidator.cs(62,21,62,52): 'ModuleScope'
src\FakeItEasy\Creation\CastleDynamicProxy\CastleDynamicProxyInterceptionValidator.cs(72,67,72,86): 'DynamicProxy'
src\FakeItEasy\Creation\CastleDynamicProxy\CastleDynamicProxyInterceptionValidator.cs(72,67,72,98): 'ModuleScope'
src\FakeItEasy\Creation\CastleDynamicProxy\DynamicProxyResources.Designer.cs(38,65,38,115): 'EditorBrowsableState'
src\FakeItEasy\Creation\CastleDynamicProxy\DynamicProxyResources.Designer.cs(38,65,38,94): 'ComponentModel'
src\FakeItEasy\Creation\CastleDynamicProxy\DynamicProxyResources.Designer.cs(53,65,53,115): 'EditorBrowsableState'
src\FakeItEasy\Creation\CastleDynamicProxy\DynamicProxyResources.Designer.cs(53,65,53,94): 'ComponentModel'
src\FakeItEasy\ExceptionMessages.Designer.cs(39,65,39,115): 'EditorBrowsableState'
src\FakeItEasy\ExceptionMessages.Designer.cs(39,65,39,94): 'ComponentModel'
src\FakeItEasy\ExceptionMessages.Designer.cs(54,65,54,115): 'EditorBrowsableState'
src\FakeItEasy\ExceptionMessages.Designer.cs(54,65,54,94): 'ComponentModel'
src\FakeItEasy\Fake.of.T.cs(50,71,50,86): 'Fake'
src\FakeItEasy\Fake.of.T.cs(60,55,60,70): 'Fake'
src\FakeItEasy.Analyzer\Resources.Designer.cs(39,65,39,115): 'EditorBrowsableState'
src\FakeItEasy.Analyzer\Resources.Designer.cs(39,65,39,94): 'ComponentModel'
src\FakeItEasy.Analyzer\Resources.Designer.cs(54,65,54,115): 'EditorBrowsableState'
src\FakeItEasy.Analyzer\Resources.Designer.cs(54,65,54,94): 'ComponentModel'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(214,76,214,108): 'DoSomethingNonVirtual'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(230,76,230,104): 'DoSomethingNonVirtual'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(246,76,246,98): 'DoSomething'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(262,76,262,94): 'DoSomething'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(278,76,278,112): 'ReturnSomethingNonVirtual'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(294,76,294,108): 'ReturnSomethingNonVirtual'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(310,76,310,102): 'ReturnSomething'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(326,76,326,98): 'ReturnSomething'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(342,79,342,110): 'SomeNonVirtualProperty'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(358,79,358,106): 'SomeNonVirtualProperty'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(374,79,374,100): 'SomeProperty'
tests\FakeItEasy.Specs\ConfigurationSpecs.cs(390,79,390,96): 'SomeProperty'
tests\FakeItEasy.Specs\ConfiguringPropertySetterSpecs.cs(53,79,53,110): 'NonConfigurableProperty'
tests\FakeItEasy.Specs\CreationOptionsSpecs.cs(861,23,861,33): 'Create'
tests\FakeItEasy.Specs\CreationSpecs.cs(195,23,195,33): 'Create'
tests\FakeItEasy.Specs\CreationSpecs.cs(200,23,200,33): 'Create'
tests\FakeItEasy.Specs\CreationSpecs.cs(205,20,205,30): 'Create'
tests\FakeItEasy.Specs\CreationSpecs.cs(210,20,210,30): 'Create'
tests\FakeItEasy.Specs\DummyCreationSpecs.cs(121,23,121,33): 'Create'
tests\FakeItEasy.Specs\DummyCreationSpecs.cs(126,20,126,30): 'Create'
```

And a few samples. You can find them, if anyone cares. Or find others!

```c#
            return (T)Sdk.Create.Fake(typeof(T));
        public IEnumerable<ICompletedFakeObjectCall> RecordedCalls => FakeItEasy.Fake.GetCalls(this.FakedObject);
        [global::System.ComponentModel.EditorBrowsableAttribute(global::System.ComponentModel.EditorBrowsableState.Advanced)]
        public IEnumerable<ICompletedFakeObjectCall> RecordedCalls => FakeItEasy.Fake.GetCalls(this.FakedObject);
```

Here, `FakeItEasy`, `global::System.ComponentModel`, and `FakeItEasy` (again!) are namespaces.
No idea why this wouldn't show up at the command line build.

thomaslevesque:
@medmondson, the superseding PR #848 created by @blairconrad has just been merged, so I'm closing this one. The squashed commit is still yours, of course :wink:. Thanks again for your work!

PR #827, opened 2016-08-14
Add FakeItEasy.Sdk.Create class to FakeItEasy.dotnet

blairconrad:
Connects to #531. Was missed from #617.

### DIFF
--- a/src/FakeItEasy.dotnet/FakeItEasy.dotnet.csproj
+++ b/src/FakeItEasy.dotnet/FakeItEasy.dotnet.csproj
@@ -582,6 +582,9 @@
     <Compile Include="..\FakeItEasy\RootModule.cs">
       <Link>RootModule.cs</Link>
     </Compile>
+    <Compile Include="..\FakeItEasy\Sdk\Create.cs">
+      <Link>Sdk\Create.cs</Link>
+    </Compile>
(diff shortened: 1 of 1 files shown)


adamralph:
:+1: thanks @blairconrad. We should really get an approval test up and running for the netstd project ASAP, since then we can do an initial baseline comparison of the approved API for the net40 and netstd projects.

blairconrad:
> We should really get an approval test up and running for the netstd project ASAP, since then we can do an initial baseline comparison of the approved API for the net40 and netstd projects.

Agreed. If only ApprovalTests and PublicApiGenerator supported .NET Standard. :frown;

adamralph:
@JakeGinnivan has anything been done yet towards supporting a .NET Standard in PublicApiGenerator?

JakeGinnivan:
Not yet. 

I hope to have `.ShouldMatchApproved()` working on netcore soonish though. Just need to dedicate some time to it. 

ShouldMatchApproved() in shouldly is compatible with ApprovalTests mostly so it's really easy to switch. I cannot see ApprovalTests moving to netcore anytime soon.

adamralph:
@JakeGinnivan thanks for the info. Does that mean "ShouldMatchApproved() in shouldly" is self contained? Or do we still need PublicApiGenerator to actually generate the files?

PR #828, opened 2016-08-14
Indicate that gems must be installed from root folder

blairconrad:
Closes #825.

### DIFF
--- a/how_to_build.md
+++ b/how_to_build.md
@@ -35,6 +35,8 @@ At the time of writing the build is only confirmed to work on Windows using the 
 
 1. Install gems:
 
+    Using a command prompt, navigate to your clone root folder and execute:
+
     `bundler install`
 
 ## Building


how_to_build.md
```diff
@@ -35,6 +35,8 @@ At the time of writing the build is only confirmed to work on Windows using the
 
 1. Install gems:
 
+    Using a command prompt, navigate to your clone root folder and execute:
```
adamralph: It's a nitpick, but I don't like the repitition of "using a command prompt" here. It's already two items above, and if we have it here, we should also have it in the item directly above. (We also repeat it in the next section.) We can do a more general clean up later, but for now I'd reduce this to something like:

> 1. Install gems (after navigating to your clone root folder):

adamralph:
> Closes #825.

#825 is also a PR. Is the intention here to supersede it? Note that my preference would be something more like the change proposed in #825.

blairconrad:
> #825 is also a PR. Is the intention here to supersede it?

Yes. GitHub doesn't recognize "supersedes" as an action, though. I should've kept "supersede" in there somewhere for the humans.

> Note that my preference would be something more like the change proposed in #825.

That sounds fine, especially if the intent is to do "a more general cleanup later".

Rescinding.

PR #831, opened 2016-08-17
Don't return null for value type after DoesNothing

### LINKED ISSUE #806: NullReferenceException when calling a method that returns a value type and has been configured with DoesNothing
This code throws a `NullReferenceException` on the call to `Baz()`:

```c#
void Main()
{
    var foo = A.Fake<IFoo>();
    A.CallTo(foo).DoesNothing();
    foo.Bar();
    foo.Baz();
}

public interface IFoo
{
    string Bar();
    int Baz();
}
```

The stack trace isn't very helpful:
```
   at Castle.Proxies.ObjectProxy_8.Baz()
   at UserQuery.Main()
   ...
```
But the fact that `Baz` throws and `Bar` doesn't is a good hint: I think the generated code tries to cast the `IInvocation.ReturnValue` to `int`, which fails because it's null.

In a way, that's a bug in Castle.Core (it should throw a more specific exception). Maybe we could do something about it, but I'm not sure what we *should* do. I discovered this bug by accident; I didn't really want to use `DoesNothing()`, what I wanted was to leave the default behavior, but I needed to specify a behavior in order to use `.Once()`. Currently, if you just want to add a callback with `.Invokes()`, there's no way to _not_ affect the default behavior, since you have to specify a behavior. Maybe we need something like `.UseDefaultBehavior()`...

blairconrad:
Fixes #806

### DIFF
--- a/src/FakeItEasy/Configuration/RuleBuilder.cs
+++ b/src/FakeItEasy/Configuration/RuleBuilder.cs
@@ -63,7 +63,7 @@ namespace FakeItEasy.Configuration
 
         public virtual IAfterCallSpecifiedConfiguration DoesNothing()
         {
-            this.RuleBeingBuilt.Applicator = x => { };
+            this.RuleBeingBuilt.Applicator = call => call.SetReturnValue(call.Method.ReturnType.GetDefaultValue());
             return this;
         }
 
--- a/src/FakeItEasy/Configuration/RuleBuilder.cs
+++ b/src/FakeItEasy/Configuration/RuleBuilder.cs
@@ -49,7 +49,7 @@ namespace FakeItEasy.Configuration
 
         public virtual IAfterCallSpecifiedConfiguration Throws(Func<IFakeObjectCall, Exception> exceptionFactory)
         {
-            this.RuleBeingBuilt.Applicator = x => { throw exceptionFactory(x); };
+            this.RuleBeingBuilt.UseApplicator(x => { throw exceptionFactory(x); });
             return this;
         }
 
@@ -63,7 +63,7 @@ namespace FakeItEasy.Configuration
 
         public virtual IAfterCallSpecifiedConfiguration DoesNothing()
         {
-            this.RuleBeingBuilt.Applicator = call => call.SetReturnValue(call.Method.ReturnType.GetDefaultValue());
+            this.RuleBeingBuilt.UseDefaultApplicator();
             return this;
         }
 
@@ -77,7 +77,7 @@ namespace FakeItEasy.Configuration
 
         public virtual IAfterCallSpecifiedConfiguration CallsBaseMethod()
         {
-            this.RuleBeingBuilt.Applicator = x => { };
+            this.RuleBeingBuilt.UseApplicator(x => { });
             this.RuleBeingBuilt.CallBaseMethod = true;
             return this;
         }
@@ -129,7 +129,7 @@ namespace FakeItEasy.Configuration
             {
                 Guard.AgainstNull(valueProducer, nameof(valueProducer));
 
-                this.ParentConfiguration.RuleBeingBuilt.Applicator = x => x.SetReturnValue(valueProducer(x));
+                this.ParentConfiguration.RuleBeingBuilt.UseApplicator(x => x.SetReturnValue(valueProducer(x)));
                 return this.ParentConfiguration;
             }
 
--- a/src/FakeItEasy/Configuration/BuildableCallRule.cs
+++ b/src/FakeItEasy/Configuration/BuildableCallRule.cs
@@ -13,40 +13,17 @@ namespace FakeItEasy.Configuration
     {
         private readonly List<WherePredicate> wherePredicates;
         private Action<IInterceptedFakeObjectCall> applicator;
-        private bool canSetApplicator;
+        private bool wasApplicatorSet;
         private Func<IFakeObjectCall, ICollection<object>> outAndRefParametersValueProducer;
         private bool canSetOutAndRefParametersValueProducer;
 
         protected BuildableCallRule()
         {
             this.Actions = new LinkedList<Action<IFakeObjectCall>>();
-            this.applicator = call => call.SetReturnValue(call.Method.ReturnType.GetDefaultValue());
-            this.canSetApplicator = true;
+            this.UseDefaultApplicator();
+            this.wasApplicatorSet = false;
             this.canSetOutAndRefParametersValueProducer = true;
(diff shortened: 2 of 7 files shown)


src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -63,7 +63,6 @@ public IVoidConfiguration WhenArgumentsMatch(Func<ArgumentCollection, bool> argu
 
         public virtual IAfterCallSpecifiedConfiguration DoesNothing()
         {
-            this.RuleBeingBuilt.Applicator = x => { };
```
adamralph: It seems so simple, yet it worries me at the same time for some reason I can't quite put my finger on.

Is there any undesired side effect here? I won't pretend to know what `this.RuleBeingBuilt.Applicator` is or does, but is there any scenario where this will change behaviour and break some existing tests? I.e. is it currently possible to chain `DoesNothing` onto some other configuration calls, which effectively clears some previous config? (Please tell me if I'm talking complete nonsense!)

blairconrad: You're not talking nonsense at all. It is a little worrisome, isn't it?

First, I gotta say that the new implementation is a lot closer to "does nothing" than before. If only I could get rid of the pesky `return`, but we need something to hang the `Repeated` off.

Okay. I'll be serious now.

Your concerns are completely reasonable. 
The `Applicator` is... the thing that does stuff. Most of the time, it's used to set a return value. Otherwise, it's the thing that `Throws` the exception.
I think that's about it, really.

And every new rule is created with an applicator that returns the default value (null or value instance), so I think that aspect of the behaviour is correct.

Although you've caused me to realize something. A while back @thomaslevesque fixed #534 to make sure we couldn't save away a rule and then add multiple behaviours to it. The change as-is will break that for `DoesNothing`. Even though there's no "normal" way to make this happen using the fluent API, my change can't stand. I'll add an extra spec and amend the fix.

Thanks!

thomaslevesque: There's another reason why this change doesn't really work : it no longer causes the call to do nothing if the call was doing something... With your current fix, if you do something like this:

```
var fake = A.Fake<SomeClass>();
A.CallTo(fake).CallsBaseMethod();
A.CallTo(() => fake.SomeMethod()).DoesNothing();
```
`DoesNothing` will _not_ prevent the call to `base.SomeMethod()`.

I think `DoesNothing` should reset the applicator to its initial value (the one set in the `BuildableCallRule` constructor).

blairconrad: @thomaslevesque, in this example, `DoesNothing` will prevent the call to `base.SomeMethod()`. It works because the repeated `CallTo` creates a new rule on which the applicator lives. The `DoesNothing` will leave the default applicator on the rule, and then it's used when the call is made.

Only if we reused the result of the first `A.CallTo` would the `DoesNothing` not override the `CallsBaseMethod`, but this is the same situation that I described above, where I break #534.

thomaslevesque: > @thomaslevesque, in this example, DoesNothing will prevent the call to base.SomeMethod(). It works because the repeated CallTo creates a new rule on which the applicator lives. The DoesNothing will leave the default applicator on the rule, and then it's used when the call is made.

Good point. I'm just back from holiday and my brained isn't fully turned back on yet :wink:

blairconrad:
Okay. `DoesNothing` will (once again) not mix with `CallsBaseMethod` or `Throws` or `Returns` when invoked on the same rule.
I added a refactoring step to avoid duplicating the "default return value logic".
Once I'd added a method to set the applicator to the default behaviour, I noticed the property/method dichotomy that I'd introduced and made `Applicator` `UseApplicator` and removed the getter altogether, since it was only used by unit tests that were covered elsewhere (and so out they went).
It's more than I expected to do, and I'm happy to change any of it.

thomaslevesque:
The changes look good to me, and I like your `UseApplicator` refactoring :+1:

blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
@adamralph, any further comments?

src/FakeItEasy/Configuration/RuleBuilder.cs
adamralph: Do we we have any specs which verify that the sequence of actions described above work in the way described?

adamralph:
I like the refactoring too, and the removal of the Gulliver's Travels tests.

Whilst we're here, I think it might be nice to lock in some of the sequential behaviour that we've mentioned in some specs, if it's not already done.

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: > Do we we have any specs which verify that the sequence of actions described above work in the way described?

I just checked, we don't seem to have such specs.

blairconrad: Are you talking about [these specs](https://github.com/blairconrad/FakeItEasy/blob/49ab38fd9fc06c4e5c00b2d2870e318e0178f9f6/tests/FakeItEasy.Specs/ConfigurationSpecs.cs#L495-L534)?

I added them when I fixed my fixed. I figured since @thomaslevesque had asked about them, I might as well add evidence that it worked the way I thought while I was at it.

thomaslevesque: Ah, I missed those, thanks @blairconrad.

blairconrad: Not at all. I'm glad everyone's keeping me on my toes!

blairconrad:
I added a new commit because I'd accidentally used "it should do"-style naming instead of "it does"-style for some `Then` steps. I'd be inclined to eventually squash into the first commit where the specs were added.

adamralph:
After squashing the 4th commit into the 1st, I think this is good to go.

BTW - `git commit --fixup` and `git rebase --autosquash` are ideal for these scenarios. See https://robots.thoughtbot.com/autosquashing-git-commits

blairconrad:
Thanks for the `commit --fixup` and `rebase --autosquash` hint, @adamralph. I used them!
I'm not entirely sure it saved me any effort, but it was an interesting experiment. I'll try it a few more times to evaluate.

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thanks, y'all!

PR #835, opened 2016-08-23
Fix exceptions when configuring unnatural fakes' void methods

### LINKED ISSUE #832: Cannot use Throws or DoesNothing to configure void method on unnatural fake
Found while investigating #806.

```c#
 [Scenario]
 public static void CallsToNonVoidMethodThrows(Fake<IFoo> fake)
 {
     "Given an unnatural fake"
         .x(() => fake = new Fake<IFoo>());

     "And I configure a void method to throw an exception"
         .x(() => fake.CallsTo(f => f.VoidMethod()).Throws<Exception>());
 }
```

throws at that `Throws` line:

```
System.InvalidOperationException
The behavior for this call has already been defined
```

It's because in `StartConfiguration.CallsTo`, we set the `rule.Applicator`.

`DoesNothing` has a similar problem.

Reinforces my belief that "unnatural fakes" are not often used.

blairconrad:
Fixes #832.

### DIFF
--- a/tests/FakeItEasy.Specs/UnnaturalFakeSpecs.cs
+++ b/tests/FakeItEasy.Specs/UnnaturalFakeSpecs.cs
@@ -1,7 +1,11 @@
 ﻿namespace FakeItEasy.Specs
 {
+    using System;
+    using FakeItEasy.Configuration;
+    using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
     using Xbehave;
+    using Xunit;
 
     public static class UnnaturalFakeSpecs
     {
@@ -13,7 +17,7 @@
         }
 
         [Scenario]
-        public static void CallsToVoidMethod(Fake<IFoo> fake, bool wasCalled)
+        public static void CallsToVoidMethodInvokes(Fake<IFoo> fake, bool wasCalled)
         {
             "Given an unnatural fake"
                 .x(() => fake = new Fake<IFoo>());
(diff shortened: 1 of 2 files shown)


thomaslevesque:
Thanks @blairconrad!

As hard I as tried, I couldn't find anything wrong in your code, so as far as I'm concerned, it's good to go :wink:

I'll wait a bit before merging, in case @adamralph wants to have a look.

blairconrad:
> As hard I as tried, I couldn't find anything wrong in your code

I appreciate your dedication!

Thanks. I'm in no rush.

tests/FakeItEasy.Specs/UnnaturalFakeSpecs.cs
```diff
@@ -59,5 +159,20 @@ public static void AnyCall(Fake<IFoo> fake, bool wasCalled)
...
+        public class Foo : IFoo
+        {
+            public bool WasVoidMethodCalled { get; private set; }
+
+            public virtual void VoidMethod()
+            {
+                this.WasVoidMethodCalled = true;
+            }
+
+            public virtual int MethodWithResult()
```
adamralph: Is this used for anything?

blairconrad: No, I suppose not!
I'd started using `IFoo`s until I wanted to test `CallsBaseMethod` and then I foolishly made a class implement `IFoo`.
I've pushed another commit to remove `IFoo` as an interface of `Foo` (which I've renamed) and clean up one Code Analysis violation.
Well spotted. 
I guess @thomaslevesque [wasn't trying hard enough](https://github.com/FakeItEasy/FakeItEasy/pull/835#issuecomment-241655071). (I kid. All deficiencies in the code are my responsibility, of course.)

adamralph:
LGTM. Happy to merge after the `autosquash` :wink:

thomaslevesque:
Thanks @blairconrad !

PR #837, opened 2016-08-23
Wrapping exception

### LINKED ISSUE #836: Wrapping fake throws TargetInvocationException instead of the original exception
The following code:
```c#
void Main()
{
    var foo = new Foo();
    var wrapper = A.Fake<IFoo>(o => o.Wrapping(foo));
    wrapper.Bar(null);
}

public interface IFoo
{
    int Bar(string s);
}

public class Foo : IFoo
{
    public int Bar(string s)
    {
        return s.Length;
    }
}
```
throws a `TargetInvocationException`, but calling `foo.Bar(null)` directly throws a `NullReferenceException`. The wrapping fake should throw the original exception.

thomaslevesque:
Fixes #836

### DIFF
--- a/src/FakeItEasy/Core/WrappedObjectRule.cs
+++ b/src/FakeItEasy/Core/WrappedObjectRule.cs
@@ -1,5 +1,7 @@
 namespace FakeItEasy.Core
 {
+    using System.Reflection;
+
     /// <summary>
     /// A call rule that applies to any call and just delegates the
     /// call to the wrapped object.
@@ -50,7 +52,17 @@ namespace FakeItEasy.Core
             Guard.AgainstNull(fakeObjectCall, nameof(fakeObjectCall));
 
             var parameters = fakeObjectCall.Arguments.GetUnderlyingArgumentsArray();
-            var valueFromWrappedInstance = fakeObjectCall.Method.Invoke(this.wrappedObject, parameters);
+            object valueFromWrappedInstance;
+            try
+            {
+                valueFromWrappedInstance = fakeObjectCall.Method.Invoke(this.wrappedObject, parameters);
+            }
+            catch (TargetInvocationException ex)
+            {
+                ex.InnerException?.Rethrow();
+                throw;
+            }
+
             fakeObjectCall.SetReturnValue(valueFromWrappedInstance);
         }
     }
--- a/src/FakeItEasy/ExceptionExtensions.cs
+++ b/src/FakeItEasy/ExceptionExtensions.cs
@@ -1,0 +1,50 @@
+namespace FakeItEasy
+{
+    using System;
+#if FEATURE_EXCEPTION_DISPATCH_INFO
+    using System.Runtime.ExceptionServices;
+#else
+    using System.Diagnostics.CodeAnalysis;
+    using System.Reflection;
+#endif
+
+    internal static class ExceptionExtensions
+    {
+#if !FEATURE_EXCEPTION_DISPATCH_INFO
+        private static readonly Action<Exception> PreserveStackTrace = CreatePreserveStackTrace();
+#endif
+
+        /// <summary>
+        /// Re-throws an exception, trying to preserve its stack trace.
+        /// </summary>
+        /// <param name="exception">The exception to rethrow.</param>
+#if FEATURE_EXCEPTION_DISPATCH_INFO
+        public static void Rethrow(this Exception exception)
+        {
+            ExceptionDispatchInfo.Capture(exception).Throw();
+        }
+#else
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Appropriate in this case, since we can't do anything but ignore the exception.")]
+        public static void Rethrow(this Exception exception)
+        {
+            try
+            {
+                PreserveStackTrace(exception);
+            }
+            catch
+            {
+            }
+
+            throw exception;
+        }
+
+        private static Action<Exception> CreatePreserveStackTrace()
+        {
+            var method = typeof(Exception).GetMethod(
+                "InternalPreserveStackTrace",
+                BindingFlags.Instance | BindingFlags.NonPublic);
+            return (Action<Exception>)Delegate.CreateDelegate(typeof(Action<Exception>), null, method);
+        }
+#endif
+    }
+}
--- a/tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
+++ b/tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
@@ -1,0 +1,178 @@
+﻿namespace FakeItEasy.Specs
+{
+    using System;
+    using System.Diagnostics.CodeAnalysis;
+    using FakeItEasy.Tests.TestHelpers;
+    using FluentAssertions;
+    using Xbehave;
+    using Xunit;
+
+    public class WrappingFakeSpecs
+    {
+        public interface IFoo
+        {
+            int NonVoidMethod(string parameter);
+
+            void VoidMethod(string parameter);
+
+            [SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "0#", Justification = "Required for test scenario")]
+            [SuppressMessage("Microsoft.Design", "CA1021:AvoidOutParameters", MessageId = "1#", Justification = "Required for test scenario")]
+            void OutAndRefMethod(ref int @ref, out int @out);
+        }
+
+        [Scenario]
+        public void NonVoidSuccess(
+            Foo realObject,
+            IFoo wrapper,
+            int result)
+        {
+            "Given a real object"
+                .x(() => realObject = new Foo());
+
+            "And a fake wrapping this object"
+                .x(() => wrapper = A.Fake<IFoo>(o => o.Wrapping(realObject)));
+
+            "When a non-void method is called on the wrapper"
+                .x(() => result = wrapper.NonVoidMethod("hello"));
+
+            "Then the real object's method is called"
+                .x(() => realObject.NonVoidMethodCalled.Should().BeTrue());
+
+            "And the wrapper returns the value returned by the real object's method"
+                .x(() => result.Should().Be(5));
+        }
+
+        [Scenario]
+        public void NonVoidException(
+            Foo realObject,
+            IFoo wrapper,
+            Exception exception)
+        {
+            "Given a real object"
+                .x(() => realObject = new Foo());
+
+            "And a fake wrapping this object"
+                .x(() => wrapper = A.Fake<IFoo>(o => o.Wrapping(realObject)));
+
+            "When a non-void method is called on the wrapper"
+                .x(() => exception = Record.Exception(() => wrapper.NonVoidMethod(null)));
+
+            "Then the real object's method is called"
+                .x(() => realObject.NonVoidMethodCalled.Should().BeTrue());
+
+            "And the wrapper throws the exception thrown by the real object's method"
+                .x(() => exception.Should().BeAnExceptionOfType<ArgumentNullException>());
+
+            "And the exception stack trace is preserved"
+                .x(() => exception.StackTrace.Should().Contain("FakeItEasy.Specs.WrappingFakeSpecs.Foo.NonVoidMethod"));
+        }
+
+        [Scenario]
+        public void VoidSuccess(
+            Foo realObject,
+            IFoo wrapper)
+        {
+            "Given a real object"
+                .x(() => realObject = new Foo());
+
+            "And a fake wrapping this object"
+                .x(() => wrapper = A.Fake<IFoo>(o => o.Wrapping(realObject)));
+
+            "When a void method is called on the wrapper"
+                .x(() => wrapper.VoidMethod("hello"));
+
+            "Then the real object's method is called"
+                .x(() => realObject.VoidMethodCalled.Should().BeTrue());
+        }
+
+        [Scenario]
+        public void VoidException(
+            Foo realObject,
+            IFoo wrapper,
+            Exception exception)
+        {
+            "Given a real object"
+                .x(() => realObject = new Foo());
+
+            "And a fake wrapping this object"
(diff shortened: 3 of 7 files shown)


src/FakeItEasy/Core/WrappedObjectRule.cs
```diff
@@ -50,8 +58,32 @@ public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
...
+            var si = new SerializationInfo(e.GetType(), new FormatterConverter());
+
+            e.GetObjectData(si, ctx);
+            mgr.RegisterObject(e, 1, si); // prepare for SetObjectData
+            mgr.DoFixups(); // ObjectManager calls SetObjectData
+
+            throw e;
+#else
+            ExceptionDispatchInfo.Capture(e).Throw();
+#endif
```
thomaslevesque: I marked the PR as [WIP] because of this ugly hack. `ExceptionDispatchInfo` is only available since .NET 4.5, so I had to resort to this solution (found [here](http://stackoverflow.com/a/2085377/98713)). Another approach involves calling an internal method (`InternalPreserveStacktrace`) via reflection... not sure which is better.

blairconrad: I thought I'd made this comment earlier. Sorry if it's a duplicate. Consider looking at [FakeManager.TryPreserveStackTrace](https://github.com/FakeItEasy/FakeItEasy/blob/master/src/FakeItEasy/Core/FakeManager.EventRule.cs#L72). Not because it's necessarily better, but because it does (a subset of) what you're doing here, and we should probably have only the one way.

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
```diff
@@ -0,0 +1,169 @@
...
+            int refParameter,
+            int outParameter)
+        {
+            "Given a real object"
+                .x(() => realObject = new Foo());
+
+            "And a fake wrapping this object"
+                .x(() => wrapper = A.Fake<IFoo>(o => o.Wrapping(realObject)));
+
+            "When a method with out and ref parameters is called is called on the wrapper"
```
blairconrad: "is called is called"

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
```diff
@@ -0,0 +1,169 @@
...
+                if (parameter == null)
+                {
+                    throw new ArgumentNullException(nameof(parameter));
+                }
+            }
+
+            public void OutAndRefMethod(ref int refParameter, out int outParameter)
+            {
+                this.OutAndRefMethodCalled = true;
+                refParameter += 1;
```
blairconrad: Why a `+=` here, instead of a `=`? It works, but in theory could make tests more brittle, or require more logic in them.

src/FakeItEasy/Core/WrappedObjectRule.cs
thomaslevesque: Ah, I thought we must already have something like that, but didn't find it. Thanks!
However, it works on the full .NET Framework, but probably not on .NET Core... We must have missed it during the port to .NET Core because of the empty catch. I'll extract it to a separate class and make it work for .NET Core.

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
thomaslevesque: oops!

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
thomaslevesque: Because it's a ref param, so I wanted to ensure that we were passing the original value. I know it's a bit strange, especially since I don't initialize the parameter before the call, so it's always 0 (I never know where to initialize things in XBehave specs). If you prefer I can change it to just `refParameter = 1`.

src/FakeItEasy/Core/WrappedObjectRule.cs
thomaslevesque: Actually, the `InternalPreserveStackTrace` trick _does_ work on .NET Core. But it's probably still better to use `ExceptionDispatchInfo`, since it's public.

thomaslevesque:
OK, I'm clearly missing something about netstd/project.json

I tried to add a conditional compilation symbol in project.json:

```
"compilationOptions": {
    "define": [ "FEATURE_EXCEPTION_DISPATCH_INFO" ]
  },
```

but it has no effect at all. I had to put it in FakeItEasy.netstd.csproj instead. So what's the point of having project.json ? Is it only for dependencies?

thomaslevesque:
OK, I made the requested changes, except for the `+=`; it kinda bothers me to overwrite a ref param without even looking at its value, and ReSharper yells at me. If you think it's important, I'll change it.

I also added spec steps to check that the exception stack trace is preserved.

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
blairconrad: That's okay, I'm content.

blairconrad:
@thomaslevesque, thanks for at least thinking about the `+=`. I'm satisfied with your reasoning.
Also, thanks for including the `FakeManager` in your change.

blairconrad:
> So what's the point of having project.json

Darned if I know. I'm seeing them have more effect in an `xproj`-based project. And they seem to be _the place_ for specifying your multi-targeting. But, yeah, the whole thing is magic to me still. I'm flailing about, hoping things work, and only rarely do I understand why they do when they do.

src/FakeItEasy/ExceptionExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// Re-throws an exception, trying to preserve its stack trace.
+        /// </summary>
+        /// <param name="exception">The exception to rethrow.</param>
+#if FEATURE_EXCEPTION_DISPATCH_INFO
+        public static void Rethrow(this Exception exception)
+        {
+            ExceptionDispatchInfo.Capture(exception).Throw();
+        }
+#else
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Appropriate in try method.")]
```
blairconrad: Nit: it's no longer a "Try*" method.

thomaslevesque: True, I'll amend

blairconrad:
@thomaslevesque, it all looks good to me. Thanks!
Pausing for a bit in case any other comments come in.

thomaslevesque:
- fixed the justification in the `[SuppressMessage]` attribute
- cached a delegate for `InternalPreserveStackTrace` to avoid doing reflection every time
- squashed the fixup commits

src/FakeItEasy/Core/WrappedObjectRule.cs
```diff
@@ -50,8 +52,16 @@ public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
             Guard.AgainstNull(fakeObjectCall, nameof(fakeObjectCall));
 
             var parameters = fakeObjectCall.Arguments.GetUnderlyingArgumentsArray();
-            var valueFromWrappedInstance = fakeObjectCall.Method.Invoke(this.wrappedObject, parameters);
-            fakeObjectCall.SetReturnValue(valueFromWrappedInstance);
+            try
+            {
+                var valueFromWrappedInstance = fakeObjectCall.Method.Invoke(this.wrappedObject, parameters);
+                fakeObjectCall.SetReturnValue(valueFromWrappedInstance);
```
adamralph: Can this line also throw a `TargetInvocationException`?

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
adamralph: > I never know where to initialize things in XBehave specs

If the initialisation is a first class part of the scenario, then it should be mentioned in the natural language as one of the steps, e.g.

```gherkin
Given a real object
  And a fake wrapping the object
  And a value of zero
 When the value is passed by reference to a method which increments it
...
```

If it's irrelevant to the natural language, you can do it anywhere, e.g. at the last moment before the value is used.

src/FakeItEasy/Core/WrappedObjectRule.cs
thomaslevesque: Good question. I don't think so, but it couldn't hurt to move it outside the `try`. I'll do it

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
thomaslevesque: In this case, no, it's not really a first class part of the scenario. It's just the initial value of a `ref` parameter. Is it okay to do it between steps?

blairconrad:
Nice "fixup!".

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
adamralph: You shouldn't really do anything outside steps. The entire body of the test should really be inside the step delegates. Doing some stuff outside them will work, other stuff won't, so I think it's much better to keep everything inside them, e.g.

```C#
"When a method with out and ref parameters is called on the wrapper"
    .x(() =>
    {
        refParameter = 0;
        wrapper.OutAndRefMethod(ref refParameter, out outParameter);
    });
```

However, if you know that the wrapped object increments the ref param, then aren't the initial and final values fundamental to the scenario?

adamralph: I.e. ignoring `out` for a moment (perhaps it's better to split the scenario?) you could write the scenario like so:

```gherkin
Given an object with a method that increments an integer passed by reference
  And a fake which wraps the object
  And an integer
 When I pass the integer to the fake method
 Then the integer is incremented
```

Or being less specific:

```gherkin
Given an object with a method that assigns an argument passed by reference
  And a fake which wraps the object
  And another object
 When I pass the second object to the fake method
 Then the object is reassigned
```

adamralph: I guess the point I'm trying to make is that the natural language should adequately describe the scenario on it's own. I.e. you should be able to take the code away and still understand what the scenario is and the steps it describes.

thomaslevesque: @adamralph thanks for the explanation.

> However, if you know that the wrapped object increments the ref param, then aren't the initial and final values fundamental to the scenario?

Yes, you're right. I'll update the specs to initialize it explicitly.

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
```diff
@@ -0,0 +1,178 @@
...
+            int refParameter,
+            int outParameter)
+        {
+            "Given a real object"
+                .x(() => realObject = new Foo());
+
+            "And a fake wrapping this object"
+                .x(() => wrapper = A.Fake<IFoo>(o => o.Wrapping(realObject)));
+
+            "And a refParameter variable with a value of 1"
```
adamralph: nitpick: it's called "ref parameter" elsewhere in the natural language.

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
```diff
@@ -0,0 +1,178 @@
...
+            "And the exception stack trace is preserved"
+                .x(() => exception.StackTrace.Should().Contain("FakeItEasy.Specs.WrappingFakeSpecs.Foo.VoidMethod"));
+        }
+
+        [Scenario]
+        public void OutAndRef(
+            Foo realObject,
+            IFoo wrapper,
+            int refParameter,
+            int outParameter)
```
adamralph: nitpick: the variables referred to as `refParameter` and `outParameter" are really just variables. The method has "out and ref parameters" and the variables are arguments for the method call. Perhaps `refVariable` and `outVariable` would be better names? (Although they still sound a bit strange.)

thomaslevesque: Yes, I could rename them. In fact, I had initially named them `x` and `y`, but FxCop complained that these names weren't meaningful enough...

adamralph: You could even call them `@ref` and `@out`.

thomaslevesque:
I made the changes in the specs WRT to the ref/out params, and squashed the fixup commits

blairconrad:
Sorry for the delay, @thomaslevesque. It looks very good to me. Thanks!

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
```diff
@@ -0,0 +1,178 @@
...
+            "When a method with out and ref parameters is called on the wrapper"
+                .x(() => wrapper.OutAndRefMethod(ref @ref, out outParameter));
+
+            "Then the real object's method is called"
+                .x(() => realObject.OutAndRefMethodCalled.Should().BeTrue());
+
+            "And the value of @ref is incremented"
+                .x(() => @ref.Should().Be(2));
+
+            "And the value of @out is set"
```
adamralph: Sorry to keep nitpicking but I guess this means you intended to rename `outParameter` to `@out`?

blairconrad: Man. I'm going to stop saying how great everything is.

thomaslevesque: > Sorry to keep nitpicking but I guess this means you intended to rename outParameter to @out?

yes ! I thought I had, but apparently not... I will amend (again)

thomaslevesque: Done

adamralph: :ok_hand:

adamralph:
Thanks @thomaslevesque!

thomaslevesque:
Thanks for the review, and for the merge!

PR #840, opened 2016-08-27
Make A.CallTo side-effect free

### LINKED ISSUE #556: A.CallTo is not side-effect free - can interfere with the usual default behaviour of a faked method
Probably this topic has come up before but...

We're using FakeItEasy 2.0.0-beta009 and I am noticing a horrible pattern in our test code.
Unlike methods which return void that just silently finish by default, methods returning Task will return null by default and blow up if awaited.
Would there be a downside of having Fakes return a completed Task by default if the method return type is `Task` (and not `Task<T>`)?
If there is too much downside to changing `Fake<T>` what about introducing a new class 'Fake2<T>' like Fake but that which would have that behavior by default?

```C#
var createDeploymentCall = A.CallTo(() =>
    worker.CreateCacheDeployment(null, null, null, null, null, null, null, A<CancellationToken>._))
    .WithAnyArguments();
createDeploymentCall.Returns(Task.Delay(0));
var deleteStageSlotCall = A.CallTo( () => azureTasks.DeleteStageSlotAsync(null, null, A<CancellationToken>._)).WithAnyArguments();
deleteStageSlotCall.Returns(Task.Delay(0));
var swapDeploymentCall = A.CallTo(() => azureTasks.SwapDeploymentAsync(null, null, null, A<CancellationToken>._)).WithAnyArguments();
swapDeploymentCall.Returns(Task.Delay(0));
```

thomaslevesque:
Fixes #556 

~~This is my first stab at it, still a work in progress.~~
~~The implementation is not yet complete, there are a few other cases that I need to look into.~~
~~The current tests pass, but I need to add new ones to validate the new behavior.~~

### DIFF
--- a/src/FakeItEasy/Configuration/RuleBuilder.cs
+++ b/src/FakeItEasy/Configuration/RuleBuilder.cs
@@ -10,6 +10,7 @@ namespace FakeItEasy.Configuration
     {
         private readonly FakeAsserter.Factory asserterFactory;
         private readonly FakeManager manager;
+        private bool wasRuleAdded;
 
         internal RuleBuilder(BuildableCallRule ruleBeingBuilt, FakeManager manager, FakeAsserter.Factory asserterFactory)
         {
@@ -48,6 +49,7 @@ namespace FakeItEasy.Configuration
 
         public virtual IAfterCallSpecifiedConfiguration Throws(Func<IFakeObjectCall, Exception> exceptionFactory)
         {
+            this.AddRuleIfNeeded();
             this.RuleBeingBuilt.UseApplicator(x => { throw exceptionFactory(x); });
             return this;
         }
@@ -62,6 +64,7 @@ namespace FakeItEasy.Configuration
 
         public virtual IAfterCallSpecifiedConfiguration DoesNothing()
         {
+            this.AddRuleIfNeeded();
             this.RuleBeingBuilt.UseDefaultApplicator();
             return this;
         }
@@ -69,13 +72,14 @@ namespace FakeItEasy.Configuration
         public virtual IVoidConfiguration Invokes(Action<IFakeObjectCall> action)
         {
             Guard.AgainstNull(action, nameof(action));
-
+            this.AddRuleIfNeeded();
             this.RuleBeingBuilt.Actions.Add(action);
             return this;
         }
 
         public virtual IAfterCallSpecifiedConfiguration CallsBaseMethod()
         {
+            this.AddRuleIfNeeded();
             this.RuleBeingBuilt.UseApplicator(x => { });
             this.RuleBeingBuilt.CallBaseMethod = true;
             return this;
@@ -85,6 +89,7 @@ namespace FakeItEasy.Configuration
         {
             Guard.AgainstNull(valueProducer, nameof(valueProducer));
 
+            this.AddRuleIfNeeded();
             this.RuleBeingBuilt.OutAndRefParametersValueProducer = valueProducer;
 
             return this;
@@ -94,7 +99,6 @@ namespace FakeItEasy.Configuration
         {
             Guard.AgainstNull(repeatConstraint, nameof(repeatConstraint));
 
-            this.manager.RemoveRule(this.RuleBeingBuilt);
             var asserter = this.asserterFactory.Invoke(this.Calls);
 
             var description = new StringBuilderOutputWriter();
@@ -103,6 +107,15 @@ namespace FakeItEasy.Configuration
             asserter.AssertWasCalled(this.Matcher.Matches, description.Builder.ToString(), repeatConstraint.Matches, repeatConstraint.ToString());
 
             return new UnorderedCallAssertion(this.manager, this.Matcher, description.Builder.ToString(), repeatConstraint);
+        }
+
+        private void AddRuleIfNeeded()
+        {
+            if (!this.wasRuleAdded)
+            {
+                this.manager.AddRuleFirst(this.RuleBeingBuilt);
+                this.wasRuleAdded = true;
+            }
         }
 
         public class ReturnValueConfiguration<TMember>
@@ -127,16 +140,14 @@ namespace FakeItEasy.Configuration
             public IAfterCallSpecifiedWithOutAndRefParametersConfiguration ReturnsLazily(Func<IFakeObjectCall, TMember> valueProducer)
             {
                 Guard.AgainstNull(valueProducer, nameof(valueProducer));
-
+                this.ParentConfiguration.AddRuleIfNeeded();
                 this.ParentConfiguration.RuleBeingBuilt.UseApplicator(x => x.SetReturnValue(valueProducer(x)));
                 return this.ParentConfiguration;
             }
 
             public IReturnValueConfiguration<TMember> Invokes(Action<IFakeObjectCall> action)
             {
-                Guard.AgainstNull(action, nameof(action));
-
-                this.ParentConfiguration.RuleBeingBuilt.Actions.Add(action);
+                this.ParentConfiguration.Invokes(action);
                 return this;
             }
 
--- a/tests/FakeItEasy.Tests/ACallToTests.cs
+++ b/tests/FakeItEasy.Tests/ACallToTests.cs
@@ -1,6 +1,8 @@
 namespace FakeItEasy.Tests
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
     using System.Linq.Expressions;
     using FakeItEasy.Configuration;
     using FluentAssertions;
@@ -16,6 +18,13 @@ namespace FakeItEasy.Tests
             this.configurationManager = A.Fake<IFakeConfigurationManager>(x => x.Wrapping(ServiceLocator.Current.Resolve<IFakeConfigurationManager>()));
             this.StubResolve(this.configurationManager);
         }
+
+        public static IEnumerable<object[]> CallSpecificationActions =>
+            TestCases.FromObject<Action<IFoo>>(
+                foo => A.CallTo(() => foo.Bar()),
+                foo => A.CallTo(() => foo.Baz()),
+                foo => A.CallToSet(() => foo.SomeProperty),
+                foo => A.CallTo(foo));
 
         [Fact]
         public void CallTo_with_void_call_should_return_configuration_from_configuration_manager()
@@ -49,6 +58,22 @@ namespace FakeItEasy.Tests
 
             // Assert
             result.Should().BeSameAs(configuration);
+        }
+
+        [Theory]
+        [MemberData(nameof(CallSpecificationActions))]
+        public void CallTo_should_not_add_rule_to_manager(Action<IFoo> action)
+        {
+            // Arrange
+            var foo = A.Fake<IFoo>();
+            var manager = Fake.GetFakeManager(foo);
+            var initialRules = manager.Rules.ToList();
+
+            // Act
+            action(foo);
+
+            // Assert
+            manager.Rules.Should().Equal(initialRules);
         }
     }
 }
--- a/tests/FakeItEasy.Tests/Configuration/AnyCallConfigurationTests.cs
+++ b/tests/FakeItEasy.Tests/Configuration/AnyCallConfigurationTests.cs
@@ -1,6 +1,8 @@
 namespace FakeItEasy.Tests.Configuration
 {
     using System;
+    using System.Collections.Generic;
+    using System.Linq;
     using FakeItEasy.Configuration;
     using FakeItEasy.Core;
     using FluentAssertions;
@@ -22,6 +24,13 @@ namespace FakeItEasy.Tests.Configuration
         [UnderTest]
         private AnyCallConfiguration configuration;
 #pragma warning restore 649
+
+        public static IEnumerable<object[]> CallSpecificationActions =>
+            TestCases.FromObject<Action<IAnyCallConfigurationWithNoReturnTypeSpecified>>(
+                configuration => configuration.WithReturnType<int>(),
+                configuration => configuration.WithNonVoidReturnType(),
+                configuration => configuration.Where(call => true),
+                configuration => configuration.WhenArgumentsMatch(args => true));
 
         [Fact]
         public void WithReturnType_should_return_configuration_from_factory()
@@ -277,6 +286,22 @@ namespace FakeItEasy.Tests.Configuration
 
             // Assert
             this.configuration.WhenArgumentsMatch(x => true).Should().BeSameAs(this.configuration);
+        }
+
+        [Theory]
+        [MemberData(nameof(CallSpecificationActions))]
+        public void Call_specification_method_should_not_add_rule_to_manager(Action<IAnyCallConfigurationWithNoReturnTypeSpecified> configurationAction)
+        {
+            // Arrange
+            var foo = A.Fake<IFoo>();
+            var manager = Fake.GetFakeManager(foo);
+            var initialRules = manager.Rules.ToList();
+
+            // Act
+            configurationAction(A.CallTo(foo));
+
+            // Assert
+            manager.Rules.Should().Equal(initialRules);
         }
 
         private IVoidArgumentValidationConfiguration StubVoidConfig()
--- a/tests/FakeItEasy.Tests/Configuration/FakeConfigurationManagerTests.cs
+++ b/tests/FakeItEasy.Tests/Configuration/FakeConfigurationManagerTests.cs
@@ -1,6 +1,7 @@
 namespace FakeItEasy.Tests.Configuration
 {
     using System;
+    using System.Collections.Generic;
     using System.Linq;
     using System.Linq.Expressions;
     using FakeItEasy.Configuration;
@@ -23,6 +24,13 @@ namespace FakeItEasy.Tests.Configuration
         {
             this.OnSetup();
         }
+
+        public static IEnumerable<object[]> CallSpecificationActions =>
+            TestCases.FromObject<Action<FakeConfigurationManagerTests, IFoo>>(
+                (@this, foo) => @this.configurationManager.CallTo(() => foo.Bar()),
+                (@this, foo) => @this.configurationManager.CallTo(() => foo.Baz()),
+                (@this, foo) => @this.configurationManager.CallToSet(() => foo.SomeProperty),
+                (@this, foo) => @this.configurationManager.CallTo(foo));
 
         // Callto
         [Fact]
@@ -145,6 +153,22 @@ namespace FakeItEasy.Tests.Configuration
 
             // Assert
             result.Should().BeSameAs(expectedConfiguration);
+        }
+
+        [Theory]
+        [MemberData(nameof(CallSpecificationActions))]
+        public void CallTo_should_not_add_rule_to_manager(Action<FakeConfigurationManagerTests, IFoo> action)
+        {
+            // Arrange
+            var foo = A.Fake<IFoo>();
+            var manager = Fake.GetFakeManager(foo);
+            var initialRules = manager.Rules.ToList();
+
+            // Act
+            action(this, foo);
+
+            // Assert
+            manager.Rules.Should().Equal(initialRules);
         }
 
         private void OnSetup()
--- a/tests/FakeItEasy.Tests/Configuration/FakeConfigurationManagerTests.cs
+++ b/tests/FakeItEasy.Tests/Configuration/FakeConfigurationManagerTests.cs
@@ -147,20 +147,6 @@ namespace FakeItEasy.Tests.Configuration
             result.Should().BeSameAs(expectedConfiguration);
         }
 
-        [Fact]
-        public void Should_add_call_rule_to_fake_manager_when_configuring_any_call()
-        {
-            // Arrange
-            var fake = A.Fake<IFoo>();
-            var manager = Fake.GetFakeManager(fake);
-
-            // Act
-            this.configurationManager.CallTo(fake);
-
-            // Assert
-            manager.AllUserRules.Single().Rule.Should().BeOfType<AnyCallCallRule>();
-        }
-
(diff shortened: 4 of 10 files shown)


src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -95,7 +100,6 @@ public UnorderedCallAssertion MustHaveHappened(Repeated repeatConstraint)
         {
             Guard.AgainstNull(repeatConstraint, nameof(repeatConstraint));
 
-            this.manager.RemoveRule(this.RuleBeingBuilt);
```
thomaslevesque: This was only necessary because the rule was added to the `FakeManager`

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -128,16 +141,14 @@ public IAfterCallSpecifiedConfiguration Throws(Func<IFakeObjectCall, Exception>
...
                 this.ParentConfiguration.RuleBeingBuilt.UseApplicator(x => x.SetReturnValue(valueProducer(x)));
                 return this.ParentConfiguration;
             }
 
             public IReturnValueConfiguration<TMember> Invokes(Action<IFakeObjectCall> action)
             {
-                Guard.AgainstNull(action, nameof(action));
-
-                this.ParentConfiguration.RuleBeingBuilt.Actions.Add(action);
+                this.ParentConfiguration.Invokes(action);
```
thomaslevesque: This was doing the same as the parent implementation from `RuleBuilder`, so I thought it might as well call the parent implementation directly.

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -148,9 +159,7 @@ public IAfterCallSpecifiedConfiguration CallsBaseMethod()
 
             public IReturnValueConfiguration<TMember> WhenArgumentsMatch(Func<ArgumentCollection, bool> argumentsPredicate)
             {
-                Guard.AgainstNull(argumentsPredicate, nameof(argumentsPredicate));
-
-                this.ParentConfiguration.RuleBeingBuilt.UsePredicateToValidateArguments(argumentsPredicate);
+                this.ParentConfiguration.WhenArgumentsMatch(argumentsPredicate);
```
thomaslevesque: Same as above, although it's not related to what I'm trying to fix in this PR (I can move it to a different PR if you prefer)

tests/FakeItEasy.Tests/Configuration/FakeConfigurationManagerTests.cs
```diff
@@ -147,20 +147,6 @@ public void Should_return_configuration_when_configuring_any_call()
...
-            var fake = A.Fake<IFoo>();
-            var manager = Fake.GetFakeManager(fake);
-
-            // Act
-            this.configurationManager.CallTo(fake);
-
-            // Assert
-            manager.AllUserRules.Single().Rule.Should().BeOfType<AnyCallCallRule>();
-        }
-
```
thomaslevesque: This test was ensuring the old behavior, so it no longer makes sense.

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
```diff
@@ -331,19 +330,6 @@ public void Assert_with_void_call_should_assert_on_assertions_produced_by_factor
...
-            this.fakeManager.AddRuleFirst(this.ruleProducedByFactory);
-
-            // Act
-            this.builder.MustHaveHappened();
-
-            // Assert
-            this.fakeManager.Rules.Should().BeEmpty();
-        }
-
-        [Fact]
```
thomaslevesque: No longer necessary.

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
```diff
@@ -363,20 +349,6 @@ public void Assert_with_function_call_should_assert_on_assertions_produced_by_fa
...
-
-            // Act
-            var returnConfig = new RuleBuilder.ReturnValueConfiguration<int>(this.builder);
-            returnConfig.MustHaveHappened();
-
-            // Assert
-            this.fakeManager.Rules.Should().BeEmpty();
-        }
-
-        [Fact]
```
thomaslevesque: No longer necessary.

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -11,6 +11,7 @@ internal class RuleBuilder
     {
         private readonly FakeAsserter.Factory asserterFactory;
         private readonly FakeManager manager;
+        private bool ruleAdded;
```
blairconrad: Consider adding a prefix such as "was" to make this sound more like a Boolean than an event. (And a corresponding change to `EnsureRuleAdded`?)

src/FakeItEasy/Configuration/RuleBuilder.cs
blairconrad: It's not such a big deal to me either way. You could consider a separate commit, I suppose.

thomaslevesque:
Thanks for the comments @blairconrad!

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -110,12 +110,12 @@ public UnorderedCallAssertion MustHaveHappened(Repeated repeatConstraint)
             return new UnorderedCallAssertion(this.manager, this.Matcher, description.Builder.ToString(), repeatConstraint);
         }
 
-        private void EnsureRuleAdded()
+        private void EnsureRuleIsAdded()
```
thomaslevesque: I hesitated beween `EnsureRuleWasAdded` and `EnsureRuleIsAdded`... I found the former a bit strange, so I picked the latter. Let me know if you think differently (I don't really care either way)

blairconrad: `AddRuleIfNeeded`?

thomaslevesque: :+1:

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: I eventually put it in another commit and cherry picked it to another branch; I'll submit a separate PR later.

thomaslevesque:
I added some tests. Because of the nature of the fix, I thought it best to write unit tests rather than specs; I guess I could *also* add specs, but they would probably be a little awkward.

As far as I'm concerned, I think the PR is ready.

tests/FakeItEasy.Tests/ACallToTests.cs
```diff
@@ -50,5 +68,22 @@ public void CallTo_with_function_call_should_return_configuration_from_configura
             // Assert
             result.Should().BeSameAs(configuration);
         }
+
+        [SuppressMessage("Microsoft.Design", "CA1062:Validate arguments of public methods", MessageId = "0", Justification = "It's not really a public method")]
```
blairconrad: I'm surprised we're not hounded by this in all the specs. I don't see a suppression.

I could support this moving to the global suppression file for all test projects, if you'd like me to add it to my PR.

thomaslevesque: > I'm surprised we're not hounded by this in all the specs. I don't see a suppression.

Maybe because the arguments in scenarios aren't really arguments, but are used more like local variables. We always assign them before using them, so it wouldn't make sense to validate them. (I'm not sure why we don't declare them as local variables, by the way)



> I could support this moving to the global suppression file for all test projects, if you'd like me to add it to my PR.

:+1:

tests/FakeItEasy.Tests/ACallToTests.cs
```diff
@@ -17,6 +20,21 @@ public ACallToTests()
...
+            {
+                var actions = new Action<IFoo>[]
+                {
+                    foo => A.CallTo(() => foo.Bar()),
+                    foo => A.CallTo(() => foo.Baz()),
+                    foo => A.CallToSet(() => foo.SomeProperty),
+                    foo => A.CallTo(foo),
+                };
+                return actions.Select(action => new object[] { action });
+            }
```
blairconrad: I was surprised that you had to write all this.
If you're interested in shortening this, we could add a `TestCases.FromObject` overload:

```c#
public static IEnumerable<object[]> FromObject<T>(params T[] cases)
{
    return cases.Select(@case => new object[] { @case });
}
```

tests/FakeItEasy.Tests/Configuration/FakeConfigurationManagerTests.cs
```diff
@@ -147,18 +164,21 @@ public void Should_return_configuration_when_configuring_any_call()
...
             // Arrange
-            var fake = A.Fake<IFoo>();
-            var manager = Fake.GetFakeManager(fake);
+            var foo = A.Fake<IFoo>();
+            var manager = Fake.GetFakeManager(foo);
+            var initialRules = manager.Rules.ToList();
 
             // Act
-            this.configurationManager.CallTo(fake);
+            action(this, foo);
```
blairconrad: I found all the `@this`es up near line 35 distracting. How do you feel about just passing `configurationManager` in here instead of `this`? I think it better reveals which object's being manipulated.

blairconrad:
Sorry for the delay, @thomaslevesque. Generally looks very good. I made  a couple of minor comments.
Two other points:

1. I may be slow, but didn't notice tests covering the behaviour of the unnatural Fakes. Probably you just need to point me at a file.
2. You mentioned the possibility of adding specs. I can live without them. Had you wanted to, though, I'm not sure they'd be completely awkward. Sample tests could just make sure that we still get `Dummies` back when adding an `A.CallTo` to a method, or make sure that the fakes are still strict. Still, I think you're fine.

tests/FakeItEasy.Tests/ACallToTests.cs
thomaslevesque: Good idea, thanks!

tests/FakeItEasy.Tests/Configuration/FakeConfigurationManagerTests.cs
thomaslevesque: That was my initial intention. But `FakeConfigurationManager` is internal, and test methods have to be public, so they can't take it as a parameter.

thomaslevesque:
> I may be slow, but didn't notice tests covering the behaviour of the unnatural Fakes. Probably you just need to point me at a file.

Ah, I forgot about those. I'll add some tests.

> You mentioned the possibility of adding specs. I can live without them. Had you wanted to, though, I'm not sure they'd be completely awkward. Sample tests could just make sure that we still get Dummies back when adding an A.CallTo to a method, or make sure that the fakes are still strict. Still, I think you're fine.

OK I'll try to add some specs.

tests/FakeItEasy.Tests/Configuration/FakeConfigurationManagerTests.cs
blairconrad: Whoops. I didn't look that deeply. I should've figured you had a good reason and tried to suss it out myself.

blairconrad:
> OK I'll try to add some specs.

I wasn't pushing! And there'd be a huge number if you're going to provide the same level of coverage you did in the tests. Me, I'd be satisfied with no specs, or more than satisfied with even a couple of them to demonstrate how the problem first appeared to @TimLovellSmith.

thomaslevesque:
> I wasn't pushing! And there'd be a huge number if you're going to provide the same level of coverage you did in the tests. Me, I'd be satisfied with no specs, or more than satisfied with even a couple of them to demonstrate how the problem first appeared to @TimLovellSmith.

Oh, I didn't really intend to achieve full coverage through the specs... just a few scenarios to demonstrate that the issue is no longer there.

thomaslevesque:
Didn't have time to get back to this yet. It'll have to wait till I get back from vacation.

tests/FakeItEasy.Tests/ACallToTests.cs
adamralph: > (I'm not sure why we don't declare them as local variables, by the way)

You can if you want, it's just more verbose.

Early versions of xbehave didn't allow arguments in `Scenario` methods unless they were all matched with `Example` values (just like xunit doesn't allow them in `Fact` methods, or in `Theory` methods unless they are all matched with `InlineData` values or similar).

Allowing them to be declared as method arguments instead of local variables was a feature proposed by an xbehave user. Since the feature was introduced it's become the idiomatic way of writing xbehave scenarios.

blairconrad: @thomaslevesque, You should be able to remove these lines now that #846 has been merged.

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
```diff
@@ -32,6 +34,30 @@ public RuleBuilderTests()
...
+            TestCases.FromObject<Action<IAnyCallConfigurationWithReturnTypeSpecified<int>>>(
+                configuration => configuration.WhenArgumentsMatch(args => true),
+                configuration => configuration.Where(call => true));
+
+        public static IEnumerable<object[]> BehaviorDefinitionActionsForVoid =>
+            TestCases.FromObject<Action<IVoidArgumentValidationConfiguration>>(
+                configuration => configuration.CallsBaseMethod(),
+                configuration => configuration.DoesNothing(),
+                configuration => configuration.Throws<Exception>(),
+                configuration => configuration.Invokes(DoNothing),
```
thomaslevesque: This is a workaround for a StyleCop bug. I had initially written it like this:

    configuration => configuration.Invokes(_ => { }),

But StyleCop was reporting a warning:

> SA1502 : CSharp.Layout : The property must not be placed on a single line. The opening and closing curly brackets must each be placed on their own line.

There was also a warning on the following line for `new object[] { }`. Apparently StyleCop gets confused by the curly brackets in an expression-bodied property...

thomaslevesque:
I amended the second commit to remove the FxCop suppressions and add tests for unnatural fakes, and I added some (minimal) specs. I also used `TestCases.FromObject` as suggested by @blairconrad. I think that's all...

thomaslevesque:
There's still one thing that bothers me, though. Ideally, we shouldn't add a rule to the `FakeManager` until it's fully configured. But the definition of "fully configured" is unclear... For instance, if we do this:

    var foo = A.Fake<IFoo>(o => o.Strict());
    A.CallTo(() => foo.Bar()).Invokes(() => { });
    int x = foo.Bar();

It works, and `foo.Bar()` returns 0. But we never configured the return value, so the rule wasn't fully configured... since it's a strict fake, shouldn't we throw an exception in this case?

We could add the rule to the `FakeManager` only when the call is configured to return a value or throw an exception, but then the `Invokes` in the above snippet will silently be ignored...

blairconrad:
> since it's a strict fake, shouldn't we throw an exception in this case

That's certainly an interpretation. Another is "you broke it, you bought it". By which I mean that as soon as you specify any behaviour for a method, you have to fully specify it.

Huh. I'm a little surprised that we can specify an `Invokes` without a `Returns` (or equivalent). I thought most of the actions on a non-void call were placed after the `Returns/Throws/CallsBaseMethod` in the API. `AssignsOutAndRefParameters` certainly seems to be in that category.

Am I crazy to think that it'd be better for `Invokes` not to be specifiable until the `Returns` or similar has been?

tests/FakeItEasy.Tests/ACallToTests.cs
```diff
@@ -1,6 +1,9 @@
 namespace FakeItEasy.Tests
 {
     using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
```
blairconrad: unused

tests/FakeItEasy.Tests/Configuration/AnyCallConfigurationTests.cs
```diff
@@ -1,6 +1,9 @@
 namespace FakeItEasy.Tests.Configuration
 {
     using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
```
blairconrad: also unused

tests/FakeItEasy.Tests/Configuration/FakeConfigurationManagerTests.cs
```diff
@@ -1,6 +1,8 @@
 namespace FakeItEasy.Tests.Configuration
 {
     using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
```
blairconrad: still unused

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
```diff
@@ -1,13 +1,15 @@
 namespace FakeItEasy.Tests.Configuration
 {
     using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
```
blairconrad: I probably don't need to make my now-traditional nitpick, but…

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
```diff
@@ -32,6 +34,30 @@ public RuleBuilderTests()
...
+                configuration => configuration.Throws<Exception>(),
+                configuration => configuration.Invokes(DoNothing),
+                configuration => configuration.AssignsOutAndRefParametersLazily(_ => new object[0]));
+
+        public static IEnumerable<object[]> BehaviorDefinitionActionsForNonVoid =>
+            TestCases.FromObject<Action<IAnyCallConfigurationWithReturnTypeSpecified<int>>>(
+                configuration => configuration.CallsBaseMethod(),
+                configuration => configuration.Throws<Exception>(),
+                configuration => configuration.Invokes(DoNothing),
+                configuration => configuration.ReturnsLazily(_ => 0));
```
blairconrad: I was going to say that I can live without `AssignsOutAndRefParametersLazily`, but in case you accidentally forgot, you could add.
Then I realized that _you can't_, because unlike `Invokes`, as we were talking about earlier, you can't assign the outs and refs until you return something. Or similar.

blairconrad: You know, so ignore my comment!

thomaslevesque:
> Am I crazy to think that it'd be better for Invokes not to be specifiable until the Returns or similar has been?

I agree that it would be better, but it's probably not easy to achieve. The problem is that for calls configured with `A.CallTo(fake)` (`IAnyCallConfigurationWithNoReturnTypeSpecified`), you can specify `Invokes`, then use `WithReturnType<T>` or `WithNonVoidReturnType` and specify a return value (or not).

thomaslevesque:
Or maybe we could just remove `ICallbackConfiguration<TInterface>` from `IReturnValueConfiguration<TMember>`'s base interfaces, and add it to `IAfterCallSpecifiedWithOutAndRefParametersConfiguration`'s base interfaces.

thomaslevesque:
Of course, it would be a breaking change...

thomaslevesque:
I fixed the redundant `using`s

blairconrad:
> Or maybe we could just remove `ICallbackConfiguration<TInterface>` from `IReturnValueConfiguration<TMember>`'s base interfaces, and add it to `IAfterCallSpecifiedWithOutAndRefParametersConfiguration`'s base interfaces.

That's closer to what I was thinking, yeah. Although we may not want it to be on exactly the same interfaces, since we should be able to combine `Invokes` with `AssignsOutAndRefParametersLazily`, and probably even specify any number of `Invokes`. (Although if the latter becomes problematic, we could drop it.)

And yes, it's breaking. Which makes me sad, as usual.

I understand the reasons not to make the change. But if we don't, then I think it's rude to force users (at runtime or via the analyzer) to specify a return value/throws/callsbasemethod in addition to the `Invokes`.
In that case, I think an `Invokes` should probably count as "completing call configuration", and we come back to figuring out what to do with the return value/throws/callsbasemethods behaviour. Me, I think the new rule configuration should reset everything, so we get default return values again. Of course, that leads me back to the idea of requiring some other action with the Invokes in order to avoid surprising users.

I guess long story short, I think the API should guide the users into the correct actions, and I'd like for Invokes to only be possible after specifying returns/throws/callsbasemethods.
But I do hate to jerk the existing users around.

thomaslevesque:
> I guess long story short, I think the API should guide the users into the correct actions, and I'd like for Invokes to only be possible after specifying returns/throws/callsbasemethods.

Agreed, but it should probably be done in a separate issue.

blairconrad:
Of course!
We should discuss elsewhere, but make a decision soonish so we know how to proceed here.

thomaslevesque:
> We should discuss elsewhere, but make a decision soonish so we know how to proceed here.

Do we need to make this decision now before we merge this PR? I was thinking of merging it now (since it's consistent with the current API) and making additional changes later if necessary.

blairconrad:
Ah, you said you were bothered by it, so I thought you wanted to resolve the ideal behaviour before moving on this.
As you say, `Invokes` won't be doing anything different than before. I agree, the `Invokes` question shouldn't be a barrier to merging.

thomaslevesque:
> Ah, you said you were bothered by it, so I thought you wanted to resolve the ideal behaviour before moving on this.

Well, this issue does bother me, but resolving it is probably out of scope for this PR. I'd rather fix it separately.

blairconrad:
The changes look good to me. Thanks, @thomaslevesque!

thomaslevesque:
Thanks for the merge @blairconrad!

blairconrad:
Always a pleasure. Of the FakeItEasy contributors, you're definitely in my top 23 favourites to merge PRs for!

PR #841, opened 2016-08-29
Suppress code analysis warning

blairconrad:
Introduced in #837.


thomaslevesque:
I was pretty sure I had fixed all warnings, but I must have introduced new ones when I renamed the parameters to `@out` and `@ref`... Thanks @blairconrad!

blairconrad:
Thanks, @thomaslevesque.

PR #842, opened 2016-08-29
Add characterization specs for self-initialized fakes

blairconrad:
I thought it might be a good idea to see how our self-initialized fakes actually function.

### DIFF
--- a/tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
+++ b/tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
@@ -2,9 +2,11 @@ namespace FakeItEasy.Specs
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
     using System.IO;
     using System.Linq;
     using FakeItEasy.SelfInitializedFakes;
+    using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
     using Xbehave;
     using Xunit;
@@ -12,6 +14,14 @@ namespace FakeItEasy.Specs
     public interface ILibraryService
     {
         int GetCount(string internationalStandardBookNumber);
+
+        string GetTitle(string internationalStandardBookNumber);
+
+        [SuppressMessage("Microsoft.Design", "CA1021:AvoidOutParameters", MessageId = "0#", Justification = "Required for test scenario")]
+        [SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "1#", Justification = "Required for test scenario")]
+        [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "ref", Justification = "There are no other consumers, so the name is safe.")]
+        [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "out", Justification = "There are no other consumers, so the name is safe.")]
+        bool TryToSetSomeOutAndRefParameters(out int @out, ref int @ref);
     }
 
     public static class SelfInitializedFakesSpecs
@@ -43,37 +53,37 @@ namespace FakeItEasy.Specs
 
             "When I use a self-initialized fake in recording mode to get the counts for book 1, 2, and 1 again"
                 .x(() =>
-                {
-                    using (var recorder = new RecordingManager(inMemoryStorage))
                     {
-                        var fakeService = A.Fake<ILibraryService>(options => options
-                            .Wrapping(realServiceWhileRecording).RecordedBy(recorder));
-
-                        countsWhileRecording = new List<int>
+                        using (var recorder = new RecordingManager(inMemoryStorage))
                         {
-                            fakeService.GetCount("1"),
-                            fakeService.GetCount("2"),
-                            fakeService.GetCount("1")
-                        };
-                    }
-                });
+                            var fakeService = A.Fake<ILibraryService>(options => options
+                                .Wrapping(realServiceWhileRecording).RecordedBy(recorder));
+
+                            countsWhileRecording = new List<int>
+                            {
+                                fakeService.GetCount("1"),
+                                fakeService.GetCount("2"),
+                                fakeService.GetCount("1")
+                            };
+                        }
+                    });
 
             "And I use a self-initialized fake in playback mode to get the counts for book 1, 2, and 1 again"
                 .x(() =>
-                {
-                    using (var recorder = new RecordingManager(inMemoryStorage))
                     {
-                        var playbackFakeService = A.Fake<ILibraryService>(options => options
-                            .Wrapping(realServiceDuringPlayback).RecordedBy(recorder));
-
-                        countsDuringPlayback = new List<int>
+                        using (var recorder = new RecordingManager(inMemoryStorage))
                         {
-                            playbackFakeService.GetCount("1"),
-                            playbackFakeService.GetCount("2"),
-                            playbackFakeService.GetCount("1")
-                        };
-                    }
-                });
+                            var playbackFakeService = A.Fake<ILibraryService>(options => options
+                                .Wrapping(realServiceDuringPlayback).RecordedBy(recorder));
+
+                            countsDuringPlayback = new List<int>
+                            {
+                                playbackFakeService.GetCount("1"),
+                                playbackFakeService.GetCount("2"),
+                                playbackFakeService.GetCount("1")
+                            };
+                        }
+                    });
 
             "Then the recording fake forwards calls to the wrapped service"
                 .x(() => A.CallTo(() => realServiceWhileRecording.GetCount("1"))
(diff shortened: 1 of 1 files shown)


blairconrad:
I may have gone overboard breaking up the first 3 commits, but figured I'd err on the side of small.
Happy to squash however people want.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
```diff
@@ -34,10 +34,10 @@ public static class SelfInitializedFakesSpecs
                         realServiceWhileRecording = A.Fake<ILibraryService>();
                         realServiceDuringPlayback = A.Fake<ILibraryService>();
 
-                        A.CallTo(() => realServiceWhileRecording.GetCount("9780345813923"))
-                            .ReturnsNextFromSequence(11, 10);
-                        A.CallTo(() => realServiceWhileRecording.GetCount("9781593078225"))
-                            .Returns(3);
+                        A.CallTo(() => realServiceWhileRecording.GetCount("1"))
+                            .ReturnsNextFromSequence(0x1A, 0x1B);
```
adamralph: why the hex's?

blairconrad: So I can have 1.a and 1.b!

blairconrad: I can change to any other scheme, if you like.

thomaslevesque:
> I may have gone overboard breaking up the first 3 commits, but figured I'd err on the side of small.

I'm glad you did! It makes it much easier to review the changes.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
```diff
@@ -91,48 +98,56 @@ public static class SelfInitializedFakesSpecs
...
+                                .Wrapping(realServiceWhileRecording).RecordedBy(recorder));
+                            countWhileRecording = fakeService.GetCount("8");
                         }
-                    })
+                    }
+                    catch
+                    {
+                        File.Delete(fileRecorderPath);
+                    }
+                })
```
thomaslevesque: Just to make sure I got this right:
- you split this step into 2 different steps
- so you no longer delete the file in a finally, because it still needs to be there for the next step
- it will still be deleted at the end of the scenario because of the `Teardown`
- you delete it in case something throws

Questions:
- Is the `catch` really necessary? We don't expect an exception to be thrown, so if it does, we want the step to fail, and the `catch` will prevent that. Shouldn't it at least rethrow the exception?
- Deleting in the `catch` is redundant, since the `Teardown` will take care of it

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
```diff
@@ -2,16 +2,26 @@ namespace FakeItEasy.Specs
...
     public interface ILibraryService
     {
         int GetCount(string internationalStandardBookNumber);
+
+        string GetTitle(string internationalStandardBookNumber);
+
+        [SuppressMessage("Microsoft.Design", "CA1021:AvoidOutParameters", MessageId = "0#", Justification = "Required for test scenario")]
+        [SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "1#", Justification = "Required for test scenario")]
+        [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "ref", Justification = "There are no other consumers, so the name is safe.")]
+        [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "out", Justification = "There are no other consumers, so the name is safe.")]
```
thomaslevesque: I think we really need to disable these warnings in test projects... They don't make sense in this context, and the suppressions just clutter the code.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
```diff
@@ -67,28 +209,181 @@ public static class SelfInitializedFakesSpecs
...
+            "And an exception"
+                .x(() => originalException = new InvalidOperationException());
+
+            "And a real service to wrap while recording"
+                .x(() => realServiceWhileRecording = A.Fake<ILibraryService>());
+
+            "And the real service throws the exception when getting the title for book 1"
+                .x(() =>
+                {
+                    realServiceWhileRecording = A.Fake<ILibraryService>();
```
thomaslevesque: Already initialized in the previous step

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
```diff
@@ -67,28 +209,181 @@ public static class SelfInitializedFakesSpecs
...
+            ILibraryService realServiceDuringPlayback,
+            Exception originalException,
+            Exception recordingException,
+            Exception playbackException)
+        {
+            "Given a call storage object"
+                .x(() => inMemoryStorage = new InMemoryStorage());
+
+            "And an exception"
+                .x(() => originalException = new InvalidOperationException());
```
thomaslevesque: This step seems a little out of place; the exception isn't really part of the preconditions. The variable could be initialized when the fake is configured, e.g. `.Throws(originalException = new InvalidOperationException())`

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
```diff
@@ -67,28 +209,181 @@ public static class SelfInitializedFakesSpecs
...
+
+                        playbackException = Record.Exception(() => fakeService.GetTitle("1"));
+                    }
+                });
+
+            "Then the recording fake throws the original exception"
+                .x(() => recordingException.Should().BeSameAs(originalException));
+
+            "But the playback fake throws a recording exception"
+                .x(() => playbackException.Should().BeAnExceptionOfType<RecordingException>());
```
thomaslevesque: The wording is a little confusing. `recordingException` is the exception thrown while recording, but `playbackException` is a `RecordingException`... Renaming the exceptions to `exceptionWhileRecording` and `exceptionWhilePlayingBack` (or anything else) would probably help.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
```diff
@@ -142,12 +436,13 @@ public static class SelfInitializedFakesSpecs
 
                         countDuringPlayback = playbackFakeService.GetCount("8");
                     }
-                });
+                })
+                .Teardown(() => File.Delete(fileRecorderPath));
 
-            "Then the recording fake returns the wrapped service's result"
+            "Then it should return the expected result while recording"
```
thomaslevesque: Didn't we decide against using "should" in BDD language?

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
blairconrad: That's a good idea. Here, or elsewhere? I (or anyone) can submit a PR before or after this.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
blairconrad: I see your point. Yes!

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
blairconrad: :blush:

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
blairconrad: Your summary sounds perfect.

Good point about the `catch`. Assuming it stays, it should rethrow.
I think the point of deleting in the `catch` is that the `.Teardown` isn't hooked up unless the step passes.
@adamralph may know better. If it _does_ get hooked up regardless of the step passing, the try/catch can go away.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
blairconrad: Yes. But I keep forgetting.

blairconrad: Hmm. Actually, on a reread, I seem to have lost the "improvements" I was making here. Rewording.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
blairconrad: I'm back. Turns out the Teardown [should execute](https://github.com/xbehave/xbehave.net/blob/ed11fcdda746edb9a111ac1b109d4d37f13a7c2a/tests/Xbehave.Test.Acceptance.Net40/TeardownFeature.cs#L146-L161) anyhow!

blairconrad:
Thanks for the comments, @thomaslevesque. I really enjoyed them.
I _think_ I addressed everything but moving the suppressions to cover all the tests.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
thomaslevesque: I'd say it deserves its own PR

thomaslevesque:
> I think I addressed everything but moving the suppressions to cover all the tests.

LGTM :+1:

I'll wait a bit before merging in case @adamralph wants to have a look.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
blairconrad: Excellent. I can spin one up after the merge.

blairconrad:
Thanks, @thomaslevesque. Lemme squash then.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
adamralph: Ah, I see. No, it's fine as is. The connection between the input and output just didn't leap out at me.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
adamralph: Agreed.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
adamralph: Do we want to remove "should" before merging? Given our policy against using it, this change (and the one below) is a slight regression.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
blairconrad: Oh, good. But also too bad, since that was my whole point! Ah well.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
adamralph: Yep, teardowns are guaranteed to execute regardless of success/failure.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
blairconrad: I totally do! And was sure I had. Grumble grumble.

blairconrad: I had! But when I rebased, the subsequent commit reverted them, which is an odd thing for it to do.
But I totally should've checked.

blairconrad:
Thanks for catching the "should" issue, @adamralph. And @thomaslevesque the first time through.
I think we're good to go.

tests/FakeItEasy.Specs/SelfInitializedFakesSpecs.cs
```diff
@@ -21,67 +31,353 @@ public static class SelfInitializedFakesSpecs
...
+
+                        playbackFakeService.GetCount("1");
+                        playbackFakeService.GetTitle("1");
+                        exception = Record.Exception(() => playbackFakeService.GetCount("1"));
+                    }
+                });
+
+            // This result demonstrates that the self-initialized fake relies on a script
+            // defined by which methods are called, and is completely inflexible with
+            // regard to the number of repetitions of the calls.
```
blairconrad: I amended this line in the last commit as well. (When fixing the "should"s.)

adamralph:
Thanks @blairconrad !

blairconrad:
Thank yous!

PR #845, opened 2016-09-01
Fix analyzer false positive when call spec is returned

### LINKED ISSUE #844: Analyzer gives false positive FakeItEasy0001 when call spec is returned from a method
For instance this code causes the issue:

```c#
private IVoidConfiguration Test()
{
    var fake = A.Fake<IDisposable>();
    return A.CallTo(() => fake.Dispose());
}
```

> error FakeItEasy0001: Unused call specification 'A.CallTo(() => fake.Dispose())'; did you forget to configure or assert the call?

thomaslevesque:
Fixes #844

### DIFF
--- a/src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
+++ b/src/FakeItEasy.Analyzer/UnusedReturnValueAnalyzer.cs
@@ -48,7 +48,7 @@ namespace FakeItEasy.Analyzer
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


blairconrad:
@thomaslevesque, looks great to me. Nice job.

Leaving for just a little bit in case @adamralph has a comment, but I think this can go in.

adamralph:
Thanks @thomaslevesque !

thomaslevesque:
Thanks for the merge!

PR #847, opened 2016-09-05
Ensure analyzer test code snippets compile

### LINKED ISSUE #843: Make analyzer tests fail when target code does not compile
Currently, one can write a test and if the code doesn't compile, the test passes. When I modify the code tested inside `UnusedReturnValueAnalyzerTests.Diagnostic_Should_Not_Be_Triggered_When_Call_Is_Configured_With_Returns` like so:

```c#
using FakeItEasy;
namespace TheNamespace
{
    class TheClass
    {
        void Test()
        {
            var foo = A.Fake<IFoo>();
            A.Call To(() => foo.Bar());
        }
    }

    interface IFoo { int Bar(); }
}
```

(I removed the `Returns`, but accidentally added a space in `A.CallTo`. It's a little contrived, but the sort of problem a copy/paste might introduce)
the test passes.

In some cases, this won't materially affect the quality of the test, but in some it will. 

Instead, we could have the test fail. With output like this:

```
Build failed. Diagnostics are:
  Test0.cs(10,22): error CS1528: Expected ; or = (cannot specify constructor arguments in declaration)
  Test0.cs(10,22): error CS1003: Syntax error, '[' expected
  Test0.cs(10,39): error CS1003: Syntax error, ']' expected
  Test0.cs(10,15): error CS0426: The type name 'Call' does not exist in the type 'A'
  Test0.cs(10,20): warning CS0168: The variable 'To' is declared but never used
```

As you might guess. I have a spike, if we're interested.

blairconrad:
Fixes #843.

The test code documents we feed into the analyzer as part of the unit
tests should be as close to real code as is practical. In particular,
the code should use FakeItEasy properly. Ensureing that the code compiles
is a good first step.

### DIFF
--- a/tests/FakeItEasy.Analyzer.Tests/Helpers/DiagnosticVerifier.Helper.cs
+++ b/tests/FakeItEasy.Analyzer.Tests/Helpers/DiagnosticVerifier.Helper.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.Linq;
+    using System.Text;
     using Microsoft.CodeAnalysis;
     using Microsoft.CodeAnalysis.CSharp;
     using Microsoft.CodeAnalysis.Diagnostics;
@@ -53,6 +54,9 @@
             foreach (var project in projects)
             {
                 var compilationWithAnalyzers = project.GetCompilationAsync().Result.WithAnalyzers(ImmutableArray.Create(analyzer));
+
+                AssertThatCompilationSucceeded(compilationWithAnalyzers);
+
                 var diags = compilationWithAnalyzers.GetAnalyzerDiagnosticsAsync().Result;
                 foreach (var diag in diags)
                 {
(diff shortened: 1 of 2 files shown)


blairconrad:
I added a check to make sure the all compilations succeeded. And then when they didn't, fixed 'em.

Here's a sample error from the command line build that I made by inserting a space in a `using` statement:

```
Starting:    FakeItEasy.Analyzer.Tests
   FakeItEasy.Analyzer.Tests.UnusedReturnValueAnalyzerTests.Diagnostic_Should_Not_Be_Triggered_When_Call_Is_Configure
d_With_Returns [FAIL]
      System.ArgumentException : Test code compilation failed. Error(s) encountered:
        Test0.cs(2,18): error CS1514: { expected
        Test0.cs(3,1): error CS1022: Type or namespace definition, or end-of-file expected
      Stack Trace:
         tests\FakeItEasy.Analyzer.Tests\Helpers\DiagnosticVerifier.Helper.cs(101,0): at FakeItEasy.Analyzer.Tests.He
lpers.DiagnosticVerifier.AssertThatCompilationSucceeded(CompilationWithAnalyzers compilationWithAnalyzers)
         tests\FakeItEasy.Analyzer.Tests\Helpers\DiagnosticVerifier.Helper.cs(58,0): at FakeItEasy.Analyzer.Tests.Hel
pers.DiagnosticVerifier.GetSortedDiagnosticsFromDocuments(DiagnosticAnalyzer analyzer, Document[] documents)
         tests\FakeItEasy.Analyzer.Tests\Helpers\DiagnosticVerifier.cs(26,0): at FakeItEasy.Analyzer.Tests.Helpers.Di
agnosticVerifier.VerifyCSharpDiagnostic(String source, DiagnosticResult[] expected)
Finished:    FakeItEasy.Analyzer.Tests
```

thomaslevesque:
Thanks @blairconrad, the change looks good to me. I just have one concern: if the default level for a diagnostic is `Error`, doesn't it cause the build to fail when we just want to check that the diagnostic was reported? Or does `compilationWithAnalyzers.Compilation.GetDiagnostics()` exclude the diagnostics produced by the analyzer?

blairconrad:
Welcome back, @thomaslevesque.

My understanding is that `Compilation.GetDiagnostics` does not include the diagnostics produced by the analyzer. That's what `CompilationWithAnalyzers` is for (which contains a `Compilation`).
Certainly in all our examples, where the analyzer would produce a diagnostic, `Compilation.GetDiagnostics` returns nothing.

PR #851, opened 2016-09-10
Optimize away some calls to DynamicMethod.CreateDelegate() 

### LINKED ISSUE #850: Optimize away Expression.Compile() for simple FakeItEasy tests?
I did some profiling of a unit test run with FakeItEasy 2.0. I noticed some 10% of CPU samples were in System.Linq.Expressions.Compiler.LambdaCompiler.Compile(). About 60% of those had FakeItEasy.ExpressionExtensions.Evaluate on the stack. Half of those were from AssertThatMemberCanBeIntercepted, the remainder from GetFakeManagerCallsMadeOn.... but that no longer exists in latest FakeItEasy code.

TimLovellSmith:
...from by 'AssertThatMemberCanBeIntercepted' during trivial A.CallTo() specifications, like

```
var fake = A.Fake<X>();
A.CallTo(() => fake.DoStuff();
```

Fixes #850

### DIFF
--- a/src/FakeItEasy/ExpressionExtensions.cs
+++ b/src/FakeItEasy/ExpressionExtensions.cs
@@ -1,6 +1,8 @@
 namespace FakeItEasy
 {
+    using System.Linq;
     using System.Linq.Expressions;
+    using System.Reflection;
 
     /// <summary>
     /// Provides extension methods for <see cref="Expression"/>.
@@ -8,12 +10,98 @@ namespace FakeItEasy
     internal static class ExpressionExtensions
     {
         /// <summary>
-        /// Evaluates an expression by compiling it into a delegate and invoking the delegate.
+        /// Evaluates an expression, potentially by compiling it into a delegate and invoking the delegate, but faster if possible.
         /// </summary>
+        /// <notes>
+        /// This method evaluates an expression, but tries to do it in a light-weight way that doesn't compile it into a delegate.
+        /// This is often going to be used in the context of finding out 'what object/value is a some (sub-)expression referring to?'
+        /// 
+        /// Key observation is that walking the expression tree and evaluating directly can be much faster for simple expressions like:
+        /// 
+        /// - constant expressions (null, 1, "easy", etc.)
+        /// 
+        /// - local variables referenced from lambda expressions, such as 'fake' in the A.CallTo line:
+        /// 
+        ///     {
+        ///         var fake = A.Fake&lt;Something&gt;();
+        ///         A.CallTo(() => fake.DoStuff());
+        ///     }
+        ///     
+        /// - Expressions that are simple member accesses (field gets, property gets) on an object or class, such as
+        ///     String.Empty
+        ///     A&lt;SomethingA&gt;.Ignored
+        ///     A&lt;SomethingA&gt;._
+        ///     myObj.someProperty
+        ///     
+        /// - Trivial method calls with no arguments (A.ToString(), B.GetType(), Factory.Create())
+        ///     
+        /// </notes>
         /// <param name="expression">The expression to be evaluated.</param>
         /// <returns>The value returned from the delegate compiled from the expression.</returns>
         public static object Evaluate(this Expression expression)
         {
+            return EvaluateOptimized(expression);
+        }
+        
+        private static object EvaluateOptimized(this Expression expression)
+        {
+            if (expression == null)
+            {
+                return null;
+            }
+
+            switch (expression.NodeType)
+            {
+                case ExpressionType.Constant:
+                    return ((ConstantExpression)expression).Value;
+
+                case ExpressionType.MemberAccess:
+                    var memberExpression = (MemberExpression)expression;
+
+                    var fieldInfo = memberExpression.Member as FieldInfo;
+                    if (fieldInfo != null)
+                    {
+                        return fieldInfo.GetValue(EvaluateOptimized(memberExpression.Expression));
+                    }
+
+                    var propertyInfo = memberExpression.Member as PropertyInfo;
+                    if (propertyInfo != null)
+                    {
+                        // index = null: this is always fine since it's a MemberAccess expression, not an IndexExpression
+                        return propertyInfo.GetValue(EvaluateOptimized(memberExpression.Expression), null);
+                    }
+
+                    break;
+
+                case ExpressionType.Call:
+                    var callExpression = (MethodCallExpression)expression;
+
+                    // for now, handling only very trivial call expressions with no arguments, like 'GetBlarg()'
+                    // because anything else might get complicated quickly (method overloads etc.)
+                    if (!callExpression.Arguments.Any())
+                    {
+                        var targetObject = EvaluateOptimized(callExpression.Object);
+                        return callExpression.Method.Invoke(targetObject, null);
+                    }
+
+                    break;
+
+                case ExpressionType.Convert:
+                    var unaryExpression = (UnaryExpression)expression;
+
+                    // for now, handling only 'boxing/casting to object' expressions like '3' used as an argument to a function which takes an object[] array parameter.
+                    if (unaryExpression.Type == typeof(object))
+                    {
+                        // in principle, we would first evaluate it without the boxing, and then box/cast to object... 
+                        // ...but EvaluateOptimized already boxes before returning, so no explicit cast needed.
+                        object obj = EvaluateOptimized(unaryExpression.Operand); // declaring an explicitly typed variable to ensure we're really boxing
+                        return obj;
+                    }
+
+                    break;
+            }
+
+            // when we didn't know how to evaluate the tree, fall back to compiling into a delegate and invoking the expression
             var lambda = Expression.Lambda(expression).Compile();
             return lambda.DynamicInvoke();
         }


src/FakeItEasy/Expressions/ParsedCallExpression.cs
```diff
@@ -34,5 +34,28 @@ internal class ParsedCallExpression
         public IEnumerable<ParsedArgumentExpression> ArgumentsExpressions { get; }
 
         public object CallTarget => this.callTarget.Value;
+
+        // Expression evaluation optimized for a very common case: 
+        // - it's a 'MemberAccess' expression on a local lambda closure (A.CallTo(() => fakeObj.doSomething(args));
```
blairconrad: I enjoyed this comment. It helps me understand things now and I'm sure will help me in the future. 
I'm not quite there, though.
I _think_ the second case tells us that we're going after some member on the fake and then the first case identifies the actual member. Is that right?

blairconrad:
I'm not super familiar with the concepts involved here, but things look promising. I'll make a few minor comments on the commits in a second and I'll take a closer look tomorrow morning.

src/FakeItEasy/Expressions/ParsedCallExpression.cs
```diff
@@ -34,5 +34,28 @@ internal class ParsedCallExpression
...
+                    var mi = me.Member;
+                    var fi = mi as FieldInfo;
+                    if (fi != null)
+                    {
+                        return fi.GetValue(FastEvaluate(me.Expression));
+                    }
+                    break;
+            }
+            return expression.Evaluate();
+        }
```
blairconrad: While I've got you here. I see you've introduced [a few style warnings](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/87#L503), and on a more "humany" note, we tend to shy away from the short (two-letter) variable names. If you're interested in editing these (especially if you're back here making other changes), I'd consider it a favour.
Otherwise we can fix them in a compensating or superseding commit.

src/FakeItEasy/Expressions/ParsedCallExpression.cs
blairconrad: If I'm not right (or even if I am), could this already helpful comment be amended to correct or clarify?

blairconrad:
I did take a slightly closer look, and it looks quite good. I'm learning things!

blairconrad:
@TimLovellSmith, I looked again, and I'm happy with the implementation. Leaving the PR open for a bit in case someone else wants to comment.

blairconrad:
@TimLovellSmith I looked again again, and realized that I _hadn't_ understood what was going on, quite, so I submitted #852 with updated comments (and fixed the style warnings while I was there). I think I've captured what's going on.

src/FakeItEasy/Expressions/ParsedCallExpression.cs
thomaslevesque: @blairconrad I'm not sure I understand your comments. From what I can tell, the goal is to get the value of `fakeObj` from `A.CallTo(() => fakeObj.doSomething(args)`

thomaslevesque: But I agree that the comment could be more explicit

blairconrad: @thomaslevesque, I did not know what I was talking about. I think I figured it out and sent #852 with (I think) more explicit and still correct comments…

thomaslevesque: :+1:

TimLovellSmith:
@blairconrad Sorry for my silence over the weekend. I'd be happy to clean up the style warnings and resubmit, or just take your PR :).

blairconrad:
@TimLovellSmith no need to apologize for the silence. You're not on the clock, as it were. And I'm perfectly capable of carrying on a conversation by myself, so I was in no way inconvenienced!

I see you're going to update for PropertyExpressions. Feel free to nick any of the changes I put in #852, if they help.

blairconrad:
Also, if you have numbers to indicate the level of speedup or whatever, it'd be great to be able to cite them when we release!

blairconrad:
@thomaslevesque, @adamralph, I'm happy with the new version.

Thanks, @TimLovellSmith.

src/FakeItEasy/Expressions/ParsedCallExpression.cs
```diff
@@ -34,5 +34,88 @@ internal class ParsedCallExpression
...
+        //    IFace PropFake {get; set; } = new A.Fake<IFace>();
+        //    IFace fieldFake = new A.Fake<IFace>();
+        // }
+        //
+        // void Test() 
+        // {
+        //     var var = new Bar();
+        //     A.CallTo(() => bar.fieldFake.DoSomething()).InvokesLazily(...orWhatever...);
+        //     A.CallTo(() => bar.propFake.DoSomething()).InvokesLazily(...orWhatever...);
+        //     A.CallTo(() => Bar.staticFake.DoSomething()).InvokesLazily(...orWhatever...);
```
thomaslevesque: Nitpick: there's no `InvokesLazily` method in FakeItEasy :wink:

thomaslevesque:
Looks good to me too, apart from the nitpick about `InvokesLazily`. I really like the detailed explanation :+1:. Thanks @TimLovellSmith!

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -72,18 +72,14 @@ private static IArgumentConstraint CreateEqualityConstraint(object expressionVal
...
-
         private IArgumentConstraint GetArgumentConstraintFromExpression(Expression expression)
         {
             object expressionValue = null;
 
             var trappedConstraints = this.argumentConstraintTrapper.TrapConstraints(() =>
             {
-                expressionValue = InvokeExpression(expression);
+                // in this ParsedArgumentExpression scenario, we get benefits from the same optimizations as ParsedCallExpression
+                expressionValue = ParsedCallExpression.FastEvaluate(expression);
```
thomaslevesque: Here be dragons... for constraint trapping to work, the expression needs to actually be evaluated. All tests still pass, but I'm not 100% sure the optimization won't break anything here.

TimLovellSmith: I convinced myself it would work, here is the argument I made.

The input to InvokeExpression is an expression tree. The requirement for correctness is that FastEvaluate evaluates the expression tree equivalently to Lambda(...).Compile().DynamicEvoke() with respect to a) results and b) side effects.

But in short: it doesn't really matter how the expression tree is evaluated as long as it is consistent with normal C# semantics. Claim: FastEvaluate is consistent with normal C# semantics.

This is easy to demonstrate because there we are definitely not altering *order of evaluation* of property getter, field getter, and constant terms because FastEvaluate never actually makes any choice about the order to evaluate in: it is a straight-line path, and so ultimately, Lambda(...).Compile() has to compile down to an equivalent sequence of get operations, which have the same side effects because they call the same field/property getters in the same order.

TimLovellSmith: Another way to look at it: supposedly this optimizing away some dynamic invoke() thing, but what does that mean! It just means we are optimizing away bits of the compiler/scripting for evaluating, not optimizing away the script that is to be evaluated - it's still all evaluated.

thomaslevesque: Ah, I think I see what you mean. It's just that it's still not completely clear in my mind what kind of expression is going to be fed to `FastEvaluate`. I'm going to re-read your detailed explanation in the comments :wink:

By the way, `FastEvaluate` should probably be moved out of `ParsedCallExpression` if you're going to use it from other places.

TimLovellSmith: @thomaslevesque

I think in general it's such a correct evaluator that it could be used anywhere correctly, but the performance effect of optimization is potentially sensitive to the scenarios. Having it hidden away and hard to find doesn't seem a bad place to start!

I see that mainly FastEvaluate is fed various expressions that are expressions inside the lambda expression parameters of A.CallTo(). What your actual expressions look like here and therefore the arguments flowing to fastEvaluate will depend on your system under test, and test framework, however, it seems often that you are configuring a fake which is stored in a local variable, or as a property/field of some class. I also notice in A.CallTo() argument specifications it's common to do little 'ignored' argument expressions like 'A<string>._, A<blarg>.Ignored' which are also very simple expressions that this optimizes the evaluation of.

I found one or two other places in the code that expressions are evaluted using compiled lambdas - but it seemed I was hardly even hitting those code paths, and so I didn't seriously consider them yet. 

Such other spots in FakeItEasy's code:
-GetConstructorArgumentsFromExpression (another possibly good candidate if you like doing lots of constructor arguments to your fakes)
-ParseInvocationExpression (perhaps useful if you like to evaluate other lambdas as part of your A.CallTo() statements?)
-ParsedArgumentExpression.get_Value (property getter, called as part of GetArgumentConstraint for 'out' and 'ref' arguments (sometimes redundantly with getting the value as part of getting the value as part for GetArgumentConstraintFromExpression))
-and some of the FakeItEasy test code :-)

thomaslevesque: > Having it hidden away and hard to find doesn't seem a bad place to start!

It should be hidden from the public API (which it is), but not hidden from internal code... It should be in a place where we can find it when we need it, and `ParsedCallExpression` is not where I would look for it. I think the best place for `FastEvaluate` would be as an extension method in `ExpressionExtensions`.

> Such other spots in FakeItEasy's code:

Yes, there are definitely more places where this could be useful

TimLovellSmith: @thomaslevesque See if you like this version better! :)

blairconrad:
From https://github.com/FakeItEasy/FakeItEasy/pull/851#discussion_r78456838:

> Having it hidden away and hard to find doesn't seem a bad place to start!

That was my thought as well, not that I'm necessarily against putting this in a common location.
My concerns were not about correctness, but about
1. diminishing returns—at what point does the custom evaluator perform less well than the built-in one (eventually it must, or the smart people at Microsoft would be doing the same thing it does), and
2. the weight of the complexity vs. the benefit—trading a little extra code for performance is okay, but if the thing gets too hard to understand, maybe it's not worth it

All that being said, the code is not yet very complicated, so I'm for what we have.

I wonder whether the beginning of the very detailed comment (which I love) still 100% applies, since it's very focused on `ParsedCallExpression`, and we're already branching out to other uses. It was the motivating example, and probably our most complicated common use case, but it almost seems like this is still an evaluator _for that use case only_. 
Or maybe it's just me. If everyone else thinks it's fine, I'm happy.

Now, complicate matters I was looking at the single remaining other use of `ExpressionExtensions.Evaluate` that we have in production code, in `ParsedArgumentExpression.Value`.
In the FakeItEasy specs, most of the time we hit this, it's a MemberAccess (probably one of those members of a closure).
That's probably a very common use case. Shall we just bite the bullet and use the optimized evaluator everywhere?

thomaslevesque:
> at what point does the custom evaluator perform less well than the built-in one (eventually it must, or the smart people at Microsoft would be doing the same thing it does)

Not necessarily. The "smart people at Microsoft" didn't write an `Evaluate` method, they just wrote a `Compile` method that we use. They don't have the same goal.

> I wonder whether the beginning of the very detailed comment (which I love) still 100% applies, since it's very focused on ParsedCallExpression, and we're already branching out to other uses. It was the motivating example, and probably our most complicated common use case, but it almost seems like this is still an evaluator for that use case only.

Agreed, the comment should probably be updated

> Shall we just bite the bullet and use the optimized evaluator everywhere?

If we're certain it doesn't alter the current behavior (and @TimLovellSmith's explanations convinced me that it doesn't), I'd say yes.

blairconrad:
> They don't have the same goal.

busted

TimLovellSmith:
@blairconrad @thomaslevesque While we're considering rewriting the comments, any nominations for better names for this func than 'EvaluateOptimizedForParsedScenarios'? :)

blairconrad:
@TimLovellSmith, `Evaluate`.

…if we take the suggestion to make it _the way_ to evaluate expressions. Then the old one just becomes a private helper.

thomaslevesque:
> @TimLovellSmith, Evaluate.
> 
> …if we take the suggestion to make it the way to evaluate expressions. Then the old one just becomes a private helper.

Exactly what i was going to suggest

src/FakeItEasy/ExpressionExtensions.cs
```diff
@@ -1,21 +1,109 @@
...
+                    }
+
+                    break;
+
+                case ExpressionType.Call:
+                    var callExpression = (MethodCallExpression)expression;
+
+                    // for now, handling only very trivial call expressions with no arguments, like 'GetBlarg()'
+                    // because anything else might get complicated quickly (method overloads etc.)
+                    if (callExpression.Arguments == null)
```
blairconrad: @TimLovellSmith, have you ever observed this being so?
ReSharper tells me that `Arguments` is not nullable, and when I run FakeItEasy's tests, I hit this line where `Arguments` is empty, but never `null`.

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -72,18 +72,13 @@ private static IArgumentConstraint CreateEqualityConstraint(object expressionVal
             return new EqualityArgumentConstraint(expressionValue);
         }
 
-        private static object InvokeExpression(Expression expression)
-        {
-            return Expression.Lambda(expression).Compile().DynamicInvoke();
-        }
-
```
blairconrad: :+1:

src/FakeItEasy/ExpressionExtensions.cs
TimLovellSmith: Interesting, I wonder how it knows? I wasn't sure from the type (ReadOnlyCollection<Argument>) that it would not return null instead of an empty collection.

thomaslevesque: > I wonder how it knows?

[External annotations](https://www.jetbrains.com/help/resharper/2016.2/Code_Analysis__External_Annotations.html)

TimLovellSmith: @thomaslevesque Ah, interesting! Will make the change.

TimLovellSmith:
BTW, I reran my tests and they still pass :)

blairconrad:
@TimLovellSmith, thanks for the additional change. Looks better. Have you also confirmed that the branch is taken now? I believe it will be, but haven't tried yet. 
I would, but am having family visit starting tomorrow and will be reducing my duties for a week and a bit. ;-)

One thing I'm curious about, as the specializations keep creeping in, where do we stand on an indication of improved execution?
Have you seen a reproducible, quantifiable drop in... anything? CPU time, memory usage,  # of allocations?

TimLovellSmith:
Yes I checked the branch is taken in some of my tests. I saw little performance delta from it because in my test suite it's not common to have method invocations that return a fake object as part of A.CallTo() - I nearly always want to capture the fake as a local variable instead.

I am tempted to construct some artifical benchmark scenarios to further explore the advantages of the specializations... but I expect little downside in adding these extra specializations because there was already the cost of a switch() statement, that's not changing, and when it does go down one of these paths I am pretty sure the optimization will be lighter on allocations.

Whether the optimization helps with your tests should be answerable by answering the question of e.g. does method GetSomeFake in A.CallTo(() => GetSomeFake(someArgs).DoSomething()) actually have some arguments or not.

TimLovellSmith:
So I wrote a little benchmarker:

```
static void Main(string[] args)
        {
            var sw = new Stopwatch();
            sw.Start();
            int i = 0;
            int lastElapsed = 0;
            while (true)
            {
                i++;
                theFake = A.Fake<IFace>();
                A.CallTo(() => GetTheFake().DoSomething()).Returns(1);
                var elapsedSoFar = sw.ElapsedMilliseconds;
                if (elapsedSoFar > lastElapsed)
                {
                    Console.WriteLine(i / (double)elapsedSoFar + "iterations per ms");
                }
            }
        }
```

And did performance comparison with FakeItEasy 2.2 as baseline.

Results: 
FakeItEasy 2.2 throughput is 2.2 iterations per second.
MasterWithOptimizations throughput is 4.3 iterations per second.
MasterWithOptimizations seems to use approx 2-3 less KB of memory (approx steady state with GCs...)

thomaslevesque:
If I read your benchmark correctly, it will never run for more than 1ms, which makes it sensible to large variations due to JIT overhead. Also, the fake is created inside the loop, so what you're measuring includes the fake creation.

thomaslevesque:
OK, I ran my own benchmark, using a benchmark I've been using for a while.

```
static void Main()
{
    var fake = A.Fake<IFoo>();
    _fake = fake;

    new TestHarness(iterationCount: 10, sampleCount: 10000, jitWarmup: true, runGCBetweenSamples: true)
    {
        { "Local variable", () => A.CallTo(() => fake.Bar()).Returns(1) },
        { "Method", () => A.CallTo(() => GetTheFake().Bar()).Returns(1) },
    }.Run().Dump();
}

private static IFoo _fake;

static IFoo GetTheFake() => _fake;

public interface IFoo
{
    int Bar();
}
```

(the `TestHarness` class can be found [here](https://gist.github.com/thomaslevesque/0eb162dddfe164fa09589dcd165f2fad))

And here are the results (average time):

- FIE 2.2.0
  - Local variable: 1.4102 ms
  - Method: 1.0482 ms

- With your optimizations:
  - Local variable: 0.1677 ms (8.4x faster)
  - Method: 0.1742 ms (6x faster)

So it looks like, at least for this scenario, your optimization is a big win :+1:

(of course, my benchmark could also be completely wrong... benchmarks are notoriously hard to get right)

thomaslevesque:
To make sure the difference wasn't due to something else, I re-ran the benchmark against a fresh build from master, with the same result. Nice work @TimLovellSmith! :clap:

TimLovellSmith:
@thomaslevesque Nope - it runs more than 1ms, but every ms it prints out the avg throughput so far.

thomaslevesque:
> @thomaslevesque Nope - it runs more than 1ms, but every ms it prints out the avg throughput so far.

Ah, yes, I misunderstood your code. But I don't think you posted your real code, because `lastElapsed` is never set in the loop so it's always 0 ;)

TimLovellSmith:
Haha, that was the real code. So lots more printing than needed. :)

blairconrad:
@TimLovellSmith, @thomaslevesque thanks for the benchmarking.
I'm not going to be able to review 'til Monday at least, but I think everything's been addressed.

I'd suggest a rebase onto the new master (probably with partial or complete squash - there's a fair amount of backtracking due to the incremental nature of the review and the enahncements) and we may be good to merge?

thomaslevesque:
> I'd suggest a rebase onto the new master (probably with partial or complete squash - there's a fair amount of backtracking due to the incremental nature of the review and the enahncements) and we may be good to merge?

:+1: for squashing everything

TimLovellSmith:
@blairconrad rebased & squashed

thomaslevesque:
Thanks @TimLovellSmith, great job :+1:

blairconrad:
Indeed, @TimLovellSmith. Great job! Thanks for the work (and rework based on demanding comments!).

PR #852, opened 2016-09-11
Avoid some calls to DynamicMethod.CreateDelegate

### LINKED ISSUE #850: Optimize away Expression.Compile() for simple FakeItEasy tests?

blairconrad:
Fixes #850. Supersedes #851.
Cleans up some style warnings and expands on the comments in the optimized code, mostly because I'd misinterpreted what was going on, and if I can do it once, I'll do it again.

### DIFF
--- a/src/FakeItEasy/Expressions/ParsedCallExpression.cs
+++ b/src/FakeItEasy/Expressions/ParsedCallExpression.cs
@@ -16,7 +16,7 @@ namespace FakeItEasy.Expressions
         {
             this.CalledMethod = calledMethod;
             this.ArgumentsExpressions = argumentsExpressions;
-            this.callTarget = new Lazy<object>(() => callTargetExpression?.Evaluate());
+            this.callTarget = new Lazy<object>(() => FastEvaluate(callTargetExpression));
         }
 
         public ParsedCallExpression(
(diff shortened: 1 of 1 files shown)


src/FakeItEasy/Expressions/ParsedCallExpression.cs
```diff
@@ -34,5 +34,42 @@ internal class ParsedCallExpression
...
+        // locating the call target described by a local lambda closure
+        // such as
+        //   A.CallTo(() => fakeObj.doSomething(args))
+        //
+        // In this example, fakeObj is a field on the anonymous
+        // type that the compiler creates to represent the closure,
+        // so the expression will be a MemberExpression where the Member is
+        // a FieldInfo. That FieldInfo can be used to find the value of
+        // fakeObj, which is represented in the expression tree by the
+        // ConstantExpression fakeObj.
```
TimLovellSmith: Slight inaccuracy: "That FieldInfo can be used to find the value of  
 47   +        // fakeObj, which is represented in the expression tree by the  
 48   +        // ConstantExpression fakeObj.  
"

Slightly inaccurate with what it says about the constant expression: the ConstantExpression's value is equal to some anonymous object: the lambda closure.

TimLovellSmith:
I've been thinking about this and I think it makes sense to update this to optimistically evaluate PropertyExpressions also as that's another common but very simple case. I'll update my PR...

blairconrad:
Closing in favour of improvements to #851.

PR #855, opened 2016-09-13
Call WhenArgumentsMatch on parent configuration

thomaslevesque:
instead of reimplementing it separately

Similar to the [change](https://github.com/FakeItEasy/FakeItEasy/pull/840/files#diff-6d9562889008c1f9bfad045e9078aba0R150) made in #840, but I moved this one to a separate PR since it wasn't related to the original PR.

### DIFF
--- a/src/FakeItEasy/Configuration/RuleBuilder.cs
+++ b/src/FakeItEasy/Configuration/RuleBuilder.cs
@@ -158,9 +158,7 @@ namespace FakeItEasy.Configuration
 
             public IReturnValueConfiguration<TMember> WhenArgumentsMatch(Func<ArgumentCollection, bool> argumentsPredicate)
             {
-                Guard.AgainstNull(argumentsPredicate, nameof(argumentsPredicate));
-
-                this.ParentConfiguration.RuleBeingBuilt.UsePredicateToValidateArguments(argumentsPredicate);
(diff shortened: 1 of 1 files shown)


src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -158,9 +158,7 @@ public IAfterCallSpecifiedConfiguration CallsBaseMethod()
 
             public IReturnValueConfiguration<TMember> WhenArgumentsMatch(Func<ArgumentCollection, bool> argumentsPredicate)
             {
-                Guard.AgainstNull(argumentsPredicate, nameof(argumentsPredicate));
```
blairconrad: I'm doing this from web, not IDE, so may be wrong, but are we losing the Guard because `ParentConfiguration` doesn't check?

thomaslevesque: `ParentConfiguration.WhenArgumentsMatch` already has the `Guard`, so it would be redundant to check it again

blairconrad: So it does. Somehow I'd got myself looking at `AnyCallConfiguration.WhenArgumentsMatch `.
At least I was right about being wrong…

blairconrad:
Thanks, @thomaslevesque!

PR #863, opened 2016-09-16
Implement all interfaces on faked classes

### LINKED ISSUE #477: have fake classes implement all interfaces, including explicitly-implemented members
Raised in  
[Stack Overflow question 'FakeItEasy DbSet / IQueryable<T> - Entity Framework 6'](http://stackoverflow.com/questions/22454174/fakeiteasy-dbset-iqueryablet-entity-framework-6) and more recently by 
lshewell in the Jabbr forum, specifically around `DbSet`'s explicit implementation of `IQueryable.Provider`, but really applies to any explicitly-implemented interface method on a class.

When faking a class that has a member that's an explicit interface implementation, then casting the fake to the interface, DynamicProxy doesn't understand that the member is overrideable.

From the StackOverflow question:

``` c#
var fakeDbSet = A.Fake<DbSet<Request>>();
A.CallTo(() => ((IQueryable<Request>)fakeDbSet).Provider).Returns(data.Provider);
```

fails because DynamicProxy doesn't understand that it's allowed to do things to `Provider`.

Workaroundable by explicitly implementing the interface by providing `options=>options.Implements(typeof(IQueryable<Request>))` to the constructor, but Ishewell reports that Moq doesn't suffer from the same limitation. My guess is that it's because Moq adds all of a faked class's interfaces to the list of interfaces to implement when having DynamicProxy create its proxy.

It seems like a good approach, and I can't think of a downside.

thomaslevesque:
Fixes #477

### DIFF
--- a/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
+++ b/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
@@ -151,7 +151,7 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
             IEnumerable<object> argumentsForConstructor,
             IInterceptor interceptor)
         {
-            var allInterfacesToImplement = GetAllInterfacesToImplement(additionalInterfacesToImplement);
+            var allInterfacesToImplement = GetAllInterfacesToImplement(typeOfProxy, additionalInterfacesToImplement);
 
             if (typeOfProxy.GetTypeInfo().IsInterface)
             {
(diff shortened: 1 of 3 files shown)


tests/FakeItEasy.Tests.ruleset
```diff
@@ -10,6 +10,7 @@
     <Rule Id="CA1020" Action="None" />
     <Rule Id="CA1021" Action="None" />
     <Rule Id="CA1025" Action="None" />
+    <Rule Id="CA1033" Action="None" />
```
thomaslevesque: This is _CA1033: Interface methods should be callable by child types_. FxCop complains because interface members are implemented explicitly without providing a way for derived classes to call them. While it's probably good advice for public APIs, I don't think we'll ever care about that in test projects.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
```diff
@@ -179,9 +179,17 @@ private static ProxyGeneratorResult GetProxyResultForValueType(Type typeOfProxy)
...
-            return additionalInterfacesToImplement.Concat(typeof(ITaggable));
+            return additionalInterfacesToImplement
+                .Union(typeOfProxy.GetInterfaces().Where(CanInterfaceBeImplemented))
+                .Concat(typeof(ITaggable));
+        }
+
+        private static bool CanInterfaceBeImplemented(Type type)
+        {
+            var info = type.GetTypeInfo();
+            return (info.IsPublic || info.IsNestedPublic) && !info.IsImport;
```
thomaslevesque: For netstd compatibility

blairconrad: Which part specifically is for netstd?

thomaslevesque: The use of `TypeInfo` rather than `Type`. I had initially used the exact code from Moq, but it didn't work on netstd.

blairconrad: Oh, thanks. I hadn't seen the intermediate step and didn't so realize.

blairconrad:
Thanks for this, @thomaslevesque!

thomaslevesque:
Thanks for the merge!

PR #865, opened 2016-09-18
Add documentation about the analyzer

### LINKED ISSUE #856: Document the analyzer
Nowhere do we mention that we have one.
We should at least put it in the docs.

thomaslevesque:
Fixes #856

### DIFF
--- a/docs/analyzer.md
+++ b/docs/analyzer.md
@@ -1,0 +1,24 @@
+# Analyzer
+
+FakeItEasy provides a Roslyn analyzer to detect incorrect usages of the library
+that cannot be prevented by the API or the compiler but will typically result in
+bugs at runtime.
+
+## Installation
+
+The analyzer works in Visual Studio 2015 Update 1 or later. It can be installed
+[from NuGet](https://www.nuget.org/packages/FakeItEasy.Analyzer) for each project
+that needs it:
+
+```
+PM> Install-Package FakeItEasy.Analyzer
+```
+
+## Diagnostics
+
+The analyzer currently provides two diagnostics:
+
+| Id             | Summary                          | Description                                                                            |
+|----------------|----------------------------------|----------------------------------------------------------------------------------------|
+| FakeItEasy0001 | Unused call specification        | Triggered when you specify a call but don't configure or assert it, making it a no-op. |
+| FakeItEasy0002 | Non-virtual member configuration | Triggered when you try to configure a non-virtual member, which cannot be faked.       |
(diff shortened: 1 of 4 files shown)


docs/analyzer.md
```diff
@@ -0,0 +1,25 @@
+# Analyzer
+
+FakeItEasy provides a Roslyn analyzer to detect incorrect usages of the library,
+which cannot be prevented by the API or the compiler but will typically result in
+bugs at runtime.
```
blairconrad: Grammar is hard, but I think ", which" should be "that".
As written, it sounds like _all_ incorrect usages of the library cannot be prevented by the API or the compiler, but I think the intent is for the analzyer to detect those incorrect usages that cannot be prevented by the API or compiler.

(If I'm not wrong, the second part of the sentence is a [nonrestrictive clause](http://www.quickanddirtytips.com/education/grammar/which-versus-that-0), and we want a restrictive one here.)

Perhaps I've misinterpreted your intent.

docs/analyzer.md
```diff
@@ -0,0 +1,25 @@
+# Analyzer
+
+FakeItEasy provides a Roslyn analyzer to detect incorrect usages of the library,
+which cannot be prevented by the API or the compiler but will typically result in
+bugs at runtime.
+
+
```
blairconrad: Nit: double newline

docs/analyzer.md
blairconrad: ", which" sounds to me like no incorrect usages of the library can be prevented by the API or the compiler.
I think we want a [restrictive clause](http://www.quickanddirtytips.com/education/grammar/which-versus-that-0) formed by "that", indicating the the analyzer is intended to help with the subset of incorrect usages that can't be prevented by the API or compiler.

Unless I've misinterpreted your intent?

blairconrad:
(fun fact: the comment I thought I had to enter twice was actually saved and applied by GitHub. It just lied about having lost it.)

docs/analyzer.md
thomaslevesque: Ah, yes,"that" sounds better. After 25 years of learning and speaking English, I *still* don't know when I should use "that" or "which"...

blairconrad: Nobody does.

blairconrad:
Thanks, @thomaslevesque. Top-notch work.

thomaslevesque:
Thanks for the merge!

docs/analyzer.md
```diff
@@ -0,0 +1,24 @@
+# Analyzer
+
+FakeItEasy provides a Roslyn analyzer to detect incorrect usages of the library
+that cannot be prevented by the API or the compiler but will typically result in
+bugs at runtime.
```
adamralph: Should we be pre-empting word wrapping, or should we let readthedocs/the user's browser handle it?

blairconrad: Are you worried that this forces linebreaks?
I'm seeing different wrapping on readthedocs.

image

adamralph: Never mind, it seems like it's ignored. Even GitHub ignores it!

adamralph: The purist in me would remove them, but it seems they have no end effect either in the docs or in GitHub's rendering.

PR #869, opened 2016-09-23
Revise comments in ExpressionExtensions to fix code analysis warnings

TimLovellSmith:


src/FakeItEasy/ExpressionExtensions.cs
```diff
@@ -14,35 +14,29 @@ internal static class ExpressionExtensions
...
             return EvaluateOptimized(expression);
         }
-        
+
+        // About the optimizations: 
+        // Key observation is that walking the expression tree and evaluating directly can be much faster for simple expressions like:
+        // - constant expressions (null, 1, "easy", etc.)
+        // - local variables referenced from lambda expressions, such as 'fake' in the A.CallTo line:
+        //     {
+        //         var fake = A.Fake&lt;Something&gt;();
```
thomaslevesque: This is no longer an XML doc comment, so you don't need to encode the `<` and `>`

thomaslevesque:
Looking good, thanks @TimLovellSmith!

> @FakeItEasy/owners, how do we feel about merging this and then the remainder of @thomaslevesque's changes?

:+1: 

There will be a conflict, but it'll be easy to fix

src/FakeItEasy/ExpressionExtensions.cs
blairconrad: Bah. Good eye.

blairconrad:
Thanks for merging, @thomaslevesque. I figured you'd be up to fixing the conflict!

src/FakeItEasy/ExpressionExtensions.cs
thomaslevesque: Would have been more useful if I had waited for the fix before merging... don't know what I was thinking :confused:

blairconrad: I'd not even noticed the escaped chars at the time.
Although when I first saw the PR, I saw this:
image

and thought they'd been unescaped. Silly of me.

thomaslevesque: Ah, you switched to a font with ligatures ? Me too, and I love it :)

PR #871, opened 2016-09-24
[WIP] Add Appveyor YML file for approval tests

### LINKED ISSUE #868: Have AppVeyor run the approve task
From https://github.com/FakeItEasy/FakeItEasy/pull/866#issuecomment-249143589.

We should have an automated test to ensure that we don't accidentally change the API of the .NET Standard version of FakeItEasy.

We could either amend the current project to also run the `approve` task, or add a second project similar to what we'd done for TeamCity.
I think we added a second build configuration to TeamCity so that it could fail and not freak out the users when a mandatory check failed on their PR.
I could go either way.

blairconrad:
Fixes #868


blairconrad:
https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/2

Fired automatically, but does not show up in the checks here. I'm going to push an update, just to see what happens.

blairconrad:
I think I lied. Now only the approval check is showing here. Maybe we can only show one AppVeyor check? They have different webhook URLs. Investigating.

blairconrad:
@FakeItEasy/owners, I'm coming up blank here. 

I see both the [Approval](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/5) and [regular](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy-au76p/build/38) AppVeyor builds firing from my commit, and we even get the commit status updates from AppVeyor, but only one status reported in this PR from AppVeyor (none yet from TeamCity, as it seems to be down/slow/sad).

However, when I check the [commit status updates received for the commit](https://api.github.com/repos/FakeItEasy/FakeItEasy/statuses/60c8bca030a4edbb7e671f32618dfb306f2f8e21), there are 4 events: a "pending" and "success" for each build. We can't directly compare against the TeamCity ones, but looking at [the latest status checks](https://api.github.com/repos/FakeItEasy/FakeItEasy/statuses/2e6b06277a5bf9f179712a0f51ba0b34ec838b34) from from #804, I see:

* only 2 events from AppVeyor, because it was built before I added the new project,
* 8! events from TeamCity. Two from each of the build configurations for each state (pending/success/pending/success per build configuration)

But looking at the events, I am failing to see something within them that would cause the two different checks to be picked up from TeamCity and not from Appveyor.

Any ideas or interest in taking a look? If not, I think the next step is to escalate to GitHub.

Thanks/sorry.

thomaslevesque:
> However, when I check the commit status updates received for the commit, there are 4 events: a "pending" and "success" for each build

Looks like both builds use the same value for `context`: `continuous-integration/appveyor/pr`
That's probably why only one of them shows up in the PR. However, I don't see an option in AppVeyor to change it... I'm pretty sure it's configurable in TeamCity but it's down (again), so I can't check.

thomaslevesque:
I posted a question in [this thread](http://help.appveyor.com/discussions/kb/4-show-multiple-statuses-in-github-pull-requests) in the Appveyor forum, which seems related.

blairconrad:
That's why we pay you the big bucks, @thomaslevesque. I'd sworn I'd looked at the contexts for TC to compare, and thought they were the same. Clearly I did not.

Thanks! I'd much rather expose my incompetence to you instead of the world at large.

thomaslevesque:
https://developer.github.com/v3/repos/statuses/

> The most recent status for each context is returned, up to 100

This seems to confirm my hypothesis

appveyor.approval.yml
```diff
@@ -0,0 +1,18 @@
+version: '{build}'
+
+environment:
+    PATH: 'C:\Ruby200\bin;%PATH%'
+
+install:
+    - cmd: bundle install
+
+build_script:
+    - cmd: build.netstd.cmd approve
```
adamralph: Rather than having 2 appveyor files, could we configure an environment variable in the web UI which is used here? The env var could be omitted for the regular project, and set to "approve" for the approval project.

thomaslevesque: Agreed. But if AppVeyor can't send status updates with different contexts, the point is moot...

adamralph: Perhaps we should bite the bullet and make `approve` part of the regular build. It may cause *a little* friction for contributors, but we can always send a superseding PR if the burden of updating the approved API is too much for them.

thomaslevesque: Yes, I think we'll have to do that anyway. We can update CONTRIBUTING.md to explain the approval process.

thomaslevesque:
FYI: https://github.com/appveyor/ci/issues/1083

adamralph:
Superseded by #884

PR #881, opened 2016-10-01
add octokit to Gemfile

adamralph:
Word of warning: I've no idea if the latest octokit works correctly for us or whether we need a specific version. The github automation part of the rakefile has never worked for me so I can't test it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,5 @@
 source 'http://rubygems.org'
-  gem 'rake'
   gem 'albacore', '0.3.6'
+  gem 'octokit'
+  gem 'rake'
   gem 'zip'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,10 +1,19 @@
 GEM
   remote: http://rubygems.org/
   specs:
+    addressable (2.4.0)
     albacore (0.3.6)
       rubyzip (< 1.0.0)
+    faraday (0.9.2)
+      multipart-post (>= 1.2, < 3)
+    multipart-post (2.0.0)
+    octokit (4.3.0)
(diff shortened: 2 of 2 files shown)


blairconrad:
Thanks for this @adamralph. 
Currently I'm getting this when I try to test your change:

```
Creating pull request...
rake aborted!
Octokit::Unauthorized: GET https://api.github.com/user: 401 - Requires authentication // See: https://developer.github.com/v3
C:/PortableApps/Documents/Source/GitHub/FakeItEasy/rakefile.rb:154:in `block in <top (required)>'
Tasks: TOP => next_version
```

It's peculiar since, as far as I can tell, I am authenticated. Will continue to investigate.

Then once this is resolved, we can address your problems with the automated steps. I don't know if you ever tried it before @thomaslevesque did, so you may be running into the same problem that he (and I) run into now: the fact that http://curl.haxx.se/ca/cacert.pem now redirects to https, making it unusable. My cached pem works great (saved at the location pointed to by `%SSL_CERT_FILE%`) ("great" in that it gets me as far as the octokit error).

The "cache the pem and add an environment variable" workaround is a little annoying though.

adamralph:
@blairconrad I imagine this because the latest version of octokit has changed. If you run `gem list octokit` that will tell you which version you were using previously (the one prior to 3.4.0 in the list). We can then set that version explicitly in the gemfile and commit that, until we figure out how to get things working with the latest octokit.

adamralph:
Superseded by #882

blairconrad:
> The "cache the pem and add an environment variable" workaround is a little annoying though.

So I made #883 to talk about it.

PR #882, opened 2016-10-01
add octokit and netrc to Gemfile

blairconrad:
Supersedes and closes #881.

We needed the netrc gem to authenticate.


blairconrad:
Thanks, @adamralph!

PR #887, opened 2016-10-02
add build suffix to pre-release packages

adamralph:
I think it's important that we increment our package number each time a new package is created in our AppVeyor NuGet feed. This PR appends the build number to it. Otherwise, each time a PR is merged, the -netstd packages are replaced in the NuGet feed, and anyone restoring them will be silently upgraded.

This will also affect pre-release packages of the regular (non -netstd) packages, but appending a build number to these is also common practice, e.g. https://www.nuget.org/packages/xunit/2.2.0-beta2-build3300.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,7 @@ version: '{build}'
 
 environment:
     PATH: 'C:\Ruby200\bin;%PATH%'
+    BUILD: $(APPVEYOR_BUILD_NUMBER)
 
 install:
     - cmd: bundle install
@@ -17,6 +18,6 @@ test: off
 
 artifacts:
     - path: .\artifacts\logs\*.*
-    - path: .\artifacts\output\*-netstd.nupkg
+    - path: .\artifacts\output\*-netstd*.nupkg
     - path: .\artifacts\output
     - path: .\artifacts\tests\*.*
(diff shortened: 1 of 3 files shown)


adamralph:
Also, if someone reports a problem in version `2.4.0-netstd` we have no easy way of knowing which of the many incarnations of `2.4.0-netstd` they are using.

blairconrad:
@adamralph, sounds like a good change. However, I wonder whether we shouldn't use '+' as the separator, [as suggested by semver](http://semver.org/#spec-item-10).

blairconrad:
Oh, scratch that. Your addition isn't quite "build metadata". And it _should_ affect version precedence.

blairconrad:
@adamralph, this is a good idea, and I think the change looks good. And will produce good output once you rebase.

thomaslevesque:
Good to go once the conflicts are fixed

appveyor.yml
```diff
@@ -17,6 +18,6 @@ test: off
 
 artifacts:
     - path: .\artifacts\logs\*.*
-    - path: .\artifacts\output\*-netstd.nupkg
+    - path: .\artifacts\output\*-netstd*.nupkg
```
adamralph: As well as the rebase, I had to amend the commit with this change.

Interestingly, GitHub did not invalidate the approval of the previous diff. :thinking: 

@FakeItEasy/owners I'd like to get your re-approval for this.

adamralph: The build artifacts looks good now: https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/18/artifacts

thomaslevesque: re-approved :+1:

blairconrad: > interestingly, GitHub did not invalidate the approval of the previous diff. :thinking:

That's disturbing.

blairconrad:
Fantastic. Thanks, @adamralph.

PR #891, opened 2016-10-02
add HTML test results

adamralph:


rakefile.netstd.rb
```diff
@@ -362,11 +362,12 @@ def run_tests(test_assemblies, command, result_dir)
 
 def run_netstd_tests(test_directories, result_dir)
```
blairconrad: I'd've thought to update `run_tests` too.

adamralph: Done (amended the commit).

PR #892, opened 2016-10-03
Add default SSL cert file

### LINKED ISSUE #883: weakness in next_version rake task
The `next_version` rake task doesn't work because of the ruby SSL bug.
Previously we'd worked around that by having the rakefile download a PEM file from http://curl.haxx.se/ca/cacert.pem unless the `SSL_CERT_FILE` environment variable was set.
Unfortunately, that URL now redirects to HTTPS, so the approach no longer works.
I've just cached a version of the PEM and permanently set my SSL_CERT_FILE environment variable. Which works fine, but is a manual step that any releaser has to perform.
So, I thought of a few alternatives, roughly sorted in my descending order of preference:

1. change the rakefile to output an error if SSL_CERT_FILE isn't set, telling users what to do. We keep talking about dropping rake. If we do that soon, we can choose another alternative.
2. do nothing for now. This is sort of like point 1, with similar reasoning, but is not quite as friendly to infrequent users of the task
3. use a different approach to creating new PRs and milestones and whatnot. We could rely on one of the other octokit flavours (although I think picking anything other than .NET would be silly), which (hopefully) understand SSL. This may even become part of our new build system.
4. use a different approach to fetch the PEM, like a powershell script or some foolishness
5. drop the task altogether. I am not sure anyone but me is a fan. Although for the year or so where I did every release, I found it to be a godsend, since I think the version-setting, and milestone-, issue- and release-creating are extremely tedious and error-prone.

What would you @FakeItEasy/owners like?

blairconrad:
Fixes #883


rakefile.netstd.rb
```diff
@@ -378,30 +378,26 @@ def restore_dependencies_for_project(project_dir)
...
+# Find an SSL cert file if necessary.
+# If ENV["SSL_CERT_FILE"] is set, takes no action.
+# Otherwise, expects the current directory to
+# contain a "cacert.pem" file that can be used
+# to enable Ruby SSL support, and sets
+# ENV["SSL_CERT_FILE"] to point at it.
+# New cert files can be had from
+#    http://curl.haxx.se/ca/cacert.pem
+# in case the current one stops working.
+def find_ssl_cert_file(default_ssl_cert_filename)
```
adamralph: Can we get rid of the whole env var thing and just rely on the file in source?

blairconrad: Absolutely.

adamralph:
Thanks @blairconrad !

PR #893, opened 2016-10-03
stop PR's incrementing build number

adamralph:
Appends a random string to the end of PR build numbers:

image

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,8 @@
 version: '{build}'
 
+pull_requests:
+  do_not_increment_build_number: true
+
 environment:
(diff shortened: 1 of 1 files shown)


blairconrad:
I don't mind the change, but why the bias against incrementing the build number?

adamralph:
It's not a huge issue. It just struck me when I force pushed the branch for #891 several times in quick succession and 3 corresponding builds appeared. Ultimately PR's are out of our control (anyone can raise them any time) so it's kind of nice to insulate the build counter against them.

blairconrad:
Fair enough.

adamralph:
A nice side effect is that it that it gives an easy distinction between PR's and branch builds:

image

PR #900, opened 2016-11-14
Upgrade to Castle.Core 4.0.0-beta002

### LINKED ISSUE #436: Adding additional attributes to a fake of an already faked type fails silently

thomaslevesque:
Fixes #436 

The only visible API change is `IFakeOptions.WithAdditionalAttributes`, which now takes a `params Expression<Func<Attribute>>[]`.

### DIFF
--- a/docs/creating-fakes.md
+++ b/docs/creating-fakes.md
@@ -27,7 +27,7 @@ When creating fakes you can, through a fluent interface, specify options for how
 
 * Specify arguments for the constructor of the faked type.
 * Specify additional interfaces that the fake should implement.
-* Assign additional custom attributes to the faked class.
+* Assign additional custom attributes to the faked type.
 * Cause a fake to have [strict mocking semantics](strict-fakes.md).
 * Configure all of a fake's methods to [use their original implementation](calling-base-methods.md).
 * Create a fake that wraps another object.
@@ -51,13 +51,9 @@ var foo = A.Fake<FooClass>(x => x.Implements(typeof(IFoo)));
 // or
 var foo = A.Fake<FooClass>(x => x.Implements<IFoo>());
 
-// Assigning custom attributes to the faked class.
-// Get a parameterless constructor for our attribute and create a builder
-var constructor = typeof(FooAttribute).GetConstructor(new Type[0]);
-var builder = new CustomAttributeBuilder(constructor, new object[0]);
-var builders = new List<CustomAttributeBuilder>() { builder };
-// foo and foo's type should both have "FooAttribute"
-var foo = A.Fake<IFoo>(x => x.WithAdditionalAttributes(builders));
+// Assigning custom attributes to the faked type.
+// foo's type should have "FooAttribute"
+var foo = A.Fake<IFoo>(x => x.WithAttributes(() => new FooAttribute()));
 
 // Create wrapper - unconfigured calls will be forwarded to wrapped
 var wrapped = new FooClass("foo", "bar");
--- a/src/FakeItEasy/Core/FakeWrapperConfigurator.cs
+++ b/src/FakeItEasy/Core/FakeWrapperConfigurator.cs
@@ -2,7 +2,7 @@ namespace FakeItEasy.Core
 {
     using System;
     using System.Collections.Generic;
-    using System.Reflection.Emit;
+    using System.Linq.Expressions;
     using FakeItEasy.Creation;
 #if FEATURE_SELF_INITIALIZED_FAKES
     using FakeItEasy.SelfInitializedFakes;
@@ -38,9 +38,9 @@ namespace FakeItEasy.Core
             return this.fakeOptions.Wrapping(wrappedInstance);
         }
 
-        public override IFakeOptions<T> WithAdditionalAttributes(IEnumerable<CustomAttributeBuilder> customAttributeBuilders)
+        public override IFakeOptions<T> WithAdditionalAttributes(params Expression<Func<Attribute>>[] attributes)
         {
-            return this.fakeOptions.WithAdditionalAttributes(customAttributeBuilders);
+            return this.fakeOptions.WithAdditionalAttributes(attributes);
         }
 
         public override IFakeOptions<T> Implements(Type interfaceType)
--- a/src/FakeItEasy/Core/FakeWrapperConfigurator.cs
+++ b/src/FakeItEasy/Core/FakeWrapperConfigurator.cs
@@ -38,9 +38,9 @@ namespace FakeItEasy.Core
             return this.fakeOptions.Wrapping(wrappedInstance);
         }
 
-        public override IFakeOptions<T> WithAdditionalAttributes(params Expression<Func<Attribute>>[] attributes)
+        public override IFakeOptions<T> WithAttributes(params Expression<Func<Attribute>>[] attributes)
         {
-            return this.fakeOptions.WithAdditionalAttributes(attributes);
+            return this.fakeOptions.WithAttributes(attributes);
         }
 
         public override IFakeOptions<T> Implements(Type interfaceType)
--- a/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
+++ b/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
@@ -5,13 +5,13 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
     using System.Diagnostics.CodeAnalysis;
     using System.Globalization;
     using System.Linq;
+    using System.Linq.Expressions;
     using System.Reflection;
-    using System.Reflection.Emit;
     using Castle.DynamicProxy;
     using FakeItEasy.Core;
 
     internal class CastleDynamicProxyGenerator
-        : IProxyGenerator
+        : FakeItEasy.Creation.IProxyGenerator
     {
         private static readonly IProxyGenerationHook ProxyGenerationHook = new InterceptEverythingHook();
         private static readonly ProxyGenerator ProxyGenerator = new ProxyGenerator();
@@ -26,15 +26,15 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
             Type typeOfProxy,
             IEnumerable<Type> additionalInterfacesToImplement,
             IEnumerable<object> argumentsForConstructor,
-            IEnumerable<CustomAttributeBuilder> customAttributeBuilders,
+            IEnumerable<Expression<Func<Attribute>>> attributes,
             IFakeCallProcessorProvider fakeCallProcessorProvider)
         {
-            Guard.AgainstNull(customAttributeBuilders, nameof(customAttributeBuilders));
+            Guard.AgainstNull(attributes, nameof(attributes));
 
             var options = CreateProxyGenerationOptions();
-            foreach (CustomAttributeBuilder builder in customAttributeBuilders)
+            foreach (var attribute in attributes)
             {
-                options.AdditionalAttributes.Add(builder);
+                options.AdditionalAttributes.Add(CustomAttributeInfo.FromExpression(attribute));
             }
 
             return GenerateProxy(typeOfProxy, options, additionalInterfacesToImplement, argumentsForConstructor, fakeCallProcessorProvider);
--- a/src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
+++ b/src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
@@ -83,7 +83,7 @@ namespace FakeItEasy.Creation
                 return this;
             }
 
-            public override IFakeOptions<T> WithAdditionalAttributes(
+            public override IFakeOptions<T> WithAttributes(
                 params Expression<Func<Attribute>>[] attributes)
             {
                 Guard.AgainstNull(attributes, nameof(attributes));
--- a/src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
+++ b/src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
@@ -2,10 +2,10 @@ namespace FakeItEasy.Creation
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq.Expressions;
 #if FEATURE_NETCORE_REFLECTION
     using System.Reflection;
 #endif
-    using System.Reflection.Emit;
     using FakeItEasy.Core;
 
     /// <summary>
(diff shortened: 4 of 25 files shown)


src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
```diff
@@ -5,13 +5,13 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
...
     using System.Linq;
+    using System.Linq.Expressions;
     using System.Reflection;
-    using System.Reflection.Emit;
     using Castle.DynamicProxy;
     using FakeItEasy.Core;
 
     internal class CastleDynamicProxyGenerator
-        : IProxyGenerator
+        : FakeItEasy.Creation.IProxyGenerator
```
thomaslevesque: This is because Castle.Core introduced a new `IProxyGenerator` interface, which conflicts with ours

src/FakeItEasy/Core/FakeWrapperConfigurator.cs
```diff
@@ -38,9 +38,9 @@ public override IFakeOptionsForWrappers<T> Wrapping(T wrappedInstance)
             return this.fakeOptions.Wrapping(wrappedInstance);
         }
 
-        public override IFakeOptions<T> WithAdditionalAttributes(IEnumerable<CustomAttributeBuilder> customAttributeBuilders)
+        public override IFakeOptions<T> WithAdditionalAttributes(params Expression<Func<Attribute>>[] customAttributeExpressions)
```
adamralph: Is `customAttributeExpressions` the best name? It feels a little like naming a `string` parameter `somethingString`. 

Usually I'd name a `Func<T>` parameter like a method, e.g. `createAttribute`, so when I invoke it I can write

```C#
var attribute = createAttribute();
```

but I'm not sure how that naming convention would translate to a collection of such objects. I also noticed that in `IProxyOptions` the property is named `AdditionalAttributes`, so perhaps even just `attributes` will work?

tests/FakeItEasy.IntegrationTests/GeneralTests.cs
```diff
@@ -51,14 +50,9 @@ public void Faked_object_with_method_that_returns_fakeable_type_should_return_fa
...
             // Arrange
-            var types = new Type[0];
-            var constructor = typeof(ForTestAttribute).GetConstructor(types);
-            var list = new object[0];
-            var attribute = new CustomAttributeBuilder(constructor, list);
-            var builder = new List<CustomAttributeBuilder> { attribute };
 
             // Act
-            var fake = A.Fake<IEmpty>(a => a.WithAdditionalAttributes(builder));
+            var fake = A.Fake<IEmpty>(a => a.WithAdditionalAttributes(() => new ForTestAttribute()));
```
adamralph: Wow, so much nicer. 😍

src/FakeItEasy/Core/FakeWrapperConfigurator.cs
thomaslevesque: I'm not particularly happy about the name either... `attributes` would work, I guess, but it's a bit misleading, since we're not actually passing attributes.

tests/FakeItEasy.IntegrationTests/GeneralTests.cs
thomaslevesque: Glad you like it ;)

src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
```diff
@@ -84,11 +84,11 @@ public override IFakeOptions<T> WithArgumentsForConstructor(IEnumerable<object>
...
 
             public override IFakeOptions<T> WithAdditionalAttributes(
-                params Expression<Func<Attribute>>[] customAttributeExpressions)
+                params Expression<Func<Attribute>>[] attributes)
             {
-                Guard.AgainstNull(customAttributeExpressions, nameof(customAttributeExpressions));
+                Guard.AgainstNull(attributes, nameof(attributes));
 
-                foreach (var expression in customAttributeExpressions)
+                foreach (var expression in attributes)
```
adamralph: `expression` → `attribute`

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
```diff
@@ -26,15 +26,15 @@ public CastleDynamicProxyGenerator(CastleDynamicProxyInterceptionValidator inter
...
-            IEnumerable<CustomAttributeBuilder> customAttributeBuilders,
+            IEnumerable<Expression<Func<Attribute>>> attributes,
             IFakeCallProcessorProvider fakeCallProcessorProvider)
         {
-            Guard.AgainstNull(customAttributeBuilders, nameof(customAttributeBuilders));
+            Guard.AgainstNull(attributes, nameof(attributes));
 
             var options = CreateProxyGenerationOptions();
-            foreach (CustomAttributeBuilder builder in customAttributeBuilders)
+            foreach (var expression in attributes)
```
adamralph: `expression` → `attribute`

src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
```diff
@@ -84,13 +84,13 @@ public override IFakeOptions<T> WithArgumentsForConstructor(IEnumerable<object>
...
 
             public override IFakeOptions<T> WithAdditionalAttributes(
-                IEnumerable<CustomAttributeBuilder> customAttributeBuilders)
+                params Expression<Func<Attribute>>[] attributes)
             {
-                Guard.AgainstNull(customAttributeBuilders, nameof(customAttributeBuilders));
+                Guard.AgainstNull(attributes, nameof(attributes));
 
-                foreach (var customAttributeBuilder in customAttributeBuilders)
+                foreach (var expression in attributes)
```
adamralph: `expression` → `attribute`

src/FakeItEasy/Creation/IFakeOptions.cs
```diff
@@ -33,9 +32,9 @@ public interface IFakeOptions : IHideObjectMembers
         /// <summary>
         /// Specifies that the fake should be created with these additional attributes.
         /// </summary>
-        /// <param name="customAttributeBuilders">The attributes to build into the proxy.</param>
+        /// <param name="attributes">Expressions that create attributes to build into the proxy.</param>
```
adamralph: nitpick: pre-existing, but would "build into" be better as "add to"?

src/FakeItEasy/Creation/IFakeOptions.of.T.cs
```diff
@@ -38,9 +37,9 @@ public interface IFakeOptions<T> : IHideObjectMembers
         /// <summary>
         /// Specifies that the fake should be created with these additional attributes.
         /// </summary>
-        /// <param name="customAttributeBuilders">The attributes to build into the proxy.</param>
+        /// <param name="attributes">Expressions that create attributes to build into the proxy.</param>
```
adamralph: nitpick: pre-existing, but would "build into" be better as "add to"?

src/FakeItEasy/Creation/IProxyGenerator.cs
```diff
@@ -19,10 +19,10 @@ internal interface IProxyGenerator
         /// <param name="typeOfProxy">The type of proxy to generate.</param>
         /// <param name="additionalInterfacesToImplement">Interfaces to be implemented by the proxy.</param>
         /// <param name="argumentsForConstructor">Arguments to pass to the constructor of the type in <paramref name="typeOfProxy" />.</param>
-        /// <param name="customAttributeBuilders">The custom attribute builders.</param>
+        /// <param name="attributes">The expressions that describe the custom attributes to build into the proxy.</param>
```
adamralph: nitpick: but would "build into" be better as "add to"?

adamralph: nitpick: "custom" seems redundant (it's not used in equivalent comments elsewhere).

src/FakeItEasy/Creation/ProxyOptions.cs
```diff
@@ -43,9 +43,9 @@ public void AddProxyConfigurationAction(Action<object> action)
             this.proxyConfigurationActions.Add(action);
         }
 
-        public void AddAttribute(CustomAttributeBuilder attribute)
+        public void AddAttribute(Expression<Func<Attribute>> customAttributeExpression)
```
adamralph: `customAttributeExpression` → `attribute`

tests/FakeItEasy.Tests.Approval.netstd/ApiApproval.ApproveApiNetStd.approved.txt
```diff
@@ -436,7 +436,7 @@ namespace FakeItEasy.Creation
         FakeItEasy.Creation.IFakeOptions ConfigureFake(System.Action<object> action);
         FakeItEasy.Creation.IFakeOptions Implements(System.Type interfaceType);
         FakeItEasy.Creation.IFakeOptions Implements<TInterface>();
-        FakeItEasy.Creation.IFakeOptions WithAdditionalAttributes(System.Collections.Generic.IEnumerable<System.Reflection.Emit.CustomAttributeBuilder> customAttributeBuilders);
+        FakeItEasy.Creation.IFakeOptions WithAdditionalAttributes(params System.Linq.Expressions.Expression<>[] attributes);
```
adamralph: While we're here, and breaking the API, shall we remove the word "additional" and make these methods `WithAttributes`? I'm not seeing which other attributes these attributes can be said to be "additional" to.

All the supporting internal methods could be changed similarly.

src/FakeItEasy/Core/FakeWrapperConfigurator.cs
thomaslevesque: "factories" seems to imply we're going to invoke them, which we're not, so it's probably misleading.

tests/FakeItEasy.Tests.Approval.netstd/ApiApproval.ApproveApiNetStd.approved.txt
thomaslevesque: Works for me. @blairconrad, any objection?

src/FakeItEasy/Core/FakeWrapperConfigurator.cs
adamralph: I don't think "factories" would be misleading, it would accurately convey the content of the expressions. However, the point is purely academic if we're not going to use that convention, and I still think it's not necessary to do so.

thomaslevesque:
> Should we squash the first three commits?

Done!
(the first four, actually, since it didn't seem to make sense to update the doc comments separately)

tests/FakeItEasy.Tests.Approval.netstd/ApiApproval.ApproveApiNetStd.approved.txt
blairconrad: No objection. It's a great idea!

docs/creating-fakes.md
```diff
@@ -52,12 +52,8 @@ var foo = A.Fake<FooClass>(x => x.Implements(typeof(IFoo)));
 var foo = A.Fake<FooClass>(x => x.Implements<IFoo>());
 
 // Assigning custom attributes to the faked class.
```
blairconrad: I feel slightly bad, but while we're here, `IFoo` is not a class. "faked t ype"?

src/FakeItEasy/Creation/IProxyOptions.cs
```diff
@@ -12,6 +12,6 @@ internal interface IProxyOptions
 
         IEnumerable<Action<object>> ProxyConfigurationActions { get; }
 
-        IEnumerable<CustomAttributeBuilder> AdditionalAttributes { get; }
+        IEnumerable<Expression<Func<Attribute>>> AdditionalAttributes { get; }
```
blairconrad: I see we've retained the "additional" here. It's only internal, so not crucial, but do we want to more closely match the language of the public API?

src/FakeItEasy/Creation/ProxyOptions.cs
```diff
@@ -3,24 +3,24 @@ namespace FakeItEasy.Creation
...
     using System.Reflection;
 #endif
-    using System.Reflection.Emit;
 
     internal class ProxyOptions : IProxyOptions
     {
         private readonly List<Type> additionalInterfacesToImplement = new List<Type>();
         private readonly List<Action<object>> proxyConfigurationActions = new List<Action<object>>();
-        private readonly List<CustomAttributeBuilder> additionalAttributes = new List<CustomAttributeBuilder>();
+        private readonly List<Expression<Func<Attribute>>> additionalAttributes = new List<Expression<Func<Attribute>>>();
```
blairconrad: I see we've retained the "additional" here. It's only internal, so not crucial, but do we want to more closely match the language of the public API?

src/FakeItEasy/Creation/IProxyOptions.cs
adamralph: Good spot, I think we should align it, as a `fixup!` for the second commit.

thomaslevesque: I considered changing it, but it was consistent with `AdditionalInterfacesToImplement`... I'll rename it if you prefer.

adamralph: If I understand `IProxyOptions` correctly then "additional" in `AdditionalInterfacesToImplement` makes sense because those interface are in addition to the originally specified interface, but with the attributes, they are not additional to anything, they are just the attributes.

thomaslevesque:
> I reviewed the fixups and they look good to squash.

Done!

adamralph:
Thanks @thomaslevesque !

PR #903, opened 2016-11-15
Make IStartConfiguration<T> internal

### LINKED ISSUE #864: Make IStartConfiguration<T> internal
This is the unnatural fake equivalent of `IFakeConfigurationManager` (which is internal). I don't think there's ever been any good reason to make it public.

There are 2 classes that implement it:
- `Fake<T>` (but its members are never called through the interface)
- `StartConfiguration<T>`, which members are called from `Fake<T>`

For consistency with `IFakeConfigurationManager`, it should probably be hidden from the public API.

thomaslevesque:
Fixes #864


adamralph:
Thanks @thomaslevesque !

Issue #904, opened 2016-11-16
Automatic cancellation support


thomaslevesque:
I have a lot of dependencies with (usually `async`) methods that accept a `CancellationToken`. When the token is cancelled, these methods throw an `OperationCanceledException`. This is the expected implicit behavior for these methods, but by default, fakes don't throw anything when they receive a call with a cancelled token. So my thinking is that this behavior could be baked into FakeItEasy as a default rule. Currently I use a method like this:

```c#
private static void ConfigureCancellation(object obj)
{
    A.CallTo(obj)
        .WhenArgumentsMatch(args => args.OfType<CancellationToken>().Any(ct => ct.IsCancellationRequested))
        .Invokes(call => call.Arguments.OfType<CancellationToken>().FirstOrDefault(ct => ct.IsCancellationRequested).ThrowIfCancellationRequested());
}
```

It works fine, but I have to remember to call it on each fake.

Note that enabling this by default would, technically, be a breaking change, although probably not one that would affect many users.

blairconrad:
> It works fine, but I have to remember to call it on each fake.

I'm not trying to quash this issue, but have you tried to build the behaviour into a [FakeOptionsBuilder](http://fakeiteasy.readthedocs.io/en/stable/implicit-creation-options/)?

blairconrad:
Also, while I'm critiquing, you might want `First` instead of `FirstOrDefault`, since there's going to be an item in the enumerable, and if there weren't you'd throw a less-helpful exception anyhow.

thomaslevesque:
> I'm not trying to quash this issue, but have you tried to build the behaviour into a `FakeOptionsBuilder`?

It wouldn't really solve anything... I want *all* fakes to have this behavior, and I don't want to create a fake options builder for each type of dependency.

By convention (with very few exceptions), a method that accepts a `CancellationToken` should always throw if the token is cancelled. What I'd like is to implicitly enforce that convention in FakeItEasy.

> Also, while I'm critiquing, you might want `First` instead of `FirstOrDefault`, since there's going to be an item in the enumerable, and if there weren't you'd throw a less-helpful exception anyhow.

Good point!

blairconrad:
> I want all fakes to have this behavior, and I don't want to create a fake options builder for each type of dependency.

I was thinking of something like

```c#
public class ThrowsWhenCancelledFakeOptionsBuilder: IFakeOptionsBuilder
{
[...]

PR #906, opened 2016-11-16
[WIP] Automatic cancellation (POC)

thomaslevesque:
Relates to #904 

For discussion

New behavior: any faked method that receives a cancelled `CancellationToken` will throw an `OperationCanceledException`, or return a cancelled `Task` if it's an async method (1). This behavior will not be overridden even if the method is explicitly configured (2).

(1) This is inconsistent with the current behavior for `Throws`. An async method configured to throw an exception will throw the exception directly, instead of returning a failed task. IMO this behavior is not ideal; the code under test might not await the task directly, and instead expect to receive a task that it can examine. Not sure how breaking it would be to change the behavior for `Throws`...

(2) The rationale for this is that it would be too easy to accidentally disable the automatic cancellation behavior. If normal call configurations could override the cancellation behavior, just doing `A.CallTo(() => foo.Bar(A<CancellationToken>._)).Returns(42)` would make the method always return 42, even if the token is cancelled. To avoid that, one would have to explicitly constrain the token to not be cancelled. The downside is that it becomes impossible to disable the automatic cancellation behavior; I think it's OK for most use cases, but we might want to add a way to disable it for specific use cases.

### DIFF
--- a/src/FakeItEasy.netstd/FakeItEasy.netstd.csproj
+++ b/src/FakeItEasy.netstd/FakeItEasy.netstd.csproj
@@ -60,6 +60,9 @@
     <Compile Include="..\FakeItEasy\Configuration\PropertySetterConfiguration.cs">
       <Link>Configuration\PropertySetterConfiguration.cs</Link>
     </Compile>
+    <Compile Include="..\FakeItEasy\Core\FakeManager.CancellationRule.cs">
+      <Link>Core\FakeManager.CancellationRule.cs</Link>
+    </Compile>
     <Compile Include="..\FakeItEasy\ExceptionExtensions.cs">
       <Link>ExceptionExtensions.cs</Link>
     </Compile>
--- a/src/FakeItEasy/Core/FakeManager.CancellationRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.CancellationRule.cs
@@ -1,0 +1,51 @@
+namespace FakeItEasy.Core
+{
+    using System;
+    using System.Linq;
+#if FEATURE_NETCORE_REFLECTION
+    using System.Reflection;
+#endif
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <content>Event rule.</content>
+    public partial class FakeManager
+    {
+#if FEATURE_BINARY_SERIALIZATION
+        [Serializable]
+#endif
+        private class CancellationRule : IFakeObjectCallRule
+        {
+            public int? NumberOfTimesToCall => null;
+
+            public bool IsApplicableTo(IFakeObjectCall fakeObjectCall) =>
+                fakeObjectCall.Arguments.OfType<CancellationToken>().Any(ct => ct.IsCancellationRequested);
+
+            public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
+            {
+                var returnType = fakeObjectCall.Method.ReturnType;
+                if (typeof(Task).GetTypeInfo().IsAssignableFrom(returnType))
+                {
+                    Task task;
+                    if (returnType == typeof(Task))
+                    {
+                        task = TaskHelper.Cancelled();
+                    }
+                    else
+                    {
+                        var taskResultType = returnType.GetTypeInfo().GetGenericArguments()[0];
+                        task = TaskHelper.Cancelled(taskResultType);
+                    }
+
+                    fakeObjectCall.SetReturnValue(task);
+                }
+                else
+                {
+                    var token =
+                        fakeObjectCall.Arguments.OfType<CancellationToken>().First(ct => ct.IsCancellationRequested);
+                    token.ThrowIfCancellationRequested();
+                }
+            }
+        }
+    }
+}
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -39,7 +39,8 @@ namespace FakeItEasy.Core
 
             this.preUserRules = new[]
                                     {
-                                        new CallRuleMetadata { Rule = new EventRule(this) }
+                                        new CallRuleMetadata { Rule = new EventRule(this) },
+                                        new CallRuleMetadata { Rule = new CancellationRule() }
                                     };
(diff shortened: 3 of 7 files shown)


blairconrad:
Sorry for the delay.

I'm always uneasy building more "smarts" into the framework, worrying about when the extra logic will make a feature cross from "feature that will help lots and lots of people" to "niche feature that will help a few and confuse others".

I don't do a lot of asynching. So I probably don't have a good sense of the benefit that the proposed changes will bring. But it seems that Tasks/async/Cancelling is the way of the future, so I'm guessing there will be more and more faked methods to which the change will apply.

I have concerns. Probably a lot due to my ignorance. Please be patient with me. Here goes.

While the big popular libraries have methods that respond to a cancelled token by throwing or returning a cancelled task as you propose, but my understanding is that really a method can do whatever is appropriate. Maybe that's just stopping. Maybe it's something else. How confident are we that the proposed behaviour will match users' needs in all cases, or at least in the majority? I'm confident in our current unconfigured `Lazy` or `Task`-returning method behaviour, but less so in this new behaviour. Maybe because I'm not familiar with it.

And regardless of the answers above, I am really nervous about the behaviour being "always on". The idea that the framework will always know better how the user's faked method should behave, and they have no recourse, frightens me.

Educate/convince me?

thomaslevesque:
> I don't do a lot of asynching. So I probably don't have a good sense of the benefit that the proposed changes will bring. But it seems that Tasks/async/Cancelling is the way of the future, so I'm guessing there will be more and more faked methods to which the change will apply.

I use async code a lot (in our apps at work, all I/O and long-running work is asynchronous), and I have `CancellationToken`s all around the place. They're the easiest way to stop a long running operation cleanly; the main alternatives are flags that you check every now and then (ugly and inconvenient), and aborting the thread, which is dangerous as it can easily leave things in an unknown state. Cancellation tokens are supported everywhere (`HttpClient.SendAsync`, `Task.Delay`, `Stream.ReadAsync`...), so any time you call a method that accepts a cancellation token, you just pass the one you received, and the cancellation will propagate seamlessly. It's a pretty elegant approach IMO. And it's not limited to async code, it can also be used in sync methods.

> While the big popular libraries have methods that respond to a cancelled token by throwing or returning a cancelled task as you propose, but my understanding is that really a method can do whatever is appropriate. Maybe that's just stopping. Maybe it's something else. How confident are we that the proposed behaviour will match users' needs in all cases, or at least in the majority? I'm confident in our current unconfigured Lazy or Task-returning method behaviour, but less so in this new behaviour. Maybe because I'm not familiar with it.

Well, it's a convention. In practice, all methods in the .NET Framework throw an `OperationCanceledException` or return a cancelled task when their cancellation token is cancelled. That being said, it's true that the cancellable method can react however it likes, including by just returning immediately (this is actually mentioned [in the MSDN docs](https://msdn.microsoft.com/en-us/library/dd997364(v=vs.110).aspx#Anchor_3)), so we should take that into account.

> And regardless of the answers above, I am really nervous about the behaviour being "always on". The idea that the framework will always know better how the user's faked method should behave, and they have no recourse, frightens me.

I didn't mean that this behavior should always be on; of course, we should add a way to disable it. What I meant is that it shouldn't be disabled just by configuring a faked method call, because it would make it difficult to *not* disable it (this the example in [this comment](https://github.com/FakeItEasy/FakeItEasy/issues/904#issuecomment-260994711)).

I understand how this proposal can seem scary; I certainly don't want it to break everyone's tests. At this point, it's just an idea open to discussion.

The new behavior could be opt-in, which would make the change much safer. For instance we could add a `ThrowsWhenCancelled` method:

    A.CallTo(() => foo.Bar(A<CancellationToken>._).ThrowsWhenCancelled().Returns(42);

And this method could also be added to `IFakeOptions`:

    var foo = A.Fake<IFoo>(o => o.ThrowsWhenCancelled());

(the same way `CallsBaseMethod` can be configured per-call or per-fake).

IMO it's a bit of a pity to have to opt-in, since well designed cancellable methods would typically throw, but I realize having this behavior enabled by default would probably break things for many people.

Side note: I'm not sure if I should write "cancelled" or "canceled". One is the British spelling and the other the American spelling, but even in the BCL, usage is inconsistent: `OperationCanceledException` (US spelling), `CancellationToken` (British spelling).

blairconrad:
> I didn't mean that this behavior should always be on; of course, we should add a way to disable it. 

Ah. I'm relieved.

>  I meant is that it shouldn't be disabled just by configuring a faked method call, 

I'm still unsure about this. I understand how it would be inconvenient, but I worry that not disabling the behaviour in this situation would lead to confused users. ("I told the method to return 42. Why is it saying it was cancelled!?") In every other situation, saying `A<CancellationToken>._` means _any `CancellationToken`_.

Taking another example you had 

```c#
A.CallTo(() => fake.MyMethod(A<CancellationToken>.That.Matches(ct => !ct.IsCancellationRequested))).Returns(42);
```

could we use template specialization to provide a nicer experience?

```c#
A.CallTo(() => fake.MyMethod(A<CancellationToken>.Uncancelled))).Returns(42);
```

It's not perfect, but maybe a step in the righter direction?

blairconrad:
Oh, for the love of Pete, there's no template specialization in C#. I should've looked this up first.

thomaslevesque:
> I'm still unsure about this. I understand how it would be inconvenient, but I worry that not disabling the behaviour in this situation would lead to confused users. ("I told the method to return 42. Why is it saying it was cancelled!?") In every other situation, saying `A<CancellationToken>._` means any `CancellationToken`.

I see your point, and it makes sense. I guess my idea is a little too disruptive. Anyway, the behavior I proposed is only useful when you specifically want to test the cancellation behavior of a component, so it's probably overkill to introduce such a drastic change. Still, I wish we could provide a better experience around cancellation scenarios.

> there's no template specialization in C#.

No, but we could have an extension method to do this:

    A.CallTo(() => fake.MyMethod(A<CancellationToken>.That.IsNotCancelled())).Returns(42);

A bit verbose for my taste, but that would work. Currently I use a `NotCancelled()` extension method that is used like this:

        A.CallTo(() => fake.MyMethod(A<CancellationToken>._)).NotCancelled().Returns(42);

PR #911, opened 2016-11-18
refactored some members to be expression bodied

adamralph:
FYI I used this little gem to do most of the heavy lifting https://marketplace.visualstudio.com/items?itemName=DustinCampbell.CEssentials

/ht @DustinCampbell

### DIFF
--- a/src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
+++ b/src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
@@ -21,10 +21,8 @@ namespace FakeItEasy.Configuration
             this.voidArgumentValidationConfigurationFactory = voidArgumentValidationConfigurationFactory;
         }
 
-        public IPropertySetterConfiguration To(TValue value)
-        {
-            return this.To(() => value);
-        }
+        public IPropertySetterConfiguration To(TValue value) =>
+            this.To(() => value);
 
         public IPropertySetterConfiguration To(Expression<Func<TValue>> valueConstraint)
         {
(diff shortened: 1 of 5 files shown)


src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
```diff
@@ -101,10 +78,8 @@ public IPropertySetterConfiguration WhenArgumentsMatch(Func<ArgumentCollection,
         }
 
         private static IPropertySetterConfiguration AsPropertySetterConfiguration(
-            IVoidConfiguration voidArgumentValidationConfiguration)
-        {
-            return new ProperySetterAdapter(voidArgumentValidationConfiguration);
-        }
+            IVoidConfiguration voidArgumentValidationConfiguration) =>
+            new ProperySetterAdapter(voidArgumentValidationConfiguration);
```
thomaslevesque: This looks a little weird, with the body aligned with the parameter... in fact, it's one of the reasons I don't like the Java-style braces, because it causes the same issue when the signature line overflows. Maybe it would be better to keep a block-style body here.

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
```diff
@@ -124,10 +99,8 @@ private ParsedCallExpression CreateSetterExpressionWithNewValue(Expression<Func<
         }
 
         private IVoidArgumentValidationConfiguration CreateArgumentValidationConfiguration(
-            ParsedCallExpression parsedSetter)
-        {
-            return this.voidArgumentValidationConfigurationFactory(parsedSetter);
-        }
+            ParsedCallExpression parsedSetter) =>
+            this.voidArgumentValidationConfigurationFactory(parsedSetter);
```
thomaslevesque: same here

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -37,14 +37,11 @@ internal RuleBuilder(BuildableCallRule ruleBeingBuilt, FakeManager manager, Fake
...
-            get
+        private RuleBuilder Then =>
+            new RuleBuilder(this.RuleBeingBuilt.CloneCallSpecification(), this.manager, this.asserterFactory)
             {
-                var newRule = this.RuleBeingBuilt.CloneCallSpecification();
-                return new RuleBuilder(newRule, this.manager, this.asserterFactory) { PreviousRule = this.RuleBeingBuilt };
-            }
-        }
+                PreviousRule = this.RuleBeingBuilt
+            };
```
thomaslevesque: In this case I find the non-expression-bodied version easier to read. The single expression does a bit too many things for my taste.

blairconrad:
1. I like the refactorings (in general) because I get to see more of this: image
2. Seriously, it mostly looks great, although I agree with @thomaslevesque's notes. Thanks!

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
adamralph: I don't mind reverting back to a block, but before I do, what's the opinion on this style?

```C#
private static IPropertySetterConfiguration AsPropertySetterConfiguration(
    IVoidConfiguration voidArgumentValidationConfiguration) =>
        new ProperySetterAdapter(voidArgumentValidationConfiguration);
```

I guess it also looks a little weird.

src/FakeItEasy/Configuration/RuleBuilder.cs
adamralph: Yeah, fair comment. I'll revert this one.

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
thomaslevesque: I like it much better, although I suspect we'll have to fight Visual Studio auto-formatting to keep it that way...

blairconrad: Ditto. If VS doesn't muck it up all the time, I could go for it.

For me (and it may be ReSharper), it keeps getting reformatted to

```c#
private static IPropertySetterConfiguration AsPropertySetterConfiguration(
        IVoidConfiguration voidArgumentValidationConfiguration) =>
    new ProperySetterAdapter(voidArgumentValidationConfiguration);
```

… which I could also live with.

adamralph: Actually, that is the way to do it. The alternative I suggested doesn't convey the nesting, but this does. And Visual Studio reformatting is happy with it. I'll push another fixup.

adamralph:
@FakeItEasy/owners I pushed another fixup. If you're still not happy with that last fixup commit, I'm still OK with reverting them back to block syntax.

thomaslevesque:
No, it's fine. OK to squash and merge!

adamralph:
Squashed!

DustinCampbell:
> FYI I used this little gem to do most of the heavy lifting https://marketplace.visualstudio.com/items?itemName=DustinCampbell.CEssentials

FWIW, this functionality is going into Visual Studio 2017. :smile:

thomaslevesque:
> Squashed!

And merged. Thanks @adamralph!

thomaslevesque:
> FWIW, this functionality is going into Visual Studio 2017. 😄

Good news!

PR #912, opened 2016-11-20
all default returns are dummies (or as close as we can come)

### LINKED ISSUE #830: Make all default return values Dummies, even after Invokes or DoesNothing
Spun off from #806.

Currently, an unconfigured method with a return value will return a `Dummy` of the appropriate type when called.

However, as soon as we start building a rule (for example, when we add an `Invokes` behaviour to the method), we create a rule that will return the "default value" for the type, as we can see in [BuildableCallRule's constructor](https://github.com/FakeItEasy/FakeItEasy/blob/40e6aefec60d17cbebe6c39c34e1089adb0881b9/src/FakeItEasy/Configuration/BuildableCallRule.cs#L23). `GetDefaultValue` [returns null for a reference type or an instance for a value type](https://github.com/FakeItEasy/FakeItEasy/blob/46becf4a6cc8d72ddefc8ca024167d29979efdde/src/FakeItEasy/TypeExtensions.cs#L33).

And then if we decide to have the method `DoesNothing`, the rule is modified so that it will [not set the return value at all](https://github.com/FakeItEasy/FakeItEasy/blob/8e950d803b0a7efc45b17cd882cd0ee7b88dda05/src/FakeItEasy/Configuration/RuleBuilder.cs#L66) (and this is the cause of the bug that @thomaslevesque found).

I think that all three situations should use the "unconfigured behaviour" and return a `Dummy` (or nothing for a void of course).

blairconrad:
Fixes #830.

### DIFF
--- a/tests/FakeItEasy.Specs/ConfigurationSpecs.cs
+++ b/tests/FakeItEasy.Specs/ConfigurationSpecs.cs
@@ -16,6 +16,10 @@ namespace FakeItEasy.Specs
             int Baz();
 
             string Bas();
+
+            IFoo Bafoo();
+
+            IFoo Bafoo(out int i);
         }
 
         [Scenario]
@@ -395,7 +399,7 @@ namespace FakeItEasy.Specs
         }
 
         [Scenario]
-        public static void DoesNothingAfterStrictVoid(
+        public static void DoesNothingAfterStrictVoidDoesNotThrow(
             IFoo fake,
             Exception exception)
         {
@@ -413,7 +417,7 @@ namespace FakeItEasy.Specs
         }
 
         [Scenario]
-        public static void DoesNothingAfterStrictValueType(
+        public static void DoesNothingAfterStrictValueTypeKeepsDefaultReturnValue(
             IFoo fake,
             int result)
         {
@@ -426,12 +430,12 @@ namespace FakeItEasy.Specs
             "When I call a value type method"
                 .x(() => result = fake.Baz());
 
-            "Then it returns a default instance of the value type"
-                .x(() => result.Should().Be(0));
+            "Then it returns the same value as an unconfigured fake"
+                .x(() => result.Should().Be(A.Fake<IFoo>().Baz()));
         }
 
         [Scenario]
-        public static void DoesNothingAfterStrictReferenceType(
+        public static void DoesNothingAfterStrictNonFakeableReferenceTypeKeepsDefaultReturnValue(
             IFoo fake,
             string result)
         {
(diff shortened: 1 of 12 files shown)


tests/FakeItEasy.Specs/ConfigurationSpecs.cs
```diff
@@ -580,11 +602,48 @@ public static void CallsBaseMethodAndDoesNothing(BaseClass fake)
...
+
+            "When I call a fakeable reference type method"
+                .x(() => result = fake.Bafoo());
+
+            "Then it returns the same value as an unconfigured fake"
+                .x(() => result.Should().Be(A.Fake<IFoo>().Bafoo()));
+        }
+
+        [Scenario]
+        public static void AssignsOutAndRefParametersForAllMethodsKeepsDefaultReturnValue(
```
blairconrad: Didn't require a change to make it pass, but I wanted to be explicit about the requirement.

tests/FakeItEasy.Tests/Core/DefaultFixtureInitializerTests.cs
```diff
@@ -132,7 +132,8 @@ public void Should_set_fake_from_sut_initializer_callback_when_available()
             var fake = A.Fake<IFoo>();
 
             A.CallTo(() => this.sutInitializer.CreateSut(A<Type>._, A<Action<Type, object>>._))
-                .Invokes(x => x.GetArgument<Action<Type, object>>(1).Invoke(typeof(IFoo), fake));
+                .Invokes(x => x.GetArgument<Action<Type, object>>(1).Invoke(typeof(IFoo), fake))
+                .Returns(null);
```
blairconrad: Without the `Returns`, our new `Invokes` behaviour returns a fake `object`, which is incompatible with the expected return type.

tests/FakeItEasy.Tests/Core/DefaultSutInitializerTests.cs
```diff
@@ -58,11 +59,10 @@ public void Should_use_same_instance_when_multiple_arguments_are_of_same_type()
         public void Should_pass_empty_options_builder_to_fake_manager_for_each_constructor_argument()
         {
             // Arrange
-            this.StubFakeManagerWithFake<IFoo>();
-            this.StubFakeManagerWithFake<object>();
             var optionsBuilders = new List<Action<IFakeOptions>>();
             A.CallTo(() => this.fakeManager.CreateFake(A<Type>._, A<Action<IFakeOptions>>._))
-                .Invokes((Type type, Action<IFakeOptions> optionBuilder) => optionsBuilders.Add(optionBuilder));
+                .Invokes((Type type, Action<IFakeOptions> optionBuilder) => optionsBuilders.Add(optionBuilder))
+                .ReturnsLazily((Type type, Action<IFakeOptions> fakeOptionBuilder) => Create.Fake(type));
```
blairconrad: Without the `ReturnsLazily`, our new `Invokes` behaviour returns a fake `object`, which is incompatible with the expected return type.
We're just simulating the "stubbing" behaviour we had before.

tests/FakeItEasy.Specs/ConfigurationSpecs.cs
```diff
@@ -665,5 +724,40 @@ public sealed override int ReturnSomething()
                 return 10;
             }
         }
+
+        public class FooFactory : DummyFactory<IFoo>, IFoo
```
thomaslevesque: The fact that the class is both an implementation of `IFoo` and a `DummyFactory<IFoo>` is a bit confusing... Make them separate perhaps?

tests/FakeItEasy.Tests/Configuration/BuildableCallRuleTests.cs
```diff
@@ -284,6 +284,16 @@ public void OutAndRefParameterProducer_should_not_be_settable_more_than_once()
...
 
+        private class DummyableClass : DummyFactory<DummyableClass>
+        {
+            private static readonly DummyableClass Instance = new DummyableClass();
+
+            protected override DummyableClass Create()
+            {
+                return Instance;
+            }
+        }
```
thomaslevesque: Separate into `DummyableClass` and `DummyableClassFactory`?

blairconrad:
Thank you, @thomaslevesque. I've made the changes and also hunted down the two other classes that followed the same pattern.

thomaslevesque:
Perfect, I'll merge after you squash the fixup commits

blairconrad:
Thanks, @thomaslevesque. Pushed. It's all up to AppVeyor now.

thomaslevesque:
Thanks @blairconrad !

PR #918, opened 2016-11-20
Remove IFakeOptionsForWrappers from .NET Standard

### LINKED ISSUE #914: Replace IFakeOptionsForWrappers with IFakeOptions in .NET Standard build
From https://github.com/FakeItEasy/FakeItEasy/issues/820#issuecomment-261774982:

> I notice that we expose `IFakeOptionsForWrappers` and `IFakeOptionsForWrappers<T>` in the netstd build, and they only expose `RecordedBy`.
> Shall we replace them with `IFakeOptions` (`<T>`) for the netstd build? If we'd remove it once we remove Self-Initializing Fakes, I think it makes sense to make the change now.

Everyone else thought it made sense too!

thomaslevesque:
Fixes #914

### DIFF
--- a/src/FakeItEasy/Core/FakeWrapperConfigurator.cs
+++ b/src/FakeItEasy/Core/FakeWrapperConfigurator.cs
@@ -13,7 +13,10 @@ namespace FakeItEasy.Core
     /// </summary>
     /// <typeparam name="T">The type of fake object generated.</typeparam>
     internal class FakeWrapperConfigurator<T>
-        : FakeOptionsBase<T>, IFakeOptionsForWrappers<T>, IFakeOptionsForWrappers
+        : FakeOptionsBase<T>
+#if FEATURE_SELF_INITIALIZED_FAKES
+        , IFakeOptionsForWrappers<T>, IFakeOptionsForWrappers
+#endif
     {
         private readonly IFakeOptions<T> fakeOptions;
 #if FEATURE_SELF_INITIALIZED_FAKES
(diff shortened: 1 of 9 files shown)


src/FakeItEasy/Core/FakeWrapperConfigurator.cs
```diff
@@ -13,7 +13,10 @@ namespace FakeItEasy.Core
     /// </summary>
     /// <typeparam name="T">The type of fake object generated.</typeparam>
     internal class FakeWrapperConfigurator<T>
-        : FakeOptionsBase<T>, IFakeOptionsForWrappers<T>, IFakeOptionsForWrappers
+        : FakeOptionsBase<T>
+#if FEATURE_SELF_INITIALIZED_FAKES
+        , IFakeOptionsForWrappers<T>, IFakeOptionsForWrappers
```
blairconrad: I am surprised StyleCop is okay with this. I'd've thought to `#if` the whole line.
I do not mind this, though.

src/FakeItEasy/Creation/IFakeOptionsForWrappers.cs
```diff
@@ -1,22 +1,18 @@
 namespace FakeItEasy.Creation
 {
-#if FEATURE_SELF_INITIALIZED_FAKES
     using FakeItEasy.SelfInitializedFakes;
-#endif
```
blairconrad: Nice touch. I would've remembered to remove these from the netstd project, but I am not convinced that I would've thought to go after the conditional compilation directives here.

blairconrad:
Thanks, @thomaslevesque!

src/FakeItEasy/Core/FakeWrapperConfigurator.cs
adamralph: It's a bit weird, and I'm also surprised StyleCop didn't complain. It's OK though.

PR #921, opened 2016-11-21
Mark all Self-Initializing Fake types and members as obsolete

### LINKED ISSUE #820: Deprecate SelfInitializedFakes
... because we generally aren't excited about the feature.

And we're planning on removing it. See #798.

blairconrad:
To be removed in 4.0.0.
Fixes #820.

### DIFF
--- a/src/FakeItEasy/Core/FakeWrapperConfigurator.cs
+++ b/src/FakeItEasy/Core/FakeWrapperConfigurator.cs
@@ -20,7 +20,9 @@ namespace FakeItEasy.Core
     {
         private readonly IFakeOptions<T> fakeOptions;
 #if FEATURE_SELF_INITIALIZED_FAKES
+#pragma warning disable CS0618 // Type or member is obsolete
         private ISelfInitializingFakeRecorder recorder;
+#pragma warning restore CS0618 // Type or member is obsolete
 #endif
 
         public FakeWrapperConfigurator(IFakeOptions<T> fakeOptions, object wrappedObject)
@@ -61,12 +63,14 @@ namespace FakeItEasy.Core
         }
 
 #if FEATURE_SELF_INITIALIZED_FAKES
+        [Obsolete("Self-initializing fakes will be removed in version 4.0.0.")]
         public IFakeOptions<T> RecordedBy(ISelfInitializingFakeRecorder fakeRecorder)
         {
             this.recorder = fakeRecorder;
             return this.fakeOptions;
         }
 
+        [Obsolete("Self-initializing fakes will be removed in version 4.0.0.")]
         IFakeOptions IFakeOptionsForWrappers.RecordedBy(ISelfInitializingFakeRecorder fakeRecorder)
         {
             return (IFakeOptions)this.RecordedBy(fakeRecorder);
@@ -89,6 +93,7 @@ namespace FakeItEasy.Core
         }
 
 #if FEATURE_SELF_INITIALIZED_FAKES
+#pragma warning disable CS0618 // Type or member is obsolete
         private static void AddRecordingRuleWhenRecorderIsSpecified(ISelfInitializingFakeRecorder recorder, FakeManager fake, WrappedObjectRule wrapperRule)
         {
             if (recorder != null)
@@ -96,6 +101,7 @@ namespace FakeItEasy.Core
                 fake.AddRuleFirst(new SelfInitializationRule(wrapperRule, recorder));
             }
         }
+#pragma warning restore CS0618 // Type or member is obsolete
 #endif
 
         private static WrappedObjectRule CreateAndAddWrapperRule(object wrappedInstance, FakeManager fake)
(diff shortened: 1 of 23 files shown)


thomaslevesque:
Thanks @blairconrad

I see a lot of `#pragma warning disable 618`. I wonder if there's a way to ignore this warning globally for members marked obsolete in the solution... Disabling it completely is probably not a good idea, since it keeps us from using obsolete members of the BCL or Castle Core, but it's annoying to have to suppress it manually in so many places.

I haven't had time to do a full review yet, I'll come back to it later.

blairconrad:
> it's annoying to have to suppress it manually in so many places.

It's true!

I did relax a little bit by disabling for entire (test) classes.

So far I have not found a way to disable obsolete warnings for obsolete things that we're responsible for.

:slightly_frowning_face:

adamralph:
AFAIK there's no way to say "ignore local obsolete warnings only". It's all or nothing. I think we've no choice but to stick with the pragmas, unless we decide to ignore warnings from other libs.

adamralph:
Wait a sec though. Shouldn't our only use of local obsoletes be in test projects, and we're already suppressing globally for them?

thomaslevesque:
> Wait a sec though. Shouldn't our only use of local obsoletes be in test projects, and we're already suppressing globally for them?

Well, we make `IFakeOptionsForWrappers` obsolete, but it's still the return type for `IFakeOptions.Wrapping`, which is not obsolete... BTW, I wonder if it's not going to be an issue (exposing an obsolete interface as the return type of a non-obsolete method). Maybe `IFakeOptionsForWrappers` itself shouldn't be made obsolete, but only the `RecordedBy` method? But we're probably going to remove that interface in the next version anyway.

thomaslevesque:
> BTW, I wonder if it's not going to be an issue (exposing an obsolete interface as the return type of a non-obsolete method)

Apparently it's not a problem:

```
void Main()
{
    IFoo foo = null;
    var bar = foo?.GetBar();
    int? x = bar?.Baz();
}

interface IFoo
{
#pragma warning disable 612
    IBar GetBar();
#pragma warning restore 612
}

[Obsolete]
interface IBar
{
    int Baz();
}
```

but only because I used `var` instead of `IBar`; so it's still a weird thing to do.

blairconrad:
Ah, this is a good point. I can make the interface non-Obsolete.
It's probably better that way.

blairconrad:
> Shouldn't our only use of local obsoletes be in test projects, and we're already suppressing globally for them?

@adamralph, I don't see any evidence that we are suppressing them.

blairconrad:
Marking as [WIP] while we discuss options.

New commits incoming. `IFakeOptionsForWrappers` will not be obsolete.
In the first commit, I just Obsolete `RecordedBy`. In the second, I expand to types (not returned by us) that are only needed for Self-Initializing Fakes.
I can go with either just the first, or first+second. Or maybe something else, if you want.

thomaslevesque:
> In the first commit, I just Obsolete RecordedBy. In the second, I expand to types (not returned by us) that are only needed for Self-Initializing Fakes.
> I can go with either just the first, or first+second. Or maybe something else, if you want.

I think we should obsolete everything except `IFakeOptionsForWrappers`, because it's not strictly related to self initialized fakes (it could conceivably have members other than `RecordedBy` that only apply to wrapping fakes).

blairconrad:
> I think we should obsolete everything except `IFakeOptionsForWrappers`

Cool! I think that's what I've done. If you think that's what I've done, I can squash the two commits.
I'm in no rush though. Take your time.

thomaslevesque:
> Cool! I think that's what I've done. If you think that's what I've done, I can squash the two commits.

I think so too. Actually I'm not sure I understand the reason for splitting the commits in the first place. OK to squash 👍

blairconrad:
> I'm not sure I understand the reason for splitting the commits in the first place

I wasn't sure whether you'd want to just obsolete `RecordedBy`, or everything (but `IFakeOptionsForWrappers`).
Obsoleting everything (but `IFakeOptionsForWrappers`) made more sense to me, but you seemed sad about the number of pragmas. So I thought you'd enjoy seeing another option. Hence the split, so you could compare.

Since `RecordedBy` is really _the_ entry point into the self-initializing fakes, obsoleting it should be enough to alert users, so if you'd liked the smaller change, with fewer pragmas, I'd've acquiesced.

But I'm happier this way. See: :smiley:

thomaslevesque:
@adamralph are you happy with the changes? i.e. everything related to SIF is marked obsolete, except the `IFakeOptionsForWrapper` (`<T>`) interfaces.

thomaslevesque:
Thanks @blairconrad. I didn't wait for @adamralph's review before merging, but I don't think there's anything really controversial in this  PR.

blairconrad:
Thanks, sir. I think you're right. And if we're wrong, we can fix it!

PR #924, opened 2016-11-28
default version suffix while Castle.Core is pre-release

adamralph:
Connects to #531

### DIFF
--- a/rakefile.rb
+++ b/rakefile.rb
@@ -13,6 +13,7 @@ solution        = "FakeItEasy.sln"
 assembly_info   = "src/CommonAssemblyInfo.cs"
 version         = IO.read(assembly_info)[/AssemblyInformationalVersion\("([^"]+)"\)/, 1]
 version_suffix  = ENV["VERSION_SUFFIX"]
+version_suffix  = version_suffix.to_s.empty? ? "-adhoc" : version_suffix
 build           = (ENV["BUILD"] || "").rjust(6, "0");
 build_suffix    = version_suffix.to_s.empty? ? "" : "-build" + build;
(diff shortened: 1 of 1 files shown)


adamralph:
This was removed when we merged the two rakefiles together, but we need to preserve it while we're taking a dep on Castle.Core pre-release. Without this, a fresh clone doesn't build.

blairconrad:
> Without this, a fresh clone doesn't build.

The freshness of the clone isn't the issue, it's the lack of `VERSION_SUFFIX` environment variable, no?

I don't mind so much the current behaviour. Everything works up 'til the packing part, and the error message is pretty clear about the problem, no?

adamralph:
Yes, it's the lack of the variable. By saying "fresh clone" I guess I was attempting to emphasise the experience for the new comer. I.e. I clone the repo, follow the how to build instructions, and I get:

> A stable release of a package should not have a prerelease dependency. Either modify the version spec of dependency "Castle.Core [4.0.0-beta002, )" or update the version field in the nuspec.
> F, [2016-11-28T07:28:56.887392 #13968] FATAL -- : Exec Failed. See Build Log For Detail
> rake aborted!
> Exec Failed. See Build Log For Detail

There's nothing obvious here which says "you need to set an environment varaible named "VERSION_SUFFIX" to *some string* in order for the build to work. It's not a great contributor experience.

thomaslevesque:
Looking good, thanks @adamralph

blairconrad:
> Yes, it's the lack of the variable. By saying "fresh clone" I guess I was attempting to emphasise the experience for the newcomer

Ah. Of course. Thanks, @adamralph.

PR #909, opened 2016-11-17
merge .NET Standard and .NET Full builds 

blairconrad:
Fixes #908.

I made many more commits than necessary, such as removing files, then moving their replacements on top instead of just renaming, to help git figure out what's really changing at each stage. I hope it helped the diffs.

I think it would be appropriate to squash all the commits later, but am happy to do whatever.

PR #925, opened 2016-11-28
Remove build.netstd.cmd from docs

ar7z1:
In #909 @blairconrad removed `build.netstd.cmd`.
In this PR I removed all mentions of `build.netstd.cmd` from docs.


blairconrad:
Yes, thanks, @ar7z1!

PR #927, opened 2016-11-28
Fixes #926. Add Invokes with 5 arguments

### LINKED ISSUE #926: Support Invokes method with 5 arguments
I'd like to call `Invokes` for method with 5 arguments ([IDataRecord.GetBytes](https://msdn.microsoft.com/en-us/library/system.data.idatarecord.getbytes)):
```csharp
A.CallTo(() => dataRecord.GetBytes(10, A<long>._, A<byte[]>._, A<int>._, A<int>._))
 .Invokes((int i, long fieldOffset, byte[] buffer, int bufferOffset, int length) => {
        new Random().NextBytes(buffer);
 }).Returns(16);
```

But `Invokes` doesn't support methods with 5 arguments :-(

ar7z1:
In this PR I added a support of 5 arguments for `Invokes` method.

Do I need to add fifth argument to `Throws`, `ReturnsLazily` and `AssignsOutAndRefParametersLazily` methods?

### DIFF
--- a/src/FakeItEasy/CallbackConfigurationExtensions.cs
+++ b/src/FakeItEasy/CallbackConfigurationExtensions.cs
@@ -117,5 +117,30 @@ namespace FakeItEasy
                     actionToInvoke(call.GetArgument<T1>(0), call.GetArgument<T2>(1), call.GetArgument<T3>(2), call.GetArgument<T4>(3));
                 });
         }
+
+        /// <summary>
+        /// Executes the specified action when a matching call is being made.
+        /// </summary>
+        /// <param name="configuration">The configuration that is extended.</param>
+        /// <param name="actionToInvoke">The <see cref="Action{T1,T2,T3,T4,T5}"/> to invoke.</param>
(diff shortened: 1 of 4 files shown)


blairconrad:
Hi, @ar7z1. Thanks for your interest in FakeItEasy.

However, we've intentionally limited the "convenience" versions of `Invokes`, `Throws`, and so on to 4 arguments. We made this decision both because of the increasing effort involved in maintaining more overloads, and because methods with a large number of arguments are often considered to be poor design.

There's no rule that says "4 is good, 5 is bad", of course, but we consider 4 arguments to be a reasonable cutoff.

You can always use `Invokes` overload that takes a `Action<IFakeObjectCall>` when you need to fake a method that takes more arguments.

blairconrad:
(I'd made a comment here about not wanting a 5+ argument overload for `Invokes`, but that we before I saw issue #926. I've moved the comment [there](https://github.com/FakeItEasy/FakeItEasy/issues/926#issuecomment-263279314).)

PR #928, opened 2016-11-28
Add a .NET Core test project for FakeItEasy.Specs

thomaslevesque:
Connects to #531 

Well, it wasn't as painful as I feared it would be :wink:

PR #929, opened 2016-11-29
Move cacert.pem up like VisualStudio wants it

blairconrad:
I'd suggested in #928 that this should be done in a separate commit or even PR.
I'd feel like a heel forcing the work on @thomaslevesque.

@thomaslevesque, I thought this might make things a little more convenient for you. If not, please ignore.


blairconrad:
Thanks, @adamralph.

PR #931, opened 2016-12-01
Automatic cancellation behavior

### LINKED ISSUE #904: Automatic cancellation support

thomaslevesque:
Fixes #904 

By default, calls that receive a cancelled `CancellationToken` argument will either:
- throw an `OperationCancelledException` (for synchronous methods)
- return a cancelled `Task` (for asynchronous methods)

This behavior will be overridden if the call is explicitly configured. To avoid this, the `CancellationToken` argument should be constrained to be not cancelled; a new `IsNotCancelled` extension method is provided as a shortcut for this scenario:

```
A.CallTo(() => foo.Bar(A<CancellationToken>.That.IsNotCancelled())).Returns(42);
```

A symmetric `IsCancelled` extension method is also available.

### DIFF
--- a/docs/default-fake-behavior.md
+++ b/docs/default-fake-behavior.md
@@ -22,11 +22,20 @@ invoked on the fake, no action will be taken by the fake. It is as if
 the body of the member were empty. If the member has a return type (or
 is a get property), the return value will depend on the type `T` of
 the member:
-  
+
 * If `T` can be made into a [Dummy](dummies.md), then a Dummy `T` will
   be returned. Note that this may be a Fake or an instance of a
   concrete, pre-existing type;
-* othewise, `default(T)` will be returned.
+* otherwise, `default(T)` will be returned.
+
+### Cancellation tokens
+
+When a faked method that accepts a `CancellationToken` receives a cancelled token
+(i.e. with `IsCancellationRequested` set to `true`), it will either:
+
+* return a cancelled task, if it is asynchronous (i.e. if it returns a `Task` or
+  `Task<T>`);
+* throw an `OperationCanceledException`, if it is synchronous.
 
 ## Examples
 
@@ -55,7 +64,7 @@ public void Members_should_return_empty_string_default_or_fake_another_fake()
 
     Assert.AreEqual(default(int), fakeLibrary.IntProperty);
 
-    Assert.AreEqual(typeof(string), fakeLibrary.StringFunction().GetType()); 
+    Assert.AreEqual(typeof(string), fakeLibrary.StringFunction().GetType());
     Assert.AreEqual(string.Empty, fakeLibrary.StringFunction());
 
     Assert.IsInstanceOfType(fakeLibrary.FakeableClassFunction(), typeof(FakeableClass));
--- a/src/FakeItEasy/Core/FakeManager.CancellationRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.CancellationRule.cs
@@ -1,0 +1,60 @@
+namespace FakeItEasy.Core
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+#if FEATURE_NETCORE_REFLECTION
+    using System.Reflection;
+#endif
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <content>Event rule.</content>
+    public partial class FakeManager
+    {
+#if FEATURE_BINARY_SERIALIZATION
+        [Serializable]
+#endif
+        private class CancellationRule : IFakeObjectCallRule
+        {
+            public int? NumberOfTimesToCall => null;
+
+            public bool IsApplicableTo(IFakeObjectCall fakeObjectCall) =>
+                GetCancelledTokens(fakeObjectCall).Any();
+
+            public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
+            {
+                var returnType = fakeObjectCall.Method.ReturnType;
+                if (typeof(Task).GetTypeInfo().IsAssignableFrom(returnType))
+                {
+                    Task task;
+                    if (returnType == typeof(Task))
+                    {
+                        task = TaskHelper.Cancelled();
+                    }
+                    else
+                    {
+                        var taskResultType = returnType.GetTypeInfo().GetGenericArguments()[0];
+                        task = TaskHelper.Cancelled(taskResultType);
+                    }
+
+                    fakeObjectCall.SetReturnValue(task);
+                }
+                else
+                {
+                    var token = GetCancelledTokens(fakeObjectCall).First();
+                    token.ThrowIfCancellationRequested();
+                }
+            }
+
+            private static IEnumerable<CancellationToken> GetCancelledTokens(IFakeObjectCall call)
+            {
+                return call.Method.GetParameters()
+                    .Select((param, index) => new { param, index })
+                    .Where(x => x.param.ParameterType == typeof(CancellationToken))
+                    .Select(x => (CancellationToken)call.Arguments[x.index])
+                    .Where(token => token.IsCancellationRequested);
+            }
+        }
+    }
+}
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -45,6 +45,7 @@ namespace FakeItEasy.Core
                                          new CallRuleMetadata { Rule = new ObjectMemberRule(this) },
                                          new CallRuleMetadata { Rule = new AutoFakePropertyRule(this) },
                                          new CallRuleMetadata { Rule = new PropertySetterRule(this) },
+                                         new CallRuleMetadata { Rule = new CancellationRule() },
                                          new CallRuleMetadata { Rule = new DefaultReturnValueRule() }
                                      };
 
--- a/src/FakeItEasy/TaskHelper.cs
+++ b/src/FakeItEasy/TaskHelper.cs
@@ -1,13 +1,51 @@
 namespace FakeItEasy
 {
+    using System;
+    using System.Collections.Concurrent;
+    using System.Reflection;
     using System.Threading.Tasks;
 
     internal static class TaskHelper
     {
+        private static readonly ConcurrentDictionary<Type, Task> CachedCancelledTasks = new ConcurrentDictionary<Type, Task>();
+
+        private static readonly MethodInfo CreateGenericCancelledTaskGenericDefinition =
+            typeof(TaskHelper).GetMethod(
+                nameof(CreateGenericCancelledTask),
+                BindingFlags.Static | BindingFlags.NonPublic);
+
         public static Task<T> FromResult<T>(T result)
         {
             var source = new TaskCompletionSource<T>();
             source.SetResult(result);
+            return source.Task;
+        }
+
+        public static Task Cancelled()
+        {
+            return Cancelled(typeof(int));
+        }
+
+        public static Task Cancelled(Type resultType)
+        {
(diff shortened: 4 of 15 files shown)


src/FakeItEasy/Core/FakeManager.CancellationRule.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+#if FEATURE_BINARY_SERIALIZATION
+        [Serializable]
+#endif
+        private class CancellationRule : IFakeObjectCallRule
+        {
+            public int? NumberOfTimesToCall => null;
+
+            public bool IsApplicableTo(IFakeObjectCall fakeObjectCall) =>
+                fakeObjectCall.Arguments.OfType<CancellationToken>().Any(ct => ct.IsCancellationRequested);
```
thomaslevesque: Oops, I just realized something... This will match any argument of type `CancellationToken`, *even if the parameter isn't declared as such* (e.g. if it's declared as `object`). I'll update it to take the actual parameter type into account.

src/FakeItEasy/Core/FakeManager.CancellationRule.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+#if FEATURE_BINARY_SERIALIZATION
+        [Serializable]
+#endif
+        private class CancellationRule : IFakeObjectCallRule
+        {
+            public int? NumberOfTimesToCall => null;
+
+            public bool IsApplicableTo(IFakeObjectCall fakeObjectCall) =>
+                GetCancellationTokens(fakeObjectCall).Any();
```
blairconrad: could the name of this method indicate that it's _cancelled_ cancellation tokens we're getting?

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -45,6 +45,7 @@ internal FakeManager(Type fakeObjectType, object proxy)
                                          new CallRuleMetadata { Rule = new ObjectMemberRule(this) },
                                          new CallRuleMetadata { Rule = new AutoFakePropertyRule(this) },
                                          new CallRuleMetadata { Rule = new PropertySetterRule(this) },
+                                         new CallRuleMetadata { Rule = new CancellationRule() },
```
blairconrad: What if a property getter has a cancellation token as an argument? Do we want the property-ness behaviour to be preserved, or the cancellation behaviour?
I'd originally imagined the cancellation rule being first in the list.

src/FakeItEasy/TaskHelper.cs
```diff
@@ -1,14 +1,59 @@
...
+
+        public static Task Cancelled<T>()
+        {
+            return Cancelled(typeof(T));
+        }
+
+        private static Task<T> CreateGenericCancelledTask<T>()
+        {
+            var source = new TaskCompletionSource<T>();
+            source.TrySetCanceled();
```
blairconrad: Is there any reason we'd expect this to return `false`? Equivalently, would `SetCanceled` be safe? I'd prefer it, unless there's a reason not to.

docs/default-fake-behavior.md
```diff
@@ -22,11 +22,20 @@ invoked on the fake, no action will be taken by the fake. It is as if
...
+* otherwise, `default(T)` will be returned.
+
+### Cancellation tokens
+
+When a faked method that accepts a `CancellationToken` receives a cancelled token
+(i.e. with `IsCancellationRequested` set to `true`), it will either:
+
+* throw an `OperationCanceledException`, if it is synchronous;
+* return a cancelled task, if it is asynchronous (i.e. if it returns a `Task` or
+  `Task<T>`).
```
blairconrad: I'd consider reversing these. Then we can concentrate on the task/no task distinction and not worry about the synchronousness of the method (which we really can't determine from where we stand).

src/FakeItEasy/Core/FakeManager.CancellationRule.cs
thomaslevesque: Ah, yes. I intended to call it `GetCancelledTokens`, not sure why I didn't.

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: My thinking was that it was very unlikely for a property to have a `CancellationToken` as an index, and if it did, it probably wouldn't mean what it usually means for a method, so I gave the `CancellationRule` a lower priority than the property rules. As for `ObjectMemberRules`, the members of `Object` don't accept cancellation tokens as argument, so there's no point in looking for them.

src/FakeItEasy/TaskHelper.cs
thomaslevesque: You're right, `SetCanceled` has no reason to fail here. I was under the impression that `TrySetCanceled` was the "right" way to do it, because they added an overload with a `CancellationToken` in 4.6 for `TrySetCanceled` but not for `SetCanceled`. It doesn't matter in this case though, I'll change it to `SetCancelled`.

thomaslevesque:
I pushed the requested changes

tests/FakeItEasy.Specs/CancellationSpecs.cs
```diff
@@ -0,0 +1,240 @@
...
+    {
+        public interface IFoo
+        {
+            int Bar();
+
+            int Bar(CancellationToken cancellationToken);
+
+            int Bar(object obj);
+
+            Task<int> BarAsync(CancellationToken cancellationToken);
```
blairconrad: I was expecting something with a plain `Task` too.

src/FakeItEasy/TaskHelper.cs
```diff
@@ -1,14 +1,59 @@
 namespace FakeItEasy
 {
+    using System;
+    using System.Collections.Concurrent;
+    using System.Reflection;
     using System.Threading.Tasks;
 
     internal static class TaskHelper
     {
+        private static readonly Lazy<Task> CachedCancelledTask = new Lazy<Task>(CreateGenericCancelledTask<int>);
```
blairconrad: Could we instead use (or add) the value at `CachedCancelledTasks[typeof(int)]`?

src/FakeItEasy/TaskHelper.cs
```diff
@@ -1,14 +1,59 @@
...
+            return CachedCancelledTasks.GetOrAdd(
+                resultType,
+                type =>
+                {
+                    var method = CreateGenericCancelledTaskGenericDefinition.MakeGenericMethod(type);
+                    return (Task)method.Invoke(null, new object[0]);
+                });
+        }
+
+        public static Task Cancelled<T>()
```
blairconrad: this is not used

src/FakeItEasy/TaskHelper.cs
thomaslevesque: > Why int? Just because we need something and int is innocuous?

Yes.

> Could we instead use (or add) the value at `CachedCancelledTasks[typeof(int)]`?

Good point. Yes, we can totally do that.

tests/FakeItEasy.Specs/CancellationSpecs.cs
thomaslevesque: OK

tests/FakeItEasy.Specs/CancellationSpecs.cs
```diff
@@ -0,0 +1,240 @@
...
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<IFoo>());
+
+            "When a method which doesn't accept a cancellation token is called"
+                .x(() => result = fake.Bar());
+
+            "Then it doesn't throw and returns the default value"
+                .x(() => result.Should().Be(0));
+        }
```
thomaslevesque: Ah, yes, I thought you were talking about another one. I included it as a sanity check, but you're right, it's covered elsewhere. I'll remove it.

thomaslevesque:
I made the requested changes. I'm not sure I'll be able to squash the last fixup commit cleanly, so if you don't mind, I think I'll squash all the specs commits together in the end.

tests/FakeItEasy.Specs/CancellationSpecs.cs
```diff
@@ -149,7 +157,7 @@ public void AsyncNonCancelledCancellationToken(IFoo fake, CancellationToken canc
                 .x(() => cancellationToken = new CancellationToken(false));
 
             "When an async method is called with this cancellation token"
-                .x(() => task = fake.BarAsync(cancellationToken));
+                .x(() => { task = fake.BarAsync(cancellationToken); });
```
blairconrad: Not a big deal, but what prompted this change (and its friends)?
It's a departure from many other of the one-line assignments in the file.

thomaslevesque: Ah, yes, I intended to comment on it but I forgot. The `x` method has 2 overloads (4 actually, but we rarely use the ones with an `IStepContext`):
- one that takes an `Action`
- and one that takes a `Func<Task>`, for async scenarios

When I use

    .x(() => task = fake.BarAsync(cancellationToken))

the expression returns a task, so it's inferred to be a `Func<Task>`. And in that case, xBehave awaits the task. This is not what we want here (we just want to assign the task, not return it), so I use a statement body to force it to be inferred as `Action`. I had already noticed the problem for cancelled tasks because it caused the steps to fail, but I thought it would be more consistent to do the same for successful tasks.

thomaslevesque: > It's a departure from many other of the one-line assignments in the file

True, but it's only necessary for task assignments.

thomaslevesque:
Squashed. I also had to reorder the commits for things to make sense (otherwise I would have had specs that used `IsNotCancelled` before it was introduced).

blairconrad:
> Ah, yes, I intended to comment on it but I forgot. The x method has 2 overloads (4 actually, but we rarely use the ones with an `IStepContext`):
>
> one that takes an Action
> and one that takes a `Func<Task>`, for async scenarios
> When I use
> 
> `.x(() => task = fake.BarAsync(cancellationToken))`
> the expression returns a task, so it's inferred to be a `Func<Task>`. And in that case, xBehave awaits the task. This is not what we want here (we just want to assign the task, not return it), so I use a statement body to
> force it to be inferred as Action. I had already noticed the problem for cancelled tasks because it caused the steps to fail, but I thought it would be more consistent to do the same for successful tasks.

Thanks! That helps!

I'll leave open for a bit in case @adamralph wants to peruse.

adamralph:
I can peruse it later today if you don't mind waiting.

thomaslevesque:
> I can peruse it later today if you don't mind waiting.

I don't mind at all :wink:

thomaslevesque:
Thanks for the merge!

PR #934, opened 2016-12-03
Trying to fix appveyor build - DON'T MERGE

thomaslevesque:


thomaslevesque:
It's ugly, but it seems to work... no idea why, though. It's as if the PATH env variable was already almost at the maximum length, and adding `C:\Ruby200\bin;` made it exceed that maximum. But in that case just doing it with `set` in cmd shouldn't change anything... unless the maximum length is an AppVeyor limitation, not a system limitation.

blairconrad:
That is peculiar, @thomaslevesque. While it'd be nice to have the `PATH` change in the `environment` section, I'd take the pragmatic way out.
Of course, now that we have only one `rakefile`, we could instead experiment with simple-targets-csx, if we're of a mind to.

thomaslevesque:
See https://github.com/appveyor/ci/issues/1202

thomaslevesque:
> Of course, now that we have only one rakefile, we could instead experiment with simple-targets-csx, if we're of a mind to.

In the long run, I'd love to move away from Rake. Maybe we should start to work on it.

adamralph:
Superseded by #935

blairconrad:
> In the long run, I'd love to move away from Rake. Maybe we should start to work on it.

Sounds good. I'd volunteer, but I'd rather focus on my Nancy/travis issue right now. Fun with vagrant!

blairconrad:
(Not fun with vagrants. That's a whole different thing.)

adamralph:
Obviously I'm all for simple-targets-csx 😉

PR #936, opened 2016-12-05
Fix code analysis warnings

thomaslevesque:
I disabled a few rules for test projects:

- CA1704:IdentifiersShouldBeSpelledCorrectly
- CA1720:IdentifiersShouldNotContainTypeNames
- CA2204:Literals should be spelled correctly

because they don't really add any value for test code, and suppressing them just adds noise.

FxCop also says to use "canceled" rather than "cancelled" ([CA1726](https://msdn.microsoft.com/en-us/library/ms182258.aspx)). I'm not sure if we want to do this or not, so it's in a separate commit, which I can remove if we don't want it.


thomaslevesque:
Note that whatever we decide about "canceled" vs. "cancelled", we'll have to stick to it, since it's part of the public API. FWIW, the BCL uses "canceled" (`OperationCanceledException`, `TaskCanceledException`, `WebExceptionStatus.RequestCanceled`...), even though it uses "cancellation" rather than "cancelation". Go figure...

adamralph:
Thanks @thomaslevesque

thomaslevesque:
Thanks for settling the issue @adamralph. We were discussing it with @blairconrad yesterday and couldn't decide 😉.

blairconrad:
Thanks for the fixups, @thomaslevesque!

PR #941, opened 2016-12-05
Improve pre-release example

blairconrad:
… because there was no good example, and it would be very easy to supply the "wrong" parameter, e.g. 3.0.0-beta003 instead of just beta003

I suppose the task could validate the input, but I am lazy. And did not think of it 'til now.


thomaslevesque:
LGTM, thanks @blairconrad

blairconrad:
Thanks, @thomaslevesque!

PR #943, opened 2016-12-09
Implement ThrowsAsync

### LINKED ISSUE #932: Add a ThrowsAsync method to return a failed task
While thinking about how [automatic cancellation](https://github.com/FakeItEasy/FakeItEasy/issues/904) should work for async methods, I realized that the current behavior for `Throws` isn't ideal for these methods. The problem is that there are two ways for a `Task`-returning method to "throw":

1. it can throw an exception directly, if its implementation is not actually asynchronous but just has a `Task` return type. A typical case is a method that just returns the result of another async method, e.g. in order to validate the arguments synchronously:

        public Task<Foo> LoadFooAsync(string path)
        {
            // the exception will be thrown immediately when the method is called
            if (path == null) throw new ArgumentNullException(nameof(path));
            return LoadFooAsyncImpl(path);
        }

        private async Task<Foo> LoadFooAsyncImpl(string path)
        {
            using (var stream = File.OpenRead(path))
            using (var reader = new StreamReader(stream))
            {
                string text = reader.ReadToEndAsync();
                return Foo.Parse(text);
            }
        }

2. or it can return a failed task, if its implementation really is asynchronous:

        public async Task<Foo> LoadFooAsync(string path)
        {
            // the exception will NOT be thrown immediately when the method is called,
            // but only when the task is awaited
            if (path == null) throw new ArgumentNullException(nameof(path));

            using (var stream = File.OpenRead(path))
            using (var reader = new StreamReader(stream))
            {
                string text = reader.ReadToEndAsync();
                return Foo.Parse(text);
            }
        }

The difference doesn't really matter in most cases, because if you directly `await` the result of the method, the behavior will be the same. But if instead you want to do something with the resulting task, it makes a big difference, because in the first case the method will throw directly and you won't get a task at all. For instance, if you do something like this:
```
LoadFooAsync(path).ContinueWith(...);
```
if `path` is null, `LoadFooAsync` will throw and `ContinueWith` won't be executed. Of course, when you have async/await, using `ContinueWith` becomes less common, but there are other scenarios where you would want to manipulate the tasks directly, e.g. parallel processing:
```
var fooTask = GetFooAsync();
var barTask = GetBarAsync();

await Task.WhenAll(fooTask, barTask);
```
In cases like that, the difference between throwing directly and returning a failed task becomes important.

OK, that was a long introduction, but I'm getting to the point :wink:

Currently, FakeItEasy only offers option 1 : configure a method to throw an exception directly. This is fine in most cases, but won't work for scenarios when the task needs to be manipulated, rather than just awaited. Of course, it's possible to explicitly configure the method to return a failed task, but it's not very convenient:
```
A.CallTo(() => fake.GetFooAsync()).Returns(Task.FromException<Foo>(new Exception("oops")));
```
(also, `FromException` is only available since .NET 4.6)

FIE already provides a shortcut to return a value directly from an async method, rather than explicitly configure it to return a completed task. We could have a similar shortcut for returning a failed task:
```
A.CallTo(() => fake.GetFooAsync()).ThrowsAsync(new Exception("oops"));
```
The name clearly indicates that it "throws asynchronously", rather than directly throw an exception.

The only drawback I can see is that it would be slightly misleading, since `ThrowsAsync` would actually configure a return value.

thomaslevesque:
Fixes #932.


tests/FakeItEasy.Specs/ThrowsAsyncSpecs.cs
```diff
@@ -0,0 +1,310 @@
...
+        [Scenario]
+        public void NoResultPassExceptionInstance(IFoo fake, Task task)
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<IFoo>());
+
+            "And an async method of the fake configured to throw asynchronously"
+                .x(() => A.CallTo(() => fake.BarAsync()).ThrowsAsync(new MyException()));
+
+            "When that method is called"
```
blairconrad: This is super-nitpicky, but I think we've been tending towards a "When I…" style of naming.
Our record isn't perfect, but we're up to about 150/239 (before your changes) conforming.
Consider rewording? I wouldn't hold up the merge for this.

thomaslevesque:
> The only other thing I see is the API approval files, which need updating.

Ah, yes. I didn't even check if the build passed :wink:

> Oh, and do we want to consider this for 3.0.0?

We could, but as you said, it's not breaking, so there's no rush. Let's just ship it when it's ready, whether it's for 3.0.0 or later.

tests/FakeItEasy.Specs/ThrowsAsyncSpecs.cs
thomaslevesque: > I think we've been tending towards a "When I…" style of naming.

Have we? That wasn't my impression; on the contrary, I've been intentionally staying away from this style... If this is "officially" what we want to do, I don't mind changing it, of course.

blairconrad: Oh, man. Maybe not! Perhaps we've been at odds. I thought I remembered that from one of @adamralph's examples or comments, but I don't have that at hand. I think the text you have reads fine.

blairconrad:
I've a mind like a sieve. [Docs](http://fakeiteasy.readthedocs.io/en/stable/throwing-exceptions/)? Sorry.

thomaslevesque:
> I've a mind like a sieve. Docs? Sorry.

Yes, I'll update the docs too.

Are we happy with the proposed API?

blairconrad:
> Yes, I'll update the docs too.

Yay!

> Are we happy with the proposed API?

I am! I know I was worried at first, but if one's in the "async" mindset, I think it makes sense, and I enjoy having it only show up on task-returning methods.

docs/throwing-exceptions.md
```diff
@@ -38,3 +38,22 @@ A.CallTo(() => fakeShop.NumberOfSweetsSoldOn(A<DateTime>._))
...
+
+- throw the exception synchronously, i.e. not actually return a `Task`
+- "throw asynchronously", i.e. return a failed task with the exception.
+
+The former is supported by the `Throws` method described above, in the same way as if the
+method was synchronous. The latter can be configured by using the `ThrowsAsync` method:
+
+```csharp
+A.CallTo(() => fakeShop.OrderSweetsAsync("cheeseburger"))
+ .ThrowsAsync(new ArgumentException("'cheeseburger' isn't a valid sweet category"));
```
blairconrad: I direct you to http://www.worksburger.com/Home.aspx

image

You still want to keep your example?

(As an aside, I have actually eaten (half of) this thing. It's better than you'd think. A little sweet. It definitely did open my eyes to "peanut butter on burgers", which is quite good!)

tests/FakeItEasy.Specs/ThrowsAsyncSpecs.cs
blairconrad: Oh! I just realized why the "I" is an option at all. The (in our case implicit) 0th part of the BDD script:

```cucumber
As a developer using FakeItEasy in my tests
Given a fake
And an async method of the fake configured to throw asynchronously
When I call that method
```

Or something. But I'm still fine with what you have!

thomaslevesque:
> You still want to keep your example?

I don't have a better idea at the moment... if you do, I'll be glad to take it ;)

blairconrad:
Sterling work, @thomaslevesque. Thanks!

thomaslevesque:
Thanks for the merge!

blairconrad:
Of course. I'm sure you'd do the same for me one day.

PR #946, opened 2016-12-13
Fix bug for constraint on ref or out argument

### LINKED ISSUE #899: Ignored/That argument constraints do not Work for ByRef-Arguments in VisualBasic
All Versions >= 2.00

See the following interface
```vb.net
Public Interface IFoo
    Sub Bar(b As Object)
    Sub BarByRef(ByRef b As Object)
End Interface
```

As soon as I want to use any argument constraint for an ByRef-Argument an exception is thrown. (System.InvalidOperationException: A<T>.Ignored, A<T>._, and A<T>.That can only be used in the context of a call specification with A.CallTo)

```vb.net
<TestMethod>
Public Sub TestMethod1()
    Dim test = A.Dummy(Of IFoo)

    test.Bar(2)
    test.BarByRef(2)

    A.CallTo(Sub() test.Bar(A(Of Object).Ignored)).MustHaveHappened()      'OK

    A.CallTo(Sub() test.BarByRef(2)).MustHaveHappened()                    'OK

    A.CallTo(Sub() test.BarByRef(A(Of Object).Ignored)).MustHaveHappened() 'Error
End Sub
```

thomaslevesque:
Fixes #899

### DIFF
--- a/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
+++ b/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
@@ -28,15 +28,18 @@ namespace FakeItEasy.Expressions
 
             var isByRefArgument = IsByRefArgument(argument);
 
-            if (isByRefArgument && IsOutArgument(argument))
-            {
-                return new OutArgumentConstraint(argument.Value);
-            }
-
-            var constraint = this.GetArgumentConstraintFromExpression(argument.Expression);
+            object argumentValue;
+            var constraint = this.GetArgumentConstraintFromExpression(argument.Expression, out argumentValue);
             if (isByRefArgument)
             {
-                constraint = new RefArgumentConstraint(constraint, argument.Value);
+                if (IsOutArgument(argument))
+                {
+                    constraint = new OutArgumentConstraint(argumentValue);
+                }
+                else
+                {
+                    constraint = new RefArgumentConstraint(constraint, argumentValue);
+                }
             }
 
             return constraint;
@@ -77,7 +80,7 @@ namespace FakeItEasy.Expressions
             return Expression.Lambda(expression).Compile().DynamicInvoke();
         }
 
-        private IArgumentConstraint GetArgumentConstraintFromExpression(Expression expression)
+        private IArgumentConstraint GetArgumentConstraintFromExpression(Expression expression, out object value)
         {
             object expressionValue = null;
 
@@ -85,6 +88,8 @@ namespace FakeItEasy.Expressions
             {
                 expressionValue = InvokeExpression(expression);
             });
+
+            value = expressionValue;
 
             return TryCreateConstraintFromTrappedConstraints(trappedConstraints.ToArray()) ?? CreateEqualityConstraint(expressionValue);
         }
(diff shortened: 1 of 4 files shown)


tests/FakeItEasy.IntegrationTests.VB/RefArgumentConstraintTests.vb
```diff
@@ -0,0 +1,38 @@
...
+
+Public Class RefArgumentConstraintTests
+
+    Public Interface IFoo
+        Function Bar(ByRef x As String) As Integer
+        Function Baz(<Out> ByRef x As String) As Integer
+    End Interface
+
+    <Fact>
+    Public Sub Constraint_On_Ref_Argument_Is_Honored()
```
blairconrad: This is such a minor comment, but the uppercase letters on each word look odd to me.
Of course, so does all of VB.NET.
Is this standard? I notice that the tests in the two sibling files use lowercase, except when referring to a symbol.

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -77,7 +80,7 @@ private static object InvokeExpression(Expression expression)
             return Expression.Lambda(expression).Compile().DynamicInvoke();
         }
 
-        private IArgumentConstraint GetArgumentConstraintFromExpression(Expression expression)
+        private IArgumentConstraint GetArgumentConstraintFromExpression(Expression expression, out object value)
```
blairconrad: `out`s always look funny to me. It's a private method, so not a huge deal.
Consider returning a compound object?
If you still prefer the `out`, I will hit the "merge" button.

tests/FakeItEasy.IntegrationTests.VB/RefArgumentConstraintTests.vb
thomaslevesque: You're right, I'll change it to lowercase

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
thomaslevesque: Well, I'd use a tuple, but we can't use C# 7 yet :wink:
We could use `Tuple<IArgumentConstraint, object>`, but IMO it's worse than `out`.
Creating a class just for this seems a bit overkill to me...

blairconrad: Fair enough. I just thought I'd present it for consideration. 🍁

thomaslevesque: In fact, I was trying to refactor the code to pass a `ParsedArgumentExpression` to `GetArgumentConstraintFromExpression`, which would make the code simpler, but unfortunately we don't have a `ParsedArgumentExpression` in `CreateParamArrayConstraint`...

thomaslevesque:
I amended the unit test commit to change the test names to lowercase, and I also added a commit to remove `ParsedArgumentExpression.Value`, which was no longer used, and was unsafe anyway.

PR #950, opened 2016-12-14
Have AppVeyor just build .NET Desktop version

blairconrad:
1. we don't need the netstd build, and
2. the netstd build appends `-adhoc` to the version number, so we'd've had to change _something_


thomaslevesque:
LGTM, thanks 👍

PR #953, opened 2016-12-23
Analyzer for VB.NET

### LINKED ISSUE #952: Add an analyzer that supports VB.NET
As discussed in https://github.com/FakeItEasy/FakeItEasy/issues/951#issuecomment-267312413

Currently the analyzer only supports C#, we should add one for VB.NET. Hopefully we can share part of the code between the two analyzers.

thomaslevesque:
Fixes #952

All the analyzer code is shared with the C# analyzer, using `#if`s.
The tests are the same, only with the subject code translated to VB.NET.


src/FakeItEasy.Analyzer/SymbolHelpers.cs
```diff
@@ -18,9 +24,12 @@ internal static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax c
...
 
-            return name == null
-                ? null
-                : context.SemanticModel.GetSymbolInfo(name).Symbol as IMethodSymbol;
+            var method =
+                name == null
+                    ? null
+                    : context.SemanticModel.GetSymbolInfo(name).Symbol as IMethodSymbol;
+
+            return method?.ReducedFrom ?? method;
```
thomaslevesque: This is the only actual code change I had to do to make it work in VB.NET. Basically, in Roslyn for VB.NET, a generic extension method is "reduced", i.e. `WhereConfigurationExtensions.Where<T>` becomes `WhereConfigurationExtensions.Where`, as if it were non-generic. It doesn't change anything for C#, where `ReducedFrom` returns null.

tests/FakeItEasy.Analyzer.Tests/NonVirtualSetupAnalyzerTests.cs
```diff
@@ -290,7 +290,7 @@ public class TheClass
         public void Test()
         {
             var foo = A.Fake<IFoo>();
-            A.CallTo(() => foo.Bar);
+            A.CallToSet(() => foo.Bar);
```
thomaslevesque: Unrelated to the PR, but I'm pretty sure it was a mistake. I can put it in a separate PR if you prefer.

tests/FakeItEasy.Analyzer.VisualBasic.Tests/FakeItEasy.Analyzer.VisualBasic.Tests.csproj
```diff
@@ -0,0 +1,179 @@
...
+      <Project>{0E26272D-045A-47C2-8928-BD9B980115D5}</Project>
+      <Name>FakeItEasy.Tests</Name>
+    </ProjectReference>
+  </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
+  <ItemGroup>
+    <Analyzer Include="..\..\packages\Microsoft.CodeAnalysis.Analyzers.1.1.0\analyzers\dotnet\cs\Microsoft.CodeAnalysis.Analyzers.dll" />
+    <Analyzer Include="..\..\packages\Microsoft.CodeAnalysis.Analyzers.1.1.0\analyzers\dotnet\cs\Microsoft.CodeAnalysis.CSharp.Analyzers.dll" />
```
thomaslevesque: This is not a mistake. Even though the project tests a VB.NET analyzer, the code is in C#, so it needs the C# analyzer. Actually, I'm not sure this analyzer (which is a "meta-analyzer" to ensure the analyzer is written correctly) is necessary, since this project doesn't *implement* an analyzer... I think it was there in the project template, and I just copied and edited the C# project. I'll see if it can be removed.

thomaslevesque: Actually, the package can't be removed, because other Microsoft.CodeAnalysis packages have a dependency on it.

thomaslevesque:
Haven't done it yet, but we might want to rename `FakeItEasy.Analyzer` to `FakeItEasy.Analyzer.CSharp`

blairconrad:
@thomaslevesque, I've not gotten through all the files, but it's looking good so far. I'll soldier on in my morning, as I find time. It's "bake for Christmas day", so it might be tough.

> Haven't done it yet, but we might want to rename FakeItEasy.Analyzer to FakeItEasy.Analyzer.CSharp

I was going suggest this as well.

tests/FakeItEasy.Analyzer.VisualBasic.Tests/NonVirtualSetupAnalyzerTests.cs
```diff
@@ -0,0 +1,378 @@
...
+End Namespace
+";
+
+            this.VerifyVisualBasicDiagnostic(
+                Test,
+                new DiagnosticResult
+                {
+                    Id = DiagnosticDefinitions.NonVirtualSetupSpecification.Id,
+                    Message =
+                        "Non virtual member 'Bar' cannot be intercepted.",
```
blairconrad: This should really be a comment on the resource that specifies the string, but a user writing in VB _may_ not be familiar with "Non virtual". I have not investigated the difficulty of doing this, but perhaps they'd prefer to hear about a "Non overridable"  member?

(Aside: I've never really looked at VB.NET before, and I'm perfectly comfortable with "abstract" and "virtual" and whatnot, but if I were a programming tyro, I think the VB syntax would be much more helpful. "Overridable" already means something to an English speaker, just like "Shared" comes closer to expressing the concept than "static". Ditto "MustOverride"/"abstract". We've been artificially raising the bar to entry! Because Java, I expect.)

thomaslevesque:
No rush, I won't be working on this again until after Christmas anyway.
Happy holidays!

Le 24 déc. 2016 05:02, "Blair Conrad" <notifications@github.com> a écrit :

@thomaslevesque <https://github.com/thomaslevesque>, I've not gotten
through all the files, but it's looking good so far. I'll soldier on in my
morning, as I find time. It's "bake for Christmas day", so it might be
tough.

Haven't done it yet, but we might want to rename FakeItEasy.Analyzer to
FakeItEasy.Analyzer.CSharp

I was going suggest this as well.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/FakeItEasy/FakeItEasy/pull/953#issuecomment-269067876>,
or mute the thread

tests/FakeItEasy.Analyzer.VisualBasic.Tests/NonVirtualSetupAnalyzerTests.cs
```diff
@@ -0,0 +1,378 @@
...
+        Function Bar() As Integer
+    End Interface
+End Namespace
+";
+
+            this.VerifyVisualBasicDiagnostic(Test);
+        }
+
+        [Fact]
+        public void Diagnostic_Should_Not_Be_Triggered_For_Virtual_Member()
```
adamralph: To continue the nitpickiness, capital letters and underscores seem to be competing here for the job of distinguishing words. Shouldn't we just use one of them? (I vote for capital letters.)

However, I also agree that this should be a separate commit adjusting both sets of tests. Or even  a separate PR.

tests/FakeItEasy.Analyzer.VisualBasic.Tests/NonVirtualSetupAnalyzerTests.cs
adamralph: Fora VB analyzer, I think we should really be using the VB terms.

adamralph:
This might be dumb comment, but can we provide a single analyzer package that does both C# and VB? I'm looking at `Microsoft.CodeAnalysis.Analyzers` which I presume does both?

adamralph:
Ignore my last comment. Now that I looked more carefully, I can see that it is indeed one package.

tests/FakeItEasy.Analyzer.VisualBasic.Tests/NonVirtualSetupAnalyzerTests.cs
thomaslevesque: > but in the user's language, we're really talking about an Overridable member. Should the test name reflect that?

Good point. I blindly copied and adapted the C# tests, but I can rename them.



> To continue the nitpickiness, capital letters and underscores seem to be competing here for the job of distinguishing words. Shouldn't we just use one of them? (I vote for capital letters.)

IMO underscores make it a little easier to read the test name as a sentence, but I don't have strong feelings either way. Whichever option we choose, I'd rather do it in a separate PR.

tests/FakeItEasy.Analyzer.VisualBasic.Tests/NonVirtualSetupAnalyzerTests.cs
```diff
@@ -0,0 +1,378 @@
...
+            this.VerifyVisualBasicDiagnostic(Test);
+        }
+
+        [Fact]
+        public void Diagnostic_Should_Not_Be_Triggered_For_Non_Sealed_Override_Method()
+        {
+            const string Test = @"Namespace FakeItEasy.Analyzer.VisualBasic.Tests
+    Class TheClass
+        Sub Test()
+            Dim foo = A.Fake(Of Foo)()
```
thomaslevesque: Good catch!

thomaslevesque:
I made a few changes:

- fixed the minor remarks by amending (comment in `SymbolHelpers.GetCalledMethodSymbol`, `Foo2` instead of `Foo`)
- moved the fix for the `CallToSet` test to a separate commit
- added specific resources for VB.NET diagnostic messages
- changed the wording of the VB.NET tests to use VB.NET terminology

Now I'll rename the `FakeItEasy.Analyzer` project to `FakeItEasy.Analyzer.CSharp`

thomaslevesque:
I made the rename, but it makes things a little weird... I wonder if I should put the common files in their own folder (or maybe even a shared project).

tests/FakeItEasy.Analyzer.VisualBasic.Tests/NonVirtualSetupAnalyzerTests.cs
blairconrad: I'm content with either snake case or Pascal. I had noticed the combo, but figured I'd hold off commenting because you were just keeping the same convention that was in use, and that as has been said twice already, maybe another PR would be a better place to fix.

blairconrad:
> I made the rename, but it makes things a little weird... I wonder if I should put the common files in their own folder (or maybe even a shared project).

It would look slightly less odd.

blairconrad:
@thomaslevesque, aside from the odd naming that you pointed out, I think everything looks good. Thanks!
I'd be content with a squash of everything but

* Fix analyzer test on CallToSet and
* Rename FakeItEasy.Analyzer to FakeItEasy.Analyzer.CSharp (which still maybe in flux, I understand)

adamralph:
What @blairconrad said.

thomaslevesque:
> I wonder if I should put the common files in their own folder (or maybe even a shared project)

Would you prefer just a folder with the common files, or a shared project? We stayed away from shared projects so far, not sure why. It makes things a bit easier because the files don't need to be added to each project separately.

thomaslevesque:
Also, in the long run we might want to switch to shared projects for the main library (after the switch from project.json to csproj)

adamralph:
I think we stayed away from shared projects because the tooling around them was horribly broken initially, but AFAIK it's been fixed now.

In the long run, we won't need a shared project for the main lib because project.json/new .csproj will do the multi-targetting for us, so we'll build for all platforms from a single project.

thomaslevesque:
> I think we stayed away from shared projects because the tooling around them was horribly broken initially, but AFAIK it's been fixed now.

Yes, I've been using them for more than two years at the day job, and it works fine (R# still gets a little confused sometimes, but it's getting better)

> In the long run, we won't need a shared project for the main lib because project.json/new .csproj will do the multi-targetting for us, so we'll build for all platforms from a single project.

Ah, yes. Good point.

thomaslevesque:
All done

blairconrad:
@thomaslevesque, looks good to me. Thanks! I do enjoy the lack of repetition in the csprojs afforded by the shares project.

src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.projitems
```diff
@@ -0,0 +1,26 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+    <HasSharedItems>true</HasSharedItems>
+    <SharedGUID>c4766c64-39e1-4023-a173-615607c40ad5</SharedGUID>
+  </PropertyGroup>
+  <PropertyGroup Label="Configuration">
+    <Import_RootNamespace>FakeItEasy.Analyzer.Shared</Import_RootNamespace>
```
adamralph: Is this correct? I noticed that in the language specific projects, the namespace we are using is still `FakeItEasy.Analyzer`.

src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.shproj
```diff
@@ -0,0 +1,13 @@
...
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>c4766c64-39e1-4023-a173-615607c40ad5</ProjectGuid>
+    <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+  </PropertyGroup>
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.Default.props" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.props" />
+  <PropertyGroup />
+  <Import Project="FakeItEasy.Analyzer.Shared.projitems" Label="Shared" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.VisualBasic.targets" />
```
adamralph: Is there a known reason that there isn't a C# equivalent of this targets file, or should we just regard this as an artifact of the mysterious intimacy of Visual Studio and MSBuild?

src/FakeItEasy.Analyzer.Shared/SymbolHelpers.cs
```diff
@@ -18,9 +24,15 @@ internal static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax c
...
 
-            return name == null
-                ? null
-                : context.SemanticModel.GetSymbolInfo(name).Symbol as IMethodSymbol;
+            var method =
+                name == null
+                    ? null
+                    : context.SemanticModel.GetSymbolInfo(name).Symbol as IMethodSymbol;
+
+            // In Roslyn for VB.NET, some generic extension method are "reduced", i.e. WhereConfigurationExtensions.Where<T>
```
adamralph: nitpick: plural "methods"

tests/FakeItEasy.Analyzer.Tests.Shared/FakeItEasy.Analyzer.Tests.Shared.projitems
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
+    <HasSharedItems>true</HasSharedItems>
+    <SharedGUID>d74665d8-d046-4c38-81d6-8aa7ead68468</SharedGUID>
+  </PropertyGroup>
+  <PropertyGroup Label="Configuration">
+    <Import_RootNamespace>FakeItEasy.Analyzer.Tests.Shared</Import_RootNamespace>
```
adamralph: Similar comment here. Naively, I'd say "Shared" should be dropped from the default namespace.

tests/FakeItEasy.Analyzer.Tests.Shared/FakeItEasy.Analyzer.Tests.Shared.shproj
```diff
@@ -0,0 +1,13 @@
...
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>d74665d8-d046-4c38-81d6-8aa7ead68468</ProjectGuid>
+    <MinimumVisualStudioVersion>14.0</MinimumVisualStudioVersion>
+  </PropertyGroup>
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.Default.props" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.props" />
+  <PropertyGroup />
+  <Import Project="FakeItEasy.Analyzer.Tests.Shared.projitems" Label="Shared" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.VisualBasic.targets" />
```
adamralph: Same comment here regarding an equivalent C# targets file.

src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.projitems
thomaslevesque: No, it's not correct. It should be `FakeItEasy.Analyzer`.

src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.shproj
thomaslevesque: No idea. I just created the shared project from the template, I didn't really look into it.

src/FakeItEasy.Analyzer.Shared/SymbolHelpers.cs
thomaslevesque: oops

tests/FakeItEasy.Analyzer.Tests.Shared/FakeItEasy.Analyzer.Tests.Shared.projitems
thomaslevesque: Agreed

tests/FakeItEasy.Analyzer.Tests.Shared/FakeItEasy.Analyzer.Tests.Shared.shproj
thomaslevesque: Still no idea ;)

src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.shproj
thomaslevesque: I tried to add the CSharp targets file: the project could no longer be loaded (to be exact, it remained in "loading" state forever)

src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.projitems
thomaslevesque: Fixed

src/FakeItEasy.Analyzer.Shared/SymbolHelpers.cs
thomaslevesque: Fixed

thomaslevesque:
Thanks for the merge!

PR #954, opened 2016-12-27
Fix integration tests for external assembly

thomaslevesque:
Some tests were failing when run from ReSharper. I think this is because it shadow copies assemblies, so the `Location` isn't the original location. Using `CodeBase` instead fixes it.

### DIFF
--- a/tests/FakeItEasy.IntegrationTests/TypeCatalogueTests.cs
+++ b/tests/FakeItEasy.IntegrationTests/TypeCatalogueTests.cs
@@ -150,7 +150,7 @@ namespace FakeItEasy.IntegrationTests
 
         private string GetPathToOriginalExternalDll()
         {
-            var executingAssemblyPath = new Uri(this.GetType().GetTypeInfo().Assembly.Location).LocalPath
+            var executingAssemblyPath = new Uri(this.GetType().GetTypeInfo().Assembly.CodeBase).LocalPath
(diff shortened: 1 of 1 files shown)


blairconrad:
Ah. I tend to disable shadow copying, if only because of the performance hit. Thanks for noticing and fixing, @thomaslevesque!

tests/FakeItEasy.IntegrationTests/TypeCatalogueTests.cs
```diff
@@ -150,7 +150,7 @@ private static string CaptureConsoleOutput(Action action)
 
         private string GetPathToOriginalExternalDll()
         {
-            var executingAssemblyPath = new Uri(this.GetType().GetTypeInfo().Assembly.Location).LocalPath
+            var executingAssemblyPath = new Uri(this.GetType().GetTypeInfo().Assembly.CodeBase).LocalPath
```
blairconrad: I had been wondering about line 161 (as I'm sure you have already). At first I thought it weird to still be `.Location`, but I it's the copied that fails to load. Okay. I got my head around it. Thanks!

thomaslevesque: To be honest, I had missed the `.Location` on line 161 :wink:
But after looking at it, I think it should stay as is.

blairconrad:
Thanks, @thomaslevesque!

PR #956, opened 2016-12-28
Ignore all per-user project settings files

thomaslevesque:
The `.user` files for the new shared projects (`.shproj`) weren't ignored; I updated the `.gitignore` to ignore all `*.*proj.user` files.


adamralph:
Thanks @thomaslevesque

PR #958, opened 2017-01-01
Fix stack overflow in CallToSet configuration

### LINKED ISSUE #957: A.CallToSet(...).Throws(...) causes a stack overflow
Observed on 3.0.0-beta002

Can be reproduced with this code:

```c#
void Main()
{
    var foo = A.Fake<IFoo>();
    A.CallToSet(() => foo.Bar).Throws(new Exception("oops"));
}

public interface IFoo
{
    int Bar { get; set; }
}
```

Current stable release (2.3.2) is not affected; I suspect the bug was introduced by the refactoring of configuration interfaces in #804.

thomaslevesque:
Fixes #957


thomaslevesque:
Thanks for the merge!

PR #961, opened 2017-01-03
Change behavior of auto properties to not override explicit configuration

### LINKED ISSUE #962: Default behavior of read/write properties should not override explicit getter or setter configuration
(Spawned from a [discussion on the chat](https://gitter.im/FakeItEasy?at=586a3b787a3f79ef5dc86a23))

I recently noticed a surprising behavior for faked read/write properties. [As documented](http://fakeiteasy.readthedocs.io/en/stable/read-write-property-behavior/), setting the value of an unconfigured property configures it to return the assigned value. But when you do that, **it also overrides any configuration made on the setter**. The same is true when you read an unconfigured property.

Neither of these code snippets throw an exception:
```csharp
    var foo = A.Fake<IFoo>();
    A.CallToSet(() => foo.Bar).To(() => A<int>.That.Not.IsEqualTo(42)).Throws(new Exception("wrong answer"));
    foo.Bar = 42; // previously configured behavior is overridden here
    foo.Bar.Dump();
    foo.Bar = 123; // should throw here, but doesn't
    foo.Bar.Dump();
```
```csharp
    var foo = A.Fake<IFoo>();
    A.CallToSet(() => foo.Bar).To(() => A<int>.That.Not.IsEqualTo(42)).Throws(new Exception("wrong answer"));
    foo.Bar.Dump(); // previously configured behavior is overridden here
    foo.Bar = 123; // should throw here, but doesn't
    foo.Bar.Dump();
```

We should probably consider changing this behavior, as it is counter-intuitive.

thomaslevesque:
This PR is an experiment to demonstrate a new behavior for "auto properties", as discussed in the chat.

Fixes #962

### DIFF
--- a/docs/read-write-property-behavior.md
+++ b/docs/read-write-property-behavior.md
@@ -1,24 +1,3 @@
-# Read/Write Property Behavior
-
-Although you can explicitly
-[specify the return value for a called property getter](specifying-return-values.md),
-there's an easier, more intuitive way to work with read/write
-properties.  By default, any
-[fakeable](what-can-be-faked.md#what-members-can-be-overridden) property
-that has both a `set` and `get` accessor behaves like you might
-expect. Setting a value and then getting the value returns the value
-that was set.
-
-```csharp
-var fakeShop = A.Fake<ICandyShop>();
-
-fakeShop.Address = "123 Fake Street";
-
-System.Console.Out.Write(fakeShop.Address); 
-// prints "123 Fake Street"
-```
-
-This behaviour can be used to 
-
-* supply values for the system under test to use (via the getter) or to
-* verify that the system under test performed the `set` action on the Fake
+The contents of this page have moved. See
+* [Default fake behavior](default-fake-behavior.md) for information about unconfigured properties and
+* [Specifying a Call to Configure](specifying-a-call-to-configure.md) for how to configure a property.
--- a/src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
@@ -10,13 +10,11 @@ namespace FakeItEasy.Core
         private class PropertyBehaviorRule
             : IFakeObjectCallRule
         {
-            private readonly FakeManager fakeManager;
             private readonly MethodInfo propertyGetter;
             private readonly MethodInfo propertySetter;
 
-            public PropertyBehaviorRule(MethodInfo propertyGetterOrSetter, FakeManager fakeManager)
+            public PropertyBehaviorRule(MethodInfo propertyGetterOrSetter)
             {
-                this.fakeManager = fakeManager;
                 var property = GetProperty(propertyGetterOrSetter);
 
                 this.propertySetter = property.GetSetMethod();
@@ -39,25 +37,18 @@ namespace FakeItEasy.Core
                 return method.IsSpecialName && method.Name.StartsWith("get_", StringComparison.Ordinal);
             }
 
+            public bool IsMatchForSetter(IFakeObjectCall fakeObjectCall) => this.IsPropertySetter(fakeObjectCall);
+
             public bool IsApplicableTo(IFakeObjectCall fakeObjectCall)
             {
-                return this.IsPropertySetter(fakeObjectCall) || this.IsPropertyGetter(fakeObjectCall);
+                return this.IsPropertyGetter(fakeObjectCall);
             }
 
             public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
             {
                 Guard.AgainstNull(fakeObjectCall, nameof(fakeObjectCall));
 
-                if (this.IsPropertyGetter(fakeObjectCall))
-                {
-                    fakeObjectCall.SetReturnValue(this.Value);
-                }
-                else
-                {
-                    this.Value = fakeObjectCall.Arguments.Last();
-                }
-
-                this.fakeManager.MoveRuleToFront(this);
+                fakeObjectCall.SetReturnValue(this.Value);
             }
 
             private static PropertyInfo GetProperty(MethodInfo propertyGetterOrSetter)
--- a/src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
@@ -32,17 +32,33 @@ namespace FakeItEasy.Core
             {
                 Guard.AgainstNull(fakeObjectCall, nameof(fakeObjectCall));
 
-                var newRule = new CallRuleMetadata
-                                  {
-                                      CalledNumberOfTimes = 1,
-                                      Rule = new PropertyBehaviorRule(fakeObjectCall.Method, this.fakeManager)
-                                                 {
-                                                     Indices = fakeObjectCall.Arguments.Take(fakeObjectCall.Arguments.Count - 1).ToArray(),
-                                                     Value = fakeObjectCall.Arguments.Last()
-                                                 }
-                                  };
+                // Setting the property adds a PropertyBehaviorRule to store the assigned value.
+                // The PropertyBehaviorRule is added at the end of user rules to avoid overriding
+                // explicit configurations of the property made with CallTo.
+                var existingRule = this.fakeManager
+                    .AllUserRules
+                    .Select(metadata => metadata.Rule)
+                    .OfType<PropertyBehaviorRule>()
+                    .LastOrDefault(rule => rule.IsMatchForSetter(fakeObjectCall));
 
-                this.fakeManager.AllUserRules.AddFirst(newRule);
+                if (existingRule != null)
+                {
+                    existingRule.Value = fakeObjectCall.Arguments.Last();
+                }
+                else
+                {
+                    var newRule = new CallRuleMetadata
+                    {
+                        CalledNumberOfTimes = 1,
+                        Rule = new PropertyBehaviorRule(fakeObjectCall.Method)
+                        {
+                            Indices = fakeObjectCall.Arguments.Take(fakeObjectCall.Arguments.Count - 1).ToArray(),
+                            Value = fakeObjectCall.Arguments.Last()
+                        }
+                    };
+
+                    this.fakeManager.AllUserRules.AddLast(newRule);
+                }
             }
         }
     }
--- a/docs/default-fake-behavior.md
+++ b/docs/default-fake-behavior.md
@@ -28,6 +28,27 @@ the member:
   concrete, pre-existing type;
 * otherwise, `default(T)` will be returned.
 
+### Read/write properties
+
+By default, any overrideable property that has both a `set` and `get` accessor,
+and that has not been explicitly configured to behave otherwise, behaves like
+you might expect. Setting a value and then getting the value returns the value
+that was set.
+
+```csharp
+var fakeShop = A.Fake<ICandyShop>();
+
+fakeShop.Address = "123 Fake Street";
+
+System.Console.Out.Write(fakeShop.Address);
+// prints "123 Fake Street"
+```
+
+This behavior can be used to
+
+* supply values for the system under test to use (via the getter) or to
+* verify that the system under test performed the `set` action on the Fake
+
 ### Cancellation tokens
 
 When a faked method that accepts a `CancellationToken` receives a canceled token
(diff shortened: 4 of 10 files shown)


src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
```diff
@@ -39,25 +39,18 @@ public static bool IsPropertyGetter(MethodInfo method)
                 return method.IsSpecialName && method.Name.StartsWith("get_", StringComparison.Ordinal);
             }
 
+            public bool IsMatchForSetter(IFakeObjectCall fakeObjectCall) => IsPropertySetter(fakeObjectCall);
```
blairconrad: is the `this` still required (encouraged)?

blairconrad: I'm not really sure why we need this method as well as `IsPropertySetter`

src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
```diff
@@ -32,17 +32,31 @@ public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
...
-                                                 }
-                                  };
-
-                this.fakeManager.AllUserRules.AddFirst(newRule);
+                var existingRule = this.fakeManager
+                    .AllUserRules
+                    .Reverse()
+                    .Select(metadata => metadata.Rule)
+                    .OfType<PropertyBehaviorRule>()
+                    .FirstOrDefault(rule => rule.IsMatchForSetter(fakeObjectCall));
```
blairconrad: it took me a while to follow all the reverses and adding to last. I believe it's correct, but find it hard to follow.

src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
thomaslevesque: > is the this still required (encouraged)?

Yes. Not sure why StyleCop doesn't detect it. Perhaps it's confused by the expression-bodied method.

> I'm not really sure why we need this method as well as IsPropertySetter

We don't really need it, but the `IsPropertySetter` instance method is private, and if I made it public, it would be strange if `IsPropertyGetter` were not also public, although it's not needed. Also, I thought `IsMatchForSetter` communicated the intent better.

thomaslevesque: But I can change it if you prefer.

src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
thomaslevesque: Yes, it's just that I wanted to search from the end. But now that I think about it, I could use `LastOrDefault` instead.

src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
blairconrad: I guess `IsMatchForSetter` makes it sound to me like we're comparing the rule against a particular setter, but perhaps that's just my brain making assumptions.
This is not such a big deal to me. I appreciate the the explanation.

src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
blairconrad: Oh, yes. `LastOrDefault` and no `Reverse` would be better. And perhaps a comment at the insertion site to indicate why we're shoving the rule at the end instead of the beginning like all the other rules?

thomaslevesque: Sure, I'll add an explanation in the comments.

src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
thomaslevesque: If you have a better name to suggest, I'll gladly take it. I didn't think too long about this one.

blairconrad: I don't think it's worth renaming.
The only alternative     I can think of that I like is `IsPropertySetter`. :wink:

thomaslevesque:
I think I made all the requested changes. I'll squash everything before merging.

src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
```diff
@@ -32,17 +32,34 @@ public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
...
-                                                 {
-                                                     Indices = fakeObjectCall.Arguments.Take(fakeObjectCall.Arguments.Count - 1).ToArray(),
-                                                     Value = fakeObjectCall.Arguments.Last()
-                                                 }
-                                  };
-
-                this.fakeManager.AllUserRules.AddFirst(newRule);
+                //// Setting the property adds a PropertyBehaviorRule to store the assigned value
+                //// The PropertyBehaviorRule is added at the end of user rules to avoid overriding
+                //// explicit configurations of the property made with CallTo.
```
blairconrad: Such a nitpick, but the `////` looks funny to me. This is the format that StyleCop suggest for commenting out code. How do you feel about just `//`?

(And while I'm here, I think you forgot a "." at the end of the first line.)

thomaslevesque:
> In #962 (comment), you'd mentioned that you were going to add documentation. Was that the docs in the code, or did you mean something to go on readthedocs to let users know about the new behaviour?

Oops! No, I intended to update the docs, not just add comments to the code.

thomaslevesque:
I updated the docs. I'm a bit uneasy about what I wrote, though, because it's not strictly true... Setting a property whose getter has been explicitly configured actually *does* something, but the effect won't be visible while the getter configuration is active (see the second snippet in https://github.com/FakeItEasy/FakeItEasy/issues/962#issuecomment-270382936). But I'm afraid that trying to explain this will confuse the reader even more... What do you think?

docs/read-write-property-behavior.md
```diff
@@ -14,11 +14,28 @@ var fakeShop = A.Fake<ICandyShop>();
 
 fakeShop.Address = "123 Fake Street";
 
-System.Console.Out.Write(fakeShop.Address); 
+System.Console.Out.Write(fakeShop.Address);
 // prints "123 Fake Street"
 ```
 
-This behaviour can be used to 
+This behavior can be used to
```
thomaslevesque: For consistency with the rest of the documents.

We have 11 occurrences of "behaviour" (all in the docs), and 271 occurrences of "behavior" (35 of them in the docs)...

blairconrad:
Sorry for the delay, @thomaslevesque. Truth is, I'd forgotten about your update. 😊
Nice catch on the "behaviour".

I understand your concern about lying in the docs, but sometimes it's better to lie about what's happening internally if it better explains the view from the outside.
I'd support merging the text you have, but want to present an alternative idea.
We could, instead of appending your explanation, at the top of the topic indicate that the behavio(u)r we describe applies only to properties that are not otherwise configured as in [Specifying Return Values](http://fakeiteasy.readthedocs.io/en/stable/specifying-return-values/), [Throwing Exceptions](http://fakeiteasy.readthedocs.io/en/stable/throwing-exceptions/) , etc., as you've indicated in your bold note.
Does that approach appeal at all?
Of course if it did, the topic would probably be better named "Default Read/Write Property Behavior". Or we could even roll it into [Default fake behavior](http://fakeiteasy.readthedocs.io/en/stable/default-fake-behavior/).

thomaslevesque:
@blairconrad I like your idea, I'll give it a try.

thomaslevesque:
> Or we could even roll it into Default fake behavior.

I think I'd rather keep it separate, but we can add a link to it in Default fake behavior.

blairconrad:
> I think I'd rather keep it separate, but we can add a link to it in Default fake behavior.

Yes. This is better.

thomaslevesque:
OK, I tried to apply your idea, but I'm not sure I like it better. I think it's useful to have an example of what happens when the default behavior no longer applies, but that example has to come *after* we explained the default behavior.

blairconrad:
Thanks, @thomaslevesque. I agree with you. As it is, I don't like it better either. If we're going to document "default read/write property behavior", I think it belongs in the docs nearer (or maybe "in" is a better approach after all) the [Default fake behavior](http://fakeiteasy.readthedocs.io/en/stable/default-fake-behavior/) topic, which supports your idea of explaining the default behavior before explaining what happens when we explicitly configure the property.

thomaslevesque:
> If we're going to document "default read/write property behavior", I think it belongs in the docs nearer (or maybe "in" is a better approach after all) the Default fake behavior topic, which supports your idea of explaining the default behavior before explaining what happens when we explicitly configure the property.

I'm not sure I understand; what are you suggesting?

blairconrad:
> I'm not sure I understand; what are you suggesting?

Move the topic for the default read/write property behaviour into [Default fake behavior](http://fakeiteasy.readthedocs.io/en/stable/default-fake-behavior/).
(Alternatively, keep the topic separate, but move it up next to "Default fake behavior" in the table of contents, but I think inlining it is better.)

Then, as you suggested, people see the default behaviour first, and then later on in the docs we talk about how to override the getter and setter with fancier behaviour.
I quick roughed something up for you to get the sense of it, in case I'm still not being clear: https://github.com/blairconrad/FakeItEasy/commit/40070b0c7f2e029a69772ef77154debf27490475

(Also, I accidentally pushed a commit to _your_ branch using that scary new GitHub "push to other people's branches" technology, but was able to reset the head of the branch and repush. Which is even scarier! I hope I didn't hurt anything.)

thomaslevesque:
Ah, I see. I wasn't completely awake when I posted this morning... Your changes look pretty good!

> (Also, I accidentally pushed a commit to your branch using that scary new GitHub "push to other people's branches" technology, but was able to reset the head of the branch and repush. Which is even scarier! I hope I didn't hurt anything.)

Oh, I don't mind you pushing to my branch. In fact, I think I'm going to merge your branch into mine, which will have the same result :wink:

thomaslevesque:
I took your commit, fixed the line endings, added back `read-write-property-behavior.md`, squashed the implementation commits together, and squashed the documentation commits together (attributed to you, since none of what I had done remained ;))

blairconrad:
>Oh, I don't mind you pushing to my branch.

Generous of you. I feel likewise.
My main concern was how easy it was for me to _accidentally_ do it!
In retrospect it makes sense. I checked out your branch and tracked it, so when I pushed without specifying a remote…

> added back `read-write-property-behavior.md`

Thanks. I'm surprised the line endings changed. I'd told emacs to visit the file "literally", which I think is supposed to leave _all_ the bytes alone. Ah well.

Also thanks for the attribution, not that it matters. It's all about the result!

blairconrad:
It looks good to me, @thomaslevesque. Thanks for the work.
Leaving for a bit again in case @adamralph wants to weigh in on the changed behavio(u)r.

thomaslevesque:
@adamralph, you're right, there should be more tests to cover the new behavior. I'll add them.

> at the black box level

What do you mean by that? specs?

adamralph:
@thomaslevesque yep, specs is what I meant. Would that be appropriate for this?

thomaslevesque:
> Would that be appropriate for this?

I think so

thomaslevesque:
I added some specs. The read/write property behavior was more or less covered in several spec classes, but none of them was specifically testing it, so I introduced a new spec class.

src/FakeItEasy/Core/FakeManager.PropertySetterRule.cs
adamralph: I guess StyleCop is complaining about the blank line after the comment when it only has `//`.

What does the comment apply to? The next line of code or the whole method? (This is exactly the ambiguity that StyleCop does't like.) If it's the next line, then the line break should be removed. If it's the whole method, then it should move out to above the method signature.

thomaslevesque: The comment applies to the next statement (the LINQ query).

I'm a bit confused... I *think* I had introduced the blank line and the `////` to make StyleCop happy, but there's no warning with `//` and no blank line, so I'll do just that.

Also, it seems I had missed @blairconrad's comment...

thomaslevesque:
I edited the comment in `PropertySetterRule` and fixed some new FxCop warnings (amended the commit).

adamralph:
Thanks @thomaslevesque !

thomaslevesque:
Thanks for the merge

PR #963, opened 2017-01-08
Add explicit dependency to fix build issue

thomaslevesque:
FakeItEasy.IntegrationTests.External.netstd sometimes fails to build when using the Rebuild command in Visual Studio. MSBuild tries to build it too early, before FakeItEasy.netstd has been built. Adding an explicit build dependency in the solution fixes the problem.


blairconrad:
Thanks, @thomaslevesque. I'm sad (and surprised) that the dependency in project.json doesn't take care of this, but glad to have the fix!

PR #967, opened 2017-01-16
Make WithNonVoidReturnType not match void methods

### LINKED ISSUE #966: WithNonVoidReturnType matches methods with void return type
Specifically, this fails:

```c#

public interface IFoo
{
    void Baz();
}

[Scenario]
public static void WithNonVoidReturnTypeAndVoidMethod(
    IFoo fake,
    bool methodWasCalled)
{
    "Given a fake with a void method"

    "When the fake is configured to set a flag for any non-void return type"
        .x(() => A.CallTo(fake).WithNonVoidReturnType().Invokes(() => methodWasCalled = true));

    "And the void method is called"
        .x(() => fake.Baz());

    "Then the flag is not set"
        .x(() => methodWasCalled.Should().BeFalse());
}
```

Found while working on external Self-Initializing Fakes project.

blairconrad:
Fixes #966.

### DIFF
--- a/src/FakeItEasy/Configuration/AnyCallCallRule.cs
+++ b/src/FakeItEasy/Configuration/AnyCallCallRule.cs
@@ -39,9 +39,9 @@ namespace FakeItEasy.Configuration
         {
             return
                 this.argumentsPredicate(fakeObjectCall.Arguments) &&
-                    (this.ApplicableToMembersWithReturnType == null
-                    || (this.ApplicableToMembersWithReturnType == fakeObjectCall.Method.ReturnType)
-                    || (this.ApplicableToAllNonVoidReturnTypes && fakeObjectCall.Method.ReturnType != typeof(void)));
+                ((this.ApplicableToMembersWithReturnType == null && !this.ApplicableToAllNonVoidReturnTypes)
+                 || (this.ApplicableToMembersWithReturnType == fakeObjectCall.Method.ReturnType)
+                 || (this.ApplicableToAllNonVoidReturnTypes && fakeObjectCall.Method.ReturnType != typeof(void)));
         }
 
         protected override BuildableCallRule CloneCallSpecificationCore() =>
--- a/tests/FakeItEasy.Specs/AnyCallConfigurationSpecs.cs
+++ b/tests/FakeItEasy.Specs/AnyCallConfigurationSpecs.cs
@@ -1,5 +1,6 @@
 ﻿namespace FakeItEasy.Specs
 {
+    using System;
     using FluentAssertions;
     using Xbehave;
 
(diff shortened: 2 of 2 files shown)


src/FakeItEasy/Configuration/AnyCallCallRule.cs
```diff
@@ -39,9 +39,9 @@ protected override bool OnIsApplicableTo(IFakeObjectCall fakeObjectCall)
         {
             return
                 this.argumentsPredicate(fakeObjectCall.Arguments) &&
-                    (this.ApplicableToMembersWithReturnType == null
-                    || (this.ApplicableToMembersWithReturnType == fakeObjectCall.Method.ReturnType)
-                    || (this.ApplicableToAllNonVoidReturnTypes && fakeObjectCall.Method.ReturnType != typeof(void)));
+                ((this.ApplicableToMembersWithReturnType == null && !this.ApplicableToAllNonVoidReturnTypes)
+                 || (this.ApplicableToMembersWithReturnType == fakeObjectCall.Method.ReturnType)
+                 || (this.ApplicableToAllNonVoidReturnTypes && fakeObjectCall.Method.ReturnType != typeof(void)));
```
thomaslevesque: This should work, but I'm worried this statement is becoming hard to understand. Maybe it would be better to change it to something like this:
```
if (this.argumentsPredicate(fakeObjectCall.Arguments))
{
    if (ApplicableToAllNonVoidReturnTypes)
    {
        return fakeObjectCall.Method.ReturnType != typeof(void);
    }
    else if (ApplicableToMembersWithReturnType != null)
    {
        return this.ApplicableToMembersWithReturnType == fakeObjectCall.Method.ReturnType;
    }
    else
    {
        return true;
    }
}
return false;
```

blairconrad:
Thanks for the comments, @thomaslevesque!

> Finally, shouldn't this fix be merged to support-2 rather than master? (as discussed on Gitter)

Oh, right. Fell out of my head. Probably because we do it differently at work.
(I'm not trying to change FIE practices, but I think that fixing in master makes more sense. The thinking is this: there's a bug that needs fixing. Wherever we end up fixing it, we'll want it to continue being fixed in the future, so it has to make it into `master` at some point. By fixing it there first, we ensure that the fix will persist, and that we won't have a situation where we end up putting a fix in an earlier version and then it's "lost" when users upgrade. Also, we have the option of leaving it broken in earlier releases, if we so choose. That last bit applies more to the work environment where (for us) every extra branch that gets the fix (and we maintain a number) means quite a bit of extra work for development, testing, documentation, project management, and so on. Here, it's just double work for the primary developer and a reviewer or two, and I think we're more than happy to repeat the work.)

I'll submit a new PR against support-2, incorporating all your suggestions and referencing this PR.

thomaslevesque:
> Probably because we do it differently at work.

Yes, me too... but I already made the mistake last time, so I knew to look out for this ;)

PR #971, opened 2017-01-17
Keep WithNonVoidReturnType from matching void

blairconrad:
An evolution of #967, but targeting our 2.x stream.

I've incorporated @thomaslevesque's suggestions from #967, specifically adding a commit to ensure that `WithNonVoidReturnType` rules describe themselves properly, and making `OnIsApplicableTo` more readable (I deviated a little from the suggested implementation, as I prefer early returns.)


blairconrad:
There we go, @thomaslevesque. I'm happy to remove the early return in `OnIsApplicableTo`, if you prefer. Or to squash anything you want, or whatever.

src/FakeItEasy/Configuration/AnyCallCallRule.cs
```diff
@@ -37,11 +37,22 @@ public override void UsePredicateToValidateArguments(Func<ArgumentCollection, bo
...
+
+            if (this.ApplicableToAllNonVoidReturnTypes)
+            {
+                return fakeObjectCall.Method.ReturnType != typeof(void);
+            }
+
+            if (this.ApplicableToMembersWithReturnType != null)
+            {
+                return this.ApplicableToMembersWithReturnType == fakeObjectCall.Method.ReturnType;
+            }
```
thomaslevesque: I'm wondering... can we have a situation where both `ApplicableToAllNonVoidReturnTypes` and `ApplicableToMembersWithReturnType != null` are true? I don't think so, but assuming it's possible, shouldn't we check `ApplicableToMembersWithReturnType` first, as it's more specific?

src/FakeItEasy/Configuration/AnyCallCallRule.cs
```diff
@@ -21,6 +21,11 @@ public override string DescriptionOfValidCall
         {
             get
             {
+                if (this.ApplicableToAllNonVoidReturnTypes)
+                {
+                    return "Any call with non-void return type to the fake object.";
+                }
```
thomaslevesque: Same concern here as in `OnIsApplicableTo`

src/FakeItEasy/Configuration/AnyCallCallRule.cs
blairconrad: I don't think that we can be in such a situation, but I like your idea to change the order of the checks. Fixup commits incoming.

thomaslevesque:
LGTM! I'll merge after you squash the fixup commits

blairconrad:
Squashed and pushed. Thanks, @thomaslevesque, vastly improving the content.

thomaslevesque:
Thanks @blairconrad !

PR #972, opened 2017-01-18
[WIP] Merge 2.3.3

blairconrad:
One possible way to merge the recent changes for 2.3.3.
I created a merge branch (merge-2.3.3) and merged 2.3.3 into it.
This is the result.
There was  version conflict, so I excluded CommonAssemblyInfo.cs.

If we don't like this approach, I could cherry-pick 7b210cafd912e81c12c5b77009a95cec40f3c0ba instead.


blairconrad:
Sadly, we're picking up changes from even before dbb80f3704c25e8385cfd2a36f1be705c813e4df. This may not be the way to go.

thomaslevesque:
With our current branching model, I don't see any benefit in merging like this rather than just cherry-picking the commits we need. This PR contains many commits that have no reason to be there (well, I'm sure there's a good *technical* reason, but conceptually, we just want to port a specific fix to master, the other commits are just noise). There will always be changes on `support-2` that we don't want on `master` (e.g. bumping version to 2.3.3). Maybe it would be different if we were using Git Flow, but in the current situation, I think cherry-picking is preferable.

blairconrad:
I concur. I know @adamralph is very keen on being able to follow the merge lines to find the hotfix changes, but I'm content with a cherry-pick. Thanks!

thomaslevesque:
> I know @adamralph is very keen on being able to follow the merge lines to find the hotfix changes

And we should be able to when we switch to gitflow ;)

PR #974, opened 2017-01-24
Upgrade to Castle.Core 4.0.0

thomaslevesque:

PR #977, opened 2017-02-08
Fix nuspec castle version

thomaslevesque:
Forgot the nuspec in #974...

Also, the `-adhoc` default version suffix was only necessary when the Castle.Core dependency was a pre-release version, so I removed it (I set the VERSION_SUFFIX to `-rc001` in AppVeyor)


blairconrad:
How embarrassing. I should've noticed the nuspec as well.

blairconrad:
Thanks, @thomaslevesque!

PR #979, opened 2017-02-10
Implement Fake.ClearRecordedCalls

### LINKED ISSUE #978: Make it possible to reset recorded calls on an object
It is sometimes useful (especially for integration testing) to have a non-trivial setup phase that might call methods on fakes, but then you want to exclude those calls from the assertions.

I propose adding the API

```c#
Fake.ClearRecordedCalls(instance);
```

So you can do things like

```c#
public void MyTest() {
    // Arrange
    var service = A.Fake<IMyService>();
    var sut = new MyClass(service);
    sut.DoSomeInitialization();
    Fake.ClearRecordedCalls(service);

    // Act
    sut.DoTheStuffThatShouldBeTested();

    // Assert
    A.CallTo(() => service.SomeServiceMethod(1, 2)).MustHaveHappened); // Ensures that this call has happened during the act phase
}
```

This used to be doable with scopes in 1.x but I can't find any way to achieve my goal in 2.x

erik-kallen:
Implement `Fake.ClearRecordedCalls`.
Fixes #978.

### DIFF
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -182,6 +182,17 @@ namespace FakeItEasy.Core
         }
 
         /// <summary>
+        /// Removes any recorded calls.
+        /// </summary>
+        internal virtual void ClearRecordedCalls()
+        {
+            ICompletedFakeObjectCall ignored;
+            while (this.recordedCalls.TryDequeue(out ignored))
+            {
+            }
+        }
+
+        /// <summary>
         /// Adds a call rule to the fake object after the specified rule.
         /// </summary>
         /// <param name="previousRule">The rule after which to add a rule.</param>
--- a/tests/FakeItEasy.Tests/Core/FakeManagerTests.cs
+++ b/tests/FakeItEasy.Tests/Core/FakeManagerTests.cs
@@ -413,6 +413,21 @@ namespace FakeItEasy.Tests.Core
         }
 
         [Fact]
+        public void Should_clear_all_recorded_calls_when_calling_clear_recorded_calls()
+        {
+            // Arrange
+            var manager = new FakeManager(typeof(int), 0);
+            manager.RecordCall(A.Dummy<ICompletedFakeObjectCall>());
+            manager.RecordCall(A.Dummy<ICompletedFakeObjectCall>());
+
+            // Act
+            manager.ClearRecordedCalls();
+
+            // Assert
+            manager.GetRecordedCalls().Should().BeEmpty();
+        }
+
+        [Fact]
         public void Should_invoke_listener_when_call_is_intercepted()
         {
             // Arrange
--- a/src/FakeItEasy/Fake.cs
+++ b/src/FakeItEasy/Fake.cs
@@ -48,6 +48,16 @@ namespace FakeItEasy
         }
 
         /// <summary>
+        /// Clears all recorded calls of the faked object.
+        /// </summary>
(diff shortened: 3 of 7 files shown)


blairconrad:
Oh, thanks, @erik-kallen. I've invited you to be a collaborator. This would allow us to assign the issue to you.
I have to run out soon, but someone will look over the contribution soon!

tests/FakeItEasy.Tests/Core/FakeManagerTests.cs
```diff
@@ -413,6 +413,21 @@ public void Should_clear_all_added_rules_when_calling_clear_user_rules()
...
+            var manager = new FakeManager(typeof(int), 0);
+            manager.RecordCall(A.Dummy<ICompletedFakeObjectCall>());
+            manager.RecordCall(A.Dummy<ICompletedFakeObjectCall>());
+
+            // Act
+            manager.ClearRecordedCalls();
+
+            // Assert
+            manager.GetRecordedCalls().Should().BeEmpty();
+        }
```
blairconrad: The new tests are great, and I think you've added them here because you found places we were testing `Fake`. Could I ask a favour, though?
We've been moving toward putting as many tests as is practical in the Specs, with even more of a focus on this when we're [talking about new or changed behaviour](https://github.com/FakeItEasy/FakeItEasy/blob/master/CONTRIBUTING.md#tests). 

How would you feel about replacing this test to the Acceptance tests? The `FakeSpecs` class looks to be a good place.
Then `FakeFacadeTests.ClearRecordedCalls_should_call_clear_recorded_calls_on_manager` could be struck.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -182,6 +182,17 @@ internal virtual void ClearUserRules()
...
 
         /// <summary>
+        /// Removes any recorded calls.
+        /// </summary>
+        internal virtual void ClearRecordedCalls()
+        {
+            ICompletedFakeObjectCall ignored;
+            while (this.recordedCalls.TryDequeue(out ignored))
+            {
+            }
```
thomaslevesque: Agreed. Replacing the queue with a new one is atomic, whereas dequeueing all items is not.

blairconrad:
Hey, @erik-kallen. It's been nearly 3 weeks since we've heard from you. Are you interested in pursuing this pull request any further? We'd be happy to provide guidance, if anything's  confusing.
Or other priorities interfere, and you'd rather not return to this, give the word and one of the owners can make the changes I've requested. You'll still get credit for your contribution.

erik-kallen:
It would probably be easier for a maintainer to implement the requested changes than to review the changes after I make them. Especially moving the test seems like I will put it in the wrong place and have to redo it again.

blairconrad:
@erik-kallen, thanks for the speedy response.

If you're worried about the amount of work _for us_, please do not. Even if you're right, we're happy to spend the time working with contributors.

If you're worried about the amount of work _for you_, then that's a different story, and I respect your position. However, I'm confident that you'd have no problems implementing the changes. If it helps, the new specs could be added [right at the bottom of FakeSpecs.cs](https://github.com/FakeItEasy/FakeItEasy/blob/master/tests/FakeItEasy.Specs/FakeSpecs.cs#L98), and the existing `Scenario`s in that file could be used as a reference for the style of method to write.

All that being said, if you prefer not to continue, that's perfectly fine. Just say the word and one of us will put the finishing touches on the PR.

PR #982, opened 2017-02-15
Removed the last references to IFakeObjectContainer

### LINKED ISSUE #757: Public docs and exception messages refer to the internal IFakeObjectContainer
See https://github.com/FakeItEasy/FakeItEasy/pull/756/files#r66874863

claudiobernasconi:
Fixes #757.

### DIFF
--- a/src/FakeItEasy/Core/DefaultExceptionThrower.cs
+++ b/src/FakeItEasy/Core/DefaultExceptionThrower.cs
@@ -78,7 +78,7 @@ namespace FakeItEasy.Core
 
                 message
                     .AppendLine()
-                    .AppendIndented("      ", "Types marked with * could not be resolved, register them in the current\r\nIFakeObjectContainer to enable these constructors.")
+                    .AppendIndented("      ", "Types marked with * could not be resolved. Please provide a Dummy Factory to enable these constructors.")
                     .AppendLine();
             }
         }
--- a/src/FakeItEasy/Creation/IFakeAndDummyManager.cs
+++ b/src/FakeItEasy/Creation/IFakeAndDummyManager.cs
@@ -12,8 +12,7 @@ namespace FakeItEasy.Creation
         /// </summary>
         /// <param name="typeOfDummy">The type of dummy to create.</param>
         /// <returns>The created dummy.</returns>
-        /// <exception cref="FakeItEasy.Core.FakeCreationException">The current IProxyGenerator is not able to generate a fake of the specified type and
-        /// the current IFakeObjectContainer does not contain the specified type.</exception>
+        /// <exception cref="FakeItEasy.Core.FakeCreationException">The current IProxyGenerator is not able to generate a dummy of the specified type.</exception>
         object CreateDummy(Type typeOfDummy);
 
         /// <summary>
--- a/src/FakeItEasy/ExceptionMessages.resx
+++ b/src/FakeItEasy/ExceptionMessages.resx
@@ -112,10 +112,10 @@
     <value>2.0</value>
   </resheader>
   <resheader name="reader">
-    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <resheader name="writer">
-    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="ApplicatorNotSetExceptionMessage" xml:space="preserve">
     <value>The Apply method of the ExpressionInterceptor may no be called before the Applicator property has been set.</value>
@@ -191,17 +191,6 @@ The following constructors failed:
   </data>
   <data name="FakeCreationExceptionDefaultMessage" xml:space="preserve">
     <value>Unable to create fake object.</value>
-  </data>
-  <data name="FailedToGenerateProxyPattern" xml:space="preserve">
-    <value>FakeItEasy failed to create fake object of type "{0}".
-
-1. The type is not registered in the current IFakeObjectContainer.
-2. The current IProxyGenerator failed to generate a proxy for the following reason:
-
-{1}</value>
-  </data>
-  <data name="UnableToCreateDummyPattern" xml:space="preserve">
-    <value>FakeItEasy was unable to create dummy of type "{0}", register it in the current IFakeObjectContainer to enable this.</value>
   </data>
   <data name="SpecifiedCallIsNotToFakedObject" xml:space="preserve">
     <value>The specified call is not made on a fake object.</value>
--- a/src/FakeItEasy/ExceptionMessages.Designer.cs
+++ b/src/FakeItEasy/ExceptionMessages.Designer.cs
@@ -162,20 +162,6 @@ namespace FakeItEasy {
         }
 
         /// <summary>
-        ///   Looks up a localized string similar to FakeItEasy failed to create fake object of type &quot;{0}&quot;.
-        ///
-        ///1. The type is not registered in the current IFakeObjectContainer.
-        ///2. The current IProxyGenerator failed to generate a proxy for the following reason:
-        ///
-        ///{1}.
-        /// </summary>
-        internal static string FailedToGenerateProxyPattern {
-            get {
-                return ResourceManager.GetString("FailedToGenerateProxyPattern", resourceCulture);
-            }
-        }
-
-        /// <summary>
         ///   Looks up a localized string similar to Unable to create fake object..
         /// </summary>
         internal static string FakeCreationExceptionDefaultMessage {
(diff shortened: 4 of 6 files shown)


src/FakeItEasy/Core/DefaultExceptionThrower.cs
```diff
@@ -78,7 +78,7 @@ private static void AppendNonTriedConstructors(IEnumerable<ResolvedConstructor>
 
                 message
                     .AppendLine()
-                    .AppendIndented("      ", "Types marked with * could not be resolved, register them in the current\r\nIFakeObjectContainer to enable these constructors.")
+                    .AppendIndented("      ", "Types marked with * could not be resolved, register them in the current\r\nIDummyValueCreationSession to enable these constructors.")
```
blairconrad: This does hit the mark in removing a reference to `IFakeObjectContainer`, but it can probably be improved.
`IDummyValueCreationSession` is internal, so an end user seeing this message will not know what to do about it.
The only thing they would know about creating Dummies comes from the [How are Dummies made?](http://fakeiteasy.readthedocs.io/en/stable/dummies/#how-are-the-dummies-made) topic (which points to [Custom Dummy Creation](http://fakeiteasy.readthedocs.io/en/stable/custom-dummy-creation/)).

So I think we might be better off suggesting the users provide a Dummy Factory for the indicated types.

src/FakeItEasy/Creation/IFakeAndDummyManager.cs
```diff
@@ -13,7 +13,7 @@ internal interface IFakeAndDummyManager
         /// <param name="typeOfDummy">The type of dummy to create.</param>
         /// <returns>The created dummy.</returns>
         /// <exception cref="FakeItEasy.Core.FakeCreationException">The current IProxyGenerator is not able to generate a fake of the specified type and
-        /// the current IFakeObjectContainer does not contain the specified type.</exception>
+        /// the current IDummyValueCreationSession does not contain the specified type.</exception>
```
blairconrad: Close, and less crucial because it's internal, but I think I'd still adjust this comment. This is a little more difficult becuase the comment had more problems than just referring to `IFakeObjectContainer`.

You can see that [CreateDummy](https://github.com/FakeItEasy/FakeItEasy/blob/56b4689254c3fe32a991a84278ba2de3b56e8bc0/src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs#L28) calls the session's [TryResolveDummyValue](https://github.com/FakeItEasy/FakeItEasy/blob/fdeb2b1f8a9e0dd363233befa97ff5fa9985722f/src/FakeItEasy/Creation/DummyValueCreationSession.cs#L38) method, which makes sure that the type isn't recursive and then defers to [TryResolveDummyValueWithAllAvailableStrategies](https://github.com/FakeItEasy/FakeItEasy/blob/fdeb2b1f8a9e0dd363233befa97ff5fa9985722f/src/FakeItEasy/Creation/DummyValueCreationSession.cs#L71), which looks for a cached Dummy resolution strategy, and if it doesn't find one, looks through [all its available strategies](https://github.com/FakeItEasy/FakeItEasy/blob/fdeb2b1f8a9e0dd363233befa97ff5fa9985722f/src/FakeItEasy/Creation/DummyValueCreationSession.cs#L29-L34) in turn, using the first one that works. This brings us to the list of strategies in [How are the Dummies made?](http://fakeiteasy.readthedocs.io/en/stable/dummies/#how-are-the-dummies-made).

But the comment makes it seem like we're all about Fake creation, and failing that doing something with some registered types in the (was `IFakeObjectContainer`, now `IDummyValueCreationSession`). And the session doesn't really contain types as such. Probably the container didn't either.

To make matters worse, this comment is on an interface, and really, how can it guarantee what implementers are going to do in order to create Dummy?
For my money, I think I'd just indicate that we throw the exception if the we're unable to create a Dummy of the requested type. It's simpler and less likely to become false in the future…

Reasonable?

claudiobernasconi:
Thank you @blairconrad for your detailed feedback. It seems like I have some space for improvement ;-). I will be working on this tonight. Don't worry; I am not that easy to scare off.

blairconrad:
>  It seems like I have some space for improvement ;-).

It was a more difficult issue than it first appeared.

> I am not that easy to scare off.

I am glad to hear it!

src/FakeItEasy/ExceptionMessages.resx
```diff
@@ -192,17 +192,6 @@ The following constructors failed:
...
-  <data name="FailedToGenerateProxyPattern" xml:space="preserve">
-    <value>FakeItEasy failed to create fake object of type "{0}".
-
-1. The type is not registered in the current IFakeObjectContainer.
-2. The current IProxyGenerator failed to generate a proxy for the following reason:
-
-{1}</value>
-  </data>
-  <data name="UnableToCreateDummyPattern" xml:space="preserve">
-    <value>FakeItEasy was unable to create dummy of type "{0}", register it in the current IFakeObjectContainer to enable this.</value>
```
claudiobernasconi: I do agree that I could split this change out into a separate commit. I am not sure about the statement that the change is not related to the PR. In the PR the goal is to remove all the references to IFakeObjectContainer and since this resource entry contains this reference I wanted to remove it. What do you think? Should I put it in a separate commit or do you really want to open another issue/PR?

blairconrad: Ha, I completely glossed over the contents of the resource entry. Disregard all my comments about this! :blush:

src/FakeItEasy/Creation/IFakeAndDummyManager.cs
claudiobernasconi: Reasonable, especially since in the next method declaration on line 25 the comment on line 24 is exactly what you proposed. So this is a much more consistent comment anyway.

blairconrad: Thanks. The only thing I'd be careful of is that in our case, it's not necessarily the proxy generator that can't make a Dummy. Well, it can't, but neither can any of our other resolution strategies; I wouldn't focus just on the proxy generator.

src/FakeItEasy/ExceptionMessages.resx
claudiobernasconi: Fine. I am currently working on this PR and will update it when it's ready.

blairconrad: I am not sure from your response whether my previous message was clear. Sorry if there was confusion.
The removals are good, and as you point out, completely related to this PR. I'd be happy to have them in this PR, and there's no need to make a separate commit.

claudiobernasconi:
@blairconrad Work is done. Alright, now I have a git problem *facepalm. Yesterday, I committed to my fix-branch and pushed that commit to my fork. Today, I committed the changes suggested by your comments. I thought it would be nice to have both changes in a single commit, so I committed using the --amend option.

I realized that I could not push my changes back to my fork because it was not on the newest revision. So I had to pull the changes down and merge them into my working copy. By now I have in my local copy of the git repo 3 commits. The one from yesterday, the one from today, and the merge.

How can I shrink this mess to a single commit and push it to my fork so that this PR gets updated? Was it a failure to try to --amend my changes from today?

Thanks for your help.

blairconrad:
@claudiobernasconi Good news. I am not a huge git expert. This means I make many many mistakes, much worse than yours, so I am experienced.

I think you can simply `git rebase -i master` your changes. In the editor that then appears, change the second and third commits to either `fixup` or `squash` (the main difference that I see between these is that with `squash`, you have an opportunity to merge and edit all the commit messages. with `fixup`, you lose the extra commit messages). After this, you will have one commit.
I would then do a 

```cmd
push origin HEAD --force-with-lease
```

which will push your commits to your GitHub repo. The `--force-with-lease` will allow you to overwrite the old commits. (Similar to `--force`, except the `-with-lease` does some checks to make sure that you "own" the remote commit you're replacing. I think. This is probably inaccurate, but the mental model has served me well.)

As for whether you made a mistake by using `--amend`, I don't think so. Sometimes during PR reviews, it's nice to see extra commits pushed to the branch, as the reviewer can see the new changes. In this case, the PR is small, and the changes well understood, so I think the amend was completely appropriate. 
In any case, if you _had_ pushed extra commits, I eventually would've asked for a squash (or fixup, whatever) anyhow!

If this makes sense, yay! If this does not make sense, perhaps it would be easier to chat over at https://gitter.im/FakeItEasy/FakeItEasy.

Also, this is a good time to let you know that we're nearly releasing 3.0.0. 3.0.0-rc001 is out now, and we'll not likely take any commits into the `master` branch until we release 3.0.0. This probably just means we'll have to hang onto your branch for another week or so. At that point, depending on what other changes arrive we _may_ ask you to rebase on the new master, but I doubt we will have any conflicts that make it necessary.

src/FakeItEasy/Core/DefaultExceptionThrower.cs
```diff
@@ -78,7 +78,7 @@ private static void AppendNonTriedConstructors(IEnumerable<ResolvedConstructor>
 
                 message
                     .AppendLine()
-                    .AppendIndented("      ", "Types marked with * could not be resolved, register them in the current\r\nIFakeObjectContainer to enable these constructors.")
+                    .AppendIndented("      ", "Types marked with * could not be resolved. Please provide a Dummy Factory to enable these constructors.")
```
blairconrad: You removed the comma splice while you were at it. :heart_eyes:

blairconrad:
Thanks, @claudiobernasconi. The changes look great. As noted, we'll have to hold onto the PR for a little bit until we release 3.0.0. (I'm afraid to say this out loud, but ideally the middle of next week.)
Also, it'll give the other owners a chance to take a peek at the changes, if they're interested.

adamralph:
Thanks @claudiobernasconi! Look out for your name in the release notes. 🏆

PR #983, opened 2017-02-17
Fix NRE when matching null arg with value-type constraint

### LINKED ISSUE #660: Debuggability: Wrong type in argument matchers results in NullReferenceException
(With FakeItEasy 2.0.0-beta009)
I have a test like below which passes.

If I accidentaly changed `A<DateTimeOffset?>` to `A<DateTimeOffset>` (non nullable type) I get a rather tricky to debug nullref:

> System.NullReferenceException occurred
>   _HResult=-2147467261
>   _message=Object reference not set to an instance of an object.
>   HResult=-2147467261
>   IsTransient=false
>   Message=Object reference not set to an instance of an object.
>   Source=FakeItEasy
>   StackTrace:
>        at FakeItEasy.Core.DefaultArgumentConstraintManager`1.MatchesConstraint.FakeItEasy.Core.IArgumentConstraint.IsValid(Object argument)
>   InnerException: 

``` c#
var provisioner = CreateProvisioner();
var operations = CreateOperations();

Deployment dep = new Deployment();
A.CallTo(() => provisioner.GetResource(A<ResourceSpec>._, A<DateTimeOffset?>._))
    .Returns(dep);

var deleteResourceCall = A.CallTo(() => operations.DeleteDeployment(dep, A<RequestData>._));

await WithTestServer(operations, provisioner)(async (server) =>
{
    string uri = CreateUri();
    var response = await server.HttpClient.DeleteAsync(uri);

    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
});

deleteResourceCall.MustHaveHappened();
```

thomaslevesque:
Fixes #660

### DIFF
--- a/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -41,6 +41,11 @@ namespace FakeItEasy.Specs
             bool Validate([Out] string value);
         }
 
+        public interface IHaveANullableParameter
+        {
+            int Bar(int? x);
+        }
+
         [Scenario]
         public static void ParameterArrays(
             ITypeWithParameterArray fake)
@@ -348,6 +353,24 @@ namespace FakeItEasy.Specs
 
             "Then it should return the default value"
                 .x(() => result.Should().BeFalse());
+        }
+
+        [Scenario]
+        public static void ThatArgumentConstraintForValueTypeWithNullArgument(
+            IHaveANullableParameter subject,
+            int result)
+        {
+            "Given a fake with a method that accepts a nullable value type parameter"
+                .x(() => subject = A.Fake<IHaveANullableParameter>());
+
+            "And a call configured for a non-nullable argument of that type"
+                .x(() => A.CallTo(() => subject.Bar(A<int>._)).Returns(42));
+
+            "When I make a call to this method with a null argument"
+                .x(() => result = subject.Bar(null));
+
+            "Then it doesn't match the configured call"
+                .x(() => result.Should().Be(0));
         }
 
         [Scenario]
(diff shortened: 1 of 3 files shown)


adamralph:
What shall we do with this one? Merge now, or wait until after 3.0.0? Given that https://github.com/FakeItEasy/FakeItEasy/pull/982 has gone in, waiting doesn't really buy us anything. If we find a showstopper bug in 3.0.0-rc001 then we'll have to release 3.0.0-rc002 anyway, so this and #982 can then become part of 3.0.0.

blairconrad:
> What shall we do with this one? Merge now, or wait until after 3.0.0?

Review first. Once we're happy, I see no reason to delay the merge. As you say, we're already committed to releasing 3.0.0 off something other than the tip of master.

adamralph:
> Review first

Naturally 😉

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -351,6 +356,24 @@ public interface IHaveAnOutParameter
...
+            "Given a fake with a method that accepts a nullable value type parameter"
+                .x(() => subject = A.Fake<IHaveANullableParameter>());
+
+            "And a call configured for a non-nullable argument of that type"
+                .x(() => A.CallTo(() => subject.Bar(A<int>._)).Returns(42));
+
+            "When I make a call to this method with a null argument"
+                .x(() => result = subject.Bar(null));
+
+            "Then it doesn't match the configured call and returns the default value"
```
blairconrad: If I were going to be super picky, I'd note that we could probably do without the compound condition here. There was no mention of return values in the "And" step above. I'd consider just dropping everything after the "call" in this step description.
I wouldn't hold things up for this, though.

blairconrad:
As usual, very nice work, @thomaslevesque. Thanks!

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
thomaslevesque: done!

blairconrad:
Thanks, @thomaslevesque!

PR #984, opened 2017-02-17
[WIP] Fix internal calls to Object members of strict fakes

### LINKED ISSUE #457: describing a call throws an exception when argument is strict
Say we setup a strict fake:

``` csharp
var a = A.Fake<SomeInterface>(s=>s.Strict()); //A strict fake!
var b = A.Fake<SomeOtherInterface>(); //strictness does not matter...
```

and then we verify the strict fake was passed as an argument to another mock:

``` csharp
//a.ToString() is called for some reason, however a is a strict fake
A.CallTo(()=>b.SomeMethod(a)).MustHaveHappened()
```

FakeItEasy throws an Exception:

``` csharp
FakeItEasy.ExpectationException : Call to non configured method "ToString" of strict fake.
```

Call Stack:

``` csharp
   at FakeItEasy.FakeOptionsBuilderExtensions.<Strict>b__0<T>(IFakeObjectCall call)
   at FakeItEasy.Configuration.BuildableCallRule.Apply(IInterceptedFakeObjectCall fakeObjectCall)
   at FakeItEasy.Core.FakeManager.FakeItEasy.Core.IFakeCallProcessor.Process(IWritableFakeObjectCall fakeObjectCall)
   at Castle.Proxies.ObjectProxy_4.ToString()
   at FakeItEasy.Expressions.ArgumentConstraints.EqualityArgumentConstraint.WriteDescription(IOutputWriter writer)
   at FakeItEasy.Expressions.ExpressionCallMatcher.AppendArgumentsListString(StringBuilder result)
   at FakeItEasy.Expressions.ExpressionCallMatcher.ToString()
   at FakeItEasy.Configuration.BuildableCallRule.WriteDescriptionOfValidCall(IOutputWriter writer)
   at FakeItEasy.Configuration.RuleBuilder.MustHaveHappened(Repeated repeatConstraint)
```

thomaslevesque:
~Fixes #457~

### DIFF
--- a/src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
@@ -15,12 +15,22 @@ namespace FakeItEasy.Core
         private class ObjectMemberRule
             : IFakeObjectCallRule
         {
+            private static readonly MethodInfo ObjectEqualsMethod = typeof(object).GetMethod("Equals", new[] { typeof(object) });
+            private static readonly MethodInfo ObjectGetHashCodeMethod = typeof(object).GetMethod("GetHashCode", new Type[] { });
+            private static readonly MethodInfo ObjectToStringMethod = typeof(object).GetMethod("ToString", new Type[] { });
+            private static readonly MethodInfo FakeObjectEqualsMethod = typeof(IFakeObject).GetMethod("Equals", new[] { typeof(object) });
+            private static readonly MethodInfo FakeObjectGetHashCodeMethod = typeof(IFakeObject).GetMethod("GetHashCode", new Type[] { });
+            private static readonly MethodInfo FakeObjectToStringMethod = typeof(IFakeObject).GetMethod("ToString", new Type[] { });
+
             private static readonly List<MethodInfo> ObjectMethods =
                 new List<MethodInfo>
                     {
-                        typeof(object).GetMethod("Equals", new[] { typeof(object) }),
-                        typeof(object).GetMethod("ToString", new Type[] { }),
-                        typeof(object).GetMethod("GetHashCode", new Type[] { })
+                        ObjectEqualsMethod,
+                        ObjectGetHashCodeMethod,
+                        ObjectToStringMethod,
+                        FakeObjectEqualsMethod,
+                        FakeObjectGetHashCodeMethod,
+                        FakeObjectToStringMethod,
                     };
 
             private readonly FakeManager fakeManager;
@@ -70,7 +80,8 @@ namespace FakeItEasy.Core
 
             private bool TryHandleGetHashCode(IInterceptedFakeObjectCall fakeObjectCall)
             {
-                if (!IsSameMethod(fakeObjectCall.Method, ObjectMethods[2]))
+                if (!IsSameMethod(fakeObjectCall.Method, ObjectGetHashCodeMethod) &&
+                    !IsSameMethod(fakeObjectCall.Method, FakeObjectGetHashCodeMethod))
                 {
                     return false;
                 }
@@ -82,7 +93,8 @@ namespace FakeItEasy.Core
 
             private bool TryHandleToString(IInterceptedFakeObjectCall fakeObjectCall)
             {
-                if (!IsSameMethod(fakeObjectCall.Method, ObjectMethods[1]))
+                if (!IsSameMethod(fakeObjectCall.Method, ObjectToStringMethod) &&
+                    !IsSameMethod(fakeObjectCall.Method, FakeObjectToStringMethod))
                 {
                     return false;
                 }
@@ -94,7 +106,8 @@ namespace FakeItEasy.Core
 
             private bool TryHandleEquals(IInterceptedFakeObjectCall fakeObjectCall)
             {
-                if (!IsSameMethod(fakeObjectCall.Method, ObjectMethods[0]))
+                if (!IsSameMethod(fakeObjectCall.Method, ObjectEqualsMethod) &&
+                    !IsSameMethod(fakeObjectCall.Method, FakeObjectEqualsMethod))
                 {
                     return false;
                 }
--- a/src/FakeItEasy/Core/IFakeObject.cs
+++ b/src/FakeItEasy/Core/IFakeObject.cs
@@ -1,0 +1,28 @@
+﻿namespace FakeItEasy.Core
+{
+    /// <summary>
+    /// This interface is used internally by FakeItEasy and is exposed only due to technical constraints. It is not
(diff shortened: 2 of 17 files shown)


adamralph:
Marking as WIP while we discuss #457.

src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
```diff
@@ -70,7 +80,8 @@ private static bool IsObjetMethod(IFakeObjectCall fakeObjectCall)
 
             private bool TryHandleGetHashCode(IInterceptedFakeObjectCall fakeObjectCall)
             {
-                if (!IsSameMethod(fakeObjectCall.Method, ObjectMethods[2]))
+                if (!IsSameMethod(fakeObjectCall.Method, ObjectGetHashCodeMethod) &&
+                    !IsSameMethod(fakeObjectCall.Method, FakeObjectGetHashCodeMethod))
```
blairconrad: Much nicer than `ObjectMethods2[i]`.

src/FakeItEasy/Core/IFakeObject.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace FakeItEasy.Core
+{
+    /// <summary>
+    /// This interface is used internally by FakeItEasy and is exposed only due to technical constraints. It is not
+    /// intended to be used directly from your code.
+    /// </summary>
```
blairconrad: Nice note. While we're at it, I wonder if a similar addition to `ITaggable` isn't a good idea.

Well, that's sort of what I wonder. What I really wonder is how I can convince everyone reading this that we should drop `ITaggable` in favour of `IFakeObject`.

src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
thomaslevesque: I thought so too... indexes were manageable with only 3 methods, but not so much with 6 methods.

src/FakeItEasy/Core/IFakeObject.cs
thomaslevesque: Yes, it would probably make sense. `ITaggable` has always felt kind of wrong to me. It looks like it's designed for a very generic usage (the name `Tag` is vague, and it's typed as `object`), but we use it for something very specific.

Of course, if we could make both interfaces internal, it would be even better.

src/FakeItEasy/FakeOptionsExtensions.cs
```diff
@@ -21,12 +21,7 @@ public static IFakeOptions<T> Strict<T>(this IFakeOptions<T> options)
         {
             Guard.AgainstNull(options, nameof(options));
 
-            Action<IFakeObjectCall> thrower = call =>
-                {
-                    throw new ExpectationException("Call to non configured method \"{0}\" of strict fake.".FormatInvariant(call.Method.Name));
-                };
-
-            return options.ConfigureFake(fake => A.CallTo(fake).Invokes(thrower));
+            return options.ConfigureFake(fake => ConfigureStrictFake(fake));
```
thomaslevesque: No, it won't work in this case. (I had written a more complete explanation, but my internet connection crashed and I'm too lazy to type it on my phone...)

blairconrad: Ah. Thanks for the explanation.

blairconrad:
Folks, what do we want to do with this PR?
I thought the solution was very good, and would be happy if we pursued this line. But if that's not the consensus, maybe the PR should be closed for rework?

adamralph:
Now that #457 has been fixed, perhaps we need to re-consider what we're trying to achieve with this PR. Is there any functional change? If not, is it refactoring/tech-debt payoff which is worth doing?

thomaslevesque:
> Now that #457 has been fixed, perhaps we need to re-consider what we're trying to achieve with this PR. Is there any functional change? If not, is it refactoring/tech-debt payoff which is worth doing?

Well, the point is to avoid exceptions when calling `Object` members from internal FakeItEasy code, without affecting the fake behavior from the user's perspective. All internal calls to these members are made via the `IFakeObject` interface instead of directly on the fake. You already solved the issue with calling `ToString` when describing a call, but there are other cases where we call `ToString`, `Equals` or `GetHashCode` internally, and these will throw for a strict fake or if the fake is explicitly configured to throw.

I think it's worth fixing those cases, but maybe my approach in this PR is overkill... We could just locate these calls and surround them with try/catch when appropriate.

blairconrad:
I'm not convinced that the approach is overkill. Locating the calls and surrounding them with try/catch may work just fine. We don't know until we try it. But I'm concerned that often `Equals` and `GetHashCode` will be called in locations that are not visible to us (inside collection code).

Also, I've become fond of `IFakeObject`. Especially as a potential replacement for `ITaggable`. But maybe that's just sentiment.

adamralph:
Can there be any functional change for users as a result of this? Or is it pure refactoring?

blairconrad:
If FakeItEasy is evaluating the equality of two fakes, perhaps because one was passed to another one as a method argument, and we want to see if the call applies. If the fake is strict, this comparison will throw. If we fix this issue, then we can still compare the object without triggering a forbidden method call.

adamralph:
@blairconrad it sounds like you've described a new bug. I assume it would be easy enough to write a spec which demonstrates it?

blairconrad:
@adamralph: https://github.com/FakeItEasy/FakeItEasy/issues/457#issuecomment-73713213

thomaslevesque:
> Can there be any functional change for users as a result of this? Or is it pure refactoring?

Well, for instance, the behavior of `A<T>.That.IsEqualTo` is changed, because it no longer uses the configured behavior of `Equals`. The same is true for `EqualityArgumentConstraint`. I'm not sure if this change is desirable...

adamralph:
> If FakeItEasy is evaluating the equality of two fakes, perhaps because one was passed to another one as a method argument, and we want to see if the call applies. If the fake is strict, this comparison will throw. If we fix this issue, then we can still compare the object without triggering a forbidden method call.

At this stage, I suggest we close this PR and raise a bug instead to represent this.

> Well, for instance, the behavior of `A<T>.That.IsEqualTo` is changed, because it no longer uses the configured behavior of `Equals`. The same is true for `EqualityArgumentConstraint`. I'm not sure if this change is desirable...

It sounds like this is not a desirable change. If `Equals` has been configured, I'd expect it to be used when making the comparison. However, we should continue this thread on the bug issue, since it's something to consider when fixing the bug.

thomaslevesque:
OK to close. I don't have time to create the issue now, I'll try to do it later.

PR #989, opened 2017-02-22
Don't consider open generic types to be instantiable

### LINKED ISSUE #523: Stop throwing first chance exceptions when CreateInstance is called on generic extension point types
Very much like #494, only with open generic types.
When we try to instantiate argument formatters, fake configurators, and dummy factories, we may try to instantiate an open generic type, causing a first chance exception.

For an example, see [Configure_should_be_null_guarded](https://github.com/FakeItEasy/FakeItEasy/blob/dbfc0289f2493ee10d0aaf325fdc8734307645b4/Source/FakeItEasy.Tests/Core/FakeConfiguratorTests.cs#L25). We try to instantiate `TestableConfigurator<T>`.

blairconrad:
Fixes #523

### DIFF
--- a/src/FakeItEasy/TypeExtensions.cs
+++ b/src/FakeItEasy/TypeExtensions.cs
@@ -36,7 +36,13 @@ namespace FakeItEasy
         [DebuggerStepThrough]
         public static bool CanBeInstantiatedAs(this Type type, Type targetType)
         {
-            return targetType.IsAssignableFrom(type) && !type.GetTypeInfo().IsAbstract;
+            if (!targetType.IsAssignableFrom(type))
+            {
+                return false;
+            }
+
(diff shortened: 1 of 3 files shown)


thomaslevesque:
LGTM. Is there a way to test this change?

blairconrad:
@thomaslevesque, good question. End-to-end testing is, I think, impossible due to our `try`s. We didn't bother with tests for #495 for this reason. 
However, that's no reason not to test the method in isolation. It's not quite as good, but better than nothing.
I've added the tests. Happy to do more, if you think there's benefit.

thomaslevesque:
perfect, thanks @blairconrad!

blairconrad:
Thank you!

PR #993, opened 2017-02-24
Ensure that calls can be described when ToString() throws

### LINKED ISSUE #457: describing a call throws an exception when argument is strict

adamralph:
fixes #457

### DIFF
--- a/src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
+++ b/src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
@@ -1,5 +1,6 @@
 namespace FakeItEasy.Expressions.ArgumentConstraints
 {
+    using System;
     using FakeItEasy.Core;
 
     internal class EqualityArgumentConstraint
@@ -32,7 +33,17 @@ namespace FakeItEasy.Expressions.ArgumentConstraints
                 return "\"{0}\"".FormatInvariant(stringValue);
             }
 
-            return this.ExpectedValue.ToString();
+            try
+            {
+                return this.ExpectedValue.ToString();
+            }
+            catch (Exception)
+            {
+                FakeManager manager = Fake.TryGetFakeManager(this.ExpectedValue);
+                return manager != null
+                    ? $"Faked {manager.FakeObjectType.FullName}"
+                    : this.ExpectedValue.GetType().ToString();
+            }
         }
 
         public void WriteDescription(IOutputWriter writer)
--- a/src/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
+++ b/src/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
@@ -17,13 +17,12 @@ namespace FakeItEasy.Core
         /// </summary>
         /// <param name="proxy">The proxy to get the manager from.</param>
         /// <returns>A fake manager.</returns>
+        /// <exception cref="ArgumentException">If <paramref name="proxy"/> is not actually a faked object.</exception>
         public FakeManager GetFakeManager(object proxy)
         {
             Guard.AgainstNull(proxy, nameof(proxy));
 
-            var taggable = AsTaggable(proxy);
-
-            var result = taggable.Tag as FakeManager;
+            FakeManager result = this.TryGetFakeManager(proxy);
 
             if (result == null)
             {
@@ -31,6 +30,20 @@ namespace FakeItEasy.Core
             }
 
             return result;
+        }
+
+        /// <summary>
+        /// Gets the fake manager associated with the proxy, if any.
+        /// </summary>
+        /// <param name="proxy">The proxy to get the manager from.</param>
+        /// <returns>The fake manager, or <c>null</c> if <paramref name="proxy"/> is not actually a faked object.</returns>
+        public FakeManager TryGetFakeManager(object proxy)
+        {
+            Guard.AgainstNull(proxy, nameof(proxy));
+
+            var taggable = AsTaggable(proxy);
+
+            return taggable.Tag as FakeManager;
         }
 
         public void TagProxy(object proxy, FakeManager manager)
--- a/src/FakeItEasy/Core/IFakeManagerAccessor.cs
+++ b/src/FakeItEasy/Core/IFakeManagerAccessor.cs
@@ -1,5 +1,7 @@
 namespace FakeItEasy.Core
 {
+    using System;
+
     /// <summary>
     /// Attaches a fake manager to the proxy so that intercepted
     /// calls can be configured.
@@ -11,7 +13,15 @@ namespace FakeItEasy.Core
         /// </summary>
         /// <param name="proxy">The proxy to get the manager from.</param>
         /// <returns>A fake manager.</returns>
+        /// <exception cref="ArgumentException">If <paramref name="proxy"/> is not actually a faked object.</exception>
         FakeManager GetFakeManager(object proxy);
+
+        /// <summary>
+        /// Gets the fake manager associated with the proxy, if any.
+        /// </summary>
+        /// <param name="proxy">The proxy to get the manager from.</param>
+        /// <returns>The fake manager, or <c>null</c> if <paramref name="proxy"/> is not actually a faked object.</returns>
+        FakeManager TryGetFakeManager(object proxy);
 
         /// <summary>
         /// Tags a proxy object, so that it can accessed later by <see cref="GetFakeManager"/>.
--- a/src/FakeItEasy/Fake.cs
+++ b/src/FakeItEasy/Fake.cs
@@ -18,6 +18,7 @@ namespace FakeItEasy
         /// </summary>
         /// <param name="fakedObject">The faked object to get the manager object for.</param>
         /// <returns>The fake object manager.</returns>
+        /// <exception cref="ArgumentException">If <paramref name="fakedObject"/> is not actually a faked object.</exception>
         [DebuggerStepThrough]
         [SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "object", Justification = "The term fake object does not refer to the type System.Object.")]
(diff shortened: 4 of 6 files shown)


src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
```diff
@@ -32,7 +33,14 @@ public override string ToString()
...
             }
 
-            return this.ExpectedValue.ToString();
+            try
+            {
+                return this.ExpectedValue.ToString();
+            }
+            catch (Exception)
+            {
+                return this.ExpectedValue.GetType().Name;
```
thomaslevesque: Shouldn't the type be prefixed by "Faked" if `ExpectedValue` is a fake? This is what the default `ToString()` implementation does for fakes.

adamralph: image


But I do know that we had no test enforcing it! I'll add one, and alter the fix.

thomaslevesque: 👍

thomaslevesque: In fact, [it *is* covered in tests](https://github.com/FakeItEasy/FakeItEasy/blob/376ba52f4126ea47e2bcaa494348612326714e4a/tests/FakeItEasy.Tests/Core/FakeManagerTests.cs#L335)... not in specs, though.

thomaslevesque: In fact, [it *is* covered in unit tests](https://github.com/FakeItEasy/FakeItEasy/blob/376ba52f4126ea47e2bcaa494348612326714e4a/tests/FakeItEasy.Tests/Core/FakeManagerTests.cs#L335)... But adding a spec for this wouldn't hurt.

adamralph:
@thomaslevesque I inserted a tech-debt pay off commit at the beginning to enforce the "Faked... " convention in the messages via a spec.

I replaced the red commit completely and I added a fixup! commit for the green commit to special case fakes.

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
```diff
@@ -32,7 +36,18 @@ public override string ToString()
...
+            try
+            {
+                return this.ExpectedValue.ToString();
+            }
+            catch (Exception)
+            {
+                var type = this.ExpectedValue.GetType();
+
+                return type.Namespace == "Castle.Proxies"
+                    ? $"Faked {type.GetTypeInfo().BaseType.FullName}"
```
blairconrad: For a faked interface, this will print "Faked System.Object", rather than the faked interface.
And it misses faked delegates altogether.

Consider trying to get the FakeManager (Unfortunately we currently only have a `GetFakeManager` that throws when we call that on a non-Fake, as far as I can tell. But I see no reason not to expose an internal accessor that isn't so cranky.)
If `ExpectedValue` has a FakeManager, we can get the faked type from `FakeObjectType` and slap a "Fake " on the front of it.

While we're at it, the specs should probably include tests for faked interfaces, classes, and delegates, just to be sure we catch all the wacky cases.

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
```diff
@@ -32,7 +36,18 @@ public override string ToString()
...
+                    : type.Name;
```
blairconrad: Why `Name` here and `FullName` above?

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
adamralph: OK, I'll give the `FakeManager` approach a try, and I'll add specs for the other types of types.

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
adamralph: For some reason FakeItEasy uses the full name for fakes, whereas non-fakes only use the name:

```C#
Console.WriteLine(new List<int>().ToString());
```
> System.Collections.Generic.List\`1[System.Int32]

```C#
Console.WriteLine(A.Fake<List<int>>().ToString());
```

> Faked System.Collections.Generic.List\`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]

blairconrad: Non-Fakes isn't quite just the name. It was the lack of namespace that tipped me off:

`System.Console.WriteLine((new List<int>()).GetType().Name);`

List`1

> System.Console.WriteLine((new List<int>()).GetType().FullName);

System.Collections.Generic.List`1[[System.Int32, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]

So just outputting `Name` still doesn't match the `ToString`, but it's perhaps a reasonable compromise.

The deeper issue, I suppose, why Fakes describe themselves differently. Maybe they shouldn't. Probably it doesn't matter. Either way, we don't need to address that question in this issue.

Thanks!

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
```diff
@@ -32,7 +33,26 @@ public override string ToString()
...
+            {
+                FakeManager manager;
+                try
+                {
+                    manager = Fake.GetFakeManager(this.ExpectedValue);
+                }
+                catch (ArgumentException)
+                {
+                    manager = null;
+                }
```
thomaslevesque: For now, the try/catch will do, but eventually we'll probably want to add a `Fake.IsFake` method and/or a `Fake.TryGetFakeManager` method. There's no point in throwing and catching an exception internally when it can easily be avoided.

adamralph: All true. I briefly investigated, but I felt like I was going down a rabbit hole so I backed out.

adamralph:
I'll squash the fixup! commits after @blairconrad has given the diff his blessing (since he also chimed in above).

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
blairconrad: I'd imagined a `TryGetFakeManager` when I made the comment. I've added one in a commit that I've [pushed to my own branch](https://github.com/blairconrad/FakeItEasy/commit/4c9c637ee35cbd2d46b1f75354f14446ad26dfae). If you like it, feel free to squash it in.
If not, I'm very happy with what you have here. I particularly like the tests. Thanks!

adamralph:
@blairconrad thanks for the fixup! commits. I've squash them into this branch. This should be ready to go now. /cc @thomaslevesque 

BTW - I was a bit unsure whether I preferred `bool TryGetFakeManager(object proxy, out FakeManager manager)` or `FakeManager TryGetFakeManager(object proxy)` but since you both seemed to prefer the latter, and since it's internal, I didn't agonise too long about it. If it were a public API I could see an interesting discussion arising.

blairconrad:
> BTW - I was a bit unsure whether I preferred `bool TryGetFakeManager(object proxy, out FakeManager manager)` or `FakeManager TryGetFakeManager(object proxy)`

I'm on the fence as well. I just didn't want to make a fuss. As you say, since it's internal, I also didn't agonise over it, and @thomaslevesque liked it, so I thought "why not?".

I also thought "in case there's controversy, I'll change the signature in an additional fixup, so @adamralph can decide whether to take 0, 1, or 2 of the changes".

adamralph:
> I'll change the signature in an additional fixup, so @adamralph can decide whether to take 0, 1, or 2 of the changes

I enjoyed that, and I almost took just the first one. 😉

blairconrad:
I believe @thomaslevesque already approved your base and the extra commits, so I'm-a merge.

Thanks, @adamralph!

PR #995, opened 2017-02-27
Use simple-targets-csx

### LINKED ISSUE #645: Switch to a build system which doesn't require extra installations
I'm getting more and more loathed to install Ruby when I'm building a new dev machine purely for the purpose of running Rake. Rake is a majestic tool, but there are several worthy competitors nowadays which can be bootstrapped locally by a `build.cmd` and/or `build.sh`. E.g. Fake, Cake, Bau.

blairconrad:
Fixes #645.

I removed the fancy milestone/release/issue/version rake targets first so their loss would be explicit. We can explore replacements, if there's interest.


build.csx
```diff
@@ -0,0 +1,225 @@
...
+targets.Add("clean", DependsOn("logsDirectory"), () => RunMsBuild("Clean"));
+
+targets.Add(
+    "restore",
+    () =>
+    {
+        Cmd(nuget, $"restore {solution} -PackagesDirectory {packagesDir}");
+        foreach (var projectDir in netstdUnitTestDirectories.Concat(netstdIntegrationTestDirectories).Concat(netstdSpecDirectories))
+        {
+            Cmd("dotnet", $"restore {projectDir}");
```
thomaslevesque: Why not `RunDotNet` here?

build.csx
```diff
@@ -0,0 +1,225 @@
...
+public static void RunDotNetTests(string[] testDirectories)
+{
+    foreach (var testDirectory in testDirectories)
+    {
+        var xml = Path.GetFullPath(Path.Combine(testsDirectory, Path.GetFileName(testDirectory) + ".TestResults.xml"));
+        RunDotNet(testDirectory, "test", $"-c Release -nologo -notrait \"explicit=yes\" -xml {xml}");
+    }
+}
+
+public static string GetVersion(string assemblyInfo)
```
thomaslevesque: Since we only do this after build, it would be easier and cleaner to get the information from the compiled assembly, rather than from the AssemblyInfo.cs file.

rakefile.rb
```diff
@@ -1,218 +0,0 @@
...
-    test_cmd.execute
-  end
-end
-
-def restore_dependencies_for_project(project_dir)
-  restore_cmd = Exec.new
-  restore_cmd.command = "dotnet"
-  restore_cmd.parameters "restore #{project_dir}"
-  restore_cmd.execute
-end
```
thomaslevesque: No more Ruby, yay !

build.cmd
```diff
@@ -1,3 +1,31 @@
-@echo off
+@echo Off
+pushd %~dp0
```
blairconrad: WHAT? Not everyone has memorized the `%~` codes? Stunning.
Sure!

build.csx
```diff
@@ -0,0 +1,225 @@
...
+
+#r "System.Xml.Linq"
+
+using System.Runtime.Serialization.Json;
+using System.Text.RegularExpressions;
+using System.Xml;
+using System.Xml.Linq;
+using static SimpleTargets;
+
+static var solutionName = "FakeItEasy";
```
blairconrad: Will do.

build.csx
```diff
@@ -0,0 +1,225 @@
...
+
+static var solution = "./" + solutionName + ".sln";
+static var packagesDir = Path.GetFullPath("packages");
+var repoUrl = "https://github.com/FakeItEasy/FakeItEasy";
+var assemblyInfo = "./src/CommonAssemblyInfo.cs";
```
blairconrad: I only added the static when needed. I was on the fence. It makes no difference to me whether they're there or not, so I'll add 'em.

build.csx
blairconrad: Lemme tell you about that. I don't know.
Also, this command is not needed. I just verified about 15 minutes ago. I was going to submit another PR, as I didn't want to change what we did here too much (just how), but I can change this PR to omit this `foreach`, if it suits.

build.csx
blairconrad: Ah. Clever. It had been done before, and I only moved to this step at the last minute. I will investigate.

build.csx
thomaslevesque: Not needed? You mean it's covered by doing `nuget restore` on the solution?

build.csx
adamralph: Why bother adding `static` at all? It just adds noise, and it a micro-optimisation at best.

build.csx
adamralph: I'm not a fan of const for anything which isn't logically constant, but I don't mind too much either way.

build.csx
adamralph: It is covered by the NuGet restore on the solution, but I think it should be separated. I.e. the nuget restore that is done in build.cmd should only restore the packages required for the build script, by way of a top level packages.config, as I do in xbehave. This restore should do everything for the solution.

Moreover, IMHO, when we switch to pure `dotnet.exe`-able projects (new csproj) we should remove this target and put a `dotnet restore` in the build instructions, to avoid having to restore every time the project is built. This fits in more closely with the typical `dotnet.exe` workflow, i.e. `dotnet restore`, `dotnet build`, and the error messages from a restore-less build tell you to run `dotnet restore` first.

build.csx
```diff
@@ -0,0 +1,225 @@
...
+
+public static void RunMsBuild(string target)
+{
+    var packagesDirOption = string.IsNullOrEmpty(packagesDir) ? "" : $"/p:NuGetPackagesDirectory={packagesDir}";
+    Cmd(
+        msBuild,
+        $"{solution} /target:{target} /p:configuration=Release /nr:false /verbosity:minimal /nologo /fl /flp:LogFile=artifacts/logs/{target}.log;Verbosity=Detailed;PerformanceSummary {packagesDirOption}");
+}
+
+public static void RunDotNet(string workingDirectory, string command, string args="")
```
adamralph: This is an almost complete replica of `Cmd`. I wonder if it would be better to have an overload for `Cmd` which takes a working directory and make the message conditional on whether it is `null` or not?

blairconrad: Can do.

build.csx
blairconrad: Done.

build.csx
```diff
@@ -2,18 +2,17 @@
 
 #r "System.Xml.Linq"
 
-using System.Runtime.Serialization.Json;
-using System.Text.RegularExpressions;
-using System.Xml;
-using System.Xml.Linq;
+//using System.Runtime.Serialization.Json;
+//using System.Xml;
+//using System.Xml.Linq;
```
thomaslevesque: Did you leave these commented lines on purpose?

build.csx
```diff
@@ -202,7 +201,9 @@ public static void RunDotNetTests(IEnumerable<string> testDirectories)
 
 public static string GetVersion()
 {
-    var version = Regex.Match(File.ReadAllText(assemblyInfo), "AssemblyInformationalVersion\\(\"([^\"]+)\"\\)").Groups[1].Value;
+    var  versionInfo = FileVersionInfo.GetVersionInfo(versionAssembly);
+    var version = versionInfo.FileVersion;
+
```
thomaslevesque: According to [this](http://stackoverflow.com/a/4466441/98713), shouldn't it be `ProductVersion` rather than `FileVersion`? Apparently `AssemblyInformationalVersion` maps to `FileVersionInfo.ProductVersion`. Not that it really matters, since we use the same value for all version attributes...

build.csx
```diff
@@ -0,0 +1,217 @@
...
+//using System.Xml.Linq;
+using static SimpleTargets;
+
+static readonly var  solutionName = "FakeItEasy";
+
+static readonly var solution = "./" + solutionName + ".sln";
+static readonly var packagesDir = Path.GetFullPath("packages");
+static readonly var repoUrl = "https://github.com/FakeItEasy/FakeItEasy";
+static readonly var versionAssembly = "./src/FakeItEasy/bin/Release/FakeItEasy.dll";
+static readonly IEnumerable<string> nuspecs = Directory.GetFiles("./src", "*.nuspec");
```
adamralph: None of these variables need to be declared with the explicit type `IEnumerable<string>`. They can all be `var`.

build.csx
```diff
@@ -0,0 +1,217 @@
+#load "packages/simple-targets-csx.5.2.0/simple-targets.csx"
+
+#r "System.Xml.Linq"
+
+//using System.Runtime.Serialization.Json;
+//using System.Xml;
+//using System.Xml.Linq;
```
adamralph: Shall we get rid of these lines?

build.csx
```diff
@@ -0,0 +1,217 @@
+#load "packages/simple-targets-csx.5.2.0/simple-targets.csx"
+
+#r "System.Xml.Linq"
```
adamralph: This appears to be redundant.

build.csx
```diff
@@ -0,0 +1,217 @@
...
+
+//using System.Runtime.Serialization.Json;
+//using System.Xml;
+//using System.Xml.Linq;
+using static SimpleTargets;
+
+static readonly var  solutionName = "FakeItEasy";
+
+static readonly var solution = "./" + solutionName + ".sln";
+static readonly var packagesDir = Path.GetFullPath("packages");
```
adamralph: Should this be `packagesDirectory`, for consistency with `logsDirectory`, `outputDirectory` and `testsDirectory`?

build.csx
```diff
@@ -0,0 +1,217 @@
...
+        if (process.ExitCode != 0)
+        {
+            throw new InvalidOperationException($"The command exited with code {process.ExitCode}.");
+        }
+    }
+}
+
+public static void RunMsBuild(string target)
+{
+    var packagesDirOption = string.IsNullOrEmpty(packagesDir) ? "" : $"/p:NuGetPackagesDirectory={packagesDir}";
```
adamralph: If `packagesDir` were to become `packagesDirectory` then `packagesDirOption` ought to change similarly.

how_to_build.md
```diff
@@ -15,52 +13,30 @@ At the time of writing the build is only confirmed to work on Windows using the
...
-    `gem update --system`
-
-1. Install bundler:
-
-    `gem install bundler`
-
-1. Install gems (after navigating to your clone root folder):
-
-    `bundler install`
-
```
adamralph: 😍

rakefile.rb
```diff
@@ -1,416 +0,0 @@
...
-
-# Load SSL cert file to enable Ruby SSL support.
-# New cert files can be had from
-#    http://curl.haxx.se/ca/cacert.pem
-# in case the current one stops working.
-def use_ssl_cert_file(ssl_cert_filename)
-  ENV["SSL_CERT_FILE"] = File.absolute_path(ssl_cert_filename)
-end
```
adamralph: 🔥 :feelsgood:

build.cmd
```diff
@@ -1,3 +1,34 @@
...
+)
+
+rem copy nuget locally
+if not exist .nuget\NuGet.exe (
+  if not exist .nuget md .nuget
+  copy %NUGET_CACHE_DIR%\NuGet.exe .nuget\NuGet.exe > nul
+)
+
+rem restore packages for build script
+.nuget\NuGet.exe restore .nuget\packages.config -PackagesDirectory .\packages -Verbosity quiet
```
thomaslevesque: Another option is to remove the `static` modifier from the methods; then we don't need to declare any variable `static`. IMO this is better than passing everything as a parameter. This is C#, sure, but it's also a script, and having to pass parameters around all the time doesn't feel very "scripty".

thomaslevesque:
> I'm not sure about all the static readonly var. I tried out how it would look if they all became var and, while the method signatures and calls are more verbose, I think I prefer it.

Another option is to remove the `static` modifier from the methods; then we don't need to declare any variable `static`. IMO this is better than passing everything as a parameter. This is C#, sure, but it's also a script, and having to pass parameters around all the time doesn't feel very "scripty".

Here's [my suggestion](https://github.com/thomaslevesque/FakeItEasy/commit/34082a577e8b0cccd6707508f8a78278d40755cf)

build.csx
blairconrad: I'd done so to highlight that they shouldn't be modified. `string[]`s could have their innards replaced more easily. @thomaslevesque had been pushing for `const` variables, so I thought it was a concern. I'm content with anything, even non-readonly vars.
So, what do we want?

* completely loosey-goosey variables and we just trust the script not to muck about with them,
* readonly variables and we trust the script not to muck about with array innards, or
* locked-down as much as we can variables?

build.csx
blairconrad: I'll change to `ProductVersion`. On your second note, if 

> `AssemblyInformationalVersion` defaults to `AssemblyFileVersion` if it is not specified. Likewise, `AssemblyInformationalVersion` and `AssemblyFileVersion` default to `AssemblyVersion` if both are not specified.

we could remove two of the attributes!

Of course, I'd rather see us pick up the version from the ReleaseNotes.md file instead and tailor the 3 values accordingly. So maybe we should leave CommonAssemblyInfo.cs alone for now.

blairconrad:
I'm happy with @thomaslevesque's [suggestion](https://github.com/FakeItEasy/FakeItEasy/pull/995#issuecomment-283006340), which appears to be "no `readonly`, no `static`, and `var` everywhere".

blairconrad:
Ah. I see @adamralph had a similar non-static, non-readonly comment (seriously, the review comments are all over the place - I keep finding them in new sections of the PR), but left the methods static and passed in the variables that needed it.

I prefer @thomaslevesque's parameterless version.  Pushing, so at least something's there before I go to work.

blairconrad:
Another great round of comments. I _think_ I addressed everything, except possibly the hypothetical "parameter vs. no parameter" donnybrook of 2017.

thomaslevesque:
LGTM 👍 

Happy to merge after you squash, unless @adamralph is unhappy with the `static`/`var`/`readonly` changes

build.csx
```diff
@@ -0,0 +1,212 @@
...
+var solution = "./" + solutionName + ".sln";
+var packagesDirectory = Path.GetFullPath("packages");
+var repoUrl = "https://github.com/FakeItEasy/FakeItEasy";
+var versionAssembly = "./src/FakeItEasy/bin/Release/FakeItEasy.dll";
+var nuspecs = Directory.GetFiles("./src", "*.nuspec");
+var gitlinks = Directory.GetDirectories("./src").Where(p => !p.EndsWith(".Shared")).Select(Path.GetFileName);
+
+var unitTests = new []
+{
+  "tests/FakeItEasy.Tests/bin/Release/FakeItEasy.Tests.dll",
```
adamralph: 😱 The indent on this line (and on all the other array initialiser argument lines) is 2 spaces instead of 4!

blairconrad:
Thanks, @thomaslevesque.

blairconrad:
> And just when you thought I was going to approve it

After my first review with you, I have never thought that again. :trollface: 

Indentation fixed. That's what I get for copying lines from a `.rb` file.

adamralph:
👍 to merge after a squash.

thomaslevesque:
Nice work @blairconrad, thanks!

blairconrad:
> Nice work @blairconrad, thanks!

Based on the comments, it was a train wreck.

You guys really improved the change. Thanks!

adamralph:
Obviously I ❤️ the change! Thanks!

adamralph:
Yep, "csharp" seemed like overreach. And everyone loves colours! 😄

blairconrad:
> And everyone loves colours!

Then why add `--no-color`? 😆

adamralph:
Some environments are lacking the love. E.g. older versions of TeamCity don't interpret the ANSI codes.

PR #998, opened 2017-03-02
Use Type.ToString() instead of Type.FullName in call descriptions

### LINKED ISSUE #994: Default fake behavior for ToString() doesn't produce the same type format as Object.ToString()
The default `ToString()` method for a non-fake object returns `GetType().ToString()`, but for a fake it returns `"Faked" <faked type>.FullName`. For non-generic types, `ToString()` and `FullName` return the same thing, but for generic types `FullName` includes the fully qualified names of type arguments, i.e.

```
Console.WriteLine(typeof(List<string>).ToString()); // System.Collections.Generic.List`1[System.String]
Console.WriteLine(typeof(List<string>).FullName); // System.Collections.Generic.List`1[[System.String, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]]
```

Fakes use the second form, which is probably too verbose for most use cases, and is inconsistent with the normal `Object.ToString()`. We should probably consider using `Type.ToString()` instead of `Type.FullName` in the default fake behavior for `ToString()`.

thomaslevesque:
Fixes #994

### DIFF
--- a/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -236,7 +236,7 @@ namespace FakeItEasy.Specs
                 .x(() => exception.Message.Should().Match(@"
 
   Assertion failed for the following call:
-    System.Collections.Generic.IDictionary`2[[System.String, *, Version=4.0.0.0, Culture=neutral, PublicKeyToken=*],[System.String, *, Version=4.0.0.0, Culture=neutral, PublicKeyToken=*]].TryGetValue(""any key"", <out parameter>)
+    System.Collections.Generic.IDictionary`2[System.String,System.String].TryGetValue(""any key"", <out parameter>)
   Expected to find it at least once but no calls were made to the fake object.
 
 "));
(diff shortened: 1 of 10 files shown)


blairconrad:
Try as I might, I got nothing to complain about. Sorry.

Thanks, @thomaslevesque!

thomaslevesque:
> Try as I might, I got nothing to complain about. Sorry.

I forgive you :wink:

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -479,27 +479,24 @@ public interface IHaveANullableParameter
...
         public static IEnumerable<object[]> StrictFakes()
         {
-            yield return new object[] { new Func<object>(() => A.Fake<object>(o => o.Strict())), "Faked " + typeof(object).FullName };
-            yield return new object[] { new Func<object>(() => A.Fake<object>(o => o.Strict())), "Faked " + typeof(object).ToString() };
-            yield return new object[] { new Func<object>(() => A.Fake<List<int>>(o => o.Strict())), "Faked " + typeof(List<int>).FullName };
-            yield return new object[] { new Func<object>(() => A.Fake<IList<int>>(o => o.Strict())), "Faked " + typeof(IList<int>).FullName };
+            yield return new object[] { new Func<object>(() => A.Fake<object>(o => o.Strict())), "Faked " + typeof(object) };
+            yield return new object[] { new Func<object>(() => A.Fake<List<int>>(o => o.Strict())), "Faked " + typeof(List<int>) };
+            yield return new object[] { new Func<object>(() => A.Fake<IList<int>>(o => o.Strict())), "Faked " + typeof(IList<int>) };
             yield return new object[] { new Func<object>(() => A.Fake<Action<int>>(o => o.Strict())), typeof(Action<int>).ToString() };
```
adamralph: Should the `ToString()` be removed from this line?

thomaslevesque: Nope. It will be passed as the fake description to `PassingAStrictFakeToAMethod`, so it needs to be a string, not a type. I omitted `ToString()` on the other lines, because it's called implicitly for the concatenation with "Faked ".

adamralph: Ah, of course. Thanks!

PR #999, opened 2017-03-03
Improved docs: Added .NET Standard 1.6 as supported platform

claudiobernasconi:
Arising from a recent question in Gitter, I send in this pull request to improve the documentation that we officially support .NET Standard 1.6.

### DIFF
--- a/docs/platform-support.md
+++ b/docs/platform-support.md
@@ -3,3 +3,4 @@
 FakeItEasy officially supports the following platforms:
 
 * .NET Framework 4.0 onwards
+* .NET Standard 1.6


thomaslevesque:
Thanks @claudiobernasconi! It looked a little silly to have a list of one target platform, it's better now :wink:

docs/platform-support.md
```diff
@@ -3,3 +3,4 @@
 FakeItEasy officially supports the following platforms:
 
 * .NET Framework 4.0 onwards
+* .NET Standard 1.6
```
adamralph: Should this line also say "onwards? Or should " onwards" be removed from the previous line?

blairconrad: I don't mind it as it is. The .NET Standards are built so that "onwards" is implied. So far the Framework 4.xes are backward compatible, but do we know that it'll continue?

But really, I wouldn't fight against either change you proposed, if you're passionate about it.

adamralph: Initially .NET Standard 2.0 was going to be breaking, but after a public backlash they decided to make it backwards compatible. But perhaps that will change with 3.0?

The .NET Framework hasn't had a breaking release in its 15 year history so I think that's less likely to happen in the future than with .NET Standard, but who knows, maybe it will happen for the first time with 5.0?

Ultimately, what the above demonstrates, is that we can't second-guess what's going to happen with future major version bumps in the platforms. That is entirely under the control of the publishers of those platforms.  I think we should state here, as we do in the NuGet package metadata, what specific versions we support. The backwards compatibility of future versions of those platforms is documented elsewhere. That means we should remove "onwards".

Phew, that's a long justification for removing the word "onwards". 😳

blairconrad:
> I think we should state here, as we do in the NuGet package metadata, what specific versions we support.

Could we instead find a way to _point at_ the information in the NuGet package metadata? I'd rather not have another thing to forget to update. ;-)

> Phew, that's a long justification for removing the word "onwards". 

Doing me proud, Ralph. Why use 2 words when 10 will do? 😀

adamralph:
> Could we instead find a way to point at the information in the NuGet package metadata? I'd rather not have another thing to forget to update. ;-)

I'd ❤️ such a solution, but nothing springs to mind.

blairconrad:
I am perhaps misunderstanding the intent, but I was thinking of something like:

------

See the "Dependencies" section of the [FakeItEasy NuGet Gallery page](https://www.nuget.org/packages/FakeItEasy) to see each released package's supported platforms and package dependencies.

adamralph:
👍 that works for me. Raised #1010

blairconrad:
This change has been released as part of [FakeItEasy 3.1.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/3.1.0).

@claudiobernasconi, look for your name in the release notes! :trophy:

PR #1001, opened 2017-03-05
Use current culture for all string formatting

### LINKED ISSUE #399: Align use of culture for string formatting
Currently there seems to be quite a mix between the usage of `CurrentCulture` and `InvariantCulture` for string formatting throughout the code base. This should be reviewed and aligned where appropriate.

blairconrad:
Fixes #399.

I built the package, running all tests, in the under the regional settings:

* ar-SA,
* fr-FR,
* he-IL,
* ka-GE,
* zh-CN, and
* cs-CZ, for @adamralph


samples/FakeItEasy.Examples/CustomArgumentValidators.cs
```diff
@@ -1,12 +1,10 @@
 namespace FakeItEasy.Examples
 {
-    using System.Globalization;
-
     public static class CustomArgumentValidators
     {
         public static string IsLongerThan(this IArgumentConstraintManager<string> validations, int length)
         {
-            return validations.Matches(x => x.Length > length, string.Format(CultureInfo.InvariantCulture, "Longer than {0}", length));
+            return validations.Matches(x => x.Length > length, "Longer than {0}", length);
```
thomaslevesque: Why not `$"Longer than {length}"`?

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
```diff
@@ -222,7 +221,7 @@ public static T Matches<T>(this IArgumentConstraintManager<T> manager, Func<T, b
         {
             Guard.AgainstNull(manager, nameof(manager));
 
-            return manager.Matches(predicate, x => x.Write(string.Format(CultureInfo.CurrentCulture, descriptionFormat, args)));
+            return manager.Matches(predicate, x => x.Write(string.Format(descriptionFormat, args)));
```
thomaslevesque: `string.Format` is redundant, it can be written as `x.Write(descriptionFormat, args)`

src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
```diff
@@ -87,7 +87,7 @@ private bool TryHandleToString(IInterceptedFakeObjectCall fakeObjectCall)
                     return false;
                 }
 
-                fakeObjectCall.SetReturnValue("Faked {0}".FormatInvariant(this.fakeManager.FakeObjectType.ToString()));
+                fakeObjectCall.SetReturnValue("Faked " + this.fakeManager.FakeObjectType);
```
thomaslevesque: Why no string interpolation here? Is it for performance, or something else?

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
```diff
@@ -30,7 +30,7 @@ public override string ToString()
             var stringValue = this.ExpectedValue as string;
             if (stringValue != null)
             {
-                return "\"{0}\"".FormatInvariant(stringValue);
+                return '"' + stringValue + '"';
```
thomaslevesque: I see several places where we enclose a value between double quotes, using various approaches. Maybe it would be worth having an `DoubleQuote` extension method...

src/FakeItEasy/OutputWriterExtensions.cs
```diff
@@ -32,7 +31,7 @@ public static IOutputWriter Write(this IOutputWriter writer, string format, para
         {
             Guard.AgainstNull(writer, nameof(writer));
 
-            writer.Write(string.Format(CultureInfo.InvariantCulture, format, args));
+            writer.Write(string.Format(format, args));
```
thomaslevesque: `writer.Write(format, args)`

adamralph:
> * cs-CZ, for @adamralph

😆

src/FakeItEasy/Creation/ProxyOptions.cs
```diff
@@ -28,11 +27,7 @@ public void AddInterfaceToImplement(Type interfaceType)
 
             if (!interfaceType.GetTypeInfo().IsInterface)
             {
-                throw new ArgumentException(
-                    string.Format(
-                        CultureInfo.CurrentCulture,
-                        "The specified type '{0}' is not an interface",
-                        interfaceType.FullNameCSharp()));
+                throw new ArgumentException($"The specified type '{interfaceType.FullNameCSharp()}' is not an interface");
```
blairconrad: Speaking of @thomaslevesque's `DoubleQuote`, perhaps a better reason to have a similar method would be to establish a consistent quoting style, rather than mechanism. There are about 10 places in the code where we use `'`s like this, instead of `"`s.
Although I think this could be a separate issue, if we care to pursue it.

CONTRIBUTING.md
```diff
@@ -54,6 +54,11 @@ Try to keep lines of code no longer than 160 characters wide. This isn't a stric
 
 Try to keep your coding style in line with the existing code. It might not exactly match your preferred style but it's better to keep things consistent. StyleCop compliance is enforced through analysis on each build and the treatment of warnings as errors. Any StyleCop settings changes or suppressions must be clearly justified.
 
+## String Formatting
+
+Unless there is a compelling reason not to, for example when serializing data to be parsed later, use the current culture when formatting strings. When string
+formatting methods have an overload that implicitly uses the current culture, opt to use it instead of specifying the culture explicitly.
```
adamralph: nitpick: this line should be part of the line above (no artificial line break).

adamralph: > When string formatting methods have an overload that implicitly uses the current culture, opt to use it instead of specifying the culture explicitly

This contradicts our code analysis ruleset. Indeed, when I run code analysis on this branch, I get plenty of these types of warnings:

> Warning	CA1305	Because the behavior of 'string.Format(string, object)' could vary based on the current user's locale settings, replace this call in 'AnyCallCallRule.DescriptionOfValidCall.get()' with a call to 'string.Format(IFormatProvider, string, params object[])'. If the result of 'string.Format(IFormatProvider, string, params object[])' will be displayed to the user, specify 'CultureInfo.CurrentCulture' as the 'IFormatProvider' parameter. Otherwise, if the result will be stored and accessed by software, such as when it is persisted to disk or to a database, specify 'CultureInfo.InvariantCulture'.

src/FakeItEasy.netstd/FakeItEasy.netstd.csproj
```diff
@@ -600,9 +600,6 @@
     <Compile Include="..\FakeItEasy\StringBuilderOutputWriter.cs">
       <Link>StringBuilderOutputWriter.cs</Link>
     </Compile>
-    <Compile Include="..\FakeItEasy\StringExtensions.cs">
-      <Link>StringExtensions.cs</Link>
-    </Compile>
```
adamralph: ❤️

src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
adamralph: I think it's fine to leave the `+` operator here. It's optimised anyway, and it does sometimes read better than the equivalent bling string.

tests/FakeItEasy.Tests/Creation/FakeOptionsConstraints.cs
```diff
@@ -1,14 +0,0 @@
...
-    using FakeItEasy.Creation;
-
-    public static class FakeOptionsConstraints
-    {
-        internal static IProxyOptions HasArgumentsForConstructor(this IArgumentConstraintManager<IProxyOptions> scope, IEnumerable<object> argumentsForConstructor)
-        {
-            return scope.Matches(x => argumentsForConstructor.SequenceEqual(x.ArgumentsForConstructor), "Constructor arguments ({0})".FormatInvariant(string.Join(", ", argumentsForConstructor.Select(x => x.ToString()).ToArray())));
-        }
-    }
-}
```
adamralph: Why has this been removed as part of this PR?

CONTRIBUTING.md
blairconrad: The previous line is already 158 characters. We like a 160 character limit, as set in `.editorconfig`. Why would we inconvenience ourselves with crazy long lines in the `.md` files?
Are you worried about ugly diffs if ever we reflow?

blairconrad: Oh, bother. I forgot to run the analysis. Thanks.

I'm happy to disable CA1305. I mean, it doesn't complain about string interpolation or implicit `ToString`s anyhow. Why should explicit `string.Format` calls be persecuted?

tests/FakeItEasy.Tests/Creation/FakeOptionsConstraints.cs
blairconrad: It wasn't used (see my [first commit](https://github.com/FakeItEasy/FakeItEasy/pull/1001/commits/806217100dfc249e24f89be04a01fa3c55afbbe2)). I preferred to remove it rather than convert the usage. I can submit a separate PR if you like.

CONTRIBUTING.md
adamralph: > We like a 160 character limit, as set in .editorconfig

That's true, but this is an artificial break in the middle of a literal, and that's not something we'd do in C# either. In MarkDown it happens to be ignored in rendering but that's coincidental. If you look through the rest of the file, we don't add line breaks like these anywhere, so this would be the first incidence of it.

adamralph: >  it doesn't complain about string interpolation or implicit ToStrings anyhow. Why should explicit string.Format calls be persecuted?

Good point. I'm 👍 to disable CA1305 as part of this PR then.

tests/FakeItEasy.Tests/Creation/FakeOptionsConstraints.cs
adamralph: Ah, in it's own commit. That's fine then. I guess it's completely unrelated to the rest of the commits, but I've no problem with it being part of this PR.

samples/FakeItEasy.Examples/CustomArgumentValidators.cs
thomaslevesque: Well, the "real" `Matches` method takes an `Action<IOutputWriter>`... there is an extension method that takes a string, and another that takes a format string and arguments :wink:
Whether the formatting is done at the call site (using string interpolation) or in the called method doesn't really matter, but I find string interpolation more readable that "classic" string formatting.

src/FakeItEasy/OutputWriterExtensions.cs
thomaslevesque: Oh, of course. I didn't realize this code was already in the `Write` method ^^

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
thomaslevesque: I don't feel strongly about it, it was just an idea... the new style looks good to me 👍

src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
thomaslevesque: > Would you like me to?

No, it's fine as it is. I was just wondering.

CONTRIBUTING.md
thomaslevesque: > Why should explicit string.Format calls be persecuted?

Because the people who designed the code analysis rules vowed to make developers' lives miserable. I can't think of any other reason :P

thomaslevesque:
> cs-CZ, for @adamralph

Does it pass the Turkey test ? :wink:

https://blog.codinghorror.com/whats-wrong-with-turkey/
http://www.moserware.com/2008/02/does-your-code-pass-turkey-test.html

tests/FakeItEasy.IntegrationTests/StringArgumentConstraints.cs
```diff
@@ -1,19 +1,17 @@
...
     using System;
-    using System.Globalization;
 
     public static class StringArgumentConstraints
     {
         public static string StartsWith(this IArgumentConstraintManager<string> scope, string beginning)
         {
-            return scope.Matches(
-                x => x.StartsWith(beginning, StringComparison.CurrentCulture), string.Format(CultureInfo.CurrentCulture, "Starts with \"{0}\"", beginning));
+            return scope.Matches(x => x.StartsWith(beginning, StringComparison.CurrentCulture), $@"Starts with ""{beginning}""");
```
blairconrad: I restored the `StringComparison.CurrentCulture` because

1. I was overeager; it's not really anything to do with this issue, and
2. removing it introduced a code analysis warning that I didn't care for

tests/FakeItEasy.Tests/Creation/FakeOptionsConstraints.cs
blairconrad: I wouldn't say completely unrelated. I _did_ remove the usage of `FormatInvariant`.

blairconrad:
> Does it pass the Turkey test ?

I never checked. AFAICT we're not changing case of strings, nor doing case-insensitive comparisons.

blairconrad:
> Does it pass the Turkey test ? 😉

Just for you, I checked. It does.

But it still does not pass the Turing test.

tests/FakeItEasy.IntegrationTests/StringArgumentConstraints.cs
thomaslevesque: Didn't the warning go away when you disabled CA1305?

thomaslevesque:
Thanks @blairconrad, looks good to me!
👍 to merge once @adamralph approves the changes.

tests/FakeItEasy.IntegrationTests/StringArgumentConstraints.cs
blairconrad: It wasn't CA1305. IIRC, [CA1307](https://msdn.microsoft.com/en-us/library/bb386080.aspx?f=255&MSPPError=-2147217396).

blairconrad:
Thanks, @thomaslevesque. I'll have to squash.

tests/FakeItEasy.IntegrationTests/StringArgumentConstraints.cs
thomaslevesque: Ah, ok. Well, I guess we could also suppress CA1307...

blairconrad: I considered, but it was only the one instance that was giving me grief, and string _comparison_ is a different beast than _formatting_, so I shied away.
I'd make a separate PR if we wanted to pursue this…

blairconrad:
I pre-squashed (at work) and would rather not redo the work later, so
I'm taking a liberty and pushing now. I copied the pre-squash branch to
https://github.com/blairconrad/FakeItEasy/tree/align-string-culture-before-squash.
Here's the latest commit: https://github.com/blairconrad/FakeItEasy/commit/981315d754df07dc36dcfdddedc20d347d290f08

CONTRIBUTING.md
adamralph: The text in markdown paragraph is a literal value. It's like a `@"..."` string in C#. Perhaps it's bad analogy.

tests/FakeItEasy.IntegrationTests/StringArgumentConstraints.cs
adamralph: I agree with separate PR. This, and #399, are only about formatting, which is a different purpose to that of comparison.

tests/FakeItEasy.Tests/Creation/FakeOptionsConstraints.cs
adamralph: Ah, of course. All good then.

src/FakeItEasy/Creation/ProxyOptions.cs
adamralph: Raised https://github.com/FakeItEasy/FakeItEasy/issues/1011

adamralph:
Thanks @blairconrad!

CONTRIBUTING.md
blairconrad: Oh. This is not an equivalence that would ever have occurred to me. Thanks for explaining.

blairconrad:
Thanks for the reviews and merges, folk!

PR #1006, opened 2017-03-09
Fix various code analysis warnings

blairconrad:
I didn't track down when they were all introduced, because I'm a little lazy.


blairconrad:
Oh, I didn't bother with the CA1305, since that should be cleared up by #1001.

blairconrad:
Thank you.

PR #1007, opened 2017-03-09
Added more detailed object exception output for proxy and its tag

rocifier:
Fixes #530.

Added more detailed object exception output for proxy and its tag when they are not recognized as a fake, and moved the exception message to the resx. Adjusted unit tests to match.

PR #1013, opened 2017-03-09
Fix custom tool for resource files

thomaslevesque:
As discussed in https://github.com/FakeItEasy/FakeItEasy/pull/1007#discussion_r105262972


thomaslevesque:
Oops... Looks like the code generated by `ResXFileCodeGenerator` doesn't work on .NET Core. We'll need to use a T4 template instead.

blairconrad:
> Oops... Looks like the code generated by ResXFileCodeGenerator doesn't work on .NET Core. We'll need to use a T4 template instead.

Yeah, I was going to say, once the issue was created. But you're too fast.

thomaslevesque:
OK, I made the changes to use T4 templates instead of `ResXFileCodeGenerator`. At this point the code isn't regenerated when the resx file changes. It's possible to run the transformation on build if we want, though.

I also have [this extension](https://marketplace.visualstudio.com/items?itemName=ThomasLevesque.AutoRunCustomTool) which can run the transformation when the source file is modified, without the need to rebuild.

thomaslevesque:
> It's possible to run the transformation on build if we want, though.

I was following these instructions: https://msdn.microsoft.com/en-us/library/ee847423.aspx

But it requires the [Modeling SDK](https://www.microsoft.com/en-us/download/details.aspx?id=48148), which isn't installed by default. Another option would be to invoke TextTransform.exe directly from a build target, but it's ugly...

src/FakeItEasy/Creation/CastleDynamicProxy/DynamicProxyResources.tt
```diff
@@ -0,0 +1,123 @@
...
+    // Keep only the first line
+    int endOfLine = value.IndexOfAny(new[] { '\r', '\n' });
+    if (endOfLine >= 0)
+    {
+        value = value.Substring(0, endOfLine);
+        truncated = true;
+    }
+    if (value.Length > 100)
+    {
+        value = value.Substring(0, 97);
```
blairconrad: you could go up to 99 if you added an ellipsis instead of "..." :wink:

src/FakeItEasy/Creation/CastleDynamicProxy/DynamicProxyResources.tt
```diff
@@ -0,0 +1,123 @@
...
+
+namespace FakeItEasy.Creation.CastleDynamicProxy {
+    using System;
+    using System.Reflection;
+
+
+    /// <summary>
+    ///   A strongly-typed resource class, for looking up localized strings, etc.
+    /// </summary>
+    // This class was auto-generated by the StronglyTypedResourceBuilder
```
blairconrad: was it?

src/FakeItEasy/Creation/CastleDynamicProxy/DynamicProxyResources.tt
```diff
@@ -0,0 +1,123 @@
...
+    // class via a tool like ResGen or Visual Studio.
+    // To add or remove a member, edit your .ResX file then rerun ResGen
+    // with the /str option, or rebuild your VS project.
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "4.0.0.0")]
```
blairconrad: Does this provide any benefit?

src/FakeItEasy/ExceptionMessages.tt
```diff
@@ -0,0 +1,122 @@
+﻿<#@ template debug="false" hostspecific="true" language="C#" #>
```
blairconrad: I was surprised to see two templates. I'd've guessed that the one would do the job for all the resxes. Then again, I've never _used_ T4, only read about it.

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -276,10 +284,26 @@
...
+  <ItemGroup>
+    <EmbeddedResource Include="Creation\CastleDynamicProxy\DynamicProxyResources.resx">
+      <LastGenOutput>DynamicProxyResources.Designer.cs</LastGenOutput>
+    </EmbeddedResource>
+    <EmbeddedResource Include="ExceptionMessages.resx">
+      <SubType>Designer</SubType>
+    </EmbeddedResource>
+  </ItemGroup>
+  <ItemGroup>
+    <Service Include="{508349B6-6B84-4DF5-91F0-309BEEBAD82D}" />
```
blairconrad: ?

blairconrad:
> the code isn't regenerated when the resx file changes

Disappointing, but probably livewithable for a bit. Since we're committing the `.cs` files, we can always build. Edits to the resxes should be rare, and requiring the manual step to rebuild should only be a minor inconvenience, no? It's an improvement on the current situation.

src/FakeItEasy/Creation/CastleDynamicProxy/DynamicProxyResources.tt
thomaslevesque: True!

src/FakeItEasy/Creation/CastleDynamicProxy/DynamicProxyResources.tt
thomaslevesque: Nope... forgot to update the comment ;)

src/FakeItEasy/Creation/CastleDynamicProxy/DynamicProxyResources.tt
thomaslevesque: I'm not sure. I _think_ it keeps StyleCop/FxCop from looking at it.

src/FakeItEasy/ExceptionMessages.tt
thomaslevesque: Well, basically, normally a T4 template produces only one output file. It might be possible to produce several files with the same template, but I don't know how. I can try to find out.

thomaslevesque:
> Disappointing, but probably livewithable for a bit. Since we're committing the .cs files, we can always build. Edits to the resxes should be rare, and requiring the manual step to rebuild should only be a minor inconvenience, no? It's an improvement on the current situation.

I guess.

In fact, I'm not sure why we bother with resx files at all, since we don't localize the messages...

blairconrad:
> In fact, I'm not sure why we bother with resx files at all, since we don't localize the messages...

My guess is that it's because we _could_ localize the messages.
Hey! You know a language that's not English. :wink:

adamralph:
Is T4 the "official" way to handle resources for project.json? What about in 2017 csproj?

adamralph:
> In fact, I'm not sure why we bother with resx files at all, since we don't localize the messages...

That sounds like a strong YAGNI argument, and it means we could drop all of this.

thomaslevesque:
> Is T4 the "official" way to handle resources for project.json?

Not that I know of. As far as I know, there's no official way. But anyway, we don't need to have the code generation in both the net40 and netstd projects (and probably shouldn't). The netstd can just include the generated file (this is what I did in this PR)

> What about in 2017 csproj?

I assume it works the same as in "classic" csproj; haven't tried yet.

blairconrad:
>  The netstd can just include the generated file (this is what I did in this PR)

I hadn't noticed. Crafty!

> That sounds like a strong YAGNI argument, and it means we could drop all of this.

Actually, I don't hate the idea of "dropping all of this". But what would we be left with?
Right now we have a resx file and a designer file, and many people will think that the latter is automatically generated from the former.
Do you think we should keep everything as is?

For a fun exercise this morning, I took a few minutes to whip up a variation of a technique I'd read about for providing localizable messages in other languages. We _could_ replace the ExceptionMessages resx and designer with this sort of thing (and similarly for the others):

```c#
namespace FakeItEasy
{
    internal static class ExceptionMessage
    {
        public static readonly NoPlaceholderMessage ArgumentNameDoesNotExist = new NoPlaceholderMessage(
            "The specified argument name does not exist in the ArgumentList.");

        public static readonly NoPlaceholderMessage FakeCreationExceptionDefault = new NoPlaceholderMessage(
            "Unable to create fake object.");

        public static readonly TwoPlaceholderMessage WrongConstructorExpressionTypeMessage = new TwoPlaceholderMessage(
            "Supplied constructor is for type {0}, but must be for {1}.");

        public static readonly NoPlaceholderMessage MethodMismatchWhenPlayingBackRecording =
            new NoPlaceholderMessage(
                "The method of the call did not match the method of the recorded call, the recorded sequence is no longer valid.");

        public static readonly NoPlaceholderMessage NumberOfOutAndRefParametersDoesNotMatchCall =
            new NoPlaceholderMessage(
                "The number of values for out and ref parameters specified does not match the number of out and ref parameters in the call.");

        public static readonly NoPlaceholderMessage NowCalledDirectly =
            new NoPlaceholderMessage(
                "The Now-method on the event raise is not meant to be called directly, only use it to register to an event on a fake object that you want to be raised.");

        public static readonly NoPlaceholderMessage NonConstructorExpressionMessage =
            new NoPlaceholderMessage("Only expression of the type ExpressionType.New (constructor calls) are accepted.");

        public static readonly NoPlaceholderMessage NoMoreRecordedCalls =
            new NoPlaceholderMessage(
                "All the recorded calls has been applied, the recorded sequence is no longer valid.");

        public static readonly NoPlaceholderMessage WrongNumberOfArgumentNames = new NoPlaceholderMessage(
            "The number of argument names does not match the number of arguments.");


        internal class NoPlaceholderMessage
        {
            private readonly string message;

            public NoPlaceholderMessage(string message)
            {
                this.message = message;
            }

            public string Format()
            {
                return this.message;
            }

            public static implicit operator string(NoPlaceholderMessage message)
            {
                return message.Format();
            }
        }

        internal class TwoPlaceholderMessage
        {
            private string message;

            public TwoPlaceholderMessage(string message)
            {
                this.message = message;
            }

            public string Format(object arg1, object arg2)
            {
                return string.Format(this.message, arg1, arg2);
            }
        }
    }
}
```

Of course, the fields could be properties. If ever we wanted to localize, it's just a matter of adding another type to hold the messages. The getters would then either returned their translated message or default back to English.

I know, we're getting perilously close to reimplementing localized resource handling…

One nice thing is that the messages are slightly easier to use, in that you're forced to pass the correct number of arguments to their `Format` method.

It's probably not worthwhile, but was a fun diversion. Thanks for listening!

thomaslevesque:
@blairconrad I like the idea!

> If ever we wanted to localize, it's just a matter of adding another type to hold the messages. The getters would then either returned their translated message or default back to English.

I'm not sure I understand that part. What "other type" would you be adding?

thomaslevesque:
Also, I would add an implicit conversion from `String` to `(No|One|Two)PlaceHolderMessage`; it would make the declarations shorter.

adamralph:
Do we have any intention to localise the messages?

thomaslevesque:
> Do we have any intention to localise the messages?

I don't know. I personally don't like localized error messages, because the translations are often pretty bad, and it makes it harder to find information on the Internet about the error. But I guess it can be helpful for developers who don't understand English very well.

adamralph:
I think the cost ("harder to find information on the Internet about the error") the benefit. I believe these messages should not be localised.

IMHO localisation is something for end user apps, i.e. UI's.

I therefore vote to drop all this complexity and inline the strings instead.

thomaslevesque:
> I therefore vote to drop all this complexity and inline the strings instead.

👍 

I'll create an issue.

thomaslevesque:
> I therefore vote to drop all this complexity and inline the strings instead.

BTW, when you mean "inline", do you mean including the string literal directly where it's used? This would cause repetition, as some of the strings are used more than once. I think I'd rather have the strings as constants in a static class, or something like what @blairconrad proposed.

adamralph:
Keeping them all centrally in that manner is OK.

blairconrad:
>  What "other type" would you be adding?

One option for localization would be to make a class per language, each extending a base class (effectively English) that would provide untranslated messages. There are other schemes, of course.

rocifier:
> One option for localization would be to make a class per language, each extending a base class (effectively English) that would provide untranslated messages. There are other schemes, of course.

Maybe we can start with just a single class, later if someone wants to localize and we decide that is valid, we could have them refactor by doing this.

blairconrad:
Sure, if ever we decide that localized messages are desirable, we can explore the mechanism at that point.

But I'm with @adamralph and @thomaslevesque. Probably more trouble than it's worth. Especially for a library that's so focused on having users write English sentences…

adamralph:
Are we dropping this in favour of #1018?

thomaslevesque:
> Are we dropping this in favour of #1018?

Looks like it. Closing.

PR #1016, opened 2017-03-10
Provide FakeManager.ClearRecordedCalls

### LINKED ISSUE #978: Make it possible to reset recorded calls on an object

blairconrad:
Fixes #978.
Closes #979.

### DIFF
--- a/tests/FakeItEasy.Tests/FakeFacadeTests.cs
+++ b/tests/FakeItEasy.Tests/FakeFacadeTests.cs
@@ -89,6 +89,18 @@ namespace FakeItEasy.Tests
         }
 
         [Fact]
+        public void ClearRecordedCalls_should_be_guarded()
+        {
+            // Arrange
+
+            // Act
+
+            // Assert
+            Expression<System.Action> call = () => this.facade.ClearRecordedCalls(A.Dummy<object>());
+            call.Should().BeNullGuarded();
+        }
+
+        [Fact]
         public void Initialize_should_call_initialize_on_fake_initializer()
         {
             // Arrange
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -18,9 +18,10 @@ namespace FakeItEasy.Core
     {
         private readonly CallRuleMetadata[] postUserRules;
         private readonly CallRuleMetadata[] preUserRules;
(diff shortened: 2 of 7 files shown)


tests/FakeItEasy.Tests/FakeFacadeTests.cs
```diff
@@ -89,6 +89,18 @@ public void ClearConfiguration_should_be_guarded()
...
 
         [Fact]
+        public void ClearRecordedCalls_should_be_guarded()
+        {
+            // Arrange
+
+            // Act
+
+            // Assert
+            Expression<System.Action> call = () => this.facade.ClearRecordedCalls(A.Dummy<object>());
```
thomaslevesque: Shouldn't this statement be in the Arrange phase?

In fact, the `.Should().BeNullGuarded()` method doesn't lend itself very well to AAA style tests, since it does the Act and Assert phases in one instruction...

blairconrad: I agree on both points.

However, the code follows along with all the other tests I've seen in which the AAA phases are spelled out in comments.

I don't mind changing, but figure if we do, we might as well sweep all of the instances. Defer?

blairconrad:
Also, I'll squash before merge.

tests/FakeItEasy.Tests/FakeFacadeTests.cs
thomaslevesque: > Defer?

Fine by me 👍

blairconrad:

thomaslevesque:
Thanks @blairconrad and @erik-kallen !

blairconrad:
Thanks for the merge, @thomaslevesque.

PR #1018, opened 2017-03-11
Extract strings from resources

### LINKED ISSUE #1017: Drop resource files and move message strings to the code
As discussed in #1013 

Error messages are currently in a resource file (.resx), but are not localized and there's no plan to localize them.

Also, the custom tool that generates code from the resource files produces code that is not compatible with .NET Standard 1.6, so it's currently disconnected. This means that adding a string to the resource file no longer updates the corresponding code file.

Since we don't need the features provided by resx files and can't use the built-in custom tool, it would make sense to convert the resources to string constants in a class. We could also use the approach suggested by @blairconrad in https://github.com/FakeItEasy/FakeItEasy/pull/1013#issuecomment-285865877

thomaslevesque:
Fixes #1017 

Note that this PR will conflict with #1007, so it's probably better to merge #1007 first.


thomaslevesque:
The last commit changes the `*PlaceholderMessage` classes to structs, to avoid unnecessary allocations.

blairconrad:
> so it's probably better to merge #1007 first.

:+1:

thomaslevesque:
> No love for the analyzer?

Oh, I forgot about it. The analyzer doesn't have the same problem; it doesn't target .NET Core, so the code generated by ResXFileCodeGenerator compiles just fine. Do you want to get rid of the .resx anyway?

blairconrad:
>  The analyzer doesn't have the same problem; it doesn't target .NET Core, so the code generated by ResXFileCodeGenerator compiles just fine. Do you want to get rid of the .resx anyway?

Of course you're right. Silly of me.

I don't mind this being a problem for Future Us.

src/FakeItEasy/Messages/ThreePlaceholderMessage.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace FakeItEasy.Messages
+{
+    internal struct ThreePlaceholderMessage
```
blairconrad: unused?

blairconrad:
Hey, what are your thoughts on [making the message types generic](https://github.com/blairconrad/FakeItEasy/commit/aa9b852ebd8c2a9487cb9b1651b487f3a61ecf80)?

src/FakeItEasy/Messages/ThreePlaceholderMessage.cs
thomaslevesque: Ah, yes. It was used before I removed the unused resource. I could remove it, but on the other hand, we might need it later... I don't mind either way.

thomaslevesque:
> Hey, what are your thoughts on making the message types generic?

I considered it. It's an interesting idea, but I don't know if it's worth it.

thomaslevesque:
Now that I think of it, we could do something even simpler:

        public static string ProxyIsSealedTypeMessage(Type type) =>
             $"The type of proxy ""{type}"" is sealed.";

This approach has several benefits:
- no need for `*PlaceholderMessage` types
- strongly typed
- named parameters (which we wouldn't have with the generic `*PlaceholderMessage` types)

blairconrad:
@thomaslevesque, beautiful!

src/FakeItEasy/Creation/CastleDynamicProxy/DynamicProxyResources.cs
```diff
@@ -0,0 +1,22 @@
...
+        public static string ProxyIsSealedTypeMessage(Type type) =>
+            $@"The type of proxy ""{type}"" is sealed.";
+
+        public static string ProxyIsValueTypeMessage(Type type) =>
+            $"The type of proxy must be an interface or a class but it was {type}.";
+
+        public static string ProxyTypeWithNoDefaultConstructorMessage(Type type) =>
+            $"No usable default constructor was found on the type {type}.";
+    }
+}
```
blairconrad: Good, overall. As a minor point, I wonder if, since all we have in here are messages, the class name shouldn't become `DynamicProxyMessages` and the members could drop `Message`. (Ditto for the members of `ExceptionMessages`.)

thomaslevesque: Agreed.

thomaslevesque: Done

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -63,6 +63,8 @@
     <Compile Include="AsyncReturnValueConfigurationExtensions.cs" />
     <Compile Include="Compatibility\MethodInfoWrapper.cs" />
     <Compile Include="Configuration\PropertyExpressionHelper.cs" />
+    <Compile Include="Creation\CastleDynamicProxy\DynamicProxyResources.cs" />
```
blairconrad: Rename the file to match the class?

thomaslevesque: Ah, I thought I had... I used R#'s rename command, which normally renames the file, but it didn't work in this case.

thomaslevesque: I amended the last commit

blairconrad:
Thanks, @thomaslevesque. 🥇

blairconrad:
Labelling as **on-hold** while we wait for #1007 to merge.

blairconrad:
Thanks, @thomaslevesque.

blairconrad:
Whoops. I didn't notice that the commits hadn't been squashed. I should've held off on the merge.
Sorry!

adamralph:
> +80 −818

😍

PR #1022, opened 2017-03-13
Strongly typed WhenArgumentsMatch

### LINKED ISSUE #1021: Add strongly typed overloads for WhenArgumentsMatch
As discussed in https://github.com/FakeItEasy/FakeItEasy/issues/1020#issuecomment-286133698

Currently we have strongly typed overloads for `Invokes`, `Throws` and `ReturnsLazily`. It would be helpful to add similar overloads for `WhenArgumentsMatch`, so that instead of this:

    .WhenArgumentsMatch(args => args.Get<T1>(0).Id == args.Get<T2>(1).OtherId)

we can write this:

    .WhenArgumentsMatch((T1 p1, T2 p2) => p1.Id == p2.OtherId)

thomaslevesque:
Fixes #1021

### DIFF
--- a/src/FakeItEasy/ArgumentValidationConfigurationExtensions.cs
+++ b/src/FakeItEasy/ArgumentValidationConfigurationExtensions.cs
@@ -1,5 +1,6 @@
 namespace FakeItEasy
 {
+    using System;
     using FakeItEasy.Configuration;
 
     /// <summary>
@@ -19,6 +20,72 @@ namespace FakeItEasy
             Guard.AgainstNull(configuration, nameof(configuration));
 
             return configuration.WhenArgumentsMatch(x => true);
+        }
+
+        /// <summary>
+        /// Configures the call to be accepted when the specified predicate returns true.
+        /// </summary>
+        /// <typeparam name="TInterface">The type of the interface.</typeparam>
+        /// <typeparam name="T1">The type of the first argument.</typeparam>
+        /// <param name="configuration">The configuration.</param>
+        /// <param name="argumentsPredicate">The argument predicate.</param>
+        /// <returns>A configuration object.</returns>
+        public static TInterface WhenArgumentsMatch<TInterface, T1>(
+            this IArgumentValidationConfiguration<TInterface> configuration,
+            Func<T1, bool> argumentsPredicate)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Configures the call to be accepted when the specified predicate returns true.
+        /// </summary>
+        /// <typeparam name="TInterface">The type of the interface.</typeparam>
+        /// <typeparam name="T1">The type of the first argument.</typeparam>
+        /// <typeparam name="T2">The type of the second argument.</typeparam>
+        /// <param name="configuration">The configuration.</param>
+        /// <param name="argumentsPredicate">The argument predicate.</param>
+        /// <returns>A configuration object.</returns>
+        public static TInterface WhenArgumentsMatch<TInterface, T1, T2>(
+            this IArgumentValidationConfiguration<TInterface> configuration,
+            Func<T1, T2, bool> argumentsPredicate)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Configures the call to be accepted when the specified predicate returns true.
+        /// </summary>
+        /// <typeparam name="TInterface">The type of the interface.</typeparam>
+        /// <typeparam name="T1">The type of the first argument.</typeparam>
+        /// <typeparam name="T2">The type of the second argument.</typeparam>
+        /// <typeparam name="T3">The type of the third argument.</typeparam>
+        /// <param name="configuration">The configuration.</param>
+        /// <param name="argumentsPredicate">The argument predicate.</param>
+        /// <returns>A configuration object.</returns>
+        public static TInterface WhenArgumentsMatch<TInterface, T1, T2, T3>(
+            this IArgumentValidationConfiguration<TInterface> configuration,
+            Func<T1, T2, T3, bool> argumentsPredicate)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// Configures the call to be accepted when the specified predicate returns true.
+        /// </summary>
+        /// <typeparam name="TInterface">The type of the interface.</typeparam>
+        /// <typeparam name="T1">The type of the first argument.</typeparam>
+        /// <typeparam name="T2">The type of the second argument.</typeparam>
+        /// <typeparam name="T3">The type of the third argument.</typeparam>
+        /// <typeparam name="T4">The type of the fourth argument.</typeparam>
+        /// <param name="configuration">The configuration.</param>
+        /// <param name="argumentsPredicate">The argument predicate.</param>
+        /// <returns>A configuration object.</returns>
+        public static TInterface WhenArgumentsMatch<TInterface, T1, T2, T3, T4>(
+            this IArgumentValidationConfiguration<TInterface> configuration,
+            Func<T1, T2, T3, T4, bool> argumentsPredicate)
+        {
+            throw new NotImplementedException();
         }
     }
 }
--- a/src/FakeItEasy/ArgumentValidationConfigurationExtensions.cs
+++ b/src/FakeItEasy/ArgumentValidationConfigurationExtensions.cs
@@ -1,6 +1,9 @@
 namespace FakeItEasy
 {
     using System;
+#if FEATURE_NETCORE_REFLECTION
+    using System.Reflection;
+#endif
     using FakeItEasy.Configuration;
 
     /// <summary>
@@ -8,6 +11,8 @@ namespace FakeItEasy
     /// </summary>
     public static class ArgumentValidationConfigurationExtensions
     {
+        private const string NameOfWhenArgumentsMatchFeature = "when arguments match";
+
         /// <summary>
         /// Specifies that a call to the configured call should be applied no matter what arguments
         /// are used in the call to the faked object.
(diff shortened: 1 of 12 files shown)


src/FakeItEasy/Configuration/ArgumentCollection.cs
```diff
@@ -68,6 +69,8 @@ public int Count
         /// </summary>
         public IEnumerable<string> ArgumentNames { get; }
 
+        internal MethodInfo Method { get; }
```
blairconrad: It feels a little weird to have a `Method` live on an `ArgumentCollection`. But I guess if I ask myself, "arguments for what?", it seems okay.

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
```diff
@@ -0,0 +1,258 @@
...
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<IFoo>());
+
+            "And a fake method with 1 parameter is configured to return a value when called with any arguments"
+                .x(() => A.CallTo(() => fake.Bar(0))
+                            .WithAnyArguments()
+                            .Returns(42));
+
+            "When the method is called with any argument"
```
blairconrad: perhaps include `argument` in the message so we know which step fails (not that it will)?

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
```diff
@@ -0,0 +1,258 @@
...
+            int Bar(int a, string b, bool c);
+
+            int Bar(int a, string b, bool c, object d);
+        }
+
+        [Scenario]
+        [InlineData(null, 0, 0)]
+        [InlineData(null, 0, 1)]
+        [InlineData(null, 0, 2)]
+        [InlineData(null, 0, 10)]
```
blairconrad: I found the `0` confusing. Consider omitting and making `result` the last parameter?

src/FakeItEasy/Configuration/ArgumentCollection.cs
thomaslevesque: Yeah, I'm not thrilled about it. But it's needed to verify the delegates signatures, and I didn't see a better way to do it. What bothers me a little is that `ArgumentCollection` has two constructors, one of which doesn't take a `MethodInfo`. It's only used in unit tests, but we could easily end up with a null method if we're not careful.

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
thomaslevesque: > I found the 0 confusing

But not the null?

> Consider omitting and making result the last parameter?

In fact, I put `argument` as the last parameter because it's the only "real" parameter. The other values will be ignored. It would be nice if we could specify only one argument...

Would it read better if I wrote it like  `[InlineData(default(IFoo), default(int), 0)]` instead?

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
thomaslevesque: Sure, I can do that.

src/FakeItEasy/Configuration/ArgumentCollection.cs
blairconrad: Here's an idea **that I have not spiked**. `Matches(IFakeObjectCall)` is our sledgehammer, essentially. It can do anything any of the other calls do.
Our internal code could always use this. Everything internal gets passed an `IFakeObjectCall` and we augment the predicates to deal with them.

Alternatively, we could at least drop the other constructor for `ArgumentCollection`.

src/FakeItEasy/ArgumentValidationConfigurationExtensions.cs
```diff
@@ -20,5 +26,127 @@ public static TInterface WithAnyArguments<TInterface>(this IArgumentValidationCo
...
+        /// Configures the call to be accepted when the specified predicate returns true.
+        /// </summary>
+        /// <typeparam name="TInterface">The type of the interface.</typeparam>
+        /// <typeparam name="T1">The type of the first argument.</typeparam>
+        /// <param name="configuration">The configuration.</param>
+        /// <param name="argumentsPredicate">The argument predicate.</param>
+        /// <returns>A configuration object.</returns>
+        public static TInterface WhenArgumentsMatch<TInterface, T1>(
+            this IArgumentValidationConfiguration<TInterface> configuration,
+            Func<T1, bool> argumentsPredicate)
```
blairconrad: Seems redundant, as one could just have `A<T>.That.Matches(…)`, but I guess it would look funny without it.

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
thomaslevesque: > You can supply only one argument:

Ah, I wasn't aware of that. Thanks!

src/FakeItEasy/Configuration/ArgumentCollection.cs
thomaslevesque: > Here's an idea that I have not spiked. Matches(IFakeObjectCall) is our sledgehammer, essentially. It can do anything any of the other calls do.
> Our internal code could always use this. Everything internal gets passed an IFakeObjectCall and we augment the predicates to deal with them.

OK, but how do I get to the `IFakeObjectCall`? It's not exposed in any way by `IArgumentValidationConfiguration<TInterface>`...

> Alternatively, we could at least drop the other constructor for ArgumentCollection.

blairconrad: > We could, but it would make the tests harder to write, since they would need to provide a `MethodInfo`

That's true, but I'd rather tests were a little harder to write than the doubt of the extra constructor. Also, I wonder how many of the tests are [Gulliver's Travels Tests](http://dhickey-ie-archive.azurewebsites.net/post/2014/03/03/gulliver-s-travels-tests.aspx)?

(I'm happy to investigate, but it'll probably be slower than you'd like.)

> OK, but how do I get to the `IFakeObjectCall`? It's not exposed in any way by `IArgumentValidationConfiguration<TInterface>`...

It's true. I told you I hadn't tried it! :grinning:
I don't have an answer. I feel like there should be a way, but I'm coming up short right now.

thomaslevesque:
I made the requested changes (amended the relevant commit).

Regarding the `ArgumentCollection.Method` property, it seems none of us has a better idea. Shall I leave it as is?

thomaslevesque:
> Regarding the ArgumentCollection.Method property, it seems none of us has a better idea. Shall I leave it as is?

Oh, I forgot about removing the constructor with no `MethodInfo`. Should it be done in this PR, or in another?

thomaslevesque:
> Oh, I forgot about removing the constructor with no MethodInfo. Should it be done in this PR, or in another?

I added a commit for this; I can always move it to another branch if we prefer to do it in another PR.

I was surprised to see that the reflection API lends itself pretty well to faking... everything seems to be abstract or virtual, which was *very* convenient for adapting the tests.

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            "When the method is called with arguments that don't satisfy the predicate"
+                .x(() => result = fake.Bar(3, "hello"));
+
+            "Then it returns the default value"
+                .x(() => result.Should().Be(0));
+        }
+
+        [Scenario]
+        public static void StronglyTypedWhenArgumentsMatchWith1ArgSuccess(IFoo fake, int result)
```
blairconrad: I'm sure it all works, but is it worthwhile checking behaviour when there's mismatched signatures?

blairconrad:
@thomaslevesque, sorry for the delay.

Very nice fake methods.

And I think that removing the methodless `ArgumentCollection` constructor in this PR was exactly the thing to do.

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
thomaslevesque: Good idea. I'll do that.

thomaslevesque:
Added 2 fixup commits

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
```diff
@@ -220,6 +263,26 @@ public static void StronglyTypedWhenArgumentsMatchWith3ArgsFailure(IFoo fake, in
         }
 
         [Scenario]
+        public static void StronglyTypedWhenArgumentsMatchWith3ArgsWrongSignature(IFoo fake, Exception exception)
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<IFoo>());
+
+            "And a fake method with 2 parameters is configured with an arguments predicate with an incompatible signature"
```
blairconrad: 3 parameters

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
```diff
@@ -254,5 +317,25 @@ public static void StronglyTypedWhenArgumentsMatchWith4ArgsFailure(IFoo fake, in
...
                 .x(() => result.Should().Be(0));
         }
+
+        [Scenario]
+        public static void StronglyTypedWhenArgumentsMatchWith4ArgsWrongSignature(IFoo fake, Exception exception)
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<IFoo>());
+
+            "And a fake method with 2 parameters is configured with an arguments predicate with an incompatible signature"
```
blairconrad: 4 parameters

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
thomaslevesque: oops

PR #1028, opened 2017-03-23
Enable test parallelization

### LINKED ISSUE #774: Enable parallel testings in Unit and Integration tests and Specs
The Specs have had parallelization disabled since we moved to xBehave.net (running on top of xUnit.net 2), and now that we moved to xUnit.net in #545, we've had to disable parallel test runs for the integration and unit tests.
Let's see if we can remove the restriction.

thomaslevesque:
Fixes #774.

As discussed with @blairconrad in chat

~~Some unit tests couldn't run in parallel because they mutated global state (`ServiceLocator.Current`). Making them part of the same test collection fixes the problem.~~ (fixed in #1029)

Specs and integration tests seem to have no problem running in parallel.

### DIFF
--- a/tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
+++ b/tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
@@ -83,7 +83,6 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="ExceptionMessagesTests.cs" />
-    <Compile Include="Properties\XunitAssemblyInfo.cs" />
     <Compile Include="TypeCatalogueTests.cs" />
     <Compile Include="BootstrapperLocatorTests.cs" />
     <Compile Include="DoubleArgumentValueFormatter.cs" />
--- a/tests/FakeItEasy.IntegrationTests/Properties/XunitAssemblyInfo.cs
+++ b/tests/FakeItEasy.IntegrationTests/Properties/XunitAssemblyInfo.cs
@@ -1,1 +1,0 @@
-[assembly: Xunit.CollectionBehavior(DisableTestParallelization = true)]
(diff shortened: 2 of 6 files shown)


blairconrad:
Nice.  But note that several test classes that extend that base don't need
to.
Also is there not a danger that the mutation of global state that these
tests do will affect tests outside the group?

thomaslevesque:
> But note that several test classes that extend that base don't need to.

Ah, I didn't notice that. But this should probably be fixed in a separate PR.

> Also is there not a danger that the mutation of global state that these tests do will affect tests outside the group?

I don't know... maybe. So far I haven't seen any test failure with this change, but maybe that's just luck.

blairconrad:
Separate is okay with me. I just thought that since it would let more tests
run in parallel, it was related. Either way.

thomaslevesque:
Just to be sure, I'm running the tests 100 times. I'm at 36 test runs, no failure so far... But again, it might be just luck. The tests that inherit `ConfigurableServiceLocatorTestBase` probably end up using a fake `ServiceLocator` sometimes, so it's bound to fail at some point.

There should be an option to say "run this collection alone, don't run any other test at the same time".

thomaslevesque:
OK, 100 runs, no failure... But I'm not satisfied. The more I look at the code, the more I'm convinced that it *should* fail... yet it doesn't.

blairconrad:
100 is a lot! Nothing is proof, but that would more than satisfy me.

Yeah. I was looking for the same thing. No dice.

thomaslevesque:
I opened https://github.com/xunit/xunit/issues/1179

thomaslevesque:
Rebased. Now there should be no concurrency issues at all.

blairconrad:
Thanks, @thomaslevesque!

PR #1031, opened 2017-03-24
Fix name warning for unindenter

blairconrad:
"unindenter" should've begun with "U". Of course that conflicts with the type name, so I just moved the instance.

### DIFF
--- a/tests/FakeItEasy.Specs/ObjectCallSpecs.cs
+++ b/tests/FakeItEasy.Specs/ObjectCallSpecs.cs
@@ -67,8 +67,6 @@
 
         public class StringOutputWriter : IOutputWriter
         {
-            private static readonly IDisposable unindenter = new Unindenter();
-
             private readonly StringBuilder builder;
 
             public StringOutputWriter()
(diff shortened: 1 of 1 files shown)


tests/FakeItEasy.Specs/ObjectCallSpecs.cs
```diff
@@ -90,7 +88,7 @@ public IOutputWriter WriteArgumentValue(object value)
             public IDisposable Indent()
             {
                 // no tests require indenting yet
-                return unindenter;
+                return Unindenter.Instance;
```
thomaslevesque: Why not just `A.Fake<IDisposable>()`? As your comment says, we don't need to indent at this point, and it would save the hassle of maintaining a class and an instance.

blairconrad: Love it. Stay tuned.

blairconrad: Or `null`!

blairconrad: Perhaps that's too far.

thomaslevesque: I guess that would work too, but a dummy is fine ;)

thomaslevesque:
Thanks @blairconrad

blairconrad:
Oh, don't thank me. It's my own foolish mistake. Thanks!

blairconrad:

PR #1034, opened 2017-03-24
Fix code analysis warning

thomaslevesque:
CA2208: Instantiate argument exceptions correctly

I remove the parameter name, because it's not one argument that is incorrect, but the fact that the arguments are mismatched. I also changed the exception message which no longer made sense.


blairconrad:

PR #1026, opened 2017-03-23
Make spec classes static

blairconrad:
Fixes #544.

PR #1035, opened 2017-03-25
Make rest of specs static

### LINKED ISSUE #544: make spec members static and re-enable rule CA1822
From https://github.com/FakeItEasy/FakeItEasy/pull/542#discussion_r40284574.

blairconrad:
Fixes #544.

### DIFF
--- a/tests/FakeItEasy.Specs/ConfigurationSpecs.cs
+++ b/tests/FakeItEasy.Specs/ConfigurationSpecs.cs
@@ -697,6 +697,7 @@ namespace FakeItEasy.Specs
 
             public void DoSomethingNonVirtual()
             {
+                this.WasCalled = true;
             }
 
             public virtual int ReturnSomething()
@@ -707,6 +708,7 @@ namespace FakeItEasy.Specs
 
             public int ReturnSomethingNonVirtual()
             {
+                this.WasCalled = true;
                 return 11;
             }
         }
--- a/tests/FakeItEasy.Specs/FakeItEasy.Specs.ruleset
+++ b/tests/FakeItEasy.Specs/FakeItEasy.Specs.ruleset
@@ -4,5 +4,6 @@
   <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
     <Rule Id="CA1003" Action="None" />
     <Rule Id="CA1009" Action="None" />
+    <Rule Id="CA1822" Action="Warning" />
   </Rules>
 </RuleSet>
--- a/tests/FakeItEasy.Specs/EventRaisingSpecs.cs
+++ b/tests/FakeItEasy.Specs/EventRaisingSpecs.cs
@@ -15,27 +15,15 @@ namespace FakeItEasy.Specs
 
     public delegate void ValueTypeEventHandler(int arg);
 
-    public class CustomEventArgs : EventArgs
+    public static class EventRaisingSpecs
     {
-    }
+        private static readonly EventArgs SampleEventArgs = new EventArgs();
 
-    public class ReferenceType
-    {
-    }
+        private static readonly CustomEventArgs SampleCustomEventArgs = new CustomEventArgs();
 
-    public class DerivedReferenceType : ReferenceType
-    {
-    }
+        private static readonly ReferenceType SampleReferenceTypeEventArgs = new ReferenceType();
 
-    public class EventRaisingSpecs
-    {
-        private readonly EventArgs eventArgs = new EventArgs();
-
-        private readonly CustomEventArgs customEventArgs = new CustomEventArgs();
-
-        private readonly ReferenceType referenceTypeEventArgs = new ReferenceType();
-
-        private readonly DerivedReferenceType derivedReferenceTypeEventArgs = new DerivedReferenceType();
+        private static readonly DerivedReferenceType SampleDerivedReferenceTypeEventArgs = new DerivedReferenceType();
 
         public interface IEvents
         {
@@ -70,7 +58,7 @@ namespace FakeItEasy.Specs
         private static object CapturedArgs2 { get; set; }
 
         [Background]
-        public void Background()
+        public static void Background()
         {
             Fake = A.Fake<IEvents>();
 
@@ -120,7 +108,7 @@ namespace FakeItEasy.Specs
         }
 
         [Scenario]
-        public void UnsubscribedEvent(
+        public static void UnsubscribedEvent(
             Exception exception)
         {
             "Given an event with no subscribers"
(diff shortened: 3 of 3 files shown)


thomaslevesque:
To be honest, I'm not sure what is the point of making everything static. Enabling CA1822 basically forces us to make all methods static if they *can* be static, including in auxiliary test classes, even if we need them to be non-static for the purpose of the tests... so we end up adding references to `this` just to make FxCop happy. Why force ourselves to go through those hoops? In this case CA1822 doesn't have *any* value, as far as I can see.

Sorry if my comments in #1026 weren't clear, but I had come to the conclusion that we just needed to re-disable CA1822.

blairconrad:
> In this case CA1822 doesn't have any value, as far as I can see.

In this case (and I think  it's largely luck, to tell the truth), it did show me that we had deficiencies in our tests, leading to fixing `CreationOptionsSpecs.cs` and `EventRaisingSpecs.cs` so they fulfill their intended purpose.

> so we end up adding references to `this` just to make FxCop happy

It's true that adding the `this` references in `ConfigurationSpecs.cs` did feel a little dirty.

I do think that 4879b53deaea72dc9b165b6bd04eeafdfb5a5c1c is worth fixing.

As for #544, we could close it **no-fix**. Since it came about largely from @adamralph's [comment](https://github.com/FakeItEasy/FakeItEasy/pull/542#r40320794)

> … I wouldn't be opposed to tighten everything up to be static wherever possible and re-enable this rule.

I'd feel better getting him to agree to closing it. In the meantime we can at least silence the warnings, as you suggest. I'll submit different PRs.

blairconrad:
@thomaslevesque, I've submitted #1036 to fix the tests in `CreationOptionSpecs` and `DummyCreationSpecs`, as I think it's not controversial.

I'll repush this branch without that change, and based on top of your ruleset modifications. That way we don't lose the work in case we want to use it.
If you and @adamralph decide that fixing #544 as written is not desirable, we can just close that issue and this PR.
If you end up wanting the fix, but with changes, I won't be able to get to it for a week or so, but you're more than welcome to slice and dice the commits any way you like, if the open PR is irksome.

blairconrad:
Oh, I took #544 out of the release notes and milestone after I reopened it. I did add #1026.
If #544 ends up being fixed, we could replace #1026 with that.

thomaslevesque:
> In this case (and I think it's largely luck, to tell the truth), it did show me that we had deficiencies in our tests, leading to fixing CreationOptionsSpecs.cs and EventRaisingSpecs.cs so they fulfill their intended purpose.

I see the problem in `CreationOptionsSpecs`, but what's wrong with `EventRaisingSpecs`?

blairconrad:
> I see the problem in `CreationOptionsSpecs`, but what's wrong with `EventRaisingSpecs`?

…me not reading what I paste again. `CreationOptionsSpecs` and `DummyCreationSpecs`. Sorry.

thomaslevesque:
Oh, of course. I should have guessed.

Closing this PR, since it's now redundant.

Thanks!

blairconrad:
> Closing this PR, since it's now redundant.

I'm not sure how it's redundant. It would still make EventRaisingSpecs static and enable CA1822 for the specs project, which would satisfy #544.

If we choose to **wont-fix** #544, then closing this PR makes sense to me.

thomaslevesque:
Ah, sorry, I misunderstood. Reopening until we decide what to do, then.

tests/FakeItEasy.Specs/ConfigurationSpecs.cs
```diff
@@ -697,6 +697,7 @@ public virtual void DoSomething()
 
             public void DoSomethingNonVirtual()
             {
+                this.WasCalled = true;
```
adamralph: Is this changed (and the one below) required as part of changing the specs to static?

tests/FakeItEasy.Specs/FakeItEasy.Specs.ruleset
```diff
@@ -4,5 +4,6 @@
   <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
     <Rule Id="CA1003" Action="None" />
     <Rule Id="CA1009" Action="None" />
+    <Rule Id="CA1822" Action="Warning" />
```
adamralph: Why reduce to warning?

tests/FakeItEasy.Specs/ConfigurationSpecs.cs
thomaslevesque: No, it's just so that the method can no longer be static, to avoid the warning. Which is exactly the kind of thing that makes me think re-enabling CA1822 isn't such a good idea, at least for test projects.

tests/FakeItEasy.Specs/FakeItEasy.Specs.ruleset
thomaslevesque: It's not reduced to warning, it's *raised* to warning. The rule is disabled in the "parent" ruleset (FakeItEasy.Tests.ruleset), and re-enabled in this one.

adamralph: Oh! Thanks. 😊

tests/FakeItEasy.Specs/ConfigurationSpecs.cs
adamralph: I must be missing something. It's a line of code, and nothing else has changed in the file.

thomaslevesque: This method was causing CA1822, because it didn't reference `this` so it could be made static. But we don't actually want it to be static. Adding this makes the warning go away.

adamralph: Thanks! Now I get it.

It sounds like it would be better to add a suppression for this method.

blairconrad:
I debated adding the suppression. I thought it was noisier. And the line I
added matched the sibling methods. But if y'all like the suppression, I'm
content.

adamralph:
I don't think it matters too much. If it were something in the main lib I'd
probably call YAGNI on the new line and insist on the suppression but in
test code I'm less concerned.

blairconrad:
Just to clarify, @adamralph, are you still interested in making restoring the warning for the specs?
I am in favour, but I think @thomaslevesque would rather not have the warning.
If you both are anti-warning, we can close the issue. Otherwise, @thomaslevesque, is there any particular misgivings you have about the warning?

adamralph:
Originally my purist mind told me we should enable it, but on reflection it probably adds a burden that provides little benefit in the case of the test projects, so I'm content with leaving it disabled for them.

thomaslevesque:
> Otherwise, @thomaslevesque, is there any particular misgivings you have about the warning?

Just what I said in my first comment on this PR ;)

PR #1036, opened 2017-03-25
Use test-specific creation methods instead of A.Fake/A.Dummy

blairconrad:
We were invoking A.Fake and A.Dummy directly, instead of calling the abstract method that had both generic and non-generic implementations.
Noticed while working on #1035.
Incidentally supports #544.


blairconrad:
Thanks for the merge!

blairconrad:

PR #1038, opened 2017-03-26
specify VS2015 in config

adamralph:
We should really specify this in config, rather than relying on the UI.


thomaslevesque:
> We should really specify this in config, rather than relying on the UI.

👍

adamralph:
For completeness, I removed the config item in the UI. I also re-ran the build to make sure all is still good.

Issue #1039, opened 2017-03-28
Automatic NuGet deployment when release is published


thomaslevesque:
Currently, when we make a new release, we manually trigger a NuGet deployment from AppVeyor. But AppVeyor also has an option to trigger a deployment when a new tag is pushed:
https://www.appveyor.com/docs/deployment/#deploy-on-tag-github-and-gitlab-only

This would make the release process even smoother, since publishing the release on GitHub would automatically build and deploy the packages.

Depends on #1046.

blairconrad:
:+1:

adamralph:
Since there was some concern in #1040 regarding automatic deployment WRT our current soluition for pre-releases, `VERSION_SUFFIX`, I suggest that we put this on holding pending:

- adopting gitflow
- adopting https://github.com/GitTools/GitVersion (which depends on adopting gitflow). This would remove the need for `VERSION_SUFFIX`.

blairconrad:
Done, @adamralph.

blairconrad:
We've adopted gitflow and gitversion. I think we can take this off-hold now.

blairconrad:
In addition to deploying when a tag is pushed, we could also have AppVeyor attach the nupkgs to the GitHub Release.

blairconrad:
This change has been released as part of [FakeItEasy 4.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/4.0.0).

Issue #1015, opened 2017-03-10
3.2.0 release


blairconrad:
Can be labelled **ready** when all other issues on this milestone are closed.

- [x] run code analysis in VS in *Release* mode and address violations (send a regular PR which must be merged before continuing)
- [x] if necessary, change `VERSION_SUFFIX` in [AppVeyor](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/settings) to  "" and initiate a build
- [x] check [build](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/158)
- edit draft release in [GitHub UI](https://github.com/FakeItEasy/FakeItEasy/releases):
    - [x] ensure completeness of release notes, including non-owner contributors, if any (move release notes forward from any pre-releases to the current release)
    - [x] attach main nupkg and/or analyzer nupkg, whichever have content to release
    - [x] publish the release
- [x] push nupkgs to NuGet using the Deploy option on the AppVeyor build
- [x] de-list pre-release or superseded buggy NuGet packages if present
- [x] tweet, mentioning contributors and post link as comment here for easy retweeting ;-)
- [x] post tweet in [Gitter](https://gitter.im/FakeItEasy/FakeItEasy)
- [x] post a link to the GitHub Release in each issue in this milestone, with thanks to contributors
- ~~run `rake next_version[new_version]` to~~
    - [x] create a pull request that changes the version in CommonAssemblyInfo.cs to the expected version (of form _xx.yy.zz_)
    - [x] create a new draft GitHub Release
    - [x] create a new milestone for the next release
    - [x] create a new issue (like this one) for the next release, adding it to the new milestone
- [x] close this milestone

adamralph:
FYI removed the link to #1014 from the checklist (copypasta artifact from #986).

blairconrad:
Oh, thanks. That sounds like the kind of error I'd make! :blush:

blairconrad:
It's been over 3 weeks since a release, and we have some useful content. What do you think about putting 3.2.0 out there?
#1030 is the only issue we have in flight right now (and I think it looks good - @adamralph, are you still waiting for changes?), but I don't mind releasing either with or without it.
Oh! Unless we want to take a stab at #1039 before the release.

Either way, I'm probably free Wednesday morning.

thomaslevesque:
Yes, I think we're ready to release once @1030 is merged.
#1039 would be a nice to have, if we can make it work.

[...]

PR #1040, opened 2017-04-03
Deploy to NuGet on tag creation

blairconrad:
I generated a new API key just for AppVeyor. 
If I read things right, this is all we need for #1039, except to update instructions in #1015.
Of course, it's completely untested…

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,3 +26,11 @@ test: off
 
 artifacts:
     - path: .\artifacts\*\*
+
+deploy:
+    - provider: NuGet
+      api_key:
+          secure: dp94z3SQBsL1aBqjCdpw4Hlb5Va0XAJQwkjHSu5kJplmxTge8los8LBdbHI72kMi
+      skip_symbols: true
+      on:
+          appveyor_repo_tag: true


blairconrad:
Yeah, that was apparently not it!

thomaslevesque:
Hey @blairconrad, welcome back !

Do we need to specify the API key? It's already in the environment configuration

blairconrad:
Thanks!

This is an excellent question, @thomaslevesque. Um. I don't know.
Then again, I prefer it in the yml to the nebulous web config.
Truth be told, I don't really understand how the deployment targets here interact with the "environments" we specify elsewhere.
The [AppVeyor deployment docs](https://www.appveyor.com/docs/deployment/#deploy-on-tag-github-and-gitlab-only) suggest a separation between inline deployments (which we have here) and manual deployment (which we've been using). If I understand correctly.

thomaslevesque:
> I assume this will work for pre-releases too? I.e. can we continue to set the VERSION_SUFFIX and then create the corresponding pre-release in GitHub, which will create the pre-release tag, which will trigger a build of that tag, which will trigger a push of the pre-release to NuGet?

I think so.

> Can we make it so that tags are built by AppVeyor, and then all that's needed is a single button click in the UI against that tag build to "release these artifacts"?

Yes, we can do that. But then it's basically the same thing we're already doing. 

> If so, is the removal of that manual step worth the risk?

It doesn't seem very risky to me... It will never deploy unless we publish a release. True, the manual step to deploy isn't much of a hassle, since it's already just a button click in AppVeyor, but I like the idea of doing it automatically when we publish.

> I guess that risk needs quantifying. What are the possible downsides of the automation? Is there an increased risk of releasing some bad permutation of tag and VERSION_SUFFIX?

I'm not sure what the downsides would be, as long as it only deploys when we add a tag. We might want to change how we handle `VERSION_SUFFIX`; for instance, it would be nice if it could be based on the tag.

blairconrad:
> I'm not sure what the downsides would be, as long as it only deploys when we add a tag. We might want to change how we handle `VERSION_SUFFIX`

Indeed. The static state terrifies me.

> for instance, it would be nice if it could be based on the tag.

I think this is achievable when we have a tagged build. Not sure how easy it would be to work with when we've not yet tagged. For example, when we want to generate releases for the AppVeyor NuGet feed with a -beta001 on them.

thomaslevesque:
> For example, when we want to generate releases for the AppVeyor NuGet feed with a -beta001 on them.

Good point. I'm not sure how we would do it.

blairconrad:
If we're interested in reducing the risk of releasing with mismatched tag and VERSION_SUFFIX, we could use the tag name to compute the version number. If `APPVEYOR_REPO_TAG_NAME` is set, it's the version.

Of course, if it's not set, then we'd still need a way to cobble together a version. It could be the way we do it now, but it seems a little silly to keep updating a version number that's not going to be used for production builds.

thomaslevesque:
Looks like we're still undecided on this. I suggest not to do it for now, we can always revisit later.

PR #1041, opened 2017-04-04
Stop implementing IStartConfiguration<T> in Fake<T>

### LINKED ISSUE #959: Remove IStartConfiguration<T> from Fake<T>'s interfaces
Relates to #864

There's absolutely no reason for `Fake<T>` to implement `IStartConfiguration<T>`. This interface is now internal anyway, and the members of `Fake<T>` are never called via the interface.

Technically, it's not a breaking change, but there's a bug the version of `PublicApiGenerator` that we use, which causes `IStartConfiguration<T>` to appear in `Fake<T>`'s implemented interfaces, even though it's internal, so it will change the approved API file. We should probably upgrade to a more recent version of PublicApiGenerator anyway.

blairconrad:
Fixes #959.

### DIFF
--- a/src/FakeItEasy/Fake.of.T.cs
+++ b/src/FakeItEasy/Fake.of.T.cs
@@ -13,7 +13,7 @@ namespace FakeItEasy
     /// FakedObject-property.
     /// </summary>
     /// <typeparam name="T">The type of the faked object.</typeparam>
-    public class Fake<T> : IStartConfiguration<T>
+    public class Fake<T> : IHideObjectMembers
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="Fake{T}"/> class.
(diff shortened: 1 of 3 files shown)


blairconrad:
Oh, I'm happy to remove the dependency on `IHideObjectMembers` as well. Hmm. Since we were picking it up via `IStartConfiguration`, I guess the outside effect of my PR is to _add_ `IHideObjectMembers` to the class, no?

thomaslevesque:
> Oh, I'm happy to remove the dependency on IHideObjectMembers as well. Hmm. Since we were picking it up via IStartConfiguration, I guess the outside effect of my PR is to add IHideObjectMembers to the class, no?

I'm not sure I understand... it already had `IHideObjectMembers` via `IStartConfiguration`, so you didn't add it (you just made it explicit).

blairconrad:
It did, but I thought that perhaps because `IStartConfiguration` was internal, the `IHideObjectMembers` didn't propagate out. But I guess not. Decompiling shows everything, and there's no concept of "internal implementation" in .NETLand, so I'll go away. Thanks.

PR #1051, opened 2017-04-10
Improve message for methods which cannot be intercepted

chrismatheson:
For newer C# programmers it is not always understood that "sealed" is the default behaviour. So hopefully this new message will explain in an actionable way how to proceed.

https://gitter.im/FakeItEasy/FakeItEasy?at=58eb83578fcce56b20ea41af

### DIFF
--- a/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
+++ b/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
@@ -34,7 +34,7 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
                 var explicitImplementation = method.Name.Contains('.');
                 if (!explicitImplementation)
                 {
-                    return "Sealed methods can not be intercepted.";
+                    return "Non-virtual methods can not be intercepted. Virtual methods are methods explicitly marked virtual, overriding methods, abstract methods, and interface methods.";
                 }
             }
 
@@ -46,7 +46,7 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
                 }
                 else
                 {
-                    return "Static methods can not be intercepted.";
+                    return "Non-virtual methods can not be intercepted. Virtual methods are methods explicitly marked virtual, overriding methods, abstract methods, and interface methods.";
                 }
             }
 
--- a/tests/FakeItEasy.IntegrationTests/ExceptionMessagesTests.cs
+++ b/tests/FakeItEasy.IntegrationTests/ExceptionMessagesTests.cs
@@ -21,7 +21,7 @@ namespace FakeItEasy.IntegrationTests
 @"
 
   The current proxy generator can not intercept the method System.Object.Equals(System.Object objA, System.Object objB) for the following reason:
-    - Static methods can not be intercepted.
+    - Non-virtual methods can not be intercepted. Virtual methods are methods explicitly marked virtual, overriding methods, abstract methods, and interface methods.
 
 ";
 
--- a/tests/FakeItEasy.Specs/ConfigurationSpecs.cs
+++ b/tests/FakeItEasy.Specs/ConfigurationSpecs.cs
@@ -267,7 +267,7 @@ namespace FakeItEasy.Specs
 
             "Then it throws a fake configuration exception"
                .x(() => exception.Should().BeAnExceptionOfType<FakeConfigurationException>()
-                   .And.Message.Should().Contain("Sealed methods can not be intercepted."));
+                   .And.Message.Should().Contain("Method should be marked as virtual, abstract or part of an Interface in oreder to be intercepted."));
         }
 
         [Scenario]
--- a/tests/FakeItEasy.Specs/ConfigurationSpecs.cs
+++ b/tests/FakeItEasy.Specs/ConfigurationSpecs.cs
@@ -267,11 +267,7 @@ namespace FakeItEasy.Specs
 
             "Then it throws a fake configuration exception"
                .x(() => exception.Should().BeAnExceptionOfType<FakeConfigurationException>()
-<<<<<<< HEAD
-                   .And.Message.Should().Contain("Method should be marked as virtual, abstract or part of an Interface in oreder to be intercepted."));
-=======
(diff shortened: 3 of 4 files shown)


blairconrad:
Hi, @chrismatheson.

Thanks for the interest. You may (or may not; I'm still not sure when notifications are sent) have noticed that the build failed, since some tests had been checking for the phrasing of the error message that you changed.

Whenever sending a PR for a project, it's a good idea to make sure that the project still builds. Our [contributing](https://github.com/FakeItEasy/FakeItEasy/blob/master/CONTRIBUTING.md) and [how to build](https://github.com/FakeItEasy/FakeItEasy/blob/master/how_to_build.md) documents can help with this.

In the meantime, I'll run over to the "files changed" section for some comments.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
```diff
@@ -34,7 +34,7 @@ private static string GetReasonForWhyMethodCanNotBeIntercepted(MethodInfo method
                 var explicitImplementation = method.Name.Contains('.');
                 if (!explicitImplementation)
                 {
-                    return "Sealed methods can not be intercepted.";
+                    return "Method should be marked as virtual, abstract or part of an Interface in oreder to be intercepted.";
```
blairconrad: I like the intent here, especially the idea of giving the user a way forward, rather than just complaining at them (and you found a way to avoid the "can not/cannot" debate, which is always fun).

I want to take a little time to trigger this error in context to see how it reads, but while I'm here, I'll make a few comments before I forget.

1. "order" is misspelled,
2. "Interface" is capitalized for some reason

But maybe don't jump right on this, as I'll probably have more to say after an experiment…

thomaslevesque: I think I'd rather keep the initial sentence, just replacing "sealed" with "non-virtual".

We *could* add some advice, but I'm not sure how to word it. Maybe something like this:

> Non-virtual methods can not be intercepted. Virtual methods are methods explicitly marked virtual, overriding methods, abstract methods and interface methods.

blairconrad: I tend to agree with you, @thomaslevesque. Looking at how the message is presented in context:

```
FakeItEasy.Specs.ConfigurationSpecs.CallToSetSealedOnFake() [02] When I start to configure a sealed property setter on the fake [FAIL]
   FakeItEasy.Configuration.FakeConfigurationException :                                                                              
                                                                                                                                      
       - Method should be marked as virtual, abstract or part of an Interface in oreder to be intercepted.                            ```
```

"Method should be marked…" doesn't really flow after the intro. Sorry, @chrismatheson, I think @thomaslevesque's version initial sentence works better in this case. 

I think your second sentence is quite good (except for the missing serial comma :wink:).
I'm a little concerned that we're going to end up turning the error messages into a .NET tutorial, but maybe that's not a bad thing.

@chrismatheson, as the person who was actually negatively affected by the original message, do you think "Non-virtual methods can not be intercepted. Virtual methods are methods explicitly marked virtual, overriding methods, abstract methods**,** and interface methods."
would've been more helpful? Helpful enough?

chrismatheson: yeah that would have worked :)

blairconrad: Cool!
Since you're invested, and have already started modifying code and tests, are you interested in updating your PR accordingly? After all, it was your feedback that will lead to an improvement one way or another!

blairconrad: @chrismatheson, I don't mean to hound, but am not sure if you saw my previous method. Are you interested in updating the PR? If so, great, and we're happy to wait/help as needed. If not, just say the word and we'll continue the work from here.
Thanks!

chrismatheson: Heya, yeah sorry, recently had our first child so everything got a little dropped on the floor :) I'll hopefully be able to get to it soonish

blairconrad: @chrismatheson,

1. Congratulations!
2. I hope you didn't drop the child on the floor! (sorry)
3. No rush at all. We're in no hurry. We just wanted to gauge your interest. Now that we know for sure, we won't hound you for at least a couple of weeks. :wink:

Thanks.

thomaslevesque: Congratulations @chrismatheson!

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
```diff
@@ -46,7 +46,7 @@ private static string GetReasonForWhyMethodCanNotBeIntercepted(MethodInfo method
                 }
                 else
                 {
-                    return "Static methods can not be intercepted.";
+                    return "Non-virtual methods can not be intercepted. Virtual methods are methods explicitly marked virtual, overriding methods, abstract methods, and interface methods.";
```
blairconrad: I would be inclined to leave this message alone. Unlike the "sealed/non-virtual" confusion, the `static` keyword is right in the code when a static method is defined, so there should be no confusion. Also, if the two messages (here and on line 37) are the same, it makes it harder to tell exactly what went wrong.

blairconrad:
Hey, @chrismatheson. In addition to the usual commits that have come into FakeItEasy/FakeItEasy, we recently changed our main branch from `master` to `develop`, so when this PR is rebased, it should be off the head of `develop`, and then the merge-to branch should be updated as well. 

Sorry for this inconvenience. If it's all too much trouble, we can take care of that bookkeeping. You'll still get credit for your contribution.

blairconrad:
There's been no movement on this PR for quite a while. I think it's been abandoned. I'll submit a superseding PR.

PR #1052, opened 2017-04-14
Make IFakeAndDummyManager a singleton 

### LINKED ISSUE #769: Register IFakeAndDummyManager as a singleton
In https://github.com/FakeItEasy/FakeItEasy/pull/768#discussion_r67624889, @thomaslevesque observed:

> I just realized that `IFakeCreatorFacade` is not registered as a singleton... and most of its dependencies are not singletons either. This means that everytime we call `A.Something`, we create new instances of all these things. I wonder if this could be improved…

Note that `IFakeCreatorFacade` and `DefaultFakeCreatorFacade` were removed in https://github.com/FakeItEasy/FakeItEasy/pull/776 so this issue now drops down to the single dependency of `DefaultFakeCreatorFacade`, which is `IFakeAndDummyManager`.

blairconrad:
Fixes #769.

I also have some ideas about continuing the work a little with some other classes. There's more duplication to be removed, I think.

### DIFF
--- a/tests/FakeItEasy.IntegrationTests/DummyTests.cs
+++ b/tests/FakeItEasy.IntegrationTests/DummyTests.cs
@@ -2,6 +2,8 @@ namespace FakeItEasy.IntegrationTests
 {
     using System;
     using System.Diagnostics.CodeAnalysis;
+    using System.Threading;
+    using System.Threading.Tasks;
     using FakeItEasy.Core;
     using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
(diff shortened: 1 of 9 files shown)


blairconrad:
Sorry for the churn (if you noticed it). I'd added some FxCop warnings and wanted to correct while I had a chance.

tests/FakeItEasy.IntegrationTests/DummyTests.cs
```diff
@@ -45,10 +47,68 @@ public void Dummy_factories_are_polled_only_once_per_dummy_type()
...
+                }
+                finally
+                {
+                    CreatedLock.Set();
+                }
+
+                return new DummyB();
+            }
+        }
+
```
thomaslevesque: It's right after lunch time here, and my brain is a bit slow right now... I have trouble following this. Is there no simpler way to test this scenario? If not, can you try to explain it in more detail?

thomaslevesque:
> Sorry for the churn (if you noticed it). I'd added some FxCop warnings and wanted to correct while I had a chance.

No problem, you made the changes before I started to review

tests/FakeItEasy.IntegrationTests/DummyTests.cs
blairconrad: I could not think of a simpler way. But I think I may have done now! I will amend sometime later today.

blairconrad: Oh, wait. That won't work.

blairconrad: Okay. I think I have a thing. I'll upload later. Gotta start the Easter baking.

blairconrad:
@thomaslevesque, thanks for the comment on the concurrency test. I think the new version is better (added as a **fixup** commit, so you can still compare the two of them, if you're inclined).

thomaslevesque:
Yes, your new approach is easier to understand. Thanks!

blairconrad:
Thanks!

blairconrad:
> Yes, your new approach is easier to understand. Thanks!

I'm so glad you commented.
I'd worked for a while even to get something that worked. Embarrassing to not have come up with something as simple as the new version.

thomaslevesque:
Thanks @blairconrad !

blairconrad:
Thank you!

Issue #1047, opened 2017-04-05
Use CLR formatting rather than C# formatting when describing types


blairconrad:
Split from #1011

blairconrad:
This issue was fixed in [Release 3.3.0](https://github.com/FakeItEasy/FakeItEasy/releases/3.3.0).

PR #1055, opened 2017-04-17
Include name of the method that can't be intercepted in the error message

### LINKED ISSUE #786: In error messages, name the method that we can't intercept on an instance
Inspired by a [StackOverflow question](http://stackoverflow.com/q/38152748/1199).

The questioner had the equivalent of

``` c#
public interface IHaveAList
{
    IEnumerable<string> GetList();
}

var fake = A.Fake<IHaveAList>();
A.CallTo(() => fake.GetList().Single())
    .Throws(new Exception());
```

And didn't understand why he was told 

```
FakeItEasy.Configuration.FakeConfigurationException: 
- Extension methods can not be intercepted since they're static.
```

The thought was the complaint was about `GetList`, when really it was about `Single`.

We should say which method (I'd almost hope for something like `Enumerable.Single<TSource>(this IEnumerable<TSource> source)`, but whatever we can get would be great.
We could maybe take a cue from `DefaultFakeObjectCallFormatter`, which has very advanced formatting rules based on whether the method is a method or a property and so on, but it's concerned with object calls and we only have `MethodInfo`s available where we build the exception message above (in `DefaultInterceptionAsserter.AssertThatMethodCanBeInterceptedOnInstance`), so we'd need to do a little work for  a full-fidelity solution.

thomaslevesque:
Fixes #786

### DIFF
--- a/src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
+++ b/src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
@@ -1,7 +1,5 @@
 namespace FakeItEasy.Core
 {
-    using System;
-    using System.Collections.Generic;
     using System.Linq;
     using System.Reflection;
     using System.Text;
@@ -32,7 +30,7 @@ namespace FakeItEasy.Core
             var builder = new StringBuilder();
 
             builder
-                .Append(this.fakeManagerAccessor.GetFakeManager(call.FakedObject).FakeObjectType)
+                .Append(this.fakeManagerAccessor.GetFakeManager(call.FakedObject).FakeObjectType.FullNameCSharp())
                 .Append(".");
 
             AppendMethodName(builder, call.Method);
@@ -44,7 +42,7 @@ namespace FakeItEasy.Core
 
         private static ArgumentValueInfo[] GetArgumentsForArgumentsList(ArgumentValueInfo[] allArguments, MethodInfo method)
         {
-            if (IsPropertySetter(method))
+            if (method.IsPropertySetter())
             {
                 return allArguments.Take(allArguments.Length - 1).ToArray();
             }
@@ -54,7 +52,7 @@ namespace FakeItEasy.Core
 
         private static void AppendArgumentListPrefix(StringBuilder builder, MethodInfo method)
         {
-            if (IsPropertyGetterOrSetter(method))
+            if (method.IsPropertyGetterOrSetter())
             {
                 builder.Append("[");
             }
@@ -66,7 +64,7 @@ namespace FakeItEasy.Core
 
         private static void AppendArgumentListSuffix(StringBuilder builder, MethodInfo method)
         {
-            if (IsPropertyGetterOrSetter(method))
+            if (method.IsPropertyGetterOrSetter())
             {
                 builder.Append("]");
             }
@@ -78,7 +76,7 @@ namespace FakeItEasy.Core
 
         private static void AppendMethodName(StringBuilder builder, MethodInfo method)
         {
-            if (IsPropertyGetterOrSetter(method))
+            if (method.IsPropertyGetterOrSetter())
             {
                 builder.Append(method.Name.Substring(4));
             }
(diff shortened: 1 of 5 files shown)


src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
```diff
@@ -90,21 +88,6 @@ private static void AppendMethodName(StringBuilder builder, MethodInfo method)
...
-
-        private static bool IsPropertySetter(MethodInfo method)
-        {
-            return method.IsSpecialName && method.Name.StartsWith("set_", StringComparison.Ordinal);
-        }
-
-        private static bool IsPropertyGetter(MethodInfo method)
-        {
-            return method.IsSpecialName && method.Name.StartsWith("get_", StringComparison.Ordinal);
-        }
```
thomaslevesque: Promoted to extension methods

src/FakeItEasy/MethodBaseExtensions.cs
```diff
@@ -18,5 +20,119 @@ public static string GetGenericArgumentsCSharp(this MethodBase method)
...
+                builder.AppendLine();
+                builder.Append("    ");
+                return;
+            }
+
+            if (argumentIndex > 0)
+            {
+                builder.Append(", ");
+            }
+        }
```
thomaslevesque: I'm not very happy about this... it's mostly the same code as in `DefaultFakeObjectCallFormatter`, with a few small differences. I considered putting the new `GetDescription` method in `IFakeObjectCallFormatter` to take advantage of the existing code, but it didn't feel right.

src/FakeItEasy/MethodBaseExtensions.cs
```diff
@@ -18,5 +20,119 @@ public static string GetGenericArgumentsCSharp(this MethodBase method)
...
+            if (IsPropertyGetterOrSetter(method))
+            {
+                builder.Append(method.Name.Substring(4));
+            }
+            else
+            {
+                builder.Append(method.Name);
+            }
+
+            builder.Append(method.GetGenericArgumentsCSharp());
```
thomaslevesque: For consistency with how calls are formatted. Will need to be changed as part of #1047

src/FakeItEasy/MethodBaseExtensions.cs
```diff
@@ -18,5 +20,119 @@ public static string GetGenericArgumentsCSharp(this MethodBase method)
...
+            }
+        }
+
+        private static void AppendParameters(StringBuilder builder, ParameterInfo[] parameters)
+        {
+            for (int i = 0; i < parameters.Length; i++)
+            {
+                AppendParameterSeparator(builder, i, parameters.Length);
+                builder
+                    .Append(parameters[i].ParameterType.FullNameCSharp())
```
thomaslevesque: For consistency with how calls are formatted. Will need to be changed as part of #1047

src/FakeItEasy/MethodBaseExtensions.cs
blairconrad: I see what you're saying.
I'm not super bothered by the duplication.

Of course, I didn't notice any tests that cover the 3+ parameter case (which I fear might look funny when it's rendered...)

blairconrad: Nor do I see property-covering tests, now that I think about it.

blairconrad:
Looks quite good, @thomaslevesque, other than my fears about rendering long calls.

src/FakeItEasy/MethodBaseExtensions.cs
thomaslevesque: Good point, I'll add some tests.

thomaslevesque: We probably don't want to split on multiple lines in this scenario. Should I add a parameter to control this? Or just assume we never want to split on multiple lines?

blairconrad: Until we need different behaviour, I would just assume we don't want to split.

src/FakeItEasy/MethodBaseExtensions.cs
```diff
@@ -18,5 +20,119 @@ public static string GetGenericArgumentsCSharp(this MethodBase method)
...
+        {
+            return method.IsSpecialName && method.Name.StartsWith("get_", StringComparison.Ordinal);
+        }
+
+        public static string GetDescription(this MethodBase method)
+        {
+            var builder = new StringBuilder();
+
+            builder
+                .Append(method.DeclaringType)
```
thomaslevesque: Just noticed that we don't include "nicely formatted" generic arguments for the declaring type in `DefaultFakeObjectCallFormatter`, so the result is a mix of CLR and C# formatting. It will go away with #1047, but in the meantime, I'll fix it both in `DefaultFakeObjectCallFormatter` and here.

thomaslevesque:
I squashed

blairconrad:
Thanks, @thomaslevesque!

PR #1057, opened 2017-04-17
Use CLR formatting for types and methods rather than C# formatting

### LINKED ISSUE #1047: Use CLR formatting rather than C# formatting when describing types

thomaslevesque:
Fixes #1047

### DIFF
--- a/src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
+++ b/src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
@@ -30,7 +30,7 @@ namespace FakeItEasy.Core
             var builder = new StringBuilder();
 
             builder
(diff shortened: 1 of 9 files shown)


thomaslevesque:
It's a little annoying that there's so much code duplication for method/call formatting... we have 3 separate implementations:
- `MethodBaseExtensions.GetDescription`: returns the signature of the method (should probably be renamed to `GetSignature`)
- `DefaultFakeObjectCallFormatter.GetDescription`: returns the description of an actual call, with argument values
- `ExpressionCallMatcher.DescriptionOfMatchingCall`: returns the description of a configured call, with argument constraints

They're all very similar, yet different enough that sharing code between them is awkward... maybe it's not worth the trouble.

thomaslevesque:
Thanks for the merge @blairconrad!

PR #1058, opened 2017-04-18
Reduce temporary objects created when making Dummies

### LINKED ISSUE #1053: Further reduce temporary objects created when making dummies
Continues #769.

blairconrad:
Fixes #1053

### DIFF
--- a/src/FakeItEasy/Creation/DummyCreationSession.cs
+++ b/src/FakeItEasy/Creation/DummyCreationSession.cs
@@ -1,0 +1,26 @@
+﻿namespace FakeItEasy.Creation
+{
+    using System;
+    using System.Collections.Generic;
+
+    internal class DummyCreationSession
+    {
+        private readonly HashSet<Type> typesCurrentlyBeingResolved = new HashSet<Type>();
+
+        public bool TryBeginToResolveType(Type typeOfDummy)
+        {
+            if (this.typesCurrentlyBeingResolved.Contains(typeOfDummy))
+            {
+                return false;
+            }
+
+            this.typesCurrentlyBeingResolved.Add(typeOfDummy);
+            return true;
+        }
+
+        public void OnSuccessfulResolve(Type typeOfDummy)
+        {
+            this.typesCurrentlyBeingResolved.Remove(typeOfDummy);
+        }
+    }
+}
--- a/src/FakeItEasy/Creation/DummyValueResolver.cs
+++ b/src/FakeItEasy/Creation/DummyValueResolver.cs
@@ -9,18 +9,18 @@ namespace FakeItEasy.Creation
     using System.Threading.Tasks;
     using FakeItEasy.Core;
 
-    internal class DummyValueCreationSession : IDummyValueCreationSession
+    internal class DummyValueResolver : IDummyValueResolver
     {
         private readonly ResolveStrategy[] strategies;
         private readonly HashSet<Type> typesCurrentlyBeingResolved;
         private readonly Dictionary<Type, ResolveStrategy> strategyCache;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="DummyValueCreationSession"/> class.
+        /// Initializes a new instance of the <see cref="DummyValueResolver"/> class.
         /// </summary>
         /// <param name="dummyFactory">The dummy factory.</param>
         /// <param name="fakeObjectCreator">The fake object creator.</param>
-        public DummyValueCreationSession(DynamicDummyFactory dummyFactory, IFakeObjectCreator fakeObjectCreator)
+        public DummyValueResolver(DynamicDummyFactory dummyFactory, IFakeObjectCreator fakeObjectCreator)
         {
             this.typesCurrentlyBeingResolved = new HashSet<Type>();
             this.strategyCache = new Dictionary<Type, ResolveStrategy>();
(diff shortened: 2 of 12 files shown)


src/FakeItEasy/Creation/DummyCreationSession.cs
```diff
@@ -0,0 +1,26 @@
...
+            this.typesCurrentlyBeingResolved.Add(typeOfDummy);
+            return true;
+        }
+
+        public void OnSuccessfulResolve(Type typeOfDummy)
+        {
+            this.typesCurrentlyBeingResolved.Remove(typeOfDummy);
+        }
+    }
+}
```
thomaslevesque: So, `DummyCreationSession` is just a wrapper around a hashset to keep track of which types are being resolved?

This means that the strategy cache will be shared between all dummy resolutions. I guess it makes sense, but can't it have unwanted side effects?

src/FakeItEasy/A.cs
```diff
@@ -44,7 +44,7 @@ public static T Fake<T>(Action<IFakeOptions<T>> optionsBuilder)
         {
             Guard.AgainstNull(optionsBuilder, nameof(optionsBuilder));
 
-            return (T)FakeAndDummyManager.CreateFake(typeof(T), options => optionsBuilder((IFakeOptions<T>)options));
+            return (T)FakeAndDummyManager.CreateFake(new DummyCreationSession(), typeof(T), options => optionsBuilder((IFakeOptions<T>)options));
```
thomaslevesque: Shouldn't it be injected at the `FakeAndDummyManager` level (i.e. one level "below")? Or maybe even at the `FakeObjectCreator` level? It seems a bit strange to me to create the `DummyCreationSession` at this level of abstraction; `A` shouldn't have to know that creating a fake (or dummy) requires a `DummyCreationSession`.

Also, moving it to a lower abstraction level means that less classes need to be modified (e.g. `A`, `Create`, and `Fake<T>` would be unaffected)

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -66,12 +65,12 @@ private bool TryResolveDummyValueWithAllAvailableStrategies(DummyCreationSession
             {
                 if (strategy.TryCreateDummyValue(session, typeOfDummy, out result))
                 {
-                    this.strategyCache.Add(typeOfDummy, strategy);
+                    this.strategyCache.TryAdd(typeOfDummy, strategy);
```
thomaslevesque: What if it fails? Just ignore it, and we'll have to resolve the correct strategy again next time?

src/FakeItEasy/Creation/DummyCreationSession.cs
thomaslevesque: Although, now that I think about it, that's the whole point of this change...

src/FakeItEasy/Creation/DummyValueResolver.cs
blairconrad: You mean if it throws? It's pretty unlikely. `typeOfDummy` isn't `null`, and we're unlikely to have `Int32.MaxValue` items…

If it returns `false`, it just means there was already a strategy registered. We don't mind competing threads finding the strategy and inserting one first. Unless you'd rather we were super strict and locked out other threads. I can try that, although the method we have to pass to `GetOrAdd` is kind of complicated, I think.

src/FakeItEasy/A.cs
blairconrad: We'd had a recent discussion about the servicelocator anti-pattern and how we'd be better off just using it at the static entry points, so I strove for that. I'm happy to inject it lower, though!

src/FakeItEasy/Creation/DummyCreationSession.cs
blairconrad: It's a big part of the point, yeah. We don't have to allocate extra cache objects, and once we figure out how to make a Dummy, we don't have to _keep_ figuring it out.

src/FakeItEasy/Creation/DummyValueResolver.cs
blairconrad: (It's `LazyThreadSafetyMode.PublicationOnly` all over again!)

thomaslevesque: > You mean if it throws?

No, I mean if it returns `false`. I'm not really worried about this, just making sure we had this case covered. I hadn't realized it would only return false if the key already existed.

src/FakeItEasy/A.cs
thomaslevesque: Well, it would make sense if we injected *all* the dependencies at the static entry points, but in this case it looks a bit inconsistent. I think it'd be cleaner to inject it lower.

src/FakeItEasy/Creation/DummyCreationSession.cs
thomaslevesque: This makes sense. The extension point discovery is done only once, so there's no reason to change strategies later.

blairconrad:
@thomaslevesque, I've pushed the change. Note that I inserted as a fixup in the second-last position in the change of commits, mostly to make the rebase easier later. As a result, the last commit was rewritten, but is effectively unchanged.

blairconrad:
Thanks, @thomaslevesque. Squashed (and built).

thomaslevesque:
Thanks!

PR #1061, opened 2017-04-20
[POC - DO NOT MERGE] Dynamic Raise

### LINKED ISSUE #987: Support Raise.With for arbitrary delegates without requiring a typeparam in the call.
**Update (@blairconrad):**  title changed after initial discussion among owners. I've left the original proposal here, but the goal of the issue has been expanded include support for all non-`EventHandler` delegates.

----
Documentation [says](http://fakeiteasy.readthedocs.io/en/stable/raising-events/) you have to specify generic type for Raise.With in case you use a **custom delegate**. And as I understand by "custom delegate" it measn everything besides EventHandler (which is not that popular for handling events besides Windows Forms). Would be nice to add support for Action and Func delegates as most of events are designed using them. Now we have to write something likes this:

`realtimeManager.ConnectionStateChanged += Raise.With<Action<ConnectionState>>(ConnectionState.Connected);`

instead of

realtimeManager.ConnectionStateChanged += Raise.With(ConnectionState.Connected);

thomaslevesque:
Fixes #987


blairconrad:
I appreciate the cleverness, but wonder whether it's worthwhile if we need to have a different name. Why not just live with the typeparams?

thomaslevesque:
> I appreciate the cleverness, but wonder whether it's worthwhile if we need to have a different name. Why not just live with the typeparams?

Because it's horribly verbose... and also because it's a fun experiment ;)

blairconrad:
I'm totally behind the "fun" part!

src/FakeItEasy/Core/DynamicRaiser.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            return this.arguments;
+        }
+
+        public override bool TryConvert(ConvertBinder binder, out object result)
+        {
+            var targetType = binder.Type;
+            if (!targetType.GetTypeInfo().IsSubclassOf(typeof(Delegate)))
+            {
+                throw new InvalidCastException($"Unable to cast object of type '{this.GetType()}' to type '{targetType}'.");
```
blairconrad: I think we now omit the single quotes when the innards can't have spaces.

src/FakeItEasy/ValueProducerSignatureHelper.cs
```diff
@@ -22,6 +24,20 @@ internal static void AssertThatValueProducerSignatureSatisfiesCallSignature(Meth
...
+            if (IsCallSignatureSatisfiedByValues(callMethod, values))
+            {
+                return;
+            }
+
+            var fakeSignature = BuildSignatureDescription(callMethod.GetParameters().Select(p => p.ParameterType));
+            var actionSignature = BuildSignatureDescription(values.Select(v => v?.GetType()));
+
+            throw new FakeConfigurationException(
+                $"The event has the signature ({fakeSignature}), but the provided arguments have types ({actionSignature}).");
```
blairconrad: Sadly, sometimes we put `()`s around signatures, and sometimes not.
And while I'm nitpicking, should the exception messages all move to `ExceptionMessages`?
(I know, I'm getting ahead of myself for comments on a proof of concept.)

tests/FakeItEasy.Specs/EventRaisingSpecs.cs
```diff
@@ -383,5 +383,132 @@ public void ActionEvent()
...
+            "When I raise the event specifying incorrect arguments"
+                .x(() => exception = Record.Exception(() => Fake.CustomEvent += Raise.WithDynamic("foo", 42, true)));
+
+            "Then it throws a FakeConfigurationException"
+                .x(() => exception.Should().BeAnExceptionOfType<FakeConfigurationException>());
+
+            "And the exception message describes the expected and actual arguments"
+                .x(() => exception.Message.Should().Be(
+                    "The event has the signature (System.Object, FakeItEasy.Specs.CustomEventArgs), but the provided arguments have types (System.String, System.Int32, System.Boolean)."));
+        }
```
blairconrad: I like the specs, but worry that the step names do not distinguish between the `WithDynamic` and `With<>` cases.

thomaslevesque:
> Overall, I'm still a little sad to have yet a 3rd way to raise events. If we didn't have With<> (should we remove it?) I'd feel better.

Well, as I said, it's just an experiment, we don't have to keep it if we don't like it. Technically, this new overload *could* replace all the others, but then we'd lose the benefits of strongly-typed overloads. And of course, this would be a breaking change.

> I'd feel even more better if we didn't have to vary the method name, but I understand the conflict.
> (And for some reason WithDynamic doesn't really thrill my brain, but I can't think of anything better.)

Same here. Maybe we could do something like `Raise.Dynamic.With(...)`. It's not great either, but it helps separate the dynamic overload from the others.

src/FakeItEasy/Core/DynamicRaiser.cs
thomaslevesque: I know, but I wanted to make it look like the "normal" `InvalidCastException`, which does include single quotes.

src/FakeItEasy/ValueProducerSignatureHelper.cs
thomaslevesque: Well, in this case I think we should include `()`, because  it's a list of parameters. Anyway, we can revisit if we decide to go ahead with this idea.

tests/FakeItEasy.Specs/EventRaisingSpecs.cs
thomaslevesque: True. I'll change it if we decide to go ahead (which seems unlikely at this point).

src/FakeItEasy/Core/DynamicRaiser.cs
blairconrad: Ah. Good idea. In the end, it might be worthwhile commenting that to keep us from forgetting why the deviation.

src/FakeItEasy/ValueProducerSignatureHelper.cs
blairconrad: Makes sense. Also, on a relook, I cannot find whatever "signature" instance I thought we had that omitted the `()`s.

tests/FakeItEasy.Specs/EventRaisingSpecs.cs
blairconrad: Unlikely? Are you not still in favour of it? Don't mind my comments; I'm always down on new things. I don't think it's a bad alternative to the typeparamed `With`, and even if we did decide having both methods was superfluous, there's no reason not to introduce the new one before deprecating and eventually removing the old…

thomaslevesque: Well, I kinda like it, but even if we remove the `With<TEventHandler>(object[])` overload, we still have an ambiguity with `With<TEventArgs>(object, TEventArgs)`, so we still have to pick a different name for the dynamic overload. So I guess it all comes down to whether we can find a satisfactory name.

thomaslevesque:
Assuming we want to keep this change:
- the new method it could replace the `With<TEventHandler>(params object[] args)` overload
- it would need a name other than `With` to avoid ambiguity with `Raise<TEventArgs>(object, TEventArgs)`

Since removing an overload would be a breaking change, we might as well revisit the existing API. We could do something like this:

```
Raise.Standard<TEventArgs>(object sender, TEventArgs args);
Raise.FreeForm(params object[] args);
```

thomaslevesque:
Closing this PR because it was opened before the switch to GitFlow, so it was to be merged into master. I'll open a new one.

PR #1064, opened 2017-04-23
IsNotNull constraint

### LINKED ISSUE #1062: Add That.IsNotNull convenience overload
As @thomaslevesque points out in https://github.com/FakeItEasy/FakeItEasy/issues/1060#issuecomment-295483516

> > (interestingly, we have a IsNull constraint, which isn't very useful since we can just say null, but no IsNotNull...)

> Of course, I forgot about `Not`... we can do `A<T>.That.Not.IsNull()`. But it doesn't read very well...

thomaslevesque:
Fixes #1062 

Also includes `IsNull` and `IsNotNull` overloads for nullable value types

### DIFF
--- a/docs/argument-constraints.md
+++ b/docs/argument-constraints.md
@@ -61,6 +61,7 @@ used. There are a few built-in matchers:
 |Matcher|Tests for|
 |:------|:--------|
 |IsNull()|`null`|
+|IsNotNull()|not `null`|
 |IsEqualTo(other)|object equality using `object.Equals`|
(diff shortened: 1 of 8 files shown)


thomaslevesque:
> Shall we augment http://fakeiteasy.readthedocs.io/en/stable/argument-constraints/#more-convenience-matchers while we're at it?

Done !

thomaslevesque:
oops... pushed to the wrong branch

PR #1065, opened 2017-04-23
Analyzer for argument constraints outside call spec

### LINKED ISSUE #1005: Diagnostic for usage of A<T> outside a call specification

thomaslevesque:
Fixes #1005


src/FakeItEasy.Analyzer.CSharp/Resources.resx
```diff
@@ -1,64 +1,5 @@
...
-    mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with 
-            : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
-            : and then encoded with base64 encoding.
-
-    mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array 
-            : using a System.ComponentModel.TypeConverter
-            : and then encoded with base64 encoding.
-    -->
```
thomaslevesque: This comment block is useless, and was causing noise in the diff because of .editorconfig

src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.shproj
```diff
@@ -9,5 +9,5 @@
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.Common.props" />
   <PropertyGroup />
   <Import Project="FakeItEasy.Analyzer.Shared.projitems" Label="Shared" />
-  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.VisualBasic.targets" />
+  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)\CodeSharing\Microsoft.CodeSharing.CSharp.targets" />
```
thomaslevesque: This was causing an issue where the "Add new item" dialog only showed VB templates. Not sure why it said VisualBasic in the first place...

src/FakeItEasy.Analyzer.Shared/NonVirtualSetupAnalyzer.cs
```diff
@@ -21,7 +21,13 @@
...
     internal class NonVirtualSetupAnalyzer : DiagnosticAnalyzer
     {
-        private static readonly ImmutableHashSet<string> DiagnosticsMap = CreateDiagnosticsMap();
+        private static readonly ImmutableHashSet<string> CallSpecMethods =
+            ImmutableHashSet.Create(
+                StringComparer.Ordinal,
+                "FakeItEasy.A.CallTo",
+                "FakeItEasy.A.CallTo`1",
+                "FakeItEasy.A.CallToSet`1",
+                "FakeItEasy.Fake`1.CallsTo`1");
```
thomaslevesque: Not related to the PR (I can move it to a different PR if you prefer), but the notion of "diagnostic map" made no sense in this case. It's a list, not a map, and we don't need a "diagnostic map", since this analyzer only produces one diagnostic.

src/FakeItEasy.Analyzer.Shared/UnusedReturnValueAnalyzer.cs
```diff
@@ -20,10 +20,25 @@ namespace FakeItEasy.Analyzer
...
+                "FakeItEasy.Fake`1.CallsTo`1",
+                "FakeItEasy.Fake`1.CallsToSet`1",
+                "FakeItEasy.Fake`1.AnyCall",
+                "FakeItEasy.ArgumentValidationConfigurationExtensions.WithAnyArguments`1",
+                "FakeItEasy.WhereConfigurationExtensions.Where`1",
+                "FakeItEasy.Configuration.IAnyCallConfigurationWithNoReturnTypeSpecified.WithReturnType`1",
+                "FakeItEasy.Configuration.IAnyCallConfigurationWithNoReturnTypeSpecified.WithNonVoidReturnType",
+                "FakeItEasy.Configuration.IArgumentValidationConfiguration`1.WhenArgumentsMatch",
+                "FakeItEasy.Configuration.IPropertySetterAnyValueConfiguration`1.To",
+                "FakeItEasy.Configuration.IWhereConfiguration`1.Where");
```
thomaslevesque: Again, not related to the PR (I can move it to a different PR if you prefer), but the notion of "diagnostic map" made no sense in this case. It's a list, not a map, and we don't need a "diagnostic map", since this analyzer only produces one diagnostic. I don't remember exactly why I had introduced this map, but now I don't see the point, and it was reused for no reason in `NonVirtualSetupAnalyzer`.

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/DiagnosticVerifier.cs
```diff
@@ -32,10 +32,7 @@ internal void VerifyCSharpDiagnostic(string source, params DiagnosticResult[] ex
         /// </summary>
         /// <returns>The diagnostic analyzer being tested.</returns>
         [SuppressMessage("Microsoft.Design", "CA1024:UsePropertiesWhereAppropriate", Justification = "It's not appropriate here")]
-        protected virtual DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer()
-        {
-            return null;
-        }
+        protected abstract DiagnosticAnalyzer GetCSharpDiagnosticAnalyzer();
```
thomaslevesque: Avoids cryptic error if you forget to override. Wasn't possible before because both `GetCSharpDiagnosticAnalyzer` and `GetVisualBasicDiagnosticAnalyzer` existed, but now only one of them exists based on compilation symbols.

tests/FakeItEasy.Analyzer.CSharp.Tests/ArgumentConstraintOutsideCallSpecAnalyzerTests.cs
```diff
@@ -0,0 +1,224 @@
...
+                "A<int>.That.IsEqualTo(42)",
+                "A<int>.That.Not.IsEqualTo(42)");
+
+        [Theory]
+        [InlineData("A<int>._")]
+        [InlineData("A<int>.Ignored")]
+        [InlineData("A<int>.That")]
+        [InlineData("A<int>.That.Not")]
+        [InlineData("A<int>.That.IsEqualTo(42)")]
+        [InlineData("A<int>.That.Not.IsEqualTo(42)")]
```
blairconrad: I probably wouldn't have thought to add the extra cases (`That`, `That.Not`). Nice.

src/FakeItEasy.Analyzer.Shared/ArgumentConstraintOutsideCallSpecAnalyzer.cs
```diff
@@ -0,0 +1,144 @@
...
+#if CSHARP
+                    case SyntaxKind.PropertyDeclaration:
+                    case SyntaxKind.MethodDeclaration:
+                    case SyntaxKind.ConstructorDeclaration:
+#elif VISUAL_BASIC
+                    case SyntaxKind.GetAccessorBlock:
+                    case SyntaxKind.SetAccessorBlock:
+                    case SyntaxKind.SubBlock:
+                    case SyntaxKind.ConstructorBlock:
+#endif
```
thomaslevesque: I just want to stop searching up the ancestor nodes when I reach the method level (whether the "method" is a method, property accessor or constructor. Unfortunately the node kinds are not the same for VB and C#. I found those using the Roslyn syntax tree visualizer in LinqPad.

thomaslevesque:
> I do have one concern. I think this test should fail, but it does not:

Oops. Looking into it.

thomaslevesque:
> I do have one concern. I think this test should fail, but it does not:

Actually, I'm not sure what this test is supposed to check... It doesn't even use the `constraint` parameter.

blairconrad:
HA! I had meant to prepare the test and then switch `foo` for `constraint`, because I was not sure what would happen. Then I, um... forgot the second part.

thomaslevesque:
> HA! I had meant to prepare the test and then switch foo for constraint, because I was not sure what would happen. Then I, um... forgot the second part.

I ran the test you intended to write, and it triggered the diagnostic as expected (I didn't actually add it to the code).

I removed the "simplification" commit and will create another PR for it.

blairconrad:
Thanks, @thomaslevesque!
Oh, man. I just remembered http://fakeiteasy.readthedocs.io/en/stable/analyzer/

blairconrad:
New PR, same issue?

thomaslevesque:
> Oh, man. I just remembered http://fakeiteasy.readthedocs.io/en/stable/analyzer/

I *always* forget the docs. I should have a checklist somewhere...

> New PR, same issue?

👍

blairconrad:
> I always forget the docs. I should have a checklist somewhere…

We could make a [PR template](https://help.github.com/articles/creating-a-pull-request-template-for-your-repository/).

thomaslevesque:
> We could make a PR template.

Yes, but the templates are mostly useful for casual contributors. We make PRs very frequently, so we would be more likely ignore the template ;)

blairconrad:
Just like me with that stop sign at the end of my street?

(I'm kidding. I'm a compulsive stopper.)

PR #1067, opened 2017-04-24
Add documentation for FakeItEasy0003 diagnostic

### LINKED ISSUE #1005: Diagnostic for usage of A<T> outside a call specification

thomaslevesque:
Fixes #1005

### DIFF
--- a/docs/analyzer.md
+++ b/docs/analyzer.md
@@ -16,9 +16,10 @@ PM> Install-Package FakeItEasy.Analyzer
 
 ## Diagnostics
 
-The analyzer currently provides two diagnostics:
+The analyzer currently provides the following diagnostics:
 
 | Id             | Summary                          | Description                                                                            |
 |----------------|----------------------------------|----------------------------------------------------------------------------------------|
 | FakeItEasy0001 | Unused call specification        | Triggered when you specify a call but don't configure or assert it, making it a no-op. |
 | FakeItEasy0002 | Non-virtual member configuration | Triggered when you try to configure a non-virtual member, which cannot be faked.       |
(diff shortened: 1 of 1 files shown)


blairconrad:
Ah. I didn't know you were around. I was just about to submit a competing PR with [this commit](https://github.com/blairconrad/FakeItEasy/commit/c5464694230ccd161d9c71769ebcfad6267b9ade).
My idea had been to link to the other articles, but I like your examples. Should we combine?

docs/analyzer.md
```diff
@@ -16,9 +16,10 @@ PM> Install-Package FakeItEasy.Analyzer
...
 ## Diagnostics
 
-The analyzer currently provides two diagnostics:
+The analyzer currently provides the following diagnostics:
 
 | Id             | Summary                          | Description                                                                            |
 |----------------|----------------------------------|----------------------------------------------------------------------------------------|
 | FakeItEasy0001 | Unused call specification        | Triggered when you specify a call but don't configure or assert it, making it a no-op. |
 | FakeItEasy0002 | Non-virtual member configuration | Triggered when you try to configure a non-virtual member, which cannot be faked.       |
+| FakeItEasy0003 | Argument constraint outside call specification | Triggered when you try to use an argument constraint (`A<T>.Ignored`, `A<T>.That...`) outside a call specification (`A.CallTo`, `Fake<T>.CallsTo`) |
```
blairconrad: aside from my comment about adding links to the other topics, how do you feel about a `&hellip;` instead of `...`? No pressure!

thomaslevesque:
Actually, I think I prefer your version. The examples are probably redundant if there are links to the relevant doc pages. But if you prefer to combine, I don't mind.

blairconrad:
I think I prefer mine (slightly). I just didn't know if that was because it was mine. :wink:. Thanks.
I'll continue continue making my commit a PR.

PR #1068, opened 2017-04-24
Add documentation for FakeItEasy0003 diagnostic

### LINKED ISSUE #1005: Diagnostic for usage of A<T> outside a call specification

blairconrad:
Fixes #1005. Closes #1067.
(Well, most of the work was done in #1065. This is just the finishing touch.)


thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you! #TeamFakeItEasy!

PR #1073, opened 2017-04-28
Fix false call setup warning when property index is a non-virtual member

### LINKED ISSUE #1072: Non-Virtual Setup Analyzer gives false positive when complex argument constraint used in indexer
Found in [the Nancy codebase](https://github.com/NancyFx/Nancy/blob/be4f8d42076e4e568a3742715437868e6c7d05af/test/Nancy.ViewEngines.Razor.Tests/TextResourceFinderFixture.cs#L68).
Can be reproduced by

```c#
namespace TheNamespace
{
    using FakeItEasy;

    public interface IHaveAnIndexer { string this[int context] { get; } }

    public class TheClass
    {
        public void Test()
        {
            var fake = A.Fake<IHaveAnIndexer>();
            A.CallTo(() => fake[A<int>._]).MustHaveHappened();
        }
    }
}
```

blairconrad:
Fixes #1072.

### DIFF
--- a/src/FakeItEasy.Analyzer.Shared/NonVirtualSetupAnalyzer.cs
+++ b/src/FakeItEasy.Analyzer.Shared/NonVirtualSetupAnalyzer.cs
@@ -73,12 +73,10 @@
 
             var invocationExpression = (T)context.Node;
 
-            var invocationParent = invocationExpression
-                                    .Ancestors()
-                                    .OfType<InvocationExpressionSyntax>()
-                                    .FirstOrDefault();
+            var invocationParent = FindEnclosingInvocation(invocationExpression);
 
-            if (!IsSetupInvocation(context, invocationParent))
+            if (invocationParent == null ||
+                !IsSetupInvocation(context, invocationParent))
             {
                 return;
             }
@@ -90,6 +88,28 @@
                 Diagnostic diagnostic = Diagnostic.Create(DiagnosticDefinitions.NonVirtualSetupSpecification, location, symbolInfo.Symbol.Name);
                 context.ReportDiagnostic(diagnostic);
             }
+        }
+
+        private static InvocationExpressionSyntax FindEnclosingInvocation<T>(T invocationExpression) where T : ExpressionSyntax
+        {
+            foreach (var ancestor in invocationExpression.Ancestors())
+            {
+#if CSHARP
+                if (ancestor is ElementAccessExpressionSyntax)
+                {
+                    // The invocation expression being analyzed is nested within a property indexer,
+                    // so it is not a call to set up a non-virtual method.
+                    return null;
+                }
+#endif
+                var enclosingInvocation = ancestor as InvocationExpressionSyntax;
+                if (enclosingInvocation != null)
+                {
+                    return enclosingInvocation;
+                }
+            }
+
+            return null;
         }
 
         private static bool IsVirtual(SymbolInfo symbolInfo)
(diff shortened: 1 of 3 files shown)


thomaslevesque:
Uh... apparently the change I suggested breaks things for VB.NET. Looking into it.

thomaslevesque:
OK, I have something that seems to work: https://github.com/thomaslevesque/FakeItEasy/commits/indexer-analyzer

blairconrad:
@thomaslevesque, your comments have merit, as always. I think maybe this should be your PR now, but I've merged your commits into mine because it was the easiest way for me to make comments. Comments forthcoming.

src/FakeItEasy.Analyzer.Shared/NonVirtualSetupAnalyzer.cs
```diff
@@ -128,9 +160,55 @@ private static bool IsMethod(SymbolInfo symbolInfo)
...
+
+            return false;
+        }
+
         private static bool IsInterfaceMember(SymbolInfo symbolInfo)
         {
             return symbolInfo.Symbol?.ContainingType?.TypeKind == TypeKind.Interface;
         }
+
+        private static InvocationExpressionSyntax FindInvocationInHierarchy(SyntaxNode node, IEnumerable<Ancestor> ancestors)
```
blairconrad: `ancestors` sounds to me like these are the actual ancestors, not the ones we require (or expect). Perhaps a name change here and for `NodeAncestors`?

Now that I notice it, we _could_ opt not to pass in `ancestors` and just use `NodeAncestors` directly.

tests/FakeItEasy.Analyzer.VisualBasic.Tests/NonOverridableSetupAnalyzerTests.cs
```diff
@@ -259,6 +259,83 @@ End Namespace
...
+            A.CallTo(Function() fake(A(Of Integer).Ignored)).MustNotHaveHappened()
+        End Sub
+    End Class
+End Namespace
+";
+            this.VerifyVisualBasicDiagnostic(Test);
+        }
+
+        [Fact]
+        public void Diagnostic_Should_Not_Be_Triggered_For_Non_Overridable_Indexer()
```
blairconrad: "Overridable_Indexer"

src/FakeItEasy.Analyzer.Shared/NonVirtualSetupAnalyzer.cs
thomaslevesque: > ancestors sounds to me like these are the actual ancestors, not the ones we require (or expect). Perhaps a name change here and for NodeAncestors?

What do you have in mind? Something like `expectedAncestors` or `requiredAncestors`?

> Now that I notice it, we could opt not to pass in ancestors and just use NodeAncestors directly.

Good point.

tests/FakeItEasy.Analyzer.VisualBasic.Tests/NonOverridableSetupAnalyzerTests.cs
thomaslevesque: Oops

src/FakeItEasy.Analyzer.Shared/NonVirtualSetupAnalyzer.cs
blairconrad: Perhaps `expectedHierarchy`?

thomaslevesque: Sounds good

thomaslevesque:
I pushed the changes to your branch

blairconrad:
Squashed!

PR #1074, opened 2017-05-04
Replace master with develop as default branch

### LINKED ISSUE #955: Adopt GitFlow
This arose from a [Gitter conversation about how we support older releases](https://gitter.im/FakeItEasy?at=586104e2c895451b7520c61a).

@adamralph proposed moving from GitHub Flow to [GitFlow](http://nvie.com/posts/a-successful-git-branching-model/#hotfix-branches). @thomaslevesque and @blairconrad agree.

We'll need to  (at least - augment, since I'm sure I've forgotten things)

- [x] create the `develop` branch (it will be cleaner to wait until we have a new tag, such as 3.0.0-beta003. I'd consider just renaming master and making a new master off 3.0.0-beta002, but I think this would disrupt people)
- [x] protect the branch in GitHub
- [x] configure AppVeyor to make it the default branch
- [ ] update all our "contributing/how to build" documentation

Those are the obvious steps, I think.

Less obvious, I think we'll need to overhaul our "next release" workflow. It's been a little problematic anyhow, I think, since we've been guessing about our next release version and then renaming or replacing milestones and releases and build issues. Let's see how we can improve things, especially now that GitFlow has more to say about releases than does GitHubFlow.
I see two potential paths forward. I'm sure there are more.

**The first** is that we can work essentially as we've been doing, except not assign versions to milestones, releases, and release issues. We could make placeholders for "next release" and assign specific version names when we need them (when we create a release branch). When we create a hotfix branch, we will also create a new set of these things. In that case, they can be tied to a version right away.

**The second** is more radical, and maybe should be split off into another discussion, but I'll introduce it here. My biases abound.

* I generally find milestones to be nothing but overhead. We work by marking an issue in-progress, and then when it's done, we add it to the release. Once we release, the milestone provides no value at all. I'd consider ditching milestones. We could reserve them for special cases, like "breaking issues to go in the next major" or "things that have to go in 4.0, even though that release isn't on the horizon (like removing Self-Initializing Fakes)", but I think we could get along without them just about all the time
* I've long thought we should keep release notes in an `.md` file in the repo. As soon as a PR that's releasenotable is merged (or as part of the PR, if the contributor is savvy), we update the file with the release note (and thanks to the contributors). If we do this, we could make our "release" step generate the GitHub release from that file. We could do this manually at first, but I've been toying with the idea of a tool that'll scrape the file and create the GitHub Release. (Longer term, _I'd_ like to see this file be the source of our current version number, but that's slightly more involved as well and, I imagine, contentious.)
* Then all we need is a release checklist. We could either always have one around and just name it when we decide to target a particular version (i.e. when we create the release branch) **or** not even make one until we create the release or hotfix branch (with a script/build target that will create the branch and checklist for us).

blairconrad:
Fixes #955.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,40 +76,42 @@ You should pick either the "personal" or "this computer" option.
 
 ## Making Changes
 
+FakeItEasy uses the git branching model known as [GitFlow](http://nvie.com/posts/a-successful-git-branching-model/). As such, all development must be performed on a ["feature branch"](https://martinfowler.com/bliki/FeatureBranch.html) created from the main development branch, which is called `develop`. To submit a change:
+
 1. [Fork](http://help.github.com/forking/) the  [FakeItEasy repository](https://github.com/FakeItEasy/FakeItEasy/) on GitHub
 1. Clone your fork locally
 1. Configure the upstream repo (`git remote add upstream git://github.com/FakeItEasy/FakeItEasy.git`)
-1. Create a local branch (`git checkout -b my-branch`)
+1. Create a local branch (`git checkout -b my-branch develop`)
 1. Work on your feature
 1. Rebase if required (see below)
 1. Run code analysis on the solution to ensure you have not introduced any violations
-1. Ensure the build succeeds (see ['How to build'](https://github.com/FakeItEasy/FakeItEasy/blob/master/how_to_build.md "How to build"))
+1. Ensure the build succeeds (see ['How to build'](how_to_build.md "How to build"))
 1. Push the branch up to GitHub (`git push origin my-branch`)
 1. Send a [pull request](https://help.github.com/articles/using-pull-requests) on GitHub
 
-You should **never** work on a clone of master and you should **never** send a pull request from master - always from a branch. The reasons for this are detailed below.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -76,48 +76,50 @@ You should pick either the "personal" or "this computer" option.
...
 1. Work on your feature
 1. Rebase if required (see below)
 1. Run code analysis on the solution to ensure you have not introduced any violations
-1. Ensure the build succeeds (see ['How to build'](https://github.com/FakeItEasy/FakeItEasy/blob/master/how_to_build.md "How to build"))
+1. Ensure the build succeeds (see ['How to build'](how_to_build.md "How to build"))
 1. Push the branch up to GitHub (`git push origin my-branch`)
 1. Send a [pull request](https://help.github.com/articles/using-pull-requests) on GitHub
 
-You should **never** work on a clone of master and you should **never** send a pull request from master - always from a branch. The reasons for this are detailed below.
+You should **never** work on a clone of develop and you should **never** send a pull request from develop - always from a branch. The reasons for this are detailed below.
```
thomaslevesque: This doesn't seem very clear to me... you clone a repo, not a branch. It should probably be rephrased (maybe something like "you should never work directly on the `develop` branch")

PR #1077, opened 2017-05-04
Reduce severity of FakeItEasy0003 to Warning

### LINKED ISSUE #1076: FakeItEasy0003 false positive when returning argument constraint from helper method
For instance this code doesn't compile because of FakeItEasy0003 error:

```
static int EvenNumber()
{
    return A<int>.That.Matches(i => i % 2 == 0);
}
```

Calling this method outside a call specification would throw, but calling it in a call specification is valid and works.

thomaslevesque:
Fixes #1076


thomaslevesque:
rebased

blairconrad:
The branch was actually successfully merged as 46169c924ca63da4fb0815f8c3a9ba6913200d63, but GitHub threw an error, resulting in this PR having a weird state. #1076 is fixed.

adamralph:
Moving the milestone to #1076

Issue #833, opened 2016-08-19
Switch analyser package to .NET Standard


adamralph:
Currently we produce a PCL Profile7 assembly (`net40`, `dotnet` and `win8`) and put it in a `dotnet` folder in the nupkg.

Whilst this works, `dotnet` (ASP.NET Core 1.0) is a deprecated moniker and we should stop using it. I think we should be using a .NET Standard moniker instead.

However, this works for now, and I don't think we should make this change until the .NET Standard tooling comes out of preview.

thomaslevesque:
This was fixed as a side effect of #1090

The analyzers now target `netstandard1.1`

blairconrad:
This change has been released as part of [FakeItEasy 3.4.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/3.4.0).

PR #1090, opened 2017-05-07
[WIP] Migrate to new csproj format

### LINKED ISSUE #991: Collapse netstd/framework packages
I know that we're eventually going to move to a full csproj-based solution anyhow, but I wonder if we wouldn't be better off collapsing the netstd projects so we have a single set of projects, not unlike [xBehave.net](https://github.com/xbehave/xbehave.net) and [SelfInitializingFakes](https://github.com/blairconrad/SelfInitializingFakes).
It may even make migrating to the new `csproj`s easier.

----
**Update** after discussion below, we've decided to wait until after we switch to VS2017's new csproj format.

When we _do_ make this switch, though, we may also want to 

- [ ]  generate a .net45 lib, which will
    - [ ] allow event handlers whose event args don't extend `EventArgs` (#1008)

### LINKED ISSUE #996: Replace StyleCop.MSBuild with StyleCop.Analyzers
StyleCop.MSBuild doesn't work with pure project.json projects, so in that respect, #991 is blocked by this. However, it may well work with the new csproj, so in that respect #991 may not be blocked by this.

While StyleCop.MSBuild _may_ work with the new csproj, I believe it's better to move to StyleCop.Analyzers. It uses the Roslyn analyzer infra, so no need for custom build targets, and no need for `Settings.StyleCop` and its custom editor (like any other Roslyn analyzer, StyleCop.Analyzers uses code analysis rulesets).

### LINKED ISSUE #1059: Migrate to VS2017 and new csproj
project.json is obsolete; it has never been very well supported in VS215, and is not supported at all in VS2017. We can probably expect support for it to be progressively dropped by most tools. VS2017 can't open the project.json projects in the FakeItEasy solution, which forces us to keep using VS2015.

We should migrate to the new csproj format, which has many benefits:
- no need to maintain separate projects for net40/netstd (see #991)
- easy to add new targets (e.g. net45, see #1008)
- much more concise than current csproj
- no need to list all source files in the project
- get rid of FakeItEasy.nuspec, packages.config and most of AssemblyInfo.cs; everything is in the csproj

thomaslevesque:
Fixes #1059
Fixes #991 
Fixes #996 

This is very much a work in progress, but I'm publishing it now to get early feedback.

Outstanding issues:

- [x] Stylecop is disabled on the migrated projects. I haven't yet looked at how to reenable it. => switch to StyleCop.Analyzers
- [x] MSB3277 warnings in "classic" projects referencing SDK-style csprojs. Need to build in diagnostic mode to get the details
- [x] The build script hasn't been updated yet
  - [x] Apparently the latest SDK doesn't let us pass xunit arguments via `dotnet test`. We should probably switch to `dotnet xunit` (EDIT: tested, works fine).
- [x] gitlink doesn't support the new csproj format (_"The default XML namespace of the project must be the MSBuild XML namespace"_) => switch to pdbgit
- [x] The approval tests fail due to `COMVisibleAttribute`. I haven't looked into it yet.
- [ ] Some projects cannot or might not need to be migrated
  - [x] Analyzer: Analyzers normally target PCL Profile 7, but it's not possible to target this TFM with the new csproj. However, it's also possible make analyzers that target netstandard1.1.
    - [x] Analyzer tests
  - [x] Approval tests: for the sake of consistency
    - [x] merge the two projects into one
  - [x] C# Examples: for the sake of consistency
  - [ ] VB.NET examples: VB.NET is not supported by the .NET Core tooling yet
  - [ ] IntegrationTests.VB: VB.NET is not supported by the .NET Core tooling yet


blairconrad:
> We should probably switch to dotnet xunit

:+1:. I like it for SelfInitializingFakes

thomaslevesque:
Nice benefit of the migration: code coverage in ReSharper now works :)

image

build.csx
```diff
@@ -148,32 +150,33 @@ targets.Add(
...
-        RunTests(unitTests);
-        RunDotNetTests(netstdUnitTestDirectories);
+        RunDotNetTests(unitTestDirectories);
+        RunTests(analyzerTestAssemblies);
     });
 
 targets.Add(
+    "analyzer",
+    DependsOn("build", "testsDirectory"),
+    () => RunTests(analyzerTestAssemblies));
```
blairconrad: Already covered in `unit`, which I think is the better place for it. I think this target can be removed.

build.csx
```diff
@@ -183,7 +186,7 @@ targets.Add(
         var version = GetVersion();
         foreach (var nuspec in nuspecs)
         {
-            Cmd(nuget, $"pack {nuspec} -Version {version} -OutputDirectory {outputDirectory} -NoPackageAnalysis");
+            Cmd("nuget", $"pack {nuspec} -Version {version} -OutputDirectory {outputDirectory} -NoPackageAnalysis");
```
blairconrad: Why this change? I see we still rely on the `.nuget/nuget.exe` elsewhere. Is there a way to rely on just one?

build.csx
```diff
@@ -266,3 +269,32 @@ public string GetVersion()
...
+        process.WaitForExit();
+        if (process.ExitCode != 0)
+        {
+            throw new InvalidOperationException($"The vswhere command exited with code {process.ExitCode}.");
+        }
+        else
+        {
+            string installationPath = process.StandardOutput.ReadToEnd().Trim();
+            if (string.IsNullOrEmpty(installationPath))
+                throw new InvalidOperationException("Visual Studio 2017 was not found");
```
blairconrad: I'm a stickler, but we have moustaches up on lines 289 and 291. Could we here as well?

build.csx
```diff
@@ -246,7 +249,7 @@ public void RunDotNetTests(IEnumerable<string> testDirectories)
     foreach (var testDirectory in testDirectories)
     {
         var xml = Path.GetFullPath(Path.Combine(testsDirectory, Path.GetFileName(testDirectory) + ".TestResults.xml"));
-        Cmd(testDirectory, "dotnet", $"test -c Release -nologo -notrait \"explicit=yes\" -xml {xml}");
+        Cmd(testDirectory, "dotnet", $"xunit -configuration Release -nologo -notrait \"explicit=yes\" -xml {xml}");
```
blairconrad: Is xunit adding the framework to the output path? I see `artifacts\tests\FakeItEasy.IntegrationTests.TestResults-net461.xml, but cannot tell how it was generated from this.

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -1,292 +1,65 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard1.6;net40</TargetFrameworks>
+    <AssemblyName>FakeItEasy</AssemblyName>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <FileAlignment>512</FileAlignment>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\..\FakeItEasy.snk</AssemblyOriginatorKeyFile>
+    <CodeAnalysisRuleSet>..\..\FakeItEasy.ruleset</CodeAnalysisRuleSet>
+    <DebugType>full</DebugType>
```
blairconrad: Should we not be pdbonly for the release build and full for debug?
I haven't checked other csprojs.

src/FakeItEasy/Properties/AssemblyInfo.cs
```diff
@@ -14,4 +14,4 @@
 [assembly: InternalsVisibleTo("FakeItEasy.IntegrationTests, PublicKey=00240000048000009400000006020000002400005253413100040000010001002f50d82092713be482c885861c984334606da0f54c78738a3dd0862fb2dfc6080f0780132cc65d88f0f0c70af74e8a53430962395bfc1a36fab08b7a2549d387e805c13cc84acd884447ec8c4dcfb6216df720f0998380f9c906b5de8141798d64661f036d47274e6ecb76c9cde5f4cf2b521040601e44b3914fbeb9f39127f9")]
 
 // Module level suppress messages.
-[assembly: SuppressMessage("Microsoft.Usage", "CA2243:AttributeStringLiteralsShouldParseCorrectly", Justification = "AssemblyInformationalVersion uses SemVer.")]
+[assembly: SuppressMessage("Microsoft.Usage", "CA2243:AttributeStringLiteralsShouldParseCorrectly", Justification = "AssemblyInformationalVersion uses SemVer.")]
```
blairconrad: Is the only change here to remove the newline?
Was it necessary?

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
```diff
@@ -119,13 +17,28 @@
...
-    <Error Condition="!Exists('..\..\packages\xunit.runner.visualstudio.2.1.0\build\net20\xunit.runner.visualstudio.props')" Text="$([System.String]::Format('$(ErrorText)', '..\..\packages\xunit.runner.visualstudio.2.1.0\build\net20\xunit.runner.visualstudio.props'))" />
-  </Target>
+
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp1.0'">
+    <DefineConstants>$(DefineConstants);FEATURE_NETCORE_REFLECTION</DefineConstants>
+    <PackageTargetFallback>$(PackageTargetFallback);dnxcore50;portable-net451+win81</PackageTargetFallback>
```
blairconrad: Do you know how many of these are still required?
I'm not that worried; just curious.

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
```diff
@@ -119,13 +17,28 @@
...
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net461'">
+    <DefineConstants>$(DefineConstants);FEATURE_BINARY_SERIALIZATION</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="FluentAssertions" Version="4.19.2" />
+    <PackageReference Include="xunit.core" Version="2.2.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
+    <PackageReference Include="Microsoft.TestPlatform.TestHost" Version="15.0.0" />
```
blairconrad: I'd hoped that most (all?) of the references above here would no longer be needed, since we reference FakeItEasy.Tests. Do you know if they are required?

build.csx
thomaslevesque: Oops. In fact, I intended to remove it from the `unit` target. I thought it would make sense to be able to test only the analyzer or only the library.

build.csx
thomaslevesque: it's a mistake. I don't know what I did here... It might have happened when I tried to use `dotnet restore` instead of nuget.

build.csx
thomaslevesque: I don't know. I'll check.

build.csx
thomaslevesque: Sure. If only we had StyleCop for csx scripts ^^

src/FakeItEasy/FakeItEasy.csproj
thomaslevesque: Right. I saw `<DebugType>full</DebugType>` in the old csproj, and didn't realize it was only for Debug. I'll change it.

src/FakeItEasy/Properties/AssemblyInfo.cs
thomaslevesque: Ah, it must be because of editorconfig. I had initially changed that file because I wanted to move some of the attributes to the csproj, but eventually rolled it back.

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
thomaslevesque: Not sure. It didn't work without it, but maybe that was before I updated FluentAssertions. I'll check.

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
thomaslevesque: Maybe they're not necessary anymore, but in general I prefer to be explicit... Anyway, I'll try without the explicit references.

build.csx
thomaslevesque: Yes, apparently it's added automatically by xunit. Nice touch!

src/FakeItEasy/FakeItEasy.csproj
thomaslevesque: I didn't specify `DebugType` in the other projects, so they use the default, which is probably fine.

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
thomaslevesque: You were right, they're not necessary anymore.

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
thomaslevesque: Again, you were right. The references are not necessary. That being said, I'd rather keep them, because having them implicitly inherited from a referenced project makes them harder to spot.

build.csx
blairconrad: I think I need to go simplify some SelfInitializingFakes tests…

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
blairconrad: Ah. Thanks for the info.

thomaslevesque:
Strange... The StyleCop.MSBuild package works in VisualStudio, but not on the command line. That was too easy to be true...

thomaslevesque:
All done.

The "Collapse approval projects into one" commit might be out of scope. If you prefer, I can move it to a separate PR.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.CSharp.csproj
```diff
@@ -0,0 +1,51 @@
...
+  <ItemGroup>
+    <EmbeddedResource Remove="VisualBasic/Resources.resx" />
+    <Compile Remove="objvb\**" />
+    <EmbeddedResource Remove="objvb\**" />
+    <None Remove="objvb\**" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <CodeAnalysisDictionary Include="..\..\FakeItEasy.Dictionary.xml" />
+    <Compile Include="../CommonAssemblyInfo.cs" />
```
blairconrad: not the end of the world, but any idea why this is a / instead of a \?

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -1,292 +1,72 @@
...
+    <DebugType>pdbonly</DebugType>
+  </PropertyGroup>
+
+  <!-- .NET Standard 1.6 -->
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard1.6'">
+    <DefineConstants>$(DefineConstants);FEATURE_NETCORE_REFLECTION;FEATURE_EXCEPTION_DISPATCH_INFO</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.6'">
```
blairconrad: I think I know the answer here, but we can't use the constants here, can we?
Seems like checking the framework all the time is double work.
For just the two it may not be so bad, but if we want to add 4.5 and others, it might become unwieldy.
Still, I bet it's not a thing.

src/FakeItEasy.Analyzer/Properties/AssemblyInfo.cs
```diff
@@ -1,11 +1,10 @@
 ﻿using System;
-using System.Reflection;
 using System.Runtime.CompilerServices;
 
-[assembly: AssemblyProduct("FakeItEasy.Analyzer.VisualBasic")]
-[assembly: AssemblyTitle("FakeItEasy.Analyzer.VisualBasic")]
-[assembly: AssemblyDescription("Diagnostic analyzers to warn about incorrect usage of FakeItEasy in VB.NET code.")]
```
blairconrad: I think we're now missing product/title/description from the analyzers.

src/FakeItEasy/FakeItEasy.csproj
thomaslevesque: Unfortunately, it's the only way to do it, as far as I can tell. Technically, we could have separate files that we include based on the `TargetFramework` (e.g. `<Import Project="FakeItEasy.$(TargetFramework).csproj" />`), but I'm not sure it would be more convenient.

src/FakeItEasy.Analyzer/Properties/AssemblyInfo.cs
thomaslevesque: Nope. The product and title are inferred from the assembly name, and the description is in the csproj.

thomaslevesque:
Thanks for the review @blairconrad!

Can I squash the fixup commits yet?

> And one about updating how_to_build.md:
> we should update how_to_build.md.

👍 
Will do

src/FakeItEasy/FakeItEasy.csproj
blairconrad: I don't think so either. Thanks.

src/FakeItEasy.Analyzer/Properties/AssemblyInfo.cs
blairconrad: I remain unconvinced.

image

blairconrad:
> Can I squash the fixup commits yet?

Oh, yes. Do.

blairconrad:
I'm showing this warning (I guess):

```
3>------ Rebuild All started: Project: FakeItEasy.IntegrationTests.VB, Configuration: Release Any CPU ------
3>  No way to resolve conflict between "mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" and "mscorlib, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e, Retargetable=Yes". Choosing "mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" arbitrarily.
3>  FakeItEasy.IntegrationTests.VB -> D:\Sandbox\FakeItEasy\tests\FakeItEasy.IntegrationTests.VB\bin\Release\FakeItEasy.IntegrationTests.VB.dll
```

It seems to be new. Any ideas?

FakeItEasy.ruleset
```diff
@@ -5,4 +5,17 @@
     <Rule Id="CA1020" Action="None" />
     <Rule Id="CA1305" Action="None" />
   </Rules>
+  <Rules AnalyzerId="StyleCop.Analyzer" RuleNamespace="StyleCop.Analyzer">
+    <Rule Id="SA1117" Action="None" />
```
blairconrad: Ironically, all our violations of this rule are in the `SuppressMessage`s.

FakeItEasy.ruleset
```diff
@@ -5,4 +5,17 @@
     <Rule Id="CA1020" Action="None" />
     <Rule Id="CA1305" Action="None" />
   </Rules>
+  <Rules AnalyzerId="StyleCop.Analyzer" RuleNamespace="StyleCop.Analyzer">
```
blairconrad: I see this section and StyleCop.Analyzer**s** in the IDE. Any idea why both? The latter has many more entries.

src/FakeItEasy.Analyzer/Properties/AssemblyInfo.cs
thomaslevesque: OK, I'll look into it

thomaslevesque:
> I'm showing this warning (I guess):

I've seen it too. I don't know what's causing it, but it doesn't seem to have any ill effect. It only shows up in the build output, not in the VS error list.

FakeItEasy.ruleset
thomaslevesque: Yes... Would you prefer to keep it enabled and fix the violations?

blairconrad: In general, I'm loathe to turn off the rules we previously had on. I'm not sure why these weren't caught before.
There are only about 8.

Me, I would be content with a "clean up the rules" issue to be performed later.

thomaslevesque: 👍

FakeItEasy.ruleset
thomaslevesque: It was a mistake. It should be StyleCop.Analyzers. Actually, I'm surprised it works with the wrong id... I guess the AnalyzerId is just ignored.

blairconrad: Woo hoo! I'm providing value!

src/FakeItEasy.Analyzer/Properties/AssemblyInfo.cs
thomaslevesque: Ah, I see... I had `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>` in the csproj (because of `CommonAssemblyInfo.cs`), so the attributes weren't generated.

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
```diff
@@ -39,7 +39,7 @@
     <ProjectReference Include="..\..\src\FakeItEasy\FakeItEasy.csproj" />
     <ProjectReference Include="..\..\tests\FakeItEasy.Tests\FakeItEasy.Tests.csproj" />
     <ProjectReference Include="..\FakeItEasy.IntegrationTests.External\FakeItEasy.IntegrationTests.External.csproj">
-      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+      <Private>false</Private>
```
thomaslevesque: There's a story here... The reason for this was to force `FakeItEasy.IntegrationTests` to depend on `FakeItEasy.IntegrationTests.External` so that `FakeItEasy.IntegrationTests.External` is always built before `FakeItEasy.IntegrationTests`. This was initially done with a dependency in the solution file. However, because of https://github.com/dotnet/cli/issues/6258, running `dotnet restore` at the solution level didn't restore packages for `FakeItEasy.IntegrationTests`. My first attempt to fix it was to remove the dependency from the solution, and to reference the project but without actually referencing the assembly. This worked fine on the command line, but sometimes failed to build in Visual Studio (in fact, it *always* failed... it had seemed to work at some point, but no longer). But in fact, it doesn't matter if we reference it in the project: since we don't actually use it, there's no reference in the generated assembly. The `<Private>false</Private>` just prevents it from being copied to the output directory.

thomaslevesque:
Re-enabling SA1117 is causing the build to fail, because of `TreatWarningsAsErrors`... I don't see any way to exclude StyleCop warnings. There is `WarningsNotAsErrors`, but we would have to list *all* StyleCop warnings (it doesn't seem to accept wildcards).

`StyleCopTreatErrorsAsWarnings`, which we were using before, has no effect, as @adamralph [noted a few months ago](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/2199).

I'm not sure how to solve this... Having the build failing because of StyleCop warnings might be acceptable for us core maintainers, but it's a sure way to discourage casual contributors.

thomaslevesque:
Oh, and apparently listing all SA warnings in `WarningsNotAsErrors` causes them to be re-enabled, even though they were disabled in the ruleset 😕

thomaslevesque:
A possible solution is to use a different ruleset for debug and release:

- debug: same as now
- release: set severity of CS* warnings to Error

But it's a pain to maintain...

thomaslevesque:
Regarding the conflict warning in FakeItEasy.IntegrationTests.VB: it's caused by FluentAssertions.Core, which references mscorlib 2.0.5.0. (thanks [MSBuildStructuredLog](https://github.com/KirillOsenkov/MSBuildStructuredLog), you saved me again!)

It's the only project that uses FluentAssertions and is not migrated to the new csproj.

thomaslevesque:
Also, the warning existed before, with a different wording:

> There was a conflict between mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089 and mscorlib, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e, Retargetable=Yes.

thomaslevesque:
> A possible solution is to use a different ruleset for debug and release:
> 
>     debug: same as now
>     release: set severity of CS* warnings to Error
> 
> But it's a pain to maintain...

I gave it a try. It works, but:
- We can create the list of warnings based on the Microsoft.CodeAnalysis.CSharp group in the rule set editor, but it only contains some of them. The first I tried to test that solution (CS0169 field is never used) wasn't in it, even though it's a common one
- We can't blindly add everything from CS0001 to CS9999, because some of them are not warnings (they have the Hidden or Info severity, e.g. "CS8019 Unnecessary using directive")
- There doesn't seem to be a complete list of compiler errors and warnings anywhere (actually a few weeks ago I even stumbled onto an error that wasn't documented anywhere).

So, we can make a list based on what we know, but it will probably be incomplete, and will have to be updated for future versions of C#...

thomaslevesque:
I pushed my experiment so that you can see what it looks like.

thomaslevesque:
> There doesn't seem to be a complete list of compiler errors and warnings anywhere

Oh, I found this: https://github.com/dotnet/roslyn/blob/master/src/Compilers/CSharp/Portable/Errors/ErrorCode.cs
This could be useful...

thomaslevesque:
OK, I did this: https://gist.github.com/thomaslevesque/0e530476ddbaadb22e827c7308fb3e45
(and amended the last commit according to the result)

We could include a target in the build script to regenerate it on demand, in case new warnings are added to the language.

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
thomaslevesque: Damn! It doesn't work for .NET Core, because the deps.json file knows about the External assembly, and fails when it tries to load it (we should probably catch that error, btw... it happens [here](https://github.com/FakeItEasy/FakeItEasy/blob/5880552eeca974d9c4ed072163e80eb5e052415d/src/FakeItEasy/Core/TypeCatalogue.cs#L82))...
I guess I'll just remove the dependency completely. The only downside is that it will be possible to run the integration tests in VS without building the External assembly first.

blairconrad:
Okay! I'm _way_ behind. And you're operating at a higher level than I am. I'll see what I can do.

thomaslevesque:
> Okay! I'm way behind. And you're operating at a higher level than I am. I'll see what I can do.

Don't worry, I'm about done ;)
I'll let you catch up now

blairconrad:
About the warnings, I may have missed a step, but if we're using the generated list of warnings to treat as errors, and we keep the StyleCop warnings as warnings, would that not put us back on the same footing as before your change? Would we need different configurations for release and debug?
I'm leery of having warnings in debug and errors in release. Best case, users do not address the warnings and then PR build fails. As much as I enjoy strictness, I'd hate to have that happen.

blairconrad:
Thanks for the explanation about the VB Integration Tests.

As far as the external integration test assembly goes, we're still in a situation where human might run integration tests in the IDE without having built the assembly, yes?
If no other solution presents itself, we could consider moving it out of the repo, building it separately, and then including it (via NuGet or just committing the DLL in this repo). It's not pretty, but would work around that.

thomaslevesque:
> About the warnings, I may have missed a step, but if we're using the generated list of warnings to treat as errors, and we keep the StyleCop warnings as warnings, would that not put us back on the same footing as before your change? Would we need different configurations for release and debug?

It puts us on the same footing as before the PR, i.e. `TreatWarningsAsErrors` in Release + `StyleCopTreatErrorsAsWarnings`.

> I'm leery of having warnings in debug and errors in release.

Why? It's what we already have...

> Best case, users do not address the warnings and then PR build fails. As much as I enjoy strictness, I'd hate to have that happen.

Not with my latest changes. If you look at the build output for my PR, the StyleCop warnings appear as warnings, not errors. But if there were any C# warnings, they would appear as errors.

thomaslevesque:
I've been thinking... My latest commit to simulate "warnings as errors" is an ugly hack. Is there really any benefit in having warnings as errors in Release? We always check for code analysis warnings before a release anyway, so we would also spot C# compiler warnings. Disabling "warnings as errors" would solve the problem and simplify the projects. What do you think?

blairconrad:
> > I'm leery of having warnings in debug and errors in release.

> Why? It's what we already have...

So we do! I did not know this. Or more likely, I forgot. Huh.

> >Best case, users do not address the warnings and then PR build fails. As much as I enjoy strictness, I'd hate to have that happen.

> Not with my latest changes. If you look at the build output for my PR, the StyleCop warnings appear as warnings, not errors. But if there were any C# warnings, they would appear as errors.

I meant the C# errors.

> I've been thinking... My latest commit to simulate "warnings as errors" is an ugly hack. Is there really any benefit in having warnings as errors in Release? We always check for code analysis warnings before a release anyway, so we would also spot C# compiler warnings. Disabling "warnings as errors" would solve the problem and simplify the projects. What do you think?

Part of me cringes, but most of me says "we've got a small group of responsible people who _will_ perform the checks you mentioned", and make corrections before things get too bad. Okay! Why not warnings?

thomaslevesque:
> Part of me cringes, but most of me says "we've got a small group of responsible people who will perform the checks you mentioned", and make corrections before things get too bad. Okay! Why not warnings?

OK, I'll remove the last commit then. It was a fun exercise, though. I even did [this](https://github.com/thomaslevesque/GenerateCSharpErrors/blob/master/CSharpErrorsAndWarnings.md) afterwards. At least now I know where to find the complete list of C# errors and warnings :wink:

build.csx
```diff
@@ -11,66 +11,57 @@ var solution = "./" + solutionName + ".sln";
...
-
-var netstdSpecDirectories = new []
+var pdbs = new []
 {
-    "tests/FakeItEasy.Specs.netstd"
+    "src/FakeItEasy/bin/Release/net40/FakeItEasy.pdb",
+    "src/FakeItEasy/bin/Release/netstandard1.6/FakeItEasy.pdb",
+    "src/FakeItEasy.Analyzer/bin/Release/netstandard1.1/FakeItEasy.Analyzer.Csharp.pdb",
+    "src/FakeItEasy.Analyzer/bin/Release/netstandard1.1/FakeItEasy.Analyzer.VisualBasic.pdb"
 };
```
blairconrad: This is fine, but we could consider just adding using `src/*/bin/Release/*/*.pdb`.

thomaslevesque: Good point.  I'll do it.

thomaslevesque: Actually, it's not that simple... `Directory.GetFiles` doesn't support globbing, so I have to do something like this:

```
var pdbs =
    (from srcDir in Directory.GetDirectories("./src")
     let binDir = Path.Combine(srcDir, "bin/Release")
     where Directory.Exists(binDir)
     from pdb in Directory.GetFiles(binDir, "*.pdb", SearchOption.AllDirectories)
     select pdb).ToArray();
```

which is:
- ugly
- way more complex
- harder to read

So, maybe the explicit list isn't that bad, after all...

blairconrad: `Directory.GetFiles` sucks!

blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
Thanks for the review!

adamralph:
FWIW, I added #833 to the "fixes" list.

Issue #1092, opened 2017-05-12
Improve handling of ReflectionTypeLoadException in TypeCatalogue


thomaslevesque:
Currently, if we get an error when scanning an assembly for extension points, we just emit a warning and dismiss the entire assembly. However, the error is often a `ReflectionTypeLoadException`, which gives us the option to be a little smarter:
- this exception exposes a `Types` property which is a list of types that were successfully loaded, so we could still scan those types (no need to dismiss the entire assembly just because one or two types couldn't be loaded)
- it also exposes a `LoaderExceptions` property with the errors that occurred when loading each type, so we could list the types that failed to load (and possibly indicate why they failed to load)

thomaslevesque:
Example of possible output:

```
Warning: FakeItEasy failed to get some types from assembly 'DynamicProxyGenAssembly2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=a621a9e7e5c32e69' while scanning for extension points. Some IArgumentValueFormatters, IDummyFactories, and IFakeOptionsBuilders in that assembly might not be available.
  Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
  The following loader exceptions occurred:
    - System.TypeLoadException: Could not load type 'Castle.Proxies.TypeWithNonConfigurableMethodProxy' from assembly 'DynamicProxyGenAssembly2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=a621a9e7e5c32e69'.
    - System.TypeLoadException: Could not load type 'Castle.Proxies.Invocations.Object_Equals_10' from assembly 'DynamicProxyGenAssembly2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=a621a9e7e5c32e69'.
```

thomaslevesque:
> (and possibly indicate why they failed to load)

Actually, no. The `InnerException`s of the exceptions from `LoaderExceptions` seem to always be null.

adamralph:
Would it be possible rename this issue to be more reflective of the functional benefit of this enhancement?

thomaslevesque:
> Would it be possible rename this issue to be more reflective of the functional benefit of this enhancement?

Better now?

adamralph:
👍 much better, thanks!

blairconrad:
This change has been released as part of [FakeItEasy 3.4.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/3.4.0).

PR #1095, opened 2017-05-12
Dynamically generate external extension point assembly

### LINKED ISSUE #1094: Generate external assembly with extension points for integration tests
As noted in https://github.com/FakeItEasy/FakeItEasy/pull/1090#issuecomment-300454071, the current assembly will not be compiled if a user runs the integration tests from the IDE without doing a command-line build, and then the tests would fail.

We could generate the assembly as part of the test fixture setup.

blairconrad:
Fixes #1094


tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        /// <summary>
+        /// Gets the path to a generated assembly that contains extension points, but is not referenced by the test assembly.
+        /// </summary>
+        public string AssemblyPath1 { get; }
+
+        /// <summary>
+        /// Gets the path to a copy of <see cref="AssemblyPath1"/>.
+        ///
+        public string AssemblyPath2 { get; }
```
thomaslevesque: Maybe `AssemblyOriginalPath` and `AssemblyCopyPath` instead? The 1 and 2 suffixes aren't very descriptive

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
```diff
@@ -0,0 +1,98 @@
...
+                "FakeItEasy.ExtensionPoints.External",
+                syntaxTrees: new[] { CSharpSyntaxTree.ParseText(assemblyContent) },
+                references: references,
+                options: new CSharpCompilationOptions(outputKind: OutputKind.DynamicallyLinkedLibrary));
+
+            this.TempAssemblyDirectory = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            Directory.CreateDirectory(this.TempAssemblyDirectory);
+
+            this.AssemblyPath1 = Emit(compilation, this.TempAssemblyDirectory, "1");
+            this.AssemblyPath2 = Emit(compilation, this.TempAssemblyDirectory, "2");
```
thomaslevesque: Why generate the assembly a second time? Couldn't you just copy the first file?

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
blairconrad: Yeah, I went back and forth on this. Will change as you suggest.

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
blairconrad: Too lazy! I'd have to make the directory and then copy.  I was of two minds about this as well. I'll update as you say.

blairconrad:
Okay, changes pushed (as a fixup). I switched to a consistently-named directory, cleaning it up beforehand, because I found that the assembly deletion was failing after the tests were run.
I'm not sure why I didn't see this the first time around.

thomaslevesque:
LGTM, thanks!

I'll let you squash before I merge. I'll need this for testing #1092.

blairconrad:
Thanks, @thomaslevesque. Squashed!

adamralph:
Moving the milestone from here to #1094

PR #1096, opened 2017-05-14
Handle case where some types can't be loaded from external assembly

### LINKED ISSUE #1092: Improve handling of ReflectionTypeLoadException in TypeCatalogue

thomaslevesque:
Fixes #1092 

The warning message looks like this:

```
Warning: FakeItEasy failed to get some types from assembly 'FakeItEasy.ExtensionPoints.External, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' while scanning for extension points. Some IArgumentValueFormatters, IDummyFactories, and IFakeOptionsBuilders in that assembly might not be available.
  System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
  1 type(s) were not loaded for the following reasons:
   - System.IO.FileNotFoundException: Could not load file or assembly 'FakeItEasy.ExtensionPoints.ExternalDependency, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.
```

Unfortunately it's not always possible to get the information about *which* types can't be loaded. The information is only present if the exception is a `TypeLoadException`.

### DIFF
--- a/src/FakeItEasy/Core/TypeCatalogue.cs
+++ b/src/FakeItEasy/Core/TypeCatalogue.cs
@@ -50,10 +50,20 @@ namespace FakeItEasy.Core
                         this.availableTypes.Add(type);
                     }
                 }
+                catch (ReflectionTypeLoadException ex)
+                {
+                    // In this case, some types failed to load.
+                    // Just keep the types that were successfully loaded, and warn about the rest.
+                    foreach (var type in ex.Types.Where(t => t != null))
+                    {
+                        this.availableTypes.Add(type);
+                    }
+
+                    WarnFailedToGetSomeTypes(assembly, ex);
+                }
                 catch (Exception ex)
                 {
                     WarnFailedToGetTypes(assembly, ex);
-                    continue;
                 }
             }
         }
@@ -73,8 +83,6 @@ namespace FakeItEasy.Core
 #if FEATURE_REFLECTION_GETASSEMBLIES
             var loadedAssemblies = AppDomain.CurrentDomain.GetAssemblies();
 #else
-            var fakeItEasyLibraryName = TypeCatalogue.FakeItEasyAssembly.GetName().Name;
-
             var context = DependencyContext.Default;
             var loadedAssemblies = context.RuntimeLibraries
                 .SelectMany(library => library.GetDefaultAssemblyNames(context))
@@ -107,7 +115,7 @@ namespace FakeItEasy.Core
         {
             foreach (var file in files)
             {
-                Assembly assembly = null;
+                Assembly assembly;
                 try
                 {
 #if FEATURE_REFLECTION_GETASSEMBLIES
@@ -148,25 +156,54 @@ namespace FakeItEasy.Core
         {
             Write(
                 ex,
-                "Warning: FakeItEasy failed to load assembly '{0}' while scanning for extension points. Any IArgumentValueFormatters, IDummyFactories, and IFakeOptionsBuilders in that assembly will not be available.",
-                path);
+                $"Warning: FakeItEasy failed to load assembly '{path}' while scanning for extension points. Any IArgumentValueFormatters, IDummyFactories, and IFakeOptionsBuilders in that assembly will not be available.");
         }
 
         private static void WarnFailedToGetTypes(Assembly assembly, Exception ex)
         {
             Write(
                 ex,
-                "Warning: FakeItEasy failed to get types from assembly '{0}' while scanning for extension points. Any IArgumentValueFormatters, IDummyFactories, and IFakeOptionsBuilders in that assembly will not be available.",
-                assembly);
+                $"Warning: FakeItEasy failed to get types from assembly '{assembly}' while scanning for extension points. Any IArgumentValueFormatters, IDummyFactories, and IFakeOptionsBuilders in that assembly will not be available.");
         }
 
-        private static void Write(Exception ex, string messageFormat, params object[] messageArgs)
+        private static void WarnFailedToGetSomeTypes(Assembly assembly, ReflectionTypeLoadException ex)
         {
             var writer = new DefaultOutputWriter(Console.Write);
-            writer.Write(messageFormat, messageArgs);
+            Write(
+                writer,
+                ex,
+                $"Warning: FakeItEasy failed to get some types from assembly '{assembly}' while scanning for extension points. Some IArgumentValueFormatters, IDummyFactories, and IFakeOptionsBuilders in that assembly might not be available.");
+
+            using (writer.Indent())
+            {
+                int notLoadedCount = ex.Types.Count(t => t == null);
+                writer.Write($"{notLoadedCount} type(s) were not loaded for the following reasons:");
+                writer.WriteLine();
+                foreach (var loaderException in ex.LoaderExceptions)
+                {
+                    writer.Write(" - ");
+                    writer.Write(loaderException.GetType());
+                    writer.Write(": ");
+                    writer.Write(loaderException.Message);
+                    writer.WriteLine();
+                }
+            }
+        }
+
+        private static void Write(Exception ex, string message)
+        {
+            var writer = new DefaultOutputWriter(Console.Write);
+            Write(writer, ex, message);
+        }
+
+        private static void Write(IOutputWriter writer, Exception ex, string message)
+        {
+            writer.Write(message);
             writer.WriteLine();
             using (writer.Indent())
             {
+                writer.Write(ex.GetType());
+                writer.Write(": ");
                 writer.Write(ex.Message);
                 writer.WriteLine();
             }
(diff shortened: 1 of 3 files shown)


src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -50,10 +50,20 @@ public void Load(IEnumerable<string> extraAssemblyFiles)
...
+                    {
+                        this.availableTypes.Add(type);
+                    }
+
+                    WarnFailedToGetSomeTypes(assembly, ex);
+                }
                 catch (Exception ex)
                 {
                     WarnFailedToGetTypes(assembly, ex);
-                    continue;
```
thomaslevesque: redundant

src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -73,8 +83,6 @@ private static IEnumerable<Assembly> GetAllAssemblies(IEnumerable<string> extraA
 #if FEATURE_REFLECTION_GETASSEMBLIES
             var loadedAssemblies = AppDomain.CurrentDomain.GetAssemblies();
 #else
-            var fakeItEasyLibraryName = TypeCatalogue.FakeItEasyAssembly.GetName().Name;
-
```
thomaslevesque: never used

src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -107,7 +115,7 @@ private static IEnumerable<Assembly> GetAssemblies(IEnumerable<string> files)
         {
             foreach (var file in files)
             {
-                Assembly assembly = null;
```
thomaslevesque: redundant

src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -148,28 +156,54 @@ private static void WarnFailedToLoadAssembly(string path, Exception ex)
...
+        private static void Write(Exception ex, string message, Action<IOutputWriter> additionalOutput = null)
         {
             var writer = new DefaultOutputWriter(Console.Write);
-            writer.Write(messageFormat, messageArgs);
+            writer.Write(message);
             writer.WriteLine();
             using (writer.Indent())
             {
+                writer.Write(ex.GetType());
+                writer.Write(": ");
```
thomaslevesque: Added this because it makes it easier to understand what the problem is. The exception type is often as useful as the message. I did it in a separate commit in case we don't want it.

src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -148,28 +156,54 @@ private static void WarnFailedToLoadAssembly(string path, Exception ex)
...
+                            writer.Write(loaderException.GetType());
+                            writer.Write(": ");
+                            writer.Write(loaderException.Message);
+                            writer.WriteLine();
+                        }
+                    }
+                });
+        }
+
+        private static void Write(Exception ex, string message, Action<IOutputWriter> additionalOutput = null)
```
thomaslevesque: The delegate parameter might not be the best way to do it... If you prefer, I could return the writer instead.

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
```diff
@@ -27,10 +29,16 @@ public ExternalAssemblyGenerator()
 
         /// <summary>
         /// Gets the path to a copy of <see cref="AssemblyOriginalPath"/>.
-        ///
+        /// </summary>
```
thomaslevesque: I missed this during the review of #1095...

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
```diff
@@ -27,10 +29,16 @@ public ExternalAssemblyGenerator()
...
         public string AssemblyCopyPath { get; }
 
+        /// <summary>
+        /// Gets the path of an assembly on which a type of FakeItEasy.IntegrationTests.External depends.
+        /// </summary>
+        public string AssemblyDependencyPath { get; }
+
         private const string AssemblyName = "FakeItEasy.ExtensionPoints.External";
+        private const string DependencyAssemblyName = "FakeItEasy.ExtensionPoints.ExternalDependency";
```
thomaslevesque: This assembly is referenced by `FakeItEasy.ExtensionPoints.External`, but won't be available at run time. This is done in order to cause a type load error when we scan the types in `FakeItEasy.ExtensionPoints.External`.

tests/FakeItEasy.IntegrationTests/TypeCatalogueTests.cs
```diff
@@ -127,6 +127,20 @@ public void Should_not_be_able_to_get_types_from_assembly_that_does_not_referenc
...
+        {
+            // Arrange
+            var catalogue = new TypeCatalogue();
+
+            // Act
+            string output = CaptureConsoleOutput(
+                () => catalogue.Load(new[] { this.externalAssemblyGenerator.AssemblyOriginalPath }));
+
+            // Assert
+            output.Should().Match("*Warning: FakeItEasy failed to get some types from assembly 'FakeItEasy.ExtensionPoints.External*' while scanning for extension points. Some IArgumentValueFormatters, IDummyFactories, and IFakeOptionsBuilders in that assembly might not be available.*");
```
blairconrad: Perhaps we should verify the rest of the message: make sure there was 1 type not found, and that it was named?

src/FakeItEasy/Core/TypeCatalogue.cs
blairconrad: I like it.

src/FakeItEasy/Core/TypeCatalogue.cs
blairconrad: Both feel a little awkward to me.
Would you think it terrible to pass the writer in? e.g.

```c#
var writer = new DefaultOutputWriter(Console.Write);
WriteError(writer, message, ex);
```

in two places, and just above follow it up with

```
using (writer.Indent())
{
    writer.Write($"{notLoadedCount} type(s) were not loaded for the following reasons:");
    writer.WriteLine();
    foreach (var loaderException in ex.LoaderExceptions)
    {
        writer.Write(" - ");
        writer.Write(loaderException.GetType());
        writer.Write(": ");
        writer.Write(loaderException.Message);
        writer.WriteLine();
    }
}
```

?

If that doesn't appeal, I think returning the writer is a little better than the delegate.

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
thomaslevesque: Agreed

tests/FakeItEasy.IntegrationTests/TypeCatalogueTests.cs
thomaslevesque: Ah, but it's *not* named in this case... The exception doesn't contain that information. The errors related to types in `DynamicProxyGenAssembly2` did contain the type name, because they were `TypeLoadException`s. In this case it's a `FileNotFoundException`.

thomaslevesque: But I can still check the rest of the message if you thing it's useful.

src/FakeItEasy/Core/TypeCatalogue.cs
thomaslevesque: In fact, we could even have a single `IOutputWriter` for the lifetime of the class. It would make things simpler.

tests/FakeItEasy.IntegrationTests/TypeCatalogueTests.cs
```diff
@@ -131,6 +131,23 @@ public void Should_not_be_able_to_get_types_from_assembly_that_does_not_referenc
...
+
+            // Act
+            string output = CaptureConsoleOutput(
+                () => catalogue.Load(new[] { this.externalAssemblyGenerator.AssemblyOriginalPath }));
+
+            // Assert
+            output.Should().Match(@"*Warning: FakeItEasy failed to get some types from assembly 'FakeItEasy.ExtensionPoints.External, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' while scanning for extension points. Some IArgumentValueFormatters, IDummyFactories, and IFakeOptionsBuilders in that assembly might not be available.
+  System.Reflection.ReflectionTypeLoadException: *.
+  1 type(s) were not loaded for the following reasons:
+   - System.IO.FileNotFoundException: *");
```
thomaslevesque: Using wildcards instead of the actual exception messages, so that it still works on non-English systems.

thomaslevesque:
Thanks for the review!

PR #1097, opened 2017-05-14
Exclude DynamicProxyGenAssembly2 when scanning for extension points

### LINKED ISSUE #1093: Suppress console output from TypeCatalogueTests
As discussed in #1097

Sometimes warnings appear on the console when running `TypeCatalogueTests`, because the type catalogue is loaded after fakes are created by other tests, which causes it to scan `DynamicProxyGenAssembly2`. This doesn't happen under normal conditions, so we should just suppress the output.

Not scanning `DynamicProxyGenAssembly2` at all has been considered, but would be a breaking change (unlikely to affect anyone, but who knows...)

thomaslevesque:
Fixes #1093

### DIFF
--- a/src/FakeItEasy/Core/TypeCatalogue.cs
+++ b/src/FakeItEasy/Core/TypeCatalogue.cs
@@ -1,4 +1,4 @@
-namespace FakeItEasy.Core
+﻿namespace FakeItEasy.Core
 {
     using System;
     using System.Collections.Generic;
@@ -91,7 +91,7 @@ namespace FakeItEasy.Core
                     // Exclude the ReflectionOnly assemblies because we may fully load them later.
                     .Where(a => !a.ReflectionOnly)
 #endif
-                    .Where(a => !a.IsDynamic)
+                    .Where(a => !a.IsDynamic && !a.Name().Equals("DynamicProxyGenAssembly2", StringComparison.OrdinalIgnoreCase))
                     .Select(a => a.Location),
                 StringComparer.OrdinalIgnoreCase);
 


src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -91,7 +91,7 @@ private static IEnumerable<Assembly> GetAllAssemblies(IEnumerable<string> extraA
                     // Exclude the ReflectionOnly assemblies because we may fully load them later.
                     .Where(a => !a.ReflectionOnly)
 #endif
-                    .Where(a => !a.IsDynamic)
+                    .Where(a => !a.IsDynamic && !a.Name().Equals("DynamicProxyGenAssembly2", StringComparison.OrdinalIgnoreCase))
```
blairconrad: I don't get it. From what I can tell, `DynamicProxyGenAssembly2` is dynamic, so I can't tell why this is needed.
It's not that I don't believe the problem exists. I'm just not sure why it does…

src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -1,4 +1,4 @@
-namespace FakeItEasy.Core
+﻿namespace FakeItEasy.Core
```
blairconrad: I see that VS2017 doesn't respect the lack of BOM (as specified by .editorconfig).
We should probably just not worry about it, no?

src/FakeItEasy/Core/TypeCatalogue.cs
thomaslevesque: Same here. I thought `!a.IsDynamic` would exclude it, but apparently it doesn't.

src/FakeItEasy/Core/TypeCatalogue.cs
thomaslevesque: Ah, yes. The .editorconfig in VS2017 is very limited, there are many features it doesn't support (e.g. final newline, trim trailing whitespace, encoding...).

I usually try not to commit this kind of change, but I missed it this time. I can change it back if you prefer.

blairconrad: > I usually try not to commit this kind of change, but I missed it this time.

Me too.

> I can change it back if you prefer.

I think worrying about this is just going to exhaust us.  And _every_ contribution from a community member is going to have the change. I don't want harass them about this.
Just let it go?

src/FakeItEasy/Core/TypeCatalogue.cs
blairconrad: Odd. I'm struggling with this issue. On the one hand, I believe that you've seen this problem. I just struggle to  understand how. And without understanding how, I have less confidence in the fix.

First, it's rare to reference `DynamicProxyGenAssembly2` before `TypeCatalogue.Load` is called.
Have you seen the error in the wild, or only while running the integration tests?
Are you able to reproduce it at will? How?
Do you know where the error was triggered? Is it the same as in #1092, where you see problems loading a particular type?

I've hacked up the code to throw when it sees `DynamicProxyGenAssembly2`, and to record its dynamicness, and it invariably says that it's dynamic.

Help me believe in the fix?

blairconrad: I think I'm working toward understanding.

I think this is happening when DynamicProxyGenAssembly2 references FakeItEasy. Certainly, I can reproduce the problem in that situation, which occurs when tests that create Fakes have already been run and then we try to load the `TypeCatalogue`. In that case, I think that DynamicProxyGenAssembly2 is being stripped out by the `IsDynamic` check, but is added back in by

```c#
            return GetAssemblies(extraAssemblyFiles.Except(loadedAssemblyFiles))
-->             .Concat(loadedAssembliesReferencingFakeItEasy)                    // <---- this line here
                .Concat(FakeItEasyAssembly)
                .Distinct();
```

And that's when we have problems.

What think you?

src/FakeItEasy/Core/TypeCatalogue.cs
thomaslevesque: 👍

thomaslevesque:
> Have you seen the error in the wild, or only while running the integration tests?

Only while running the integration tests, and only about 1 time out of 5 or 10

src/FakeItEasy/Core/TypeCatalogue.cs
thomaslevesque: > Have you seen the error in the wild, or only while running the integration tests?

Only while running the integration tests, and only about 1 time out of 5 or 10

> Do you know where the error was triggered? Is it the same as in #1092, where you see problems loading a particular type?

I don't know where exactly. It is what prompted #1092, but it's not the same.

> I think this is happening when DynamicProxyGenAssembly2 references FakeItEasy

Doesn't it always reference FakeItEasy? As soon as we create a fake, at least?

> What think you?

This might be the explanation. I'll have a look later today.

blairconrad:
It does always reference FIE, but not until we make a Fake, and
TypeCatalogue loads first. Except in the integration tests. Or possibly
some other contrived example. Like if someone were to use Dynamic Proxy to
create a Dummy Factory. Which is unlikely, but would be a reason not to fix
this issue..

src/FakeItEasy/Core/TypeCatalogue.cs
blairconrad: It does always reference FIE, but not until we make a Fake, and
TypeCatalogue loads first. Except in the integration tests. Or possibly
some other contrived example. Like if someone were to use Dynamic Proxy to
create a Dummy Factory. Which is unlikely, but would be a reason not to fix
this issue…

thomaslevesque: OK, I had a closer look, and I now have a better understanding of what is happening. We only filter out dynamic assemblies when we [build the list of already loaded assembly files](https://github.com/FakeItEasy/FakeItEasy/blob/a9ba9a42d14db7171c2beac3a04702a4bee8d272/src/FakeItEasy/Core/TypeCatalogue.cs#L87-L96). But we still [scan those assemblies](https://github.com/FakeItEasy/FakeItEasy/blob/a9ba9a42d14db7171c2beac3a04702a4bee8d272/src/FakeItEasy/Core/TypeCatalogue.cs#L101)... Which means my fix doesn't actually fix anything: it just prevents us from loading `DynamicProxyGenAssembly2` again from a file, which is impossible anyway.

I think the only reason we're filtering out dynamic assemblies is because they don't have a `Location`, and accessing that property would cause a `NotSupportedException`. But we don't exclude them otherwise, and we still try to load extension points from them.

Technically, it would be a breaking change to stop doing that. If a user generates extension points with Castle.Core before using FakeItEasy, they will be in `DynamicProxyGenAssembly2 `, and the user will expect them to be loaded. So we can't even filter out `DynamicProxyGenAssembly2`. Admittedly, this is a pretty unlikely scenario... We *could* assume nobody is doing that now, and make it an unsupported scenario. Or we could do nothing, and let the warnings show up sometimes in the integration tests (or we could capture the console output so that they don't show up in the build output).

What do you think?

blairconrad: Ah, good. Your reasoning about what's going on (and why), and the effectiveness of your change, duplicated mine exactly.

Me, I'd lean toward

> capture the console output so that they don't show up in the build output

thomaslevesque: > Me, I'd lean toward
> 
> >    capture the console output so that they don't show up in the build output

👍 

I'll do just that, then

thomaslevesque:
Closing this PR, because the branch name no longer makes sense, and I'm not keeping anything from it anyway.

PR #1098, opened 2017-05-15
Capture output of integration tests

### LINKED ISSUE #1093: Suppress console output from TypeCatalogueTests

thomaslevesque:
to suppress warnings when trying to load types from `DynamicProxyGenAssembly2`.

Fixes #1093 
Supersedes #1097

### DIFF
--- a/tests/FakeItEasy.IntegrationTests/TypeCatalogueTests.cs
+++ b/tests/FakeItEasy.IntegrationTests/TypeCatalogueTests.cs
@@ -11,6 +11,10 @@
     using FluentAssertions;
     using Xunit;
 
+    // Note: The console output is captured in all tests in this class, even when we don't need to use it.
+    // This is done in order to suppress a warning that sometimes occurs when these tests run after a fake
+    // has already been created. This situation normally happens only during tests, not during normal usage
+    // of the library.
     public class TypeCatalogueTests : IClassFixture<ExternalAssemblyGenerator>
     {
         public TypeCatalogueTests(ExternalAssemblyGenerator externalAssemblyGenerator)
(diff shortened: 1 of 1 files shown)


thomaslevesque:
> Long story short (too late!), I had to leave a comment here to change my review comment about a comment from "approve" to "comment".

Which didn't actually remove your "approve" review :wink:

thomaslevesque:
> The only thing I'd suggest is a class-level comment explaining why we always capture the output even if we're not going to use it, since someone coming in later (e.g. me, next week) will have no clue.

👍

thomaslevesque:
Done (amended the commit)

The `////` is because StyleCop doesn't like it when a so-called "single-line comment" (starting with `//`) is followed by a blank like (this is possibly the most useless StyleCop rule ever).

blairconrad:
Thanks for the update!

Oh, yes. I remember a previous discussion about that rule and `////`. I've been content with just deleting the empty line but if you prefer the `////` and the extra line, that's fine too. I'll give you a little while to comment.

thomaslevesque:
> Oh, yes. I remember a previous discussion about that rule and ////. I've been content with just deleting the empty line but if you prefer the //// and the extra line, that's fine too. I'll give you a little while to comment.

Well, I don't really like deleting the empty line, because it makes it look like the comment applies to the constructor. Alternatively, I could use a multiline comment if you prefer.

blairconrad:
Oh, sorry. I didn't pay that much attention to the location. I'd've put the comment right above the class.
But what you have now is fine. I'm not going to give you too much grief!

thomaslevesque:
> I'd've put the comment right above the class.

That works for me. I'll do it.

blairconrad:
Thanks, @thomaslevesque!

adamralph:
Removing from 3.4.0 milestone (#1093 is on the milestone).

PR #1099, opened 2017-05-16
Publish artifacts even when the build fails

blairconrad:
Because if we don't, it's sometimes hard to find out why the build fails.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,6 +19,9 @@ build_script:
         & .\build.cmd
     }
 
+on_failure:
+  - ps: $root = (Resolve-Path artifacts); [IO.Directory]::GetFiles($root.Path, '*.*', 'AllDirectories') | % { Push-AppveyorArtifact $_ -FileName $_.Substring($root.Path.Length + 1) -DeploymentName failed-build }
+
 nuget:
     disable_publish_on_pr: true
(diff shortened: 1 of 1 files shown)


blairconrad:
[Fancy](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/243-sayfabhp/artifacts).

blairconrad:
And when [the build passes](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy), nothing bad happens!

thomaslevesque:
Nice! But there might be an easier way: https://github.com/appveyor/ci/issues/401#issuecomment-301649481

thomaslevesque:
> Nice! But there might be an easier way: appveyor/ci#401 (comment)

In fact, I'm not sure it helps, since it only pushes one artifact. Your solution is probably better, and it's what the docs suggest.

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thanks, @thomaslevesque. Happy build-beaking!

blairconrad:
This change has been released as part of [FakeItEasy 3.4.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/3.4.0).

PR #1101, opened 2017-05-16
Fix PdbGit

### LINKED ISSUE #1100: The packaged PDB doesn't contain source server information
This is because `dotnet xunit` rebuilds all the assemblies before testing, which overwrites the PDB enriched by PdbGit.

We should run PdbGit *after* the tests, right before making the NuGet package, and/or update to `dotnet-xunit` 2.3.0-beta2, which has a `-nobuild` option.

thomaslevesque:
Fixes #1100

The second commit changes the order of build targets. Whether we keep it or not is open to discussion.

### DIFF
--- a/build.csx
+++ b/build.csx
@@ -286,7 +286,7 @@ class DotnetTestSuite : TestSuite
     public override void Execute()
     {
         var xml = Path.GetFullPath(Path.Combine(testsDirectory, Path.GetFileName(this.TestDirectory) + ".TestResults.xml"));
-        Cmd(this.TestDirectory, "dotnet", $"xunit -configuration Release -nologo -notrait \"explicit=yes\" -xml {xml}");
+        Cmd(this.TestDirectory, "dotnet", $"xunit -configuration Release -nologo -notrait \"explicit=yes\" -xml {xml} -nobuild");
     }
 }
 
--- a/build.csx
+++ b/build.csx
@@ -61,7 +61,7 @@ static var testsDirectory = "./artifacts/tests";
 // targets
 var targets = new TargetDictionary();
 
-targets.Add("default", DependsOn("pdbgit", "unit", "integ", "spec", "approve", "pack"));
+targets.Add("default", DependsOn("unit", "integ", "spec", "approve", "pack"));
 
 targets.Add("outputDirectory", () => Directory.CreateDirectory(outputDirectory));
 
@@ -71,7 +71,7 @@ targets.Add("logsDirectory", () => Directory.CreateDirectory(logsDirectory));
 
 targets.Add("testsDirectory", () => Directory.CreateDirectory(testsDirectory));
 
-targets.Add("build", DependsOn("clean", "outputDirectory", "restore"), () => RunMsBuild("Build"));
+targets.Add("build", DependsOn("clean", "restore"), () => RunMsBuild("Build"));
 
 targets.Add(
     "coverity",
@@ -153,7 +153,7 @@ targets.Add(
 
 targets.Add(
     "pack",
-    DependsOn("build", "outputDirectory"),
+    DependsOn("build", "outputDirectory", "pdbgit"),
     () =>
     {
         var version = GetVersion();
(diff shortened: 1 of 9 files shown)


build.csx
```diff
@@ -61,7 +61,7 @@ static var testsDirectory = "./artifacts/tests";
 // targets
 var targets = new TargetDictionary();
 
-targets.Add("default", DependsOn("pdbgit", "unit", "integ", "spec", "approve", "pack"));
+targets.Add("default", DependsOn("unit", "integ", "spec", "approve", "pdbgit", "pack"));
```
blairconrad: An alternative would be to have pack depend on pdbgit, especially if our model is "there's no point in packing something that hasn't been pdbgitted".

blairconrad:
It don't work, @thomaslevesque!

thomaslevesque:
NuGet timeout again...

build.csx
thomaslevesque: 👍

thomaslevesque:
OK, it's not just NuGet timeouts after all

> Detection failed! Please ensure you're using 'xunit.core' v2.3 beta 2 or later

thomaslevesque:
I can repro locally after deleting my packages/bin/obj folders

build.csx
```diff
@@ -153,7 +153,7 @@ targets.Add(
 
 targets.Add(
     "pack",
-    DependsOn("build", "outputDirectory"),
+    DependsOn("build", "pdbgit"),
```
blairconrad: I liked the removal of `outputDirectory` until I realized that `build` doesn't actually use it. Maybe keep here and remove from the `build` dependencies?

build.csx
```diff
@@ -286,7 +286,7 @@ class DotnetTestSuite : TestSuite
     public override void Execute()
     {
         var xml = Path.GetFullPath(Path.Combine(testsDirectory, Path.GetFileName(this.TestDirectory) + ".TestResults.xml"));
-        Cmd(this.TestDirectory, "dotnet", $"xunit -configuration Release -nologo -notrait \"explicit=yes\" -xml {xml}");
+        Cmd(this.TestDirectory, "dotnet", $"xunit -configuration Release -nologo -notrait \"explicit=yes\" -xml {xml} -nobuild");
```
blairconrad: I thought about this late, but the command might be easier to read if the shorter options come first. Maybe before "-notrait"? This is _such_ a minor deal.
I wouldn't've mentioned it if I wasn't already commenting on the outputDirectory issue.

Oh. While I'm here, I see that there are no quotes around `{xml}`, which might cause us problems if a user has spaces in their path. But this is technically a separate issue.

build.csx
blairconrad: Huh. I removed it from here and tried a build, and it all worked - I think nuget is making the directory itself.
If we're brave, we could drop the target.

thomaslevesque: Why not just add `outputDirectory` as a dependency of `pack`? It's the only target that uses this directory.

build.csx
thomaslevesque: > But this is technically a separate issue.

Yes, I'd rather do it separately

build.csx
blairconrad: Yeah, that was what my first comment intended, but then I noticed that it wasn't _required_. Still, having the dependency is intention-revealing.

blairconrad:
Oh, I wonder if 1952439e69c2b44acec71a61e6b8bbc2f5e41f62 and 29f7388846a18b95480b8e535e0161da2099cfbf shouldn't be squashed together, since the first commit is actively detrimental without the second…

No big deal, really.

thomaslevesque:
> Oh, I wonder if 1952439 and 29f7388 shouldn't be squashed together, since the first commit is actively detrimental without the second…
> 
> No big deal, really.

Make sense. It's no trouble, since I need to `rebase -i` anyway.

blairconrad:
> Make sense. It's no trouble, since I need to `rebase -i` anyway.

Yeah. Had you not been fixupping anyhow, I wouldn't've asked.

Thanks, @thomaslevesque.

PR #1103, opened 2017-05-18
[WIP] What is going on?

blairconrad:
Hey. I was playing around with the fixture initializer and found something that I can not figure out.

2d6e57a, which just removes a file, breaks the build FakeItEasy.Tests build with 

```
Builders\BuilderForExpression.cs(6,43): error CS0246: The type or namespace name 'TestDataBuilder<,>' could not be found (are you missing a using directive or an assembly reference?) [D:\Sandbox\FakeItEasy\tests\FakeItEasy.Tests\FakeItEasy.Tests.csproj]
Builders\BuilderForExpression.cs(30,39): error CS0115: 'BuilderForExpression.Build()': no suitable method found to override [D:\Sandbox\FakeItEasy\tests\FakeItEasy.Tests\FakeItEasy.Tests.csproj]
Builders\BuilderForParsedArgumentExpression.cs(10,57): error CS0246: The type or namespace name 'TestDataBuilder<,>' could not be found (are you missing a using directive or an assembly reference?) [D:\Sandbox\FakeItEasy\tests\FakeItEasy.Tests\FakeItEasy.Tests.csproj]
Builders\BuilderForParsedArgumentExpression.cs(48,53): error CS0115: 'BuilderForParsedArgumentExpression.Build()': no suitable method found to override [D:\Sandbox\FakeItEasy\tests\FakeItEasy.Tests\FakeItEasy.Tests.csproj]
```

Any ideas?

I tried _just_ applying that commit to `develop`, and it doesn't break the build. I'm flummoxed.


thomaslevesque:
I don't understand it either...

I can repro on my machine. If you comment the whole file instead of deleting it, it works fine. It's as if the presence of this file was necessary for the build...

blairconrad:
It's super weird.
If only I understood which MS people to at-mention. 

Thanks for checking it out.

thomaslevesque:
The more I look into it, the more it looks like a compiler bug...

- rename the file to anything else (still with a .cs extension): success
- rename the file to anything else (without a .cs extension): failure
- remove everything in the file: success
- remove the file : failure
- also remove another test fixture file: failure, with many more unexplainable errors
- move the file to the parent folder: failure

So:
- it's not the content of the file that matters, since the build succeeds when it's empty
- it's not the name of the file that matters, since we can rename it (as long as it has the .cs extension)
- the presence of the file (whatever its name is) in the Core folder matters

thomaslevesque:
Also, it only happens for the netcoreapp1.0 target

thomaslevesque:
> If only I understood which MS people to at-mention.

I'm not sure, but if it's a compiler bug, we can probably open an issue on https://github.com/dotnet/roslyn

blairconrad:
I knew you'd know things.
It's a bit big, but is kind of an _easy_ repro…

thomaslevesque:
Maybe it's not the compiler after all, but rather MSBuild...

I ran the build through [MSBuild Structured Log Viewer](https://github.com/KirillOsenkov/MSBuildStructuredLog), and noticed that while the `net461` target uses all the source files it's supposed to, the `netcoreapp1.0` one does not...

image

(the `SelfInitializedFakes` files are explicitly excluded in the csproj for `netcoreapp1.0`)

blairconrad:
Whoa.

thomaslevesque:
Also, renaming TestDataBuilder.cs to ATestDataBuilder.cs, or TestCases.cs to ATestCases.cs, fixes the problem...

thomaslevesque:
Looks like a bug in the file globbing algorithm... but it works fine for net461, and I can't see anything that would explain the difference.

adamralph:
There are indeed problems with the globbing that is used by `<Compile Remove="..." />` - https://github.com/dotnet/project-system/issues/2134

thomaslevesque:
Moved to https://github.com/Microsoft/msbuild/issues/2069

blairconrad:
Closing, since Microsoft/msbuild#2069 exists

thomaslevesque:
For future reference, replacing this:

    <Compile Remove="SelfInitializedFakes/*.cs" />

with either of these lines:

    <Compile Remove="SelfInitializedFakes\*.cs" />
    <Compile Remove="SelfInitializedFakes/**/*.cs" />

fixes the problem

PR #1105, opened 2017-05-25
Make MustHaveHappened build call description lazily

### LINKED ISSUE #796: Have RuleBuilder.MustHaveHappened build call description lazily
Currently the description is built even before checking to see whether the call happened.
If we delayed the description-building, we could
1. maybe save some time? If people had complicated descriptions that never needed to be written, and
2. support dynamic call descriptions, perhaps that included information about the supplied argument that did not match

See more on this in my recent discussions with @dennisdoomen [in Gitter](https://gitter.im/FakeItEasy/FakeItEasy?at=577bc1b1ef3990705bb60293).

With _very hacky_ experiment [in my repo](https://github.com/blairconrad/FakeItEasy/commit/0493991054351809457599106e2398a9eb7e0510).

blairconrad:
Fixes #796

### DIFF
--- a/src/FakeItEasy/Configuration/RuleBuilder.cs
+++ b/src/FakeItEasy/Configuration/RuleBuilder.cs
@@ -143,7 +143,7 @@ namespace FakeItEasy.Configuration
 
             var asserter = this.asserterFactory.Invoke(this.Calls);
 
-            asserter.AssertWasCalled(this.Matcher.Matches, this.RuleBeingBuilt.WriteDescriptionOfValidCall, repeatConstraint.Matches, repeatConstraint.ToString());
+            asserter.AssertWasCalled(this.Matcher.Matches, this.RuleBeingBuilt.WriteDescriptionOfValidCall, repeatConstraint);
 
             return new UnorderedCallAssertion(this.manager, this.Matcher, this.RuleBeingBuilt.WriteDescriptionOfValidCall, repeatConstraint);
         }
--- a/src/FakeItEasy/Configuration/RuleBuilder.cs
+++ b/src/FakeItEasy/Configuration/RuleBuilder.cs
@@ -143,12 +143,9 @@ namespace FakeItEasy.Configuration
 
             var asserter = this.asserterFactory.Invoke(this.Calls);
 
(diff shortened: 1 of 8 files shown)


src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -143,12 +143,9 @@ public UnorderedCallAssertion MustHaveHappened(Repeated repeatConstraint)
 
             var asserter = this.asserterFactory.Invoke(this.Calls);
 
-            var description = new StringBuilderOutputWriter();
-            this.RuleBeingBuilt.WriteDescriptionOfValidCall(description);
+            asserter.AssertWasCalled(this.Matcher.Matches, this.RuleBeingBuilt.WriteDescriptionOfValidCall, repeatConstraint.Matches, repeatConstraint.ToString());
```
thomaslevesque: Nitpick: couldn't we pass `repeatConstraint` directly instead of its string representation?

blairconrad: Thanks, @thomaslevesque. Yeah, we could. I noticed the same thing, but figured it was _technically_ outside the mandate of the PR, so I didn't do it at that time, but it was a close thing. I'm happy to do now. 
I've added a commit (and rebased).

src/FakeItEasy/Core/IFakeAsserter.cs
```diff
@@ -4,6 +4,6 @@ namespace FakeItEasy.Core
 
     internal interface IFakeAsserter
     {
-        void AssertWasCalled(Func<IFakeObjectCall, bool> callPredicate, Action<IOutputWriter> callDescriber, Func<int, bool> repeatPredicate, string repeatDescription);
+        void AssertWasCalled(Func<IFakeObjectCall, bool> callPredicate, Action<IOutputWriter> callDescriber, Repeated repeatConstraint);
```
thomaslevesque: Ah, I hadn't realized it would also remove one parameter. Even better!

thomaslevesque:
Nice work, thanks @blairconrad!

blairconrad:
Thanks, Big T.

PR #1106, opened 2017-05-25
Fix some typos in the Argument Constraints docs

### LINKED ISSUE #1104: Typo in "Always place Ignored and That inside A.CallTo"
http://fakeiteasy.readthedocs.io/en/stable/argument-constraints/#always-place-ignored-and-that-inside-acallto:

> The Ignored (and _) and That matchers must be placed within the expression inside the  A.CallTo call. This is because these special constraint methods do not return an actual matcher object. They tell FakeItEasy how to match the parameter via a special event that's fired **then** the constraint method is invoked. FakeItEasy only listens to the events in the context of an A.CallTo.

I added emphasis on the "then" that should be "when"

blairconrad:
Fixes #1104.

I fixed up a few other typos while I was here.

### DIFF
--- a/docs/argument-constraints.md
+++ b/docs/argument-constraints.md
@@ -15,7 +15,7 @@ public interface IFoo
 }
 ```
 
-Then the arguments to Bar can be constrained used to limit call matching:
+Then the arguments to Bar can be constrained to limit call matching:
 ```csharp
 var foo = A.Fake<IFoo>();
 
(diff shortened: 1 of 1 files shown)


docs/argument-constraints.md
```diff
@@ -85,8 +85,8 @@ A<string>.That.Matches(s => s.Length == 3 && s[1] == 'X');
 ```
 
 FakeItEasy will evaluate the predicate against any supplied
-argument. The predicate can be supplied as an `Expression<Func<T,
-bool>>` or as a `Func<T, bool>`. FakeItEasy can generate a description
+argument. The predicate can be supplied as an `Expression<Func<T, bool>>`
```
blairconrad: it just seemed wrong to break the line in the middle of the typeparams

thomaslevesque:
Thanks @blairconrad!

PR #1108, opened 2017-06-06
Add an assembly targeting .NET 4.5

### LINKED ISSUE #1008: stop requiring event arguments to extend EventArgs in event raisers for .NET 4.5
Equivalent of #361, for .NET 4.5.

Depends on #991.

blairconrad:
Fixes #1008.

I know we're a little concerned about the lack of testing for the .NET 4.0 assembly, and how having both 4.0 and 4.5 may lead to extra maintenance, but there's nothing preventing us from removing 4.0 support in the future.


tests/FakeItEasy.Tests.Approval/ApiApproval.cs
```diff
@@ -15,28 +15,38 @@ public class ApiApproval
         [Fact]
         [UseReporter(typeof(DiffReporter))]
         [MethodImpl(MethodImplOptions.NoInlining)]
-        public void ApproveApi()
+        public void ApproveApi45()
         {
-            Approvals.Verify(PublicApiGenerator.GetPublicApi(typeof(A).Assembly));
+            ApproveApi("net45");
```
blairconrad: I know, it's a little less efficient than just using the assembly that's in memory. I can revert, if you like.

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -49,6 +49,25 @@
     <PackageReference Include="System.Runtime.Loader" Version="4.0.0" />
   </ItemGroup>
 
+  <!-- .NET 4.5 -->
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net45'">
+    <DefineConstants>$(DefineConstants);FEATURE_BINARY_SERIALIZATION;FEATURE_REFLECTION_GETASSEMBLIES;FEATURE_SELF_INITIALIZED_FAKES</DefineConstants>
```
blairconrad: note the lack of `FEATURE_EVENT_ARGS_MUST_EXTEND_EVENTARGS`

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
thomaslevesque: No, it's fine; it's better to do it the same way for all assemblies, when possible. But the project reference to FakeItEasy should be removed from this project, since it's no longer necessary.

blairconrad: Of course!

blairconrad:
It was a good comment. I've removed the reference and the superfluous "content" entries from the csproj.

thomaslevesque:
Thanks, @blairconrad 👍

PR #1112, opened 2017-06-09
Use GitVersion to generate version numbers

### LINKED ISSUE #1046: Adopt GitVersion
This will remove the need for:

- `VERSION_SUFFIX` to build pre-releases
- bumping the minor version or patch (only major bumps required)

Depends on #955.

blairconrad:
Fixes #1046.

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,0 +1,2 @@
+ignore:
+  commits-before: 2017-01-01T00:00:00
--- a/build.csx
+++ b/build.csx
@@ -8,10 +8,10 @@ using static SimpleTargets;
 var solutionName = "FakeItEasy";
 
 var solution = "./" + solutionName + ".sln";
+var versionInfoFile = "./src/VersionInfo.cs";
 var packagesDirectory = Path.GetFullPath("packages");
 var repoUrl = "https://github.com/FakeItEasy/FakeItEasy";
 var coverityProjectUrl = "https://scan.coverity.com/builds?project=FakeItEasy%2FFakeItEasy";
-var versionAssembly = "./src/FakeItEasy/bin/Release/net40/FakeItEasy.dll";
 var nuspecs = Directory.GetFiles("./src", "*.nuspec");
 var pdbs = new []
 {
@@ -46,6 +46,7 @@ var testSuites = new Dictionary<string, TestSuite[]>
 
 // tool locations
 var vswhere = "./packages/vswhere.1.0.62/tools/vswhere.exe";
+var gitversion = "./packages/GitVersion.CommandLine.4.0.0-beta0012/tools/GitVersion.exe";
 var msBuild = $"{GetVSLocation()}/MSBuild/15.0/Bin/MSBuild.exe";
 var nuget = "./.nuget/NuGet.exe";
 var pdbGit = "./packages/pdbGit.3.0.41/tools/PdbGit.exe";
@@ -71,7 +72,9 @@ targets.Add("logsDirectory", () => Directory.CreateDirectory(logsDirectory));
 
 targets.Add("testsDirectory", () => Directory.CreateDirectory(testsDirectory));
 
-targets.Add("build", DependsOn("clean", "restore"), () => RunMsBuild("Build"));
+targets.Add("build", DependsOn("clean", "restore", "versionInfoFile"), () => RunMsBuild("Build"));
+
+targets.Add("versionInfoFile", () => Cmd(gitversion, $"/updateAssemblyInfo {versionInfoFile} /ensureAssemblyInfo"));
 
 targets.Add(
     "coverity",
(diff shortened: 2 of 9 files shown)


GitVersion.yml
```diff
@@ -0,0 +1,2 @@
+ignore:
+  commits-before: 2017-01-01T00:00:00
```
blairconrad: saves us from the dreaded 3.6.0

build.csx
```diff
@@ -46,8 +46,9 @@ var testSuites = new Dictionary<string, TestSuite[]>
 
 // tool locations
 var vswhere = "./packages/vswhere.1.0.62/tools/vswhere.exe";
+var gitversion = "./packages/GitVersion.CommandLine.4.0.0-beta0012/tools/GitVersion.exe";
 var msBuild = $"{GetVSLocation()}/MSBuild/15.0/Bin/MSBuild.exe";
-var nuget = "./.nuget/NuGet.exe";
+var nuget = @".\.nuget\NuGet.exe";
```
blairconrad: otherwise `cmd` doesn't understand how to run nuget

build.csx
```diff
@@ -98,8 +100,8 @@ targets.Add(
             var form = new MultipartFormDataContent();
             form.Add(new StringContent(coverityToken), @"""token""");
             form.Add(new StringContent(coverityEmail), @"""email""");
-            form.Add(new StringContent(version), @"""version""");
-            form.Add(new StringContent($"Build {version} ({repoCommitId})"), @"""description""");
+            form.Add(new StringContent(repoCommitId), @"""version""");
+            form.Add(new StringContent($"Build ({repoCommitId})"), @"""description""");
```
blairconrad: Rather than jump through hoops figuring out the version, I figured we could just use the commit ID.
Happy to look for an alternative if you don't like it.

build.csx
```diff
@@ -156,10 +158,9 @@ targets.Add(
     DependsOn("build", "outputDirectory", "pdbgit"),
     () =>
     {
-        var version = GetVersion();
         foreach (var nuspec in nuspecs)
         {
-            Cmd(nuget, $"pack {nuspec} -Version {version} -OutputDirectory {outputDirectory} -NoPackageAnalysis");
+            Cmd(gitversion, $@"/exec cmd /execargs ""/c {nuget} pack {nuspec} -Version %GitVersion_NuGetVersionV2% -OutputDirectory {outputDirectory} -NoPackageAnalysis");
```
blairconrad: We could do. I'll add a commit for your consideration.

blairconrad:
> I think I understand most of it, but I'm still a bit worried. It's not completely clear to me how GitVersion determines the version suffix to use.

It does seem a little bit magic.

> Apparently, on develop it uses something tlike `alpha0062`. Where does this `62` come from?

I think these log messages explain it:

```
2017-06-10 21:23:04               INFO [06/10/17 21:23:04:56] Found multiple base versions which will produce the same SemVer (3.4.0), taking oldest source for commit counting (Git tag '3.3.0')
2017-06-10 21:23:04               INFO [06/10/17 21:23:04:58] Base version used: Git tag '3.3.0': 3.3.0 with commit count source 59072ae83010d0eb49af4aca3bbe832bcc01014e (Incremented: 3.4.0)
2017-06-10 21:23:04             INFO [06/10/17 21:23:04:59] End: Calculating base versions (Took: 4,619.06ms)
2017-06-10 21:23:04             INFO [06/10/17 21:23:04:63] 62 commits found between 59072ae83010d0eb49af4aca3bbe832bcc01014e and 7691a81232a6f973c0aa62b166388c70f0fa47f5
```

> On a release branch, it's `beta0001`. Again, how does it know it's the first beta? When will it increment to `beta0002`?

It knows it's the first beta by looking at the release branch. Nothing's significant's happened on it since we released. 

It increments to beta0002 once we add commits after tagging with beta0001 or similar. Check out [Minor Release Branches](https://gitversion.readthedocs.io/en/latest/git-branching-strategies/gitflow-examples/#minor-release-branches). I've tried it out. It works!

> Also, since VersionInfo.cs is not commited, trying to build from VS before the VersionInfo.cs file is generated will build a 0.0.0.0 version. Is this really what we want?

Me, I have no qualms about it whatsoever. I find it preferable to many alternatives, such as a committed file that is then modified by the build process.

> In any case, I think we should wait for @adamralph's feedback since he knows GitVersion better than we do.

Sure. There's no rush. In fact, I'd be as happy with other systems for calculating the version. However, since we've committed to following GitFlow and will be creating the release branches anyhow, and GitVersion  appears to generate reasonable version numbers, I'm willing to give it a go.

blairconrad:
@thomaslevesque, now getting version from GitVersion and calling `nuget pack` with that value. I also reverted to using a version in the Coverity upload.
Added this change as a separate commit, for your reviewing ease. I'd squash before merge.

blairconrad:
@thomaslevesque, it's a little inefficient. We could instead run it once and capture all the output variables into an internal structure by parsing the JSON (which is a small pain, if we don't want to take a dependency), then remember that we've that command and access the variable we need when we want it.

On the other hand, GitVersion caches the version information in `.git/gitversion_cache` after the first run on a particular git commit, so the subsequent invocations are, as you say, pretty fast. In this case, I'd be inclined to lean on the cache and retain the simpler `build.csx` file.

thomaslevesque:
> On the other hand, GitVersion caches the version information in .git/gitversion_cache after the first run on a particular git commit, so the subsequent invocations are, as you say, pretty fast. In this case, I'd be inclined to lean on the cache and retain the simpler build.csx file.

Ah, I didn't know about the cache. Indeed, it's probably best to rely on it then.

blairconrad:
Squashed, since @adamralph would never have been confused by the intermediate step anyhow.

thomaslevesque:
Let's see how it goes. Thanks @blairconrad!

PR #1114, opened 2017-06-11
Fix build when VS2015 is not installed

### LINKED ISSUE #1113: Build script doesn't work on a machine without VS2015
This is because the path to csi.exe is hard-coded as `"%ProgramFiles(x86)%\MSBuild\14.0\Bin\csi.exe"` in build.cmd.

thomaslevesque:
The solution requires VS2017 anyway, so we shouldn't rely on VS2015 in any way.

Fixes #1113

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -5,7 +5,7 @@ pushd %~dp0
 setlocal
 
 rem options
-set NUGET_VERSION=3.5.0
+set NUGET_VERSION=4.1.0
 
 rem determine nuget cache dir
 set NUGET_CACHE_DIR=%LocalAppData%\.nuget\v%NUGET_VERSION%
@@ -19,16 +19,14 @@ if not exist %NUGET_CACHE_DIR%\NuGet.exe (
 )
 
 rem copy nuget locally
-if not exist .nuget\NuGet.exe (
-  if not exist .nuget md .nuget
-  copy %NUGET_CACHE_DIR%\NuGet.exe .nuget\NuGet.exe > nul
-)
+if not exist .nuget md .nuget
+copy /Y %NUGET_CACHE_DIR%\NuGet.exe .nuget\NuGet.exe > nul
 
 rem restore packages for build script
 .nuget\NuGet.exe restore .\packages.config -PackagesDirectory .\packages -Verbosity quiet
 
 rem run build script
-"%ProgramFiles(x86)%\MSBuild\14.0\Bin\csi.exe" .\build.csx %*
+".\packages\Microsoft.Net.Compilers.2.2.0\tools\csi.exe" .\build.csx %*
 
 rem return to original working directory
 popd
--- a/FakeItEasy.sln
+++ b/FakeItEasy.sln
@@ -1,7 +1,7 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.26403.7
-MinimumVisualStudioVersion = 14.0.0.0
+MinimumVisualStudioVersion = 15.0.0.0
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{A6A7B553-5359-4FB9-B19C-8A23004F49DB}"
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
--- a/build.csx
+++ b/build.csx
@@ -127,7 +127,7 @@ targets.Add(
     "restore",
     () =>
     {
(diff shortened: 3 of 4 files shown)


build.cmd
```diff
@@ -19,16 +19,18 @@ if not exist %NUGET_CACHE_DIR%\NuGet.exe (
...
-  copy %NUGET_CACHE_DIR%\NuGet.exe .nuget\NuGet.exe > nul
-)
+if not exist .nuget md .nuget
+copy %NUGET_CACHE_DIR%\NuGet.exe .nuget\NuGet.exe > nul
 
 rem restore packages for build script
 .nuget\NuGet.exe restore .\packages.config -PackagesDirectory .\packages -Verbosity quiet
 
+rem find Visual Studio and MSBuild installation paths
+FOR /F "delims=" %%i IN ('.\packages\vswhere.1.0.62\tools\vswhere.exe -property installationPath') DO set VS_PATH=%%i
```
blairconrad: If we have more than one entry, we'll keep overwriting `VS_PATH`. How do we know we're getting the right one?

Perhaps we should narrow down the version and ask for Roslyn instead of MBuild? This is closer to what we do in the build.csx.

```
vswhere.exe -nologo -latest -property installationPath -requires Microsoft.VisualStudio.Component.Roslyn.Compiler -version [15,16)
```

Even if you keep the loop, at least we'll make sure we get a VS15.

build.cmd
```diff
@@ -19,16 +19,18 @@ if not exist %NUGET_CACHE_DIR%\NuGet.exe (
 )
 
 rem copy nuget locally
-if not exist .nuget\NuGet.exe (
-  if not exist .nuget md .nuget
-  copy %NUGET_CACHE_DIR%\NuGet.exe .nuget\NuGet.exe > nul
-)
+if not exist .nuget md .nuget
+copy %NUGET_CACHE_DIR%\NuGet.exe .nuget\NuGet.exe > nul
```
thomaslevesque: Ah, yes, I wanted to add a comment about this but forgot.

> why not `/Y` instead of `> nul`?

Well, you're the one who [wrote it](https://github.com/thomaslevesque/FakeItEasy/blob/cb2db7b0700508e542fd09ad768d55345b7f9681/build.cmd#L24), so I thought you'd know :wink:
I just left it as it was. I can change it if it makes sense.

> (my preference) why do we bother copying to the local directory at all? We're always making sure there's something in the cache directory. We could just run from there.

I wondered about this too. I'm not sure I see the point of copying either. The only reason I can see is that this way, the csx script doesn't need to know about the cache location. If we want to run directly from the cache dir, we need to somehow tell the script where to find NuGet (maybe via an environment variable). This will make it harder to run the csx manually, without using the cmd, which is probably not a big issue, since we're unlikely to want to do that.

build.cmd
thomaslevesque: Good point. I'll change it.

thomaslevesque: Now that I think about it, there might be a better option. Csi.exe is included in the [Microsoft.Net.Compilers package](https://www.nuget.org/packages/Microsoft.Net.Compilers/), so we could use this instead of looking for csi.exe in the VS2017 installed. I just tried it, it works fine.

build.cmd
thomaslevesque: In fact, we still need `> null` to avoid the "1 file(s) copied." output.

thomaslevesque:
I added an extra commit that is somewhat unrelated, so I can move it to a separate PR if you prefer. I realized that we where restoring the packages for the solution to a local packages directory which was not even used during the build.

build.cmd
blairconrad: >  Csi.exe is included in the Microsoft.Net.Compilers package, so we could use this instead of looking for csi.exe in the VS2017 installed.

It _is_ much cleaner, but I'm a little torn. We're pulling down 22.5 MB that we likely don't need. (Only once, I suppose.)

thomaslevesque: True. Well, technically, we're only pulling 6.58 MB, the rest is the result of extracting the nupkg...
On the other hand, with my extra commit regarding the packages directory, we're saving quite a lot of disk space ;)

blairconrad: I was thinking more of network than storage, so good point about the 6.58.

It still feels a little odd to add the package dependency when we _know_ there's a Roslyn just sitting on the disk. Sort of like running GitVersion twice, when we could've just cached variables from the first run. :wink:

Okay. If you want to leave this and the nuget copy as is, I can merge after a squash. Or I'm happy to wait for any change.

blairconrad: > However, this is the same as with GitVersion

Yes. That's the joke I was trying (not well) to make. ☹️

blairconrad:
Thanks, @thomaslevesque!

PR #1115, opened 2017-06-12
Better error message when faking inaccessible methods

### LINKED ISSUE #341: improve error message when trying to configure inaccessible methods
Spun off from #338. This doesn't really apply until that issue is built into a release, but I want to create this one before I forget.

When trying to configure a method (via `A.CallTo`, for example) that is not accessible to DynamicProxy, the user will be confronted with a message very much like the following:

> Methods not accessible to DynamicProxyGenAssembly2 can not be intercepted.

To contrast, when attempting to fake a type that is not accessible, the user will get a much more descriptive error message, such as

> Can not create proxy for type IInternal because it is not accessible. Make it public, or internal and mark your assembly with [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")] attribute, because assembly FakeItEasyQuestions is not strong-named.

With variants based on whether the assembly containing the inaccessible type is strong-named or not, and whether _it_ references DynamicProxy.

It would be nice to be so helpful when trying to configure inaccessible methods, but the infrastructure for building the better messages lives in Castle.Core and is very _type_ oriented. The best bet would be to ask the Castle people if they'd let us extend Castle.Core to have access to the message-generating code.

blairconrad:
Fixes #341.

### DIFF
--- a/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
+++ b/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
@@ -55,12 +55,10 @@
                 return "Non-virtual members can not be intercepted. Only interface members and virtual, overriding, and abstract members can be intercepted.";
             }
 
-            if (!Castle.DynamicProxy.Internal.InternalsUtil.IsAccessible(method))
+            string message;
+            if (!Castle.DynamicProxy.ProxyUtil.IsAccessible(method, out message))
             {
-                return string.Concat(
-                    "Methods not accessible to ",
-                    Castle.DynamicProxy.ModuleScope.DEFAULT_ASSEMBLY_NAME,
-                    " can not be intercepted.");
+                return message;
             }
 
             return null;
--- a/src/FakeItEasy/FakeItEasy.csproj
+++ b/src/FakeItEasy/FakeItEasy.csproj
@@ -43,7 +43,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.6'">
-    <PackageReference Include="Castle.Core" Version="4.0.0" />
+    <PackageReference Include="Castle.Core" Version="4.1.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="1.0.0" />
     <PackageReference Include="System.Collections.Concurrent" Version="4.3.0" />
     <PackageReference Include="System.Runtime.Loader" Version="4.0.0" />
@@ -59,7 +59,7 @@
     <PackageReference Include="ILMerge" Version="2.14.1208">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Castle.Core" Version="4.0.0">
+    <PackageReference Include="Castle.Core" Version="4.1.0">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
   </ItemGroup>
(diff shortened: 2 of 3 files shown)


blairconrad:
I put the "DO NOT MERGE" in the title because although we're just getting a new minor release of Castle.Core (4.1.0, from 4.0.0), the [release notes indicate](https://github.com/castleproject/Core/blob/v4.1.0/CHANGELOG.md#410-2017-06-11) that there's a breaking change: `AllowPartiallyTrustedCallersAttribute` was removed.
However, as far as I can read, this attribute [was not applied in 4.0.0 for .NET 3.5., 4.0, 4.5, or even .NET Core](https://github.com/castleproject/Core/pull/268/files#diff-04c6e90faac2675aa89e2176d2eec7d8), so I'm not sure it's a breaking change at all. I think we're safe to pick up Castle.Core 4.1.0.
@jonorossi and/or @fir3pho3nixx, can you check my reasoning?

fir3pho3nixx:
@blairconrad - If you download the source code for Castle Core v4.0.0.0 you can find the `FEATURE_APTCA` build symbol wrapping the `AllowPartiallyTrustedCallers` attribute in the CommonAssemblyInfo.cs in the `buildscripts` folder. 

```csharp 
#if FEATURE_SECURITY_PERMISSIONS && FEATURE_APTCA
[assembly: AllowPartiallyTrustedCallers]
#endif
```
It was never defined in the `Settings.proj` in the root along with all the other build symbols so according to me it was never compiled into the output.

I also reflected into the assemblies to check and could not find any trace of it, so I think you are OK to merge this.

blairconrad:
@fir3pho3nixx thanks for confirming!

thomaslevesque:
Thanks @blairconrad  👍

fir3pho3nixx:
@blairconrad - Mainly tracking purposes, please see https://github.com/castleproject/Windsor/issues/248 to get some background.

blairconrad:
Oh, thanks, @fir3pho3nixx. I think I follow. I mean, I still don't understand anything about partial trust, but that's another problem!

fir3pho3nixx:
@blairconrad - No worries, in simple terms CAS in net20 because it was littered with `side effects` which for example could be cascaded down from machine configurations and various other sources(in a corporate environment you can understand how this could be a problem if admins are moving the cheese underneath you between deploys). In asp.net 4.0 they implemented appdomain isolation and dis-allowed partially trusted non-web hosting environments calling into web environment use cases whilst improving the framework to make this more relevant by creating security transparent code(in other words they are trying to make security easier for the consumer). You could also have consumers doing stupid shit like trying to launch winforms controls(more trusted, gains winhandles and has access to ex-win32) in MVC controllers(which are less trusted should not be native at all).  You have to remember silverlight and sandboxing was also thing during this time, so that had influence also. This is basically microsoft's fancy way of saying we messed up and we want to make things better for our users.

jonorossi:
> the release notes indicate that there's a breaking change: AllowPartiallyTrustedCallersAttribute was removed.
> However, as far as I can read, this attribute was not applied in 4.0.0 for .NET 3.5., 4.0, 4.5, or even .NET Core, so I'm not sure it's a breaking change at all. I think we're safe to pick up Castle.Core 4.1.0.

> @fir3pho3nixx thanks for confirming! (Although I wonder a little why the change was marked as breaking, but that's between you and @jonorossi!)

It was marked as breaking because of Castle's history, there was a time when everyone built from subversion "trunk" themselves and we never made releases because trunk was always stable and things were great. I wanted to capture that detail in the release notes for anyone still doing that (although no longer recommended) or upgrading, however we obviously didn't go v5.0 because of it. Apologies for the confusion, I sometimes forget that users reading the release notes don't have all the context 😉 .

PR #1119, opened 2017-06-13
Don't expose Not after Not in argument constraint

### LINKED ISSUE #1063: Don't expose Not after Not in argument constraint
With the current API it's possible to write argument constraint like this:

    A<int>.That.Not.Not.Not.Not.Not.Not.IsEqualTo(42)

Of course, it's a stupid thing to do, and presumably nobody would do this, but it's annoying that `Not` is even valid and shows up in Intellisense after `Not`. We could easily prevent it.

Suggested solution: introduce a `INegatableArgumentContraintManager<T>` interface that inherits from `IArgumentConstraintManager<T>` and exposes the `Not` property, and return it only from `A<T>.That`.

This is a breaking change, so it will have to wait until the next major release.

thomaslevesque:
Fixes #1063

### DIFF
--- a/src/FakeItEasy/IArgumentConstraintManager.cs
+++ b/src/FakeItEasy/IArgumentConstraintManager.cs
@@ -1,21 +1,13 @@
-namespace FakeItEasy
+﻿namespace FakeItEasy
 {
     using System;
-    using System.Diagnostics.CodeAnalysis;
 
     /// <summary>
     /// Manages attaching of argument constraints.
     /// </summary>
     /// <typeparam name="T">The type of argument to constrain.</typeparam>
-    [SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1623:PropertySummaryDocumentationMustMatchAccessors", Justification = "Fluent API.")]
     public interface IArgumentConstraintManager<T> : IHideObjectMembers
     {
-        /// <summary>
-        /// Inverts the logic of the matches method.
-        /// </summary>
-        [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "Not", Justification = "Part of the fluent syntax.")]
-        IArgumentConstraintManager<T> Not { get; }
-
         /// <summary>
         /// Constrains the argument with a predicate.
         /// </summary>
--- a/src/FakeItEasy/RootModule.cs
+++ b/src/FakeItEasy/RootModule.cs
@@ -188,7 +188,7 @@ namespace FakeItEasy
         private class ArgumentConstraintManagerFactory
             : IArgumentConstraintManagerFactory
         {
-            public IArgumentConstraintManager<T> Create<T>()
+            public INegatableArgumentConstraintManager<T> Create<T>()
             {
                 return new DefaultArgumentConstraintManager<T>(ArgumentConstraintTrap.ReportTrappedConstraint);
             }
--- a/src/FakeItEasy/A.of.T.cs
+++ b/src/FakeItEasy/A.of.T.cs
@@ -15,7 +15,7 @@ namespace FakeItEasy
         /// Gets an argument constraint object that will be used to constrain a method call argument.
         /// </summary>
         [SuppressMessage("Microsoft.Design", "CA1000:DoNotDeclareStaticMembersOnGenericTypes", Justification = "This is a special case where the type parameter acts as an entry point into the fluent api.")]
-        public static IArgumentConstraintManager<T> That =>
+        public static INegatableArgumentConstraintManager<T> That =>
             ServiceLocator.Current.Resolve<IArgumentConstraintManagerFactory>().Create<T>();
 
         /// <summary>
--- a/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
+++ b/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
@@ -1,4 +1,4 @@
-namespace FakeItEasy
+﻿namespace FakeItEasy
 {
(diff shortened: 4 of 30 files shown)


src/FakeItEasy/IArgumentConstraintManager.cs
```diff
@@ -7,21 +7,28 @@ namespace FakeItEasy
...
+    /// </summary>
+    /// <typeparam name="T">The type of argument to constrain.</typeparam>
+    public interface INegatableArgumentConstraintManager<T> : IArgumentConstraintManager<T>
+    {
+        /// <summary>
+        /// Inverts the logic of the matches method.
+        /// </summary>
+        [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "Not", Justification = "Part of the fluent syntax.")]
+        [SuppressMessage("StyleCop.CSharp.DocumentationRules", "SA1623:PropertySummaryDocumentationMustMatchAccessors", Justification = "Fluent API.")]
+        IArgumentConstraintManager<T> Not { get; }
```
blairconrad: Should we move `IsNotNull` and `IsNull` here as well? Otherwise I can constrain like this:

```c#
A<IList>.That.Not.IsNull(); // which isn't the end of the world I suppose, and has existed for a while, but plain old IsNotNull is preferable
A<IList>.That.Not.IsNotNull(); // yuck
```

src/FakeItEasy/IArgumentConstraintManager.cs
```diff
@@ -7,21 +7,28 @@ namespace FakeItEasy
...
     }
+
+    /// <summary>
+    /// Enables negation of argument constraints.
+    /// </summary>
+    /// <typeparam name="T">The type of argument to constrain.</typeparam>
+    public interface INegatableArgumentConstraintManager<T> : IArgumentConstraintManager<T>
+    {
+        /// <summary>
+        /// Inverts the logic of the matches method.
```
blairconrad: Sounds like there's a method called `matches`.  Which was maybe supposed to be `Matches`, but even that doesn't really apply.
Perhaps "of the subsequent constraint"?

src/FakeItEasy/IArgumentConstraintManager.cs
```diff
@@ -7,21 +7,28 @@ namespace FakeItEasy
...
         /// <param name="descriptionWriter">An action that will be write a description of the constraint.</param>
         /// <returns>A dummy argument value.</returns>
         T Matches(Func<T, bool> predicate, Action<IOutputWriter> descriptionWriter);
     }
+
+    /// <summary>
+    /// Enables negation of argument constraints.
+    /// </summary>
+    /// <typeparam name="T">The type of argument to constrain.</typeparam>
+    public interface INegatableArgumentConstraintManager<T> : IArgumentConstraintManager<T>
```
thomaslevesque: I kept them together because they were so closely related, but sure, I can move it to a separate file.

src/FakeItEasy/IArgumentConstraintManager.cs
thomaslevesque: Ah, yes. I moved the property from the base interface and didn't update the comment. I like your suggestion.

src/FakeItEasy/RootModule.cs
```diff
@@ -188,7 +188,7 @@ public bool TryCreateFakeObject(DummyCreationSession session, Type typeOfFake, D
         private class ArgumentConstraintManagerFactory
             : IArgumentConstraintManagerFactory
```
thomaslevesque: > One option would be to rename to include `Negative`

You mean `Negatable`, right? `NegatableArgumentConstraintManagerFactory` is quite a mouthful, and I don't think the fact that it creates a *negatable* argument constraint manager is relevant. I'd rather keep it as is.

src/FakeItEasy/IArgumentConstraintManager.cs
thomaslevesque: This might make sense. Let me look into it.

src/FakeItEasy/RootModule.cs
blairconrad: Of course I meant `Negatable`. I shouldn't review code and what confusing TV at the same time. :blush:

src/FakeItEasy/IArgumentConstraintManager.cs
thomaslevesque: We have the same problem with `IsNotCanceled`... but since it only applies to `CancellationToken`s, we can't put it in the interface.

thomaslevesque: Moving `IsNotNull` to `INegatableArgumentConstraintManager<T>` makes sense, I guess, but moving `IsNull` as well would unnecessarily break existing code using `Not.IsNull()`. It's also a bit weird to have some of the constraints on the interface, and some of them as extension methods. I wonder if it's worth the trouble at all.

thomaslevesque: A better option would probably be to change the extension methods to extend `INegatableArgumentConstraintManager<T>` instead of `IArgumentConstraintManager<T>`

blairconrad: An elegant approach.

thomaslevesque:
Squashed

thomaslevesque:
Oh, wait. Wrong message in the second commit

thomaslevesque:
Fixed

blairconrad:
Thanks, @thomaslevesque!

PR #1125, opened 2017-06-15
Deprecate initializefixture

### LINKED ISSUE #1102: Deprecate automatic fake injection into systems under test (InitializeFixture)
In preparation of #992

blairconrad:
Fixes #1102


src/FakeItEasy/FakeFacade.cs
```diff
@@ -50,6 +51,7 @@ public virtual void ClearRecordedCalls(object fakedObject)
             manager.ClearRecordedCalls();
         }
 
+        [Obsolete("Test fixture initialization will be removed in version 5.0.0.")]
```
thomaslevesque: This method isn't publicly visible, so I don't think it needs the `Obsolete` attribute.

tests/FakeItEasy.Tests/FakeFacadeTests.cs
```diff
@@ -106,7 +106,9 @@ public void Initialize_should_be_guarded()
             // Act
 
             // Assert
-             Expression<System.Action> call = () => this.facade.InitializeFixture(new object());
+#pragma warning disable CS0618 // Type or member is obsolete
+            Expression<System.Action> call = () => this.facade.InitializeFixture(new object());
+#pragma warning restore CS0618 // Type or member is obsolete
```
thomaslevesque: If `FakeFacade.InitializeFixture` isn't marked obsolete, this warning suppression isn't necessary.

src/FakeItEasy/FakeFacade.cs
blairconrad: I was thinking to keep us from starting to use it accidentally. But that _does_ seem unlikely. Removing the attribute.

blairconrad:
Thanks. Squashed. (To this day, still typing `git rebase -i master` and then replacing with `develop`.)

PR #1135, opened 2017-06-16
Remove ITaggable

### LINKED ISSUE #1012: Get rid of ITaggable
This interface is intended for internal use only, but due to an issue in Castle.Core, it has to be exposed publicly. As @blairconrad suggested in the chat, we could probably get rid of it and use a `ConditionalWeakTable<object, FakeManager>` instead.

thomaslevesque:
Fixes #1012


src/FakeItEasy/Core/FakeManagerProvider.cs
```diff
@@ -81,13 +81,19 @@ public void EnsureInitialized(object proxy)
...
+                    this.fakeManagerAccessor.SetFakeManager(proxy, this.initializedFakeManager);
 
                     this.ApplyInitialConfiguration(proxy);
                 }
             }
         }
 
+        public void EnsureManagerIsRegistered()
+        {
+            var accessor = ServiceLocator.Current.Resolve<IFakeManagerAccessor>();
```
blairconrad: How 'bout we just make `this.fakeManagerAccessor` serializable? Everything works except for `FakeManagerProviderTests.Should_be_able_to_fetch_a_serialized_and_deserialized_initialized_provider`, but I think that's because of all the fakes _it's_ made of. And I think the test is redundant.

src/FakeItEasy/Core/IFakeCallProcessorProvider.cs
```diff
@@ -26,9 +26,11 @@ internal interface IFakeCallProcessorProvider
         IFakeCallProcessor Fetch(object proxy);
 
         /// <summary>
-        /// Ensures that the <see cref="IFakeCallProcessor"/> is initialized (and can be used via the proxy's tag later on).
+        /// Ensures that the <see cref="IFakeCallProcessor"/> is initialized (and can be used to retrieve the proxy's fake call processor later on).
         /// </summary>
         /// <param name="proxy">The corresponding proxy object of the new <see cref="IFakeCallProcessor"/>.</param>
         void EnsureInitialized(object proxy);
+
+        void EnsureManagerIsRegistered();
```
thomaslevesque: We totally should. I'll add some doc comments.

src/FakeItEasy/Core/FakeManagerProvider.cs
thomaslevesque: I think I tried, but it didn't work. But I changed other things since then. I'll see what I can do. It would definitely be cleaner.

thomaslevesque: OK, I agree that this test probably doesn't bring any value. Let's drop it, and serialize the `fakeManagerAccessor`.

thomaslevesque:
All fixed. I'll probably squash everything before merging, since the individual commits aren't really meaningful on their own.

blairconrad:
That sounds great. Thanks!

thomaslevesque:
Rebased & squashed

blairconrad:
Thanks, @thomaslevesque!

PR #1136, opened 2017-06-17
Generate nuspec dependencies from project file

### LINKED ISSUE #1133: Implement system to ensure nupkg references are consistent with assembly references
Inspired by #1126, which noticed that we'd released FakeItEasy 3.4.0,  whose assembly relied on Castle.Core 4.1.0, as a nupkg with a reference (for .NET Standard 1.6) to Castle.Core 4.0.0.

@holytshirt suggested using `dotnet pack` on the project instead of `nuget pack`, which is probably the ideal situation if it works, but @thomaslevesque thinks that this will create a separate symbols package instead of including the symbols in the nupkg. I'm also unsure that we'll get the right package versions, but I've not explored.

Another question I have about "dotnet pack" is how to create the analyzer package, including both C# and VB analyzers.

If it turns out that using "dotnet pack" isn't feasible, we could perhaps [expand the solution I used for SelfIntitializingFakes](https://github.com/blairconrad/SelfInitializingFakes/pull/2), where we read the referenced version from the csproj and use it to replace a property defined in the nuspec.
We'd need to be a little more sophisticated in FakeItEasy, though, since we have more dependencies and differing numbers of dependencies between frameworks (although so far we don't have different versions of dependencies for different frameworks).

thomaslevesque:
Fixes #1133 

- FakeItEasy.nuspec is renamed to FakeItEasy.nuspec.template, with no dependencies specified
- The build script generates FakeItEasy.nuspec from the template, using FakeItEasy.csproj to fill the dependencies
- The dependencies are located using the `Condition="'$(TargetFramework)' == '{tfm}'"` attribute. _This is a bit crude, and will break if we use more complex conditions_
- Build-only dependencies are identified by the presence of the `<PrivateAssets>` element. _This is also an oversimplification, because `<PrivateAssets>` can have values other than `all`, but it works fine with the current bits._

### DIFF
--- a/build.csx
+++ b/build.csx
@@ -1,8 +1,10 @@
 #load "packages/simple-targets-csx.5.2.0/simple-targets.csx"
 
 #r "System.Net.Http"
+#r "System.Xml.Linq"
 
 using System.Net.Http;
+using System.Xml.Linq;
 using static SimpleTargets;
 
 var solutionName = "FakeItEasy";
@@ -12,7 +14,15 @@ var versionInfoFile = "./src/VersionInfo.cs";
 var packagesDirectory = Path.GetFullPath("packages");
 var repoUrl = "https://github.com/FakeItEasy/FakeItEasy";
 var coverityProjectUrl = "https://scan.coverity.com/builds?project=FakeItEasy%2FFakeItEasy";
-var nuspecs = Directory.GetFiles("./src", "*.nuspec");
+var fakeItEasyProjectPath = "./src/FakeItEasy/FakeItEasy.csproj";
+var fakeItEasyNuspecTemplatePath = "./src/FakeItEasy.nuspec.template";
+var fakeItEasyNuspecPath = "./src/FakeItEasy.nuspec";
+var analyzerNuspecPath = "./src/FakeItEasy.Analyzer.nuspec";
+var nuspecs = new[] {
+    fakeItEasyNuspecPath,
+    analyzerNuspecPath,
+};
+
 var pdbs = new []
 {
     "src/FakeItEasy/bin/Release/net40/FakeItEasy.pdb",
@@ -155,8 +165,12 @@ targets.Add(
     () => RunTests("approve"));
 
 targets.Add(
+    "nuspec",
+    () => GenerateNuspec(fakeItEasyProjectPath, fakeItEasyNuspecTemplatePath, fakeItEasyNuspecPath));
+
+targets.Add(
     "pack",
-    DependsOn("build", "outputDirectory", "pdbgit"),
+    DependsOn("build", "outputDirectory", "pdbgit", "nuspec"),
     () =>
     {
         var version = ReadCmdOutput(".", gitversion, "/showvariable NuGetVersionV2");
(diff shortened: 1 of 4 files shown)


thomaslevesque:
My initial approach was to add a custom attribute on the `<PackageReference>` elements, like this:

```xml
<PackageReference Include="Castle.Core" Version="4.1.0" fie:NuGetDependencyFor="netstandard1.6" />
```

This would be more robust, but unfortunately MSBuild is pretty strict and rejects unrecognized elements and attributes... Alternatively, we could use the `Label` attribute, but that's not really what it's designed for.

thomaslevesque:
btw, the old nuspec was using TFMs like `.NETFramework4.0` or `.NETStandard1.6`. The new one uses `net40` and `netstandard1.6`. The result is the same, and it makes things easier because it matches the target frameworks in the csproj.

build.csx
```diff
@@ -12,7 +14,14 @@ var versionInfoFile = "./src/VersionInfo.cs";
 var packagesDirectory = Path.GetFullPath("packages");
 var repoUrl = "https://github.com/FakeItEasy/FakeItEasy";
 var coverityProjectUrl = "https://scan.coverity.com/builds?project=FakeItEasy%2FFakeItEasy";
-var nuspecs = Directory.GetFiles("./src", "*.nuspec");
+var nuspecs = new[] {
+    "./src/FakeItEasy.nuspec",
+    "./src/FakeItEasy.Analyzer.nuspec",
+};
```
thomaslevesque: This is because `FakeItEasy.nuspec` doesn't exist at this point. We could also evaluate it lazily, but if FakeItEasy.nuspec doesn't exist, it will just be ignored, so I thought it would be better to specify the files explicitly.

src/FakeItEasy.nuspec.template
```diff
@@ -13,16 +13,7 @@
...
-      <group targetFramework=".NETFramework4.0" />
-      <group targetFramework=".NETFramework4.5" />
-      <group targetFramework=".NETStandard1.6">
-        <dependency id="Microsoft.Extensions.DependencyModel" version="1.0.0" />
-        <dependency id="System.Collections.Concurrent" version="4.0.12" />
-        <dependency id="System.Runtime.Loader" version="4.0.0" />
-        <dependency id="Castle.Core" version="4.1.0" />
-      </group>
-    </dependencies>
+    <dependencies />
```
blairconrad: Pretty.

Out of curiosity, did you notice that System.Collections.Concurrent got pumped from 4.0.12 to 4.3.0?
You made the change in be8e8cfd94e62bba0ccdc11975c287977889397c, and we never updated the nuspec!
So I guess we're getting a bonus fix.

build.csx
```diff
@@ -12,7 +14,14 @@ var versionInfoFile = "./src/VersionInfo.cs";
...
 var repoUrl = "https://github.com/FakeItEasy/FakeItEasy";
 var coverityProjectUrl = "https://scan.coverity.com/builds?project=FakeItEasy%2FFakeItEasy";
-var nuspecs = Directory.GetFiles("./src", "*.nuspec");
+var nuspecs = new[] {
+    "./src/FakeItEasy.nuspec",
+    "./src/FakeItEasy.Analyzer.nuspec",
+};
+var fakeItEasyProjectPath = "./src/FakeItEasy/FakeItEasy.csproj";
+var fakeItEasyNuspecTemplatePath = "./src/FakeItEasy.nuspec.template";
+var fakeItEasyNuspecPath = "./src/FakeItEasy.nuspec";
```
thomaslevesque: Indeed

src/FakeItEasy.nuspec.template
thomaslevesque: No, I hadn't noticed. Strange that it didn't break like it did for Castle.Core.

build.csx
thomaslevesque: It was a very small change, so I just amended the commit.

thomaslevesque:
@blairconrad you're not worried about this solution being brittle in the face of future changes to the csproj?

src/FakeItEasy.nuspec.template
blairconrad: Probably everyone just had System.Collections.Concurrent 4.3.0 anyhow.

blairconrad:
@thomaslevesque, I'm hoping that https://github.com/NuGet/Home/issues/5440 is fixed before the .csproj changes! 
I'm content if you are.

thomaslevesque:
> I'm content if you are.

I am!

blairconrad:
Okay. I'll rename and merge before you rebase anything else. :wink:

PR #1137, opened 2017-06-17
Fix: assigning ref param overwrites recorded argument

### LINKED ISSUE #1124: Setting out or ref argument values masks incoming argument values used to verify calls

thomaslevesque:
Fixes #1124


src/FakeItEasy/Creation/CastleDynamicProxy/CastleInvocationCallAdapter.cs
```diff
@@ -26,19 +28,12 @@ internal class CastleInvocationCallAdapter
         public CastleInvocationCallAdapter(IInvocation invocation)
         {
             this.invocation = invocation;
+            this.originalArguments = invocation.Arguments.ToArray();
             this.Method = invocation.Method;
-
             this.Arguments = new ArgumentCollection(invocation.Arguments, this.Method);
-
-            SequenceNumberManager.RecordSequenceNumber(this);
```
thomaslevesque: I moved this to `FakeManager.RecordCall`, because it's the completed call that must be recorded, and it's now a separate object.

PR #1138, opened 2017-06-17
Fix: assigning ref param overwrites recorded argument

### LINKED ISSUE #1124: Setting out or ref argument values masks incoming argument values used to verify calls

thomaslevesque:
Fixes #1124

### DIFF
--- a/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
+++ b/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
@@ -1,0 +1,25 @@
+﻿namespace FakeItEasy.Core
+{
+    using System.Linq;
+    using System.Reflection;
+    using FakeItEasy.Configuration;
+
+    internal class CompletedFakeObjectCall : ICompletedFakeObjectCall
+    {
+        public CompletedFakeObjectCall(object fakedObject, MethodInfo method, object[] arguments, object returnValue)
+        {
+            this.FakedObject = fakedObject;
+            this.Method = method;
+            this.Arguments = new ArgumentCollection(arguments, method);
+            this.ReturnValue = returnValue;
+        }
+
+        public object ReturnValue { get; }
+
+        public MethodInfo Method { get; }
+
+        public ArgumentCollection Arguments { get; }
+
+        public object FakedObject { get; }
+    }
+}
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -1,4 +1,4 @@
-namespace FakeItEasy.Core
+﻿namespace FakeItEasy.Core
 {
     using System;
     using System.Collections.Concurrent;
@@ -170,6 +170,7 @@ namespace FakeItEasy.Core
         /// <param name="call">The call to remember.</param>
         internal void RecordCall(ICompletedFakeObjectCall call)
         {
+            SequenceNumberManager.RecordSequenceNumber(call);
             this.recordedCalls.Enqueue(call);
         }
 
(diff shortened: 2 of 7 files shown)


src/FakeItEasy/SelfInitializedFakes/SelfInitializationRule.cs
```diff
@@ -55,7 +56,12 @@ public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
             if (this.recorder.IsRecording)
             {
                 this.wrappedRule.Apply(fakeObjectCall);
-                this.recorder.RecordCall(fakeObjectCall.AsReadOnly());
+                var completedFakeObjectCall = new CompletedFakeObjectCall(
+                    fakeObjectCall.FakedObject,
+                    fakeObjectCall.Method,
+                    fakeObjectCall.Arguments.ToArray(),
+                    fakeObjectCall.AsReadOnly().ReturnValue);
+                this.recorder.RecordCall(completedFakeObjectCall);
```
thomaslevesque: The recorder expects a call that contains the post-call arguments (in order to be able to replay it with the appropriate out and ref arguments), but `AsReadOnly()` returns a call that contains the incoming arguments. So I'm explicitly creating a call that has the expected arguments. It's a little ugly, but it's not for long, since SelfInitializedFakes are being removed in 4.0.0.

btw, @blairconrad, your SelfInitializedFakes library will probably be affected by this fix as well. We could extend `ICompletedFakeObjectCall` to include both the incoming and outgoing arguments, if necessary.

src/FakeItEasy/Core/CompletedFakeObjectCall.cs
```diff
@@ -0,0 +1,25 @@
...
+            this.Method = method;
+            this.Arguments = new ArgumentCollection(arguments, method);
+            this.ReturnValue = returnValue;
+        }
+
+        public object ReturnValue { get; }
+
+        public MethodInfo Method { get; }
+
+        public ArgumentCollection Arguments { get; }
```
blairconrad: It's a little troubling that `Arguments` means two different things, depending on the type of the underlying object implementing `IFakeObjectCall`. I'm not sure what to do about it, though.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -170,6 +170,7 @@ internal IEnumerable<ICompletedFakeObjectCall> GetRecordedCalls()
         /// <param name="call">The call to remember.</param>
         internal void RecordCall(ICompletedFakeObjectCall call)
         {
+            SequenceNumberManager.RecordSequenceNumber(call);
```
blairconrad: This is an excellent change.

tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
```diff
@@ -157,5 +157,31 @@ public static void AssignOutAndRefParameters(IHaveARef subject, string outValue)
...
+            "And the configured method is called"
+                .x(() => subject.MightReturnAKnownValue(ref refValue));
+
+            "The the assertion that a call with the expected value has happened succeeds"
+                .x(() =>
+                {
+                    string expectedValue = Condition;
+                    A.CallTo(() => subject.MightReturnAKnownValue(ref expectedValue)).MustHaveHappened();
+                });
+        }
```
blairconrad: Since the delegate fakes needed an additional change, I think we should have a spec for them too.

src/FakeItEasy/Core/CompletedFakeObjectCall.cs
thomaslevesque: True. Maybe we should change the API in 4.0.0 to expose both.

tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
thomaslevesque: Good point. I'll add it.

thomaslevesque: Done

blairconrad:
:+1:

blairconrad:
Thanks, @thomaslevesque!

PR #1141, opened 2017-06-18
Hotfix 3.4.2

### LINKED ISSUE #1124: Setting out or ref argument values masks incoming argument values used to verify calls

thomaslevesque:
Fixes #1124

### DIFF
--- a/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
+++ b/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
@@ -1,0 +1,25 @@
+﻿namespace FakeItEasy.Core
+{
+    using System.Linq;
+    using System.Reflection;
+    using FakeItEasy.Configuration;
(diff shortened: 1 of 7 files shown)


blairconrad:
Hey, @thomaslevesque. Looks great. Thanks.
I'm delaying only because I didn't know if we're ready to release. I am not sure I have the time tomorrow morning.
If you have cycles, I say merge and go for it. Otherwise, we'll see what time I wake up.

thomaslevesque:
> I'm delaying only because I didn't know if we're ready to release

Well, it's a hotfix, and the only fix has been merged. I don't think we need anything else before releasing.

> If you have cycles, I say merge and go for it. Otherwise, we'll see what time I wake up.

👍 
I'll do it today if I can.

blairconrad:
Thanks, @thomaslevesque!

PR #1142, opened 2017-06-18
Merge hotfix 3.4.2 into develop

thomaslevesque:

PR #1143, opened 2017-06-19
Merge hotfix 3.4.2 into develop

thomaslevesque:
Supersedes #1142

### DIFF
--- a/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
+++ b/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
@@ -1,0 +1,25 @@
+﻿namespace FakeItEasy.Core
+{
+    using System.Linq;
+    using System.Reflection;
+    using FakeItEasy.Configuration;
+
+    internal class CompletedFakeObjectCall : ICompletedFakeObjectCall
+    {
+        public CompletedFakeObjectCall(object fakedObject, MethodInfo method, object[] arguments, object returnValue)
+        {
+            this.FakedObject = fakedObject;
+            this.Method = method;
+            this.Arguments = new ArgumentCollection(arguments, method);
+            this.ReturnValue = returnValue;
+        }
+
+        public object ReturnValue { get; }
+
+        public MethodInfo Method { get; }
+
+        public ArgumentCollection Arguments { get; }
+
+        public object FakedObject { get; }
+    }
+}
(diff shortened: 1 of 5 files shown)


blairconrad:
I'm surprised to see "[Merge pull request #1130 from FakeItEasy/hotfix-3.4.1  …](https://github.com/FakeItEasy/FakeItEasy/pull/1143/commits/f2d27a24531078dc0fc2220c84fdb852122adacb)" in the commit list.
I guess it makes sense, though…

thomaslevesque:
> I'm surprised to see "Merge pull request #1130 from FakeItEasy/hotfix-3.4.1 …" in the commit list.

That's because we didn't merge hotfix-3.4.1 into develop (because there was nothing to merge)

blairconrad:
> That's because we didn't merge hotfix-3.4.1 into develop (because there was nothing to merge)

fca1bf0e3d2a1676c748fff4d710f6b024fb7448

blairconrad:
I think it's because while we _did_ merge hotfix-3.4.1 into develop, we also merged into master, and that's what got tagged, so the merge to master is the commit that we're no merging, and it's a no-op because the content made it into develop already.

I think Git is doing the right thing. I just think it's confusing.

blairconrad:
On that note, if we merge your PR, and then later release 3.4.3, I think when we try to merge hotfix-3.4.3 into develop, the missing commit will come back!

thomaslevesque:
> On that note, if we merge your PR, and then later release 3.4.3, I think when we try to merge hotfix-3.4.3 into develop, the missing commit will come back!

Oh... I think you're right. I wanted to leave no trace of this commit, which is why I did it like this. OK, let me try again, then.

blairconrad:
The branches are getting wide, but I guess this is the best we can do if we want to use GitHub to merge:
image

I still think we're going to see the missing commit come back if we make a 3.4.3, but I do not know how we can possibly avoid that.

Shall I go ahead, then?

thomaslevesque:
> I still think we're going to see the missing commit come back if we make a 3.4.3, but I do not know how we can possibly avoid that.

Why? It's part of this merge, so it shouldn't come back later.

> Shall I go ahead, then?

👍

blairconrad:
> Why? It's part of this merge, so it shouldn't come back later.

Ah, yes. I see. Sorry.

thomaslevesque:
Thanks for the merge!

PR #1145, opened 2017-06-20
Support assigning out and refs without setting return value

### LINKED ISSUE #1134: Allow AssignsOutAndRefParameters on a non-void member without first specifying Returns
The return value should remain a `Dummy`.

This was spun off from https://github.com/FakeItEasy/FakeItEasy/issues/854#issuecomment-308392749.

Labelled **breaking** because we'll change the signature on some of the API methods.

blairconrad:
Fixes #1134.


tests/FakeItEasy.Specs/AssignsOutAndRefParametersSpecs.cs
```diff
@@ -162,6 +162,30 @@ public static void AssignOutAndRefParameters(IHaveARef subject, string outValue)
...
+
+            "When the faked delegate is configured to assign the out and ref parameters without setting the return value"
+                .x(() => A.CallTo(() => subject(1, ref refValue, out outValue))
+                    .AssignsOutAndRefParameters(43, 100));
+
+            "And I call the faked delegate"
+                .x(() => result = subject(1, ref refValue, out outValue));
+
+            "Then it returns a Dummy value"
+                .x(() => result.Should().Be(A.Dummy<int>()));
```
thomaslevesque: It would be better to test with a reference type, because `A.Dummy<int>()` returns 0, which is the same as if it wasn't actually returning a dummy. With `string` as the return value, we could test that it returns an empty string rather than null.

blairconrad:
Excellent comment, @thomaslevesque. I added a Dummy. I wanted some changes to the spec's layout to support this, so I just ditched the original commit and pushed two new ones. I hope you can still keep track of the involved change to the production code.

blairconrad:
Thank you.

PR #1149, opened 2017-06-21
Split C# and VB analyzers into separate packages, use dotnet pack to build packages

### LINKED ISSUE #1148: Split C# and VB.NET analyzers into two separate packages
We can't use `dotnet pack` to create the analyzer package, because it doesn't support packaging multiple projects into a single package. However, we don't actually need to have the two analyzers in a single package, because the vast majority of users will only ever need one. So we could use `dotnet pack` to build a package for each analyzer.

Since current users of the analyzer have installed the `FakeItEasy.Analyzer` package, we need to keep generating this package so that current users still get the updates. However, it will now be a meta-package, with no content, and dependencies on the two language-specific packages.

So the packages would look like this:

- FakeItEasy.Analyzer.CSharp
  - contains: FakeItEasy.Analyzer.CSharp.dll/pdb
  - depends on: Roslyn
- FakeItEasy.Analyzer.VisualBasic
  - contains: FakeItEasy.Analyzer.VisualBasic.dll/pdb
  - depends on: Roslyn
- FakeItEasy.Analyzer
  - contains: nothing
  - depends on: FakeItEasy.Analyzer.CSharp, FakeItEasy.Analyzer.VisualBasic

thomaslevesque:
Fixes #1148

### DIFF
--- a/build.csx
+++ b/build.csx
@@ -14,15 +14,21 @@ var versionInfoFile = "./src/VersionInfo.cs";
 var packagesDirectory = Path.GetFullPath("packages");
 var repoUrl = "https://github.com/FakeItEasy/FakeItEasy";
 var coverityProjectUrl = "https://scan.coverity.com/builds?project=FakeItEasy%2FFakeItEasy";
-var fakeItEasyProjectPath = "./src/FakeItEasy/FakeItEasy.csproj";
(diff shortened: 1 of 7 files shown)


blairconrad:
Thanks, @thomaslevesque.

PR #1150, opened 2017-06-21
Raise for free-form events

### LINKED ISSUE #987: Support Raise.With for arbitrary delegates without requiring a typeparam in the call.

thomaslevesque:
Fixes #987 
Supersedes #1061

### DIFF
--- a/docs/raising-events.md
+++ b/docs/raising-events.md
@@ -5,7 +5,7 @@ an event defined:
 
 ```csharp
 public interface IRobot
-{ 
+{
     event EventHandler FellInLove;
 }
 ```
@@ -18,12 +18,12 @@ raising with empty event args.
 
 ```csharp
 var robot = A.Fake<IRobot>();
-            
+
 robot.FellInLove += (s, e) =>
     {
         Console.WriteLine("Yay!");
     };
-         
+
 // Raise the event!
 robot.FellInLove += Raise.With(someEventArgs); // the "sender" will be robot
 
@@ -34,12 +34,12 @@ robot.FellInLove += Raise.WithEmpty(); // sender will be robot, args will be Eve
 robot.FellInLove += Raise.With(sender: robot, e: someEventArgs);
 ```
 
-Events of type `EventHandler<TEventArgs>` can be raised in exactly the same way. 
+Events of type `EventHandler<TEventArgs>` can be raised in exactly the same way.
 
-Note that as of version 3.1.0, for the .NET Standard version of the FakeItEasy library, `TEventArgs` need not extend `EventArgs` to support the above syntax. The .NET 4.0 library still requires `TEventArgs` to extend `EventArgs`.
+Note that for the .NET 4.0 version of the FakeItEasy library, `TEventArgs` needs to extend `EventArgs` to support the above syntax.
 
-If an event is defined using a **custom delegate**, then `Raise` needs
-a typeparam to help it out:
+It is also possible to raise events defined using a **custom delegate** (a.k.a
+"free-form delegate"), like these:
 
 ```csharp
 public delegate void CustomEventHandler(object sender, CustomEventArgs e);
@@ -47,7 +47,20 @@ public delegate void FreeformEventHandler(int count);
 …
 event CustomEventHandler CustomEvent;
 event FreeformEventHandler FreeformEvent;
-…
+```
+
+To do that, you can either:
+
+* use `Raise.FreeForm.With`, which automatically infers the correct delegate type:
+
+```csharp
+fake.CustomEvent += Raise.FreeForm.With(fake, sampleCustomEventArgs);
+fake.FreeformEvent += Raise.FreeForm.With(7);
+```
+
+* or specify the delegate type explicitly as a type parameter to `Raise.With`:
+
+```csharp
 fake.CustomEvent += Raise.With<CustomEventHandler>(fake, sampleCustomEventArgs);
 fake.FreeformEvent += Raise.With<FreeformHandler>(7);
 ```
--- a/src/FakeItEasy/Raise.cs
+++ b/src/FakeItEasy/Raise.cs
@@ -69,5 +69,21 @@ namespace FakeItEasy
         {
             return new DelegateRaiser<TEventHandler>(arguments, ArgumentProviderMap);
         }
+
+        /// <summary>
+        /// Allows the developer to raise an event with a non-standard signature on a faked object.
+        /// </summary>
+        public static class FreeForm
+        {
+            /// <summary>
+            /// Raises an event with non-standard signature, resolving the actual delegate type dynamically.
+            /// </summary>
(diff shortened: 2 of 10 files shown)


src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -1,92 +1,92 @@
...
+  <ItemGroup Condition="'$(TargetFramework)' == 'net40'">
+    <PackageReference Include="ILMerge" Version="2.14.1208" PrivateAssets="all" />
+    <PackageReference Include="Castle.Core" Version="4.1.0" PrivateAssets="all" />
+  </ItemGroup>
+
+  <Target Name="ILMerge40" AfterTargets="Build" Condition="'$(TargetFramework)' == 'net40'">
+    <Exec Command="&quot;$(NuGetPackageRoot)\ilmerge\2.14.1208\tools\ILmerge.exe&quot; /keyfile:..\..\FakeItEasy.snk /lib:$(OutputPath) /targetplatform:&quot;v4,$(MSBuildProgramFiles32)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.0&quot; /internalize:&quot;..\ILMerge.Internalize.Exclude.txt&quot; /out:@(MainAssembly) /log:$(OutputPath)ILMerge.log &quot;@(IntermediateAssembly)&quot; &quot;$(OutputPath)Castle.Core.dll&quot;" />
+  </Target>
+
+</Project>
```
blairconrad: Dude, you changed every line!
I'd revert.

src/FakeItEasy/Raise.cs
```diff
@@ -69,5 +69,21 @@ public static TEventHandler With<TEventHandler>(params object[] arguments)
         {
             return new DelegateRaiser<TEventHandler>(arguments, ArgumentProviderMap);
         }
```
blairconrad: Is this not going away? Or is that another issue?

blairconrad:
The changes look generally good, except where noted.
If I can make a meta-comment, the commit order threw me. I didn't expect to see the "implement" before the specs!
There's no harm. I was just confused is all.

src/FakeItEasy/FakeItEasy.csproj
thomaslevesque: Oops. Don't know what happened, it wasn't like that before I rebased... Probably just a line-ending change. I'll fix it.

src/FakeItEasy/Raise.cs
thomaslevesque: We could make it go away, but it's unrelated to this change. Since it's breaking, I think we should do it separately.

thomaslevesque:
> If I can make a meta-comment, the commit order threw me. I didn't expect to see the "implement" before the specs!

Yeah, I don't always strictly apply TDD ;)
The commit order reflect the order in which I did things.

src/FakeItEasy/Raise.cs
blairconrad: Fair enough.

thomaslevesque:
> I forgot. Docs!

Ah. Good thing you never forget about those, because I always do...

blairconrad:
But I did forget! I only last-minute unforgot!

docs/raising-events.md
```diff
@@ -34,20 +34,33 @@ robot.FellInLove += Raise.WithEmpty(); // sender will be robot, args will be Eve
 robot.FellInLove += Raise.With(sender: robot, e: someEventArgs);
 ```
 
-Events of type `EventHandler<TEventArgs>` can be raised in exactly the same way. 
+Events of type `EventHandler<TEventArgs>` can be raised in exactly the same way.
 
-Note that as of version 3.1.0, for the .NET Standard version of the FakeItEasy library, `TEventArgs` need not extend `EventArgs` to support the above syntax. The .NET 4.0 library still requires `TEventArgs` to extend `EventArgs`.
+Note that as of version 3.1.0, for the .NET Standard and .NET 4.5 versions of the FakeItEasy library, `TEventArgs` need not extend `EventArgs` to support the above syntax. The .NET 4.0 library still requires `TEventArgs` to extend `EventArgs`.
```
blairconrad: Good point about .NET 4.5 version not needing `TEventArgs` to extend `EventArgs`, but we didn't release a .NET 4.5 library until 3.4! Wording is hard, but perhaps something like

> Note that for the .NET 4.5 version of the FakeItEasy library, and the .NET Standard version starting with release 3.1.0, `TEventArgs` need not extend `EventArgs` to support the above syntax. All releases of the .NET 4.0 library still require `TEventArgs` to extend `EventArgs`.

docs/raising-events.md
```diff
@@ -34,20 +34,33 @@ robot.FellInLove += Raise.WithEmpty(); // sender will be robot, args will be Eve
...
+To do that, you can either:
+
+* use `Raise.FreeForm.With`, which automatically infers the correct delegate type:
+
+```csharp
+fake.CustomEvent += Raise.FreeForm.With(fake, sampleCustomEventArgs);
+fake.FreeformEvent += Raise.FreeForm.With(7);
+```
+
+* specify the delegate type explicitly as a type parameter to `Raise.With`:
```
blairconrad: I think this would read better with an initial "or": 

* or specify the delegate type…

docs/raising-events.md
thomaslevesque: Ah, yes. But why mention release 3.1.0 at all? The docs are for the latest version, no?

docs/raising-events.md
thomaslevesque: 👍

docs/raising-events.md
blairconrad: This is a good point. A holdover from when we had them in the wiki perhaps.
Of course, even when docs are versioned, sometimes it's nice to indicate when a feature became available or changed.

Still, FakeItEasy should be upgradeable on a whim, and the .NET Standard behaviour has been there for over 3 releases! I think you're right. We could just  drop the versions.

thomaslevesque: I simplified this paragraph. Since we're no longer mentioning a specific version, there's no point in saying that there's no constraint in 4.5 and netstd. We just need to say that there is a constraint in 4.0.

PR #1151, opened 2017-06-21
Deploy to NuGet on new tag

### LINKED ISSUE #1039: Automatic NuGet deployment when release is published

blairconrad:
Fixes #1039.
I expect. I'm not really sure how to test this without actually merging and then tagging something.
Wanna give it a go and release a 4.0.0 alpha? :wink:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,3 +29,11 @@ test: off
 
 artifacts:
     - path: .\artifacts\*\*
+
+deploy:
+    - provider: NuGet
+      api_key:
+          secure: BVzuxe9khaakJDNLAhtUM2w5G3gAkoOasCDAilstAtpLuCdFNfu3jrf0KrI5B1Bo
+      skip_symbols: true
+      on:
+          appveyor_repo_tag: true


thomaslevesque:
> I'm not really sure how to test this without actually merging and then tagging something.

Well, we could make a dummy repo just to test it, pushing to some other deployment provider instead of NuGet...

> Wanna give it a go and release a 4.0.0 alpha? 😉

We could do that, too.

thomaslevesque:
If we do this, we'll need to update the release checklist. "edit draft release" should come before "tag and push"

blairconrad:
> Well, we could make a dummy repo just to test it, pushing to some other deployment provider instead of NuGet...

But that proves not that much. I mean, it tests that the thing is hooked up, but won't give us a proper release or even check that I encrypted the API key properly. We do have this from my AppVeyor build:
https://ci.appveyor.com/project/blairconrad/fakeiteasy/build/499#L659. It shows that the config was understood, and the action was skipped _for the right reason_.

> If we do this, we'll need to update the release checklist. "edit draft release" should come before "tag and push"

Oh? I don't mind, but is there a pressing need? Just to reduce the turnaround time after the build goes to NuGet?

thomaslevesque:
> But that proves not that much.

Well, at least it will prove that deploy is indeed triggered, and if something goes wrong, at least we don't push anything to NuGet. I'm trying it now, I'll let you know how it goes.

> Oh? I don't mind, but is there a pressing need? Just to reduce the turnaround time after the build goes to NuGet?

Just so that the release notes are up to date before we deploy

blairconrad:
> Just so that the release notes are up to date before we deploy

Fair enough. They aren't visible at that point, but I've no problems with them being up to date before deploy. Heck. I like them to be up to date _all the time_.

So if we want to release an alpha, I guess we just tag develop, yes? No need to create a branch or merge anything. (We don't need to do this right away or anything.)

thomaslevesque:
> They aren't visible at that point

Good point. *I* always see them, so I tend to forget other people don't... OK, so maybe there's no need to change the current checklist.

> So if we want to release an alpha, I guess we just tag develop, yes? No need to create a branch or merge anything. (We don't need to do this right away or anything.)

I think so

blairconrad:
>  so maybe there's no need to change the current checklist.

too late!

thomaslevesque:
Thanks @blairconrad. Now let's cross our fingers...

blairconrad:
Pre-crossed. Thanks for the merge.

PR #1156, opened 2017-06-22
Nullable constraint analyzer

### LINKED ISSUE #1060: Diagnostic for argument constraint with non-nullable type instead of nullable type
Split from #1005. Inspired by #660.

Assuming you have this interface:
```c#
public interface IFoo
{
    int Bar(int? x);
}
```
And a call configuration like this:
```c#
A.CallTo(() => foo.Bar(A<int>._)).Returns(42);
```
It's easy to miss the fact that the constraint won't match when the argument is `null` (it used to throw a NRE, which at least indicated something was wrong, but now it silently rejects the null argument). It's especially confusing because when you write `A<int>._`, you usually mean "match anything". The analyzer could detect this and emit a warning.

(I wish we could omit the type and just say `A.CallTo(() => foo.Bar(A._)`, which is more concise and easier to get right. Java's type inference is a little better in that regard. This syntax could also be achieved if we had a ["bottom type"](https://en.wikipedia.org/wiki/Bottom_type) in C#.)

thomaslevesque:
Fixes #1060 

WIP because only the analyzer is done; I also intend to add a code fix.

### DIFF
--- a/docs/analyzer.md
+++ b/docs/analyzer.md
@@ -16,10 +16,11 @@ PM> Install-Package FakeItEasy.Analyzer
 
 ## Diagnostics
 
-The analyzer currently provides two diagnostics:
+The analyzer currently provides the following diagnostics:
 
-| Id             | Summary                               | Description                                                                                                                                           |
-|----------------|---------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------|
-| FakeItEasy0001 | Unused call specification             | Triggered when you specify a call but don't configure or assert it, making it a no-op.                                                                |
-| FakeItEasy0002 | Non-virtual member configuration      | Triggered when you try to configure a non-virtual member, which cannot be faked.                                                                      |
-| FakeItEasy0003 | Argument constraint outside call spec | Triggered when you try to use an [argument constraint](argument-constraints.md) outside of a [call specification](specifying-a-call-to-configure.md). |
+| Id             | Summary                                  | Description                                                                                                                                           |
+|----------------|------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------|
+| FakeItEasy0001 | Unused call specification                | Triggered when you specify a call but don't configure or assert it, making it a no-op.                                                                |
+| FakeItEasy0002 | Non-virtual member configuration         | Triggered when you try to configure a non-virtual member, which cannot be faked.                                                                      |
+| FakeItEasy0003 | Argument constraint outside call spec    | Triggered when you try to use an [argument constraint](argument-constraints.md) outside of a [call specification](specifying-a-call-to-configure.md). |
+| FakeItEasy0004 | Argument constraint nullability mismatch | Triggered when you use a non-nullable [argument constraint](argument-constraints.md) for a nullable parameter. Calls where the argument is null won't be matched. If this is intentional, consider using `A<T?>.That.IsNotNull()` instead. If it's not, make the argument constraint nullable (`A<T?>`). |
--- a/src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchAnalyzer.cs
+++ b/src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchAnalyzer.cs
@@ -1,5 +1,6 @@
 ﻿namespace FakeItEasy.Analyzer
 {
+    using System.Collections.Generic;
     using System.Collections.Immutable;
     using Microsoft.CodeAnalysis;
 #if CSHARP
@@ -37,32 +38,32 @@
             }
 
             var argument = completeConstraint.Parent as ArgumentSyntax;
+#if CSHARP
+            var argumentList = argument?.Parent as BaseArgumentListSyntax;
+#elif VISUAL_BASIC
             var argumentList = argument?.Parent as ArgumentListSyntax;
-            var invocation = argumentList?.Parent as InvocationExpressionSyntax;
-            if (invocation == null)
+#endif
+
+            if (argumentList == null)
             {
                 return;
             }
 
-            // TODO: could be an indexer, not a method
-            var method = SymbolHelpers.GetCalledMethodSymbol(invocation, context);
-            if (method == null)
+            if (!TryGetNameAndParameters(context, argumentList.Parent, out string methodOrIndexerName, out var parameters))
             {
                 return;
             }
 
             int index = argumentList.Arguments.IndexOf(argument);
-            if (index < 0 || index >= method.Parameters.Length)
+            if (index < 0 || index >= parameters.Count)
             {
                 return;
             }
 
-            var parameter = method.Parameters[index];
-            var parameterType = parameter.Type as INamedTypeSymbol;
-            if (parameterType != null && parameterType.IsNullable())
+            var parameter = parameters[index];
+            if (parameter.Type is INamedTypeSymbol parameterType && parameterType.IsNullable())
             {
-                var nullableParameterType = (INamedTypeSymbol)parameter.Type;
-                var nonNullableParameterType = nullableParameterType.TypeArguments[0];
+                var nonNullableParameterType = parameterType.TypeArguments[0];
                 if (constraintType.Equals(nonNullableParameterType))
                 {
                     var descriptor = DiagnosticDefinitions.ArgumentConstraintNullabilityMismatch;
@@ -70,10 +71,64 @@
                         descriptor,
                         completeConstraint.GetLocation(),
                         parameter.Name,
-                        method.GetDecoratedName(),
+                        methodOrIndexerName,
                         completeConstraint.ToString());
                     context.ReportDiagnostic(diagnostic);
                 }
+            }
+        }
+
+        private static bool TryGetNameAndParameters(
+            SyntaxNodeAnalysisContext context,
+            SyntaxNode node,
+            out string methodOrIndexerName,
+            out IReadOnlyList<IParameterSymbol> parameters)
+        {
+            methodOrIndexerName = null;
+            parameters = null;
+
+            switch (node)
+            {
+                case InvocationExpressionSyntax invocation:
+                {
+                    var method = SymbolHelpers.GetCalledMethodSymbol(invocation, context);
+                    if (method == null)
+                    {
+#if VISUAL_BASIC
+                        var indexer = SymbolHelpers.GetAccessedIndexerSymbol(invocation, context);
+                        if (indexer == null)
+                        {
+                            return false;
+                        }
+
+                        methodOrIndexerName = indexer.Name;
+                        parameters = indexer.Parameters;
+                        return true;
+#else
+                        return false;
+#endif
+                    }
+
+                    parameters = method.Parameters;
+                    methodOrIndexerName = method.GetDecoratedName();
+                    return true;
+                }
+#if CSHARP
+                case ElementAccessExpressionSyntax elementAccess:
+                {
+                    var indexer = SymbolHelpers.GetAccessedIndexerSymbol(elementAccess, context);
+                    if (indexer == null)
+                    {
+                        return false;
+                    }
+
+                    methodOrIndexerName = indexer.Name;
+                    parameters = indexer.Parameters;
+                    return true;
+                }
+#endif
+                default:
+                    return false;
             }
         }
     }
--- a/src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchAnalyzer.cs
+++ b/src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchAnalyzer.cs
@@ -1,0 +1,80 @@
+﻿namespace FakeItEasy.Analyzer
+{
+    using System.Collections.Immutable;
+    using Microsoft.CodeAnalysis;
+#if CSHARP
+    using Microsoft.CodeAnalysis.CSharp.Syntax;
+#endif
+    using Microsoft.CodeAnalysis.Diagnostics;
+#if VISUAL_BASIC
+    using Microsoft.CodeAnalysis.VisualBasic.Syntax;
+#endif
+
+#if CSHARP
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+#elif VISUAL_BASIC
+    [DiagnosticAnalyzer(LanguageNames.VisualBasic)]
+#endif
+    public class ArgumentConstraintNullabilityMismatchAnalyzer : ArgumentConstraintAnalyzerBase
+    {
+        private static readonly ImmutableHashSet<string> SupportedArgumentConstraintProperties =
+            ImmutableHashSet.Create(
+                "FakeItEasy.A`1._",
+                "FakeItEasy.A`1.Ignored");
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } =
+            ImmutableArray.Create(DiagnosticDefinitions.ArgumentConstraintNullabilityMismatch);
+
+        protected override bool IsSupportedArgumentConstraintProperty(string fullName) =>
+            SupportedArgumentConstraintProperties.Contains(fullName);
+
+        protected override void AnalyzeArgumentConstraintCore(SyntaxNodeAnalysisContext context, SyntaxNode completeConstraint)
+        {
+            var constraintType = context.SemanticModel.GetTypeInfo(completeConstraint).Type as INamedTypeSymbol;
+            if (constraintType == null || !constraintType.IsValueType || constraintType.IsNullable())
+            {
+                return;
+            }
+
+            var argument = completeConstraint.Parent as ArgumentSyntax;
+            var argumentList = argument?.Parent as ArgumentListSyntax;
+            var invocation = argumentList?.Parent as InvocationExpressionSyntax;
+            if (invocation == null)
+            {
+                return;
+            }
+
+            // TODO: could be an indexer, not a method
+            var method = SymbolHelpers.GetCalledMethodSymbol(invocation, context);
+            if (method == null)
+            {
+                return;
+            }
+
+            int index = argumentList.Arguments.IndexOf(argument);
+            if (index < 0 || index >= method.Parameters.Length)
+            {
+                return;
+            }
+
+            var parameter = method.Parameters[index];
+            var parameterType = parameter.Type as INamedTypeSymbol;
+            if (parameterType != null && parameterType.IsNullable())
+            {
+                var nullableParameterType = (INamedTypeSymbol)parameter.Type;
+                var nonNullableParameterType = nullableParameterType.TypeArguments[0];
+                if (constraintType.Equals(nonNullableParameterType))
+                {
+                    var descriptor = DiagnosticDefinitions.ArgumentConstraintNullabilityMismatch;
+                    var diagnostic = Diagnostic.Create(
+                        descriptor,
+                        completeConstraint.GetLocation(),
+                        parameter.Name,
+                        method.GetDecoratedName(),
+                        completeConstraint.ToString());
+                    context.ReportDiagnostic(diagnostic);
+                }
+            }
+        }
+    }
+}
--- a/src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchCodeFixProvider.cs
+++ b/src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchCodeFixProvider.cs
@@ -1,0 +1,141 @@
+﻿namespace FakeItEasy.Analyzer
+{
+    using System.Collections.Immutable;
+    using System.Linq;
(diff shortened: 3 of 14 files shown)


tests/FakeItEasy.Analyzer.CSharp.Tests/ArgumentConstraintNullabilityMismatchAnalyzerTests.cs
```diff
@@ -0,0 +1,127 @@
...
+    using FakeItEasy.Tests;
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.Diagnostics;
+    using Xunit;
+
+    public class ArgumentConstraintNullabilityMismatchAnalyzerTests : DiagnosticVerifier
+    {
+        private const string CodeTemplate = @"using FakeItEasy;
+namespace TheNamespace
+{{
```
blairconrad: I saw this and thought my extension was acting up again. Then I saw that it was a template.

src/FakeItEasy.Analyzer/VisualBasic/Resources.resx
```diff
@@ -94,4 +94,13 @@
...
     <comment>The title of the diagnostic.</comment>
   </data>
+    <data name="ArgumentConstraintNullabilityMismatchTitle" xml:space="preserve">
+    <value>Argument constraint nullability mismatch</value>
+  </data>
+    <data name="ArgumentConstraintNullabilityMismatchDescription" xml:space="preserve">
+    <value>A non-nullable argument constraint is used for a nullable parameter.</value>
+  </data>
+    <data name="ArgumentConstraintNullabilityMismatchMessageFormat" xml:space="preserve">
+    <value>Parameter '{0}' of method '{1}' is nullable, but argument constraint '{2}' is not. Calls where this argument is Nothing will not be matched.</value>
```
blairconrad: Not "nothingable"? I guess not.

src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchCodeFixProvider.cs
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+        private static MemberAccessExpressionSyntax SimpleMemberAccess(ExpressionSyntax expression, SimpleNameSyntax name)
+        {
+#if CSHARP
+            return SyntaxFactory.MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression, expression, name);
+#elif VISUAL_BASIC
+            return SyntaxFactory.SimpleMemberAccessExpression(expression, name);
+#endif
+        }
```
thomaslevesque: I don't know why the Roslyn API is different for C# and VB.NET in this case... there's no good reason IMO.

src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchCodeFixProvider.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                // The T? type
+                var nullableConstraintType = SyntaxFactory.NullableType(constraintType);
+
+                // Replace T node with T? and return updated document
+                var newRoot = root.ReplaceNode(constraintType, nullableConstraintType);
+                return context.Document.WithSyntaxRoot(newRoot);
+            }
+
+            return context.Document;
```
thomaslevesque: If the node with the diagnostic doesn't match our expectations, we just return the original document. Not sure if this is the correct way to handle it, but I don't think it's possible anyway.

src/FakeItEasy.Analyzer/SymbolHelpers.cs
```diff
@@ -21,20 +21,38 @@ internal static class SymbolHelpers
...
         internal static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax call, SyntaxNodeAnalysisContext context)
         {
-            var name = (call?.Expression as MemberAccessExpressionSyntax)?.Name
-                ?? call?.Expression as IdentifierNameSyntax;
-
-            var method =
-                name == null
-                    ? null
-                    : context.SemanticModel.GetSymbolInfo(name).Symbol as IMethodSymbol;
+            var method = context.SemanticModel.GetSymbolInfo(call).Symbol as IMethodSymbol;
```
thomaslevesque: This change and the one in `GetAccessedPropertySymbol` are unrelated to this PR, but I realized while writing `GetAccessedIndexerSymbol` that they could be simplified. The changes are in a separate commit, so I can easily move them to a separate PR if you prefer.

src/FakeItEasy.Analyzer/SymbolHelpers.cs
```diff
@@ -21,20 +21,38 @@ internal static class SymbolHelpers
...
+        /// </summary>
+        /// <param name="invocation">The invocation expression that the symbol is required.</param>
+        /// <param name="context">Current context.</param>
+        /// <returns>The symbol for the invocation call.</returns>
+        internal static IPropertySymbol GetAccessedIndexerSymbol(InvocationExpressionSyntax invocation, SyntaxNodeAnalysisContext context)
+        {
+            return context.SemanticModel.GetSymbolInfo(invocation).Symbol as IPropertySymbol;
+        }
+#endif
+
```
thomaslevesque: The difference is because in C#, an indexer access is an `ElementAccessExpressionSyntax`, while in VB.NET it's an `InvocationExpressionSyntax` (which makes sense, since it looks exactly the same as a method call).

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/CodeFixVerifier.Helper.cs
```diff
@@ -0,0 +1,84 @@
...
+    using Microsoft.CodeAnalysis;
+    using Microsoft.CodeAnalysis.CodeActions;
+    using Microsoft.CodeAnalysis.Formatting;
+    using Microsoft.CodeAnalysis.Simplification;
+
+    /// <summary>
+    /// Diagnostic Producer class with extra methods dealing with applying codefixes
+    /// All methods are static
+    /// </summary>
+    public abstract partial class CodeFixVerifier
```
thomaslevesque: This class comes from the default template for analyzer projects. I just made the same adjustments as in `DiagnosticVerifier` (i.e. fix code analysis warnings and use FluentAssertions)

tests/FakeItEasy.Analyzer.VisualBasic.Tests/ArgumentConstraintNullabilityMismatchAnalyzerTests.cs
```diff
@@ -0,0 +1,231 @@
...
+        {
+            string completeConstraint = "A(Of Integer).Ignored";
+            string call = $"foo.Item({completeConstraint})";
+            string code = string.Format(CodeTemplate, call);
+
+            string fixedConstraint = "A(Of Integer?).That.IsNotNull()";
+            string fixedCall = $"foo.Item({fixedConstraint})";
+            string fixedCode = string.Format(CodeTemplate, fixedCall);
+            this.VerifyVisualBasicFix(code, fixedCode, codeFixIndex: 1);
+        }
```
thomaslevesque: There are more tests in VB.NET because there are two way to invoke indexers: using the default property (`foo(index)`) or using a named indexer (`foo.Item(index)`).

blairconrad:
@thomaslevesque, I've not forgotten this. I just need lots of focus to understand analyzer code, so I've been putting it off until I clear some other obligations and can devote some real time to it.

thomaslevesque:
> @thomaslevesque, I've not forgotten this. I just need lots of focus to understand analyzer code, so I've been putting it off until I clear some other obligations and can devote some real time to it.

Don't worry, there's no rush

tests/FakeItEasy.Analyzer.CSharp.Tests/ArgumentConstraintNullabilityMismatchAnalyzerTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            string completeConstraint = $"A<int>.{constraint}";
+            string call = $"foo.NonNullableParam({completeConstraint})";
+            string code = string.Format(CodeTemplate, call);
+
+            this.VerifyCSharpDiagnostic(code);
+        }
+
+        [Theory]
+        [MemberData(nameof(SupportedConstraints))]
+        public void Diagnostic_should_not_be_triggered_when_non_nullable_constraint_is_used_for_other_nullable_parameter(string constraint)
```
blairconrad: It's never occurred to me to supply a `A<int>._` for a `long?` (or even `long`) parameter. 
Is there any reason _not_ to warn? It won't match a `null`.

(On that note, neither would it match a non-`null`. Maybe we should have a diagnostic for the case where the _type_ of the constraint is mismatched!)
(Oh, I guess that's a good reason not to trigger this diagnostic. Even aligning the nullability won't cause a match.)

src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchCodeFixProvider.cs
```diff
@@ -0,0 +1,141 @@
...
+                diagnostic);
+
+            return CompletedTask;
+        }
+
+        private static MemberAccessExpressionSyntax GetConstraintNode(Diagnostic diagnostic, SyntaxNode root)
+        {
+            // getInnermostNodeForTie: true to disambiguate between ArgumentSyntax and
+            // MemberAccessExpressionSyntax, which have the same source span. We want the
+            // MemberAccessExpressionSyntax node, which is the innermost one.
```
blairconrad: Great comment. Thanks.

src/FakeItEasy.Analyzer/VisualBasic/Resources.resx
```diff
@@ -103,4 +103,10 @@
     <data name="ArgumentConstraintNullabilityMismatchMessageFormat" xml:space="preserve">
     <value>Parameter '{0}' of method '{1}' is nullable, but argument constraint '{2}' is not. Calls where this argument is Nothing will not be matched.</value>
   </data>
+  <data name="MakeConstraintNullableCodeFixTitle">
+    <value>Change constraint type to nullable</value>
+  </data>
+  <data name="MakeNotNullConstraintCodeFixTitle">
+    <value>Change to 'not Nothing' constraint</value>
```
blairconrad: I'm sure you did this intentionally, but we'll still end up with a constraint that says `IsNotNull`.
I worry about confusing the VB developers.
Is this a situation where we're better off using the actual method name rather than an English description of the outcome? Maybe both?

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/CodeFixVerifier.Helper.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        /// <summary>
+        /// Apply the inputted CodeAction to the inputted document.
```
blairconrad: I did not know that "inputted" was an English word. The things one learns!
It hits me like a punch to the eye.
I'm not suggesting a change.

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/CodeFixVerifier.cs
```diff
@@ -0,0 +1,127 @@
...
+    public abstract partial class CodeFixVerifier : DiagnosticVerifier
+    {
+#if CSHARP
+        /// <summary>
+        /// Called to test a C# codefix when applied on the inputted string as a source
+        /// </summary>
+        /// <param name="oldSource">A class in the form of a string before the CodeFix was applied to it</param>
+        /// <param name="newSource">A class in the form of a string after the CodeFix was applied to it</param>
+        /// <param name="codeFixIndex">Index determining which codefix to apply if there are multiple</param>
+        /// <param name="allowNewCompilerDiagnostics">A bool controlling whether or not the test will fail if the CodeFix introduces other warnings after being applied</param>
```
blairconrad: The text of this comment would prompt me to supply a `true` value when I wanted the test to fail if a new warning appears, but this is not the case, right?

src/FakeItEasy.Analyzer/SymbolHelpers.cs
blairconrad: I'm content with where they are. Thanks for the note!

src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchAnalyzer.cs
```diff
@@ -0,0 +1,137 @@
...
+            if (index < 0 || index >= parameters.Count)
+            {
+                return;
+            }
+
+            var parameter = parameters[index];
+            var parameterType = parameter.Type as INamedTypeSymbol;
+            if (parameterType != null && parameterType.IsNullable())
+            {
+                var nullableParameterType = (INamedTypeSymbol)parameter.Type;
```
blairconrad: This is no different from `parameterType`, if I read right.

docs/analyzer.md
```diff
@@ -16,10 +16,11 @@ PM> Install-Package FakeItEasy.Analyzer
 
 ## Diagnostics
 
-The analyzer currently provides two diagnostics:
+The analyzer currently provides the following diagnostics:
```
blairconrad: crafty!

docs/analyzer.md
```diff
@@ -16,10 +16,11 @@ PM> Install-Package FakeItEasy.Analyzer
...
+| FakeItEasy0001 | Unused call specification                | Triggered when you specify a call but don't configure or assert it, making it a no-op.                                                                |
+| FakeItEasy0002 | Non-virtual member configuration         | Triggered when you try to configure a non-virtual member, which cannot be faked.                                                                      |
+| FakeItEasy0003 | Argument constraint outside call spec    | Triggered when you try to use an [argument constraint](argument-constraints.md) outside of a [call specification](specifying-a-call-to-configure.md). |
+| FakeItEasy0004 | Argument constraint nullability mismatch | Triggered when you use a non-nullable [argument constraint](argument-constraints.md) for a nullable parameter. Calls where the argument is null won't be matched. If this is intentional, consider using `A<T?>.That.IsNotNull()` instead. If it's not, make the argument constraint nullable (`A<T?>`). |
```
blairconrad: After a quick search, I did no better than you. Let's live with it for now.
OH! OH! I thought of a way:

```html
<table>
<tr>
…
```

😉

tests/FakeItEasy.Analyzer.CSharp.Tests/ArgumentConstraintNullabilityMismatchAnalyzerTests.cs
thomaslevesque: > It's never occurred to me to supply a `A<int>._` for a `long?` (or even long) parameter.
> Is there any reason not to warn? It won't match a null.

Good point. I wanted to avoid false positives, so I made the analyzer quite restrictive on when it was triggered. But I don't remember my exact reasoning when I did this... something to do with not having to deal with implicit conversions, I think. But maybe it should be triggered in that case as well.

> (On that note, neither would it match a non-null. Maybe we should have a diagnostic for the case where the type of the constraint is mismatched!)

Oh, I had never thought of that. Currently `A<int>._` won't match if the actual parameter type is `long`. Yes, we should totally add a diagnostic for this.

> (Oh, I guess that's a good reason not to trigger this diagnostic. Even aligning the nullability won't cause a match.)

Indeed!

src/FakeItEasy.Analyzer/VisualBasic/Resources.resx
thomaslevesque: I don't know... We could add a `IsNotNothing` constraint, just for VB developers ;). But it would also appear in intellisense in C# code...

What about this?

> Change to 'not Nothing' constraint (IsNotNull)

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/CodeFixVerifier.Helper.cs
thomaslevesque: Oh, that's ugly... it comes from the project template, I hadn't even seen it.

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/CodeFixVerifier.cs
thomaslevesque: I have no idea. As I said, I didn't write that code. I can investigate, if you want.

src/FakeItEasy.Analyzer/ArgumentConstraintNullabilityMismatchAnalyzer.cs
thomaslevesque: Oh, yes. I don't know what I was thinking.

docs/analyzer.md
thomaslevesque: Well, I could live with an HTML table, but is it supported on readthedocs?

blairconrad: I assumed so, since it is in markdown. I just tried it using mkdocs and the readthedocs theme and see

image

generated from

```html
<table>
  <tr>
    <td>Id</td>
    <td>Summary</td>
    <td>Description</td>
  </tr>
  <tr>
    <td>FakeItEasy0001</td>
    <td>Unused call specification</td>
    <td>Triggered when you specify a call but don't configure or assert it, making it a no-op.</td>
  </tr>
  <tr>
    <td>FakeItEasy0002</td>
    <td>Non-virtual member configuration</td>
    <td>Triggered when you try to configure a non-virtual member, which cannot be faked.</td>
  </tr>
  <tr>
    <td>FakeItEasy0003</td>
    <td>Argument constraint outside call spec</td>
    <td>
        Triggered when you try to use an <a href="../argument-constraints">argument constraint</a> outside of a
        <a href="../specifying-a-call-to-configure">call specification</a>.
    </td>
  </tr>
  <tr>
    <td>FakeItEasy0004</td>
    <td>Argument constraint nullability mismatch</td>
    <td>
      Triggered when you use a non-nullable [argument constraint](argument-constraints.md) for a nullable parameter. 
      Calls where the argument is null won't be matched. If this is intentional, consider using <code>A&lt;T?&gt;.That.IsNotNull()</code> instead.
      If it's not, make the argument constraint nullable (<code>A&lt;T?&gt;</code>).
    </td>
  </tr>
</table>
```

Note that I did have to replace all the markdown _inside_ the table with HTML. We may be better off just sticking with the very long line.

tests/FakeItEasy.Analyzer.CSharp.Tests/ArgumentConstraintNullabilityMismatchAnalyzerTests.cs
blairconrad: I created #1163.

src/FakeItEasy.Analyzer/VisualBasic/Resources.resx
blairconrad: Love it.

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/CodeFixVerifier.Helper.cs
blairconrad: Yeah, I knew it wasn't you.

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/CodeFixVerifier.cs
blairconrad: It's fine. The parameter name is clear, and we have example usage.

docs/analyzer.md
thomaslevesque: > We may be better off just sticking with the very long line.

Yes, I think so too...

thomaslevesque:
> We have the diagnostic working on method calls and indexers. What about properties with indexes?

Oh, I forgot this comment. I think it already works; I'll add a test.

thomaslevesque:
> Oh, I forgot this comment. I think it already works; I'll add a test.

Well, I'll add a test for VB, at least. C# can consume named indexers, but not declare them, so it makes the test much harder to write...

thomaslevesque:
OK, I tried with a non-default named indexer, and it works as expected, but it's really the same as explicitly specifying the indexer name, which is already covered in the `*_for_named_indexer` tests. I don't think adding more tests for this adds any value.

blairconrad:
Nice, @thomaslevesque. Rebase away!

thomaslevesque:
Done

blairconrad:
Thanks, @thomaslevesque.

thomaslevesque:
Thanks for the merge, and the good comments ;)

PR #1160, opened 2017-07-03
Fix race condition in call assertion

### LINKED ISSUE #1159: Sometimes an expectation fails, but the error message lists the expected call.
From StackOverflow: https://stackoverflow.com/q/44884418/98713

Consider this code:

```csharp
void Main()
{
	var foo = A.Fake<IFoo>();
	
	Task.Run(() => foo.Bar("hello"));
	
	A.CallTo(() => foo.Bar(A<string>._)).MustHaveHappened();
}

public interface IFoo
{
	void Bar(string s);
}
```

Sometimes the expectation will succeed, sometimes it will fail, which is expected because there's a race condition, but when it fails, it will sometimes include the expected call in the error message:

```
    UserQuery+IFoo.Bar(s: <Ignored>)
    1: UserQuery+IFoo.Bar(s: "hello")
```

This is inconsistent; the call must have happened after we checked the expectation, but before we generated the error message.

thomaslevesque:
Fixes #1159

A simpler fix would have been to just call `.ToList()` on the calls in the `FakeAsserter` constructor, but it would have created unnecessary copies of the call list, which could possibly be large.

### DIFF
--- a/src/FakeItEasy/Core/FakeAsserter.cs
+++ b/src/FakeItEasy/Core/FakeAsserter.cs
@@ -23,13 +23,18 @@ namespace FakeItEasy.Core
         public virtual void AssertWasCalled(
             Func<ICompletedFakeObjectCall, bool> callPredicate, Action<IOutputWriter> callDescriber, Repeated repeatConstraint)
         {
-            var matchedCallCount = this.calls.Count(callPredicate);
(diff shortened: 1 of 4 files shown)


thomaslevesque:
Thanks for the merge!

PR #1166, opened 2017-07-09
Use 3 digits in version pre-release number

blairconrad:
Otherwise the package comes out as FakeItEasy.4.0.0-beta0001.

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,2 +1,3 @@
 ignore:
   commits-before: 2017-01-01T00:00:00
+legacy-semver-padding: 3


thomaslevesque:
> Otherwise the package comes out as FakeItEasy.4.0.0-beta0001.

Oh! I thought that's what we were already doing, but apparently not. Good job spotting it early 👍

thomaslevesque:
Thanks @blairconrad !

blairconrad:
Yes. "early". I completely did not push the tag and watch the build start and cancel the build and delete the tag, contravening all practices of good public tag management. Because I would never do that.

thomaslevesque:
> Yes "early". I completely did not push the tag and watch the build start and cancel the build and delete the tag, contravening all practices of good public tag management. Because I would never do that.

LOL!
Well, I didn't notice, so it was early enough ;)

blairconrad:
Thanks for the speedy review and merge!

PR #1168, opened 2017-07-09
Fix analyzer dependencies

### LINKED ISSUE #1167: 4.0.0-beta001 analyzers pull too many dependencies
The 4.0.0-beta001 analyzers have the following dependencies:

- Microsoft.CodeAnalysis.CSharp.Workspaces (>= 1.1.0)
- Microsoft.Composition (>= 1.0.27)
- NETStandard.Library (>= 1.6.1)

So when they're installed, it adds these dependencies (and their own dependencies) to the project, which is not desirable. The dependencies don't actually need to be exposed by the analyzer packages, because Visual Studio already provides them.

thomaslevesque:
Fixes #1167

### DIFF
--- a/src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.CSharp.csproj
+++ b/src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.CSharp.csproj
@@ -48,10 +48,11 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
Thanks for the merge!

PR #1173, opened 2017-07-28
Automatically attach artifacts to GitHub Release

### LINKED ISSUE #1171: automatically attach artifacts to GitHub release
Much like how we automatically publish nupkgs to nuget.org, it'd be nice to, on tagged builds, automatically attach our nupkgs to the GitHub release.

I experimented and found that the AppVeyor GitHub release deployment option overwrites the release description (where we keep the release notes), and potentially the pre-release and published statuses.

We may need to implement a workaround like described in  [the Upload release build artefacts discussion topic](http://help.appveyor.com/discussions/problems/5548-upload-release-build-artefacts) or perhaps just use the GitHub API's [Upload a release asset](https://developer.github.com/v3/repos/releases/#upload-a-release-asset) call directly.

blairconrad:
- only when building a tag
- will attach to release with same name as tag
- errors out if no such release is found
- attaches all the nupkgs

Fixes #1171.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,10 @@ environment:
         secure: 9iw6ajDvltEymr75DhlRbH8X+2olJ8mbWLPqyYMmleg=
     COVERITY_TOKEN:
         secure: S9XU2JhyUK9XstLG3CN2vB5M8ce8zYZp5VECx3kvx24=
+    GITHUB_TOKEN:
+        secure: 5dFninlYVqNF98Pk9ykQwPU2pBFeQmbMOiFG2iSkg+hhcwF3UfmnDB07lvOwYtez
+    NUGET_API_KEY:
+        secure: BVzuxe9khaakJDNLAhtUM2w5G3gAkoOasCDAilstAtpLuCdFNfu3jrf0KrI5B1Bo
 
 build_script:
 - cmd: if defined APPVEYOR_SCHEDULED_BUILD (build.cmd coverity) else (build.cmd)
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -10,6 +10,8 @@ environment:
         secure: 9iw6ajDvltEymr75DhlRbH8X+2olJ8mbWLPqyYMmleg=
     COVERITY_TOKEN:
         secure: S9XU2JhyUK9XstLG3CN2vB5M8ce8zYZp5VECx3kvx24=
+    GITHUB_TOKEN:
+        secure: 2nIBgxQYZTgGx99hpMX9QUA5J8IxOLD/8krdtdYdpu6uoXkIBxYg1zsemU+dasby
```
blairconrad: This is my key, if it matters to you. I didn't even look at creating one from the organization.

appveyor.yml
```diff
@@ -32,3 +34,35 @@ deploy:
       skip_symbols: true
       on:
           appveyor_repo_tag: true
```
blairconrad: Alternatively, we could combine `deploy` and `after_deploy` into one `deploy_script`. The NuGet API key could move up to `environment`, and pushing to NuGet could be a one-liner inside the `ForEach-Object` below.

appveyor.yml
blairconrad: On that note, it may have been encrypted for blairconrad/FakeItEasy. Let me replace it with a new one.

blairconrad:
@thomaslevesque, I amended the commit as you asked. Sucked the code out into `deploy.ps1`. I appears to not be called for PRs, but was called from a pushed branch on [my own appveyor build project](https://ci.appveyor.com/project/blairconrad/fakeiteasy/build/572).
I rolled the NuGet deploy into the script.
Can't really test this without releasing a package. I _think_ we're good. Worst case, the deployment fails, right?

thomaslevesque:
> Can't really test this without releasing a package. I think we're good. Worst case, the deployment fails, right

Yes, I don't think anything bad can happen. Let's try this and see how it goes.

thomaslevesque:
Nice work @blairconrad, thanks!

blairconrad:
Thanks!

PR #1175, opened 2017-07-28
Update Analyzer description to point to language-specific nupkgs

### LINKED ISSUE #1174: [Analyzer] Update FakeItEasy.Analyzer description to point toward C#/VB.NET packages
Most people will need either FakeItEasy.Analyzer.VisualBasic or FakeItEasy.Analyzer.CSharp, but not both. Tell them they could save a few bytes on the download by relying on one of the language-specific packages.

blairconrad:
Fixes #1174.


thomaslevesque:
Thanks @blairconrad !

PR #1178, opened 2017-08-02
Don't allow nested argument constraints

### LINKED ISSUE #1177: Argument constraint That (or Ignored, or _), when nested deeper in A.CallTo, misreports what's being matched
Found while investigating  [FakeItEasy - How to verify nested arguments value C#](https://stackoverflow.com/questions/45411178/fakeiteasy-how-to-verify-nested-arguments-value-c-sharp). 

Consider this test:

```c#
public class NestedThat
{
    public class Dto
    {
        public string Id;
    }

    public interface IService
    {
        void DoIt(Dto dto);
    }

    [Test]
    public void Test()
    {
        var fake = A.Fake<IService>();
        fake.DoIt(new Dto { Id = "some id" });

        A.CallTo(() => fake.DoIt(new Dto { Id = A<string>._ })).MustHaveHappened();
    }
}
```

The author would expect a passing test, but it fails, which isn't horrible, since the use of `A<string>._` inside the `Dto constructor` is illegal. However, the failure does not lead the reader to the fact that the use of `._` (and I'm sure any variant of `That`) is bad. Here's the output:

```
Test 'FakeItEasyPlay.NestedThat.Test' failed: FakeItEasy.ExpectationException : 

    FakeItEasyPlay.NestedThat+IService.DoIt(dto: <Ignored>)
    1: FakeItEasyPlay.NestedThat+IService.DoIt(dto: FakeItEasyPlay.NestedThat+Dto)

	at FakeItEasy.Core.FakeAsserter.AssertWasCalled(Func`2 callPredicate, String callDescription, Func`2 repeatPredicate, String repeatDescription)
	at FakeItEasy.Configuration.RuleBuilder.MustHaveHappened(Repeated repeatConstraint)
	at FakeItEasy.AssertConfigurationExtensions.MustHaveHappened(IAssertConfiguration configuration)
	NestedThat.cs(19,0): at FakeItEasyPlay.NestedThat.Test()
```

In addition to not highlighting the illegal `._`, we are misreporting the expected argument to `DoIt` as being equivalent to `A<Dto>._`.

It may be difficult (impossible) to throw correct error in all cases, but in this one, where we're capturing  a `That` constraint to match against a parameter of a different type, I'd hope we'd be able to do _something_.

thomaslevesque:
Fixes #1177

### DIFF
--- a/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
+++ b/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
@@ -1,12 +1,11 @@
-namespace FakeItEasy.Expressions
+﻿namespace FakeItEasy.Expressions
 {
     using System;
     using System.Collections.Generic;
     using System.Linq;
     using System.Linq.Expressions;
-#if FEATURE_NETCORE_REFLECTION
     using System.Reflection;
-#endif
+    using System.Runtime.CompilerServices;
     using FakeItEasy.Core;
     using FakeItEasy.Expressions.ArgumentConstraints;
 
@@ -80,8 +79,61 @@ namespace FakeItEasy.Expressions
             return Expression.Lambda(expression).Compile().DynamicInvoke();
         }
 
+        private static void CheckArgumentExpressionIsValid(Expression expression)
+        {
+            expression = GetExpressionWithoutBoxing(expression);
+
+            if (expression is MemberExpression)
+            {
+                // It's A._, or A.Ignore, or some other property/field, so it's safe.
+                return;
+            }
+
+            var visitor = new ArgumentConstraintExpressionVisitor();
+            var methodCallExpression = expression as MethodCallExpression;
+            if (methodCallExpression == null)
+            {
+                // An unknown kind of expression - could be a constructor, or almost anything else. Play it safe and
+                // check it out.
+                visitor.Visit(expression);
+            }
+            else
+            {
+                // A method call. It might be A<T>.That.Matches, or one of the other extension methods, so don't
+                // check the method node itself. Instead, look at all the arguments (except the first, if it's an
+                // extension method).
+                int argumentsToSkip = methodCallExpression.Method.IsDefined(typeof(ExtensionAttribute), false) ? 1 : 0;
+                foreach (var argument in methodCallExpression.Arguments.Skip(argumentsToSkip))
+                {
+                    visitor.Visit(argument);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Removes the explicit conversion introduced in a Linq expression by implicit boxing.
+        /// </summary>
+        /// <param name="expression">The expression from which to remove the boxing.</param>
+        /// <returns>The original expression, if no boxing is happening, or the expression that would be converted.</returns>
+        private static Expression GetExpressionWithoutBoxing(Expression expression)
+        {
+            if (expression.NodeType == ExpressionType.Convert &&
+                (expression.Type == typeof(object) || expression.Type.IsNullable() || expression.Type.GetTypeInfo().IsInterface))
+            {
+                var conversion = (UnaryExpression)expression;
+                if (conversion.Operand.Type.GetTypeInfo().IsValueType)
+                {
+                    return conversion.Operand;
+                }
+            }
+
+            return expression;
+        }
+
         private IArgumentConstraint GetArgumentConstraintFromExpression(Expression expression, out object value)
         {
+            CheckArgumentExpressionIsValid(expression);
+
             object expressionValue = null;
 
             var trappedConstraints = this.argumentConstraintTrapper.TrapConstraints(() =>
@@ -105,6 +157,35 @@ namespace FakeItEasy.Expressions
             }
 
             return new AggregateArgumentConstraint(result);
+        }
+
+        private class ArgumentConstraintExpressionVisitor : ExpressionVisitor
+        {
+            protected override Expression VisitMember(MemberExpression node)
+            {
+                if (IsMemberOfA(node.Member))
+                {
+                    throw new InvalidOperationException(ExceptionMessages.ArgumentConstraintCannotBeNestedInArgument);
+                }
+
+                return base.VisitMember(node);
+            }
+
+            private static bool IsMemberOfA(MemberInfo member)
+            {
+                return GetGenericTypeDefinition(member.DeclaringType) == typeof(A<>);
+            }
+
+            private static Type GetGenericTypeDefinition(Type type)
+            {
+                var typeInfo = type.GetTypeInfo();
+                if (typeInfo.IsGenericType && !typeInfo.IsGenericTypeDefinition)
+                {
+                    return typeInfo.GetGenericTypeDefinition();
+                }
+
+                return type;
+            }
         }
     }
 }
--- a/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -3,7 +3,6 @@ namespace FakeItEasy.Specs
     using System;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
-    using System.Linq;
     using System.Runtime.InteropServices;
     using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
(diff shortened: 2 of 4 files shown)


tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -498,6 +497,76 @@ public static void UnusedArgumentMatcherDescriptionNotUsed(IHaveNoGenericParamet
...
+        }
+
+        [Scenario]
+        public static void PassingNestedUnderscoreConstraintToAMethod(
+            IHaveOneGenericParameter fake, Exception exception)
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<IHaveOneGenericParameter>());
+
+            "When I try to configure a method of the fake with a Ignored constraint nested in an argument"
```
thomaslevesque: Ah, yes... I copied & pasted, and forgot to change the text

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -498,6 +497,76 @@ public static void UnusedArgumentMatcherDescriptionNotUsed(IHaveNoGenericParamet
...
+        }
+
+        [Scenario]
+        public static void PassingNestedThatMatchesConstraintToAMethod(
+            IHaveOneGenericParameter fake, Exception exception)
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<IHaveOneGenericParameter>());
+
+            "When I try to configure a method of the fake with a Ignored constraint nested in an argument"
```
thomaslevesque: Indeed!

thomaslevesque:
I pushed a fixup commit.

> I'm not sure I follow the very fine details for the expression walking

The visitor visits the expression by "drilling down" its nodes. If I find the `A.That.Matches(...)` node as an argument constraint, I will also find the `A.That` node later. Since I need to walk the call chain to find if something is called on `A<T>`, I keep track of the nodes I already checked, so that I don't report them again as argument constraints when these nodes are visited later.

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -106,5 +116,95 @@ private IArgumentConstraint CreateParamArrayConstraint(NewArrayExpression expres
 
             return new AggregateArgumentConstraint(result);
         }
+
+        private class ArgumentConstraintExpressionVisitor : ExpressionVisitor
+        {
+            private readonly List<Expression> constraintExpressions;
+            private readonly HashSet<Expression> visitedExpressions;
```
blairconrad: You did. Thanks! Postorder traversal. I didn't realize.

blairconrad:
> I pushed a fixup commit.

Yeah! It was nice.

> > I'm not sure I follow the very fine details for the expression walking

> The visitor visits the expression by "drilling down" its nodes. If I find the `A.That.Matches(...)` node as an argument constraint, I will also find the `A.That` node later. Since I need to walk the call chain to find if something is called on `A<T>`, I keep track of the nodes I already checked, so that I don't report them again as argument constraints when these nodes are visited later.

Yeah. It's more the bouncing between invoking on members and on methods and just on anything. I believe in it.

blairconrad:
Oh, yeah. 

1. analyzer - do you want a different issue?
2. docs

thomaslevesque:
Looks like I missed some of your comments... I'll look into it tomorrow.

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -80,8 +78,20 @@ private static object InvokeExpression(Expression expression)
             return Expression.Lambda(expression).Compile().DynamicInvoke();
         }
 
+        private static void CheckArgumentExpressionIsValid(Expression expression)
+        {
+            var visitor = new ArgumentConstraintExpressionVisitor();
+            visitor.Visit(expression);
+            if (visitor.ConstraintExpressions.Any(e => e != expression))
+            {
+                throw new InvalidOperationException(ExceptionMessages.ArgumentConstraintCannotBeNestedInArgument);
```
thomaslevesque: That was my initial intention, but then I realized that getting a sensible representation of the expression was far from trivial; we don't have access to the original source code, as in the analyzer. And anyway, should we show the expression in C#, VB, something else?

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -106,5 +116,95 @@ private IArgumentConstraint CreateParamArrayConstraint(NewArrayExpression expres
 
             return new AggregateArgumentConstraint(result);
         }
+
+        private class ArgumentConstraintExpressionVisitor : ExpressionVisitor
+        {
+            private readonly List<Expression> constraintExpressions;
```
thomaslevesque: Yes, that's an option too. We could even throw as soon as we find an invalid argument constraint. I can change it if you prefer.

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
thomaslevesque: > Postorder traversal

If you say so ;)

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
blairconrad: I'm not sure how easy it is to get a sensible representation of the expression (but I'd think that extracting out the `That`, `_`, or `Ignored` would be sufficient). As for language, if just extracting the member name isn't enough (or isn't feasible), I don't suppose getting the "CLR formatting" is an option with an expression, is it? Like we did in #1047. Perhaps one would have to compile the method to do so?

As a cheap alternative, we could augment the static exception message we have above. Maybe something like 

>An argument constraint, such as That, Ignored, or _, cannot be nested in an argument.

But if we have a `That` inside a `That`, maybe it doesn't help much.

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
blairconrad: I just thought it would leave us with slightly simpler code and would be (probably imperceptibly) faster in the (very rare) case where a user makes this error. Your choice.

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
blairconrad: I did, but I'm not so sure now. It may be someotherorder traversal, but you double visit because of the shenanigans that I didn't entirely understand - you're forcing a descent down the tree in some cases.

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
thomaslevesque: `Expression.ToString()` returns things like ``A`1.That.Matches(x => True)``, which isn't great, but is consistent with the CLR formatting used elsewhere... But when we do the same from VB.NET, we get something like this: ``() => Convert(A`1.That).Matches(a0 => Invoke(() => True))``, which is much uglier.

The cheap alternative sounds reasonable to me.

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
thomaslevesque: > you're forcing a descent down the tree in some cases.

Yes. But now that I think about it, there might be an easier way. Let me try something.

thomaslevesque: OK, now I avoid the double visit. Better?

thomaslevesque:
> analyzer - do you want a different issue?

Yes, I think we can treat it separately. I'll open an issue.

>     docs

I just pushed a new commit to mention the restriction on nested argument constraints.

blairconrad:
Thanks, @thomaslevesque. I like the docs. The examples really help illustrate a concept that's hard to describe.

Sorry for the delay. I've been busy, but also struggling to understand the code. I believe in it (feel free to squash at any time), but worry that it's hard to comprehend, and to modify.
I was debugging through to try to understand how it worked, and while I was doing that, I tried a different approach. Would you mind taking a look at it? To me, it's easier to understand, but maybe it's just me.
https://github.com/blairconrad/FakeItEasy/tree/fix-nested-arg-constraint

thomaslevesque:
Well, I kind of liked my approach, but it's true that yours is easier to understand, which is probably more important in the long run. Do you want to push it to my branch, or should I fixup my commits?

blairconrad:
> I kind of liked my approach

I know. I feel bad saying "try it this way instead", but I did find the logic hard to follow.
If you want to keep your code, perhaps you could add a comment explaining exactly 
why the `candidateExpression` ends up being the value we need. I _think_ it's because `base.VisitMethodCall` traverses the method call's arguments, but we then exit `NestedArgumentConstraintExpressionVisitor. VisitMethodCall` before `ExpressionVisitor` visits the `A<T>.That` (in `A<T>.That.Matches). But even writing this, I'm not sure.

I'll push an update to my commit (I saw a typo). If you want to take my code, I don't care whether you cherry-pick or just take the content and improve it or whatever. Commit credit isn't important. :wink:

Oh. While reading this morning, I saw some comments I'd make on your original code… I'm rushing out the door, but I'll try to get a few in, in case you want to keep yours.

thomaslevesque:
> Oh. While reading this morning, I saw some comments I'd make on your original code… I'm rushing out the door, but I'll try to get a few in, in case you want to keep yours.

Don't bother, I'm convinced by your solution. Thinking about it again, I realize I no longer understand mine, which is a pretty clear sign that it'll be hard to maintain.

blairconrad:
@thomaslevesque, thanks for the updates. I'm happy to merge after a rebase!

thomaslevesque:
> I'm happy to merge after a rebase!

Done!

blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
Thanks for the review and suggestions 👍

PR #1181, opened 2017-08-08
Include nuget path when pushing

blairconrad:
Supports #1171.


blairconrad:
(I'm so embarrassed.)

blairconrad:
Thanks, @thomaslevesque. I have high hopes this time (like the previous time, I suppose).
I fudged my API KEY and ran the commands this time. The error I get from NuGet suggests that it's only unhappy about the key.

Issue #1170, opened 2017-07-26
Release 4.1.0


blairconrad:
Can be labelled **ready** when all other issues on this milestone are closed.

- [x] ~~run code analysis in VS in *Release* mode and address all coding and style violations (send a regular PR which must be merged before continuing)~~
- [x] ensure completeness of draft [release notes](https://github.com/FakeItEasy/FakeItEasy/releases), including non-owner contributors, if any (move release notes forward from any pre-releases to the current release)
- [x] run `prepare_release.cmd x.y.z` to:
    - change "vNext" names to appropriate x.y.z version for the next GitHub Release, milestone, and release checklist (this issue)
    - create a new draft GitHub Release, milestone, and release checklist
    - create release/x.y.z branch on upstream, branching from develop
    - submit a PR to merge the release/x.y.z branch to master
- [x] once release-x.y.z has been merged, tag master and push to upstream. This will push the nupkgs to [nuget.org](https://www.nuget.org/profiles/FakeItEasy) and attach all nupkgs from the tagged build to the GitHub Release.
- [x] de-list pre-release or superseded buggy NuGet packages if present
- [x] publish the GitHub release
- [x] tweet, mentioning contributors, and post link as comment here for easy retweeting ;-)
- [x] post tweet in [Gitter](https://gitter.im/FakeItEasy/FakeItEasy)
- [x] post a link to the GitHub Release in each issue in this milestone, with thanks to contributors
- [ ] close this issue
- [ ] close the [milestone](https://github.com/FakeItEasy/FakeItEasy/milestones?state=open) for this release

blairconrad:
Hey. As an opportunity to test #1171, what do you think about doing a low-key, unpublicized alpha release? I'd rather find out that the GitHub integration isn't going to work on 4.1.0-alpha001 than on a real release.

thomaslevesque:
Sounds good 👍

blairconrad:
Ha! That didn't work. I removed the draft release and will try again later.

blairconrad:
(4th time's the charm: https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/396)

blairconrad:
Marked as **ready** since everything else in the milestone is done, and it's been a while since we had a release. What say you, owners? Ready for 4.1.0?

thomaslevesque:
> Ready for 4.1.0?

Yes, I think it's ready for release.

blairconrad:
[...]

PR #1182, opened 2017-08-08
Use new nuget API key and better server URL

blairconrad:
Supports #1171.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ environment:
     GITHUB_TOKEN:
         secure: 5dFninlYVqNF98Pk9ykQwPU2pBFeQmbMOiFG2iSkg+hhcwF3UfmnDB07lvOwYtez
     NUGET_API_KEY:
-        secure: BVzuxe9khaakJDNLAhtUM2w5G3gAkoOasCDAilstAtpLuCdFNfu3jrf0KrI5B1Bo
+        secure: ynOI8xETJ5V28b5ij2huIKbWRvqJU8QtLtFf8FLTJCWbxmcVP/OPpcF4B8jcOOKW
 
 build_script:
 - cmd: if defined APPVEYOR_SCHEDULED_BUILD (build.cmd coverity) else (build.cmd)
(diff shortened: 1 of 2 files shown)


blairconrad:
I though maybe the transfer of the API key from the deploy area to an environment variable had invalidated it somehow, so regenerated one. That still failed.
Then Iooked up the push API and saw that there was a better preferred Source setting. So I deployed used it and then deployed https://www.nuget.org/packages/FakeItEasy.Analyzer.VisualBasic/4.1.0-alpha003 from my home machine. It's now delisted.

I _really_ hope this is the last time. Sorry for the churn.

blairconrad:
In retrospect, once we decided it was okay to create an unpublicized release, I should have just run the darn script at home until it worked and then submitted a PR. Ah well. Next time.

blairconrad:
Huzzah. Thanks for putting up with my incompetence!
https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/396#L683

thomaslevesque:
Well done @blairconrad! Now releasing will be easier than ever.

thomaslevesque:
I'm worried about one thing, though; in the previous builds, the deployment script had several errors, but the build was still considered successful. It looks like the errors don't make the script fail, they're just ignored.

blairconrad:
I'm not so sure about "well done", but thanks.
I've updated the instructions in #1170.
At some point, if we become confident in the workflow, we could have the deployment also publish the GitHub release, but for now I don't mind running eyeballs over it one last time.

thomaslevesque:
> I'm not so sure about "well done", but thanks.

Well, you made it work 😉 

> At some point, if we become confident in the workflow, we could have the deployment also publish the GitHub release, but for now I don't mind running eyeballs over it one last time.

Agreed.

blairconrad:
> in the previous builds, the deployment script had several errors, but the build was still considered successful.

A good point. We can try to add robustification.
It was in my head as well. I expect we'll catch any failures due to lack of nuget packages (and the e-mails that we get with that) and also because we close the release ourselves, but there's no reason not to have the computer help us.

Simulating the various failures and catching them may be tricky, but we can take a stab. I've made #1183.

Issue #1163, opened 2017-07-07
Create diagnostic and code fix for mismatched argument constraint types


blairconrad:
From https://github.com/FakeItEasy/FakeItEasy/pull/1156#discussion_r125862320

Most type mismatches will be detected by the compiler, but if the types can be coerced, the compiler lets us down.

@blairconrad:
> Maybe we should have a diagnostic for the case where the type of the constraint is mismatched!

@thomasleveque:
> Oh, I had never thought of that. Currently `A<int>._` won't match if the actual parameter type is `long`. Yes, we should totally add a diagnostic for this.

An alternative would be to make the argument constraints more forgiving, but that seems like a tonne of work.

thomaslevesque:
To clarify, here's an example:

```csharp
public interface IFoo
{
    string Bar(long x);
}

...

var foo = A.Fake<IFoo>();
A.CallTo(() => foo.Bar(A<int>._)).Returns("hello");
```
It compiles, because `int` is implicitly converted to `long`, but no call will ever match the constraint because it's of the wrong type.

blairconrad:
Self-assigning. I might as well get in on this analyzer action.

thomaslevesque:
> Self-assigning. I might as well get in on this analyzer action.

👍 

Rather than creating a completely new analyzer, we should probably consider adding the new diagnostic to the existing `ArgumentConstraintNullabilityMismatchAnalyzer`, since most of the code will be identical (a single analyzer can produce multiple diagnostics). Of course, it will have to be merged first... I'll make the necessary changes now so you can merge it.

blairconrad:
There's no rush at all. I already have a small proof of concept built off your branch.
Also, I may not jump onto the meat of the issue right away. I'm being greedy by taking it, but there's no particular rush for this. No big deal if it waits until after 4.0, IMO.

thomaslevesque:
> No big deal if it waits until after 4.0, IMO.

Sure, it's not a breaking change anyway (well, it might cause warnings on existing broken code, but that doesn't really count as breaking)

blairconrad:
I'd be inclined to make the diagnostic an error.

thomaslevesque:
> I'd be inclined to make the diagnostic an error.

Yes, that makes sense.

thomaslevesque:
Closed by #1192

blairconrad:
[...]

PR #1192, opened 2017-08-25
Add diagnostic FakeItEasy0005

blairconrad:
Contributes to #1163 

[WIP] because

1. no code fix
2. @thomaslevesque had suggested adding the diagnostic to `ArgumentConstraintNullabilityMismatchAnalyzer` (which would have to be renamed). I didn't  yet do this. For now I have:
    - diagnostic tests. Critique welcome.
    - a separate analyzer. Much of the code is shared with `ArgumentConstraintNullabilityMismatchAnalyzer`. I'll work toward collapsing the classes as time allows.


src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchAnalyzer.cs
```diff
@@ -0,0 +1,137 @@
...
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; } =
+            ImmutableArray.Create(DiagnosticDefinitions.ArgumentConstraintTypeMismatch);
+
+        protected override bool IsSupportedArgumentConstraintProperty(string fullName) =>
+            SupportedArgumentConstraintProperties.Contains(fullName);
+
+        protected override void AnalyzeArgumentConstraintCore(SyntaxNodeAnalysisContext context, SyntaxNode completeConstraint)
+        {
+            var constraintType = context.SemanticModel.GetTypeInfo(completeConstraint).Type as INamedTypeSymbol;
+            if (constraintType == null || !constraintType.IsValueType)
```
thomaslevesque: Why should it be limited to value types only? I can think of cases where the problem would occur for reference types as well. For instance, a constraint of type `Bar` is passed to a method that accepts a `Foo`, and `Bar` isn't a subclass of `Foo`. Of course, it would be a compilation error anyway... unless there's an implicit conversion. But we would never match the constraint, because the method would actually receive a `Foo`, not a `Bar`.

blairconrad: I'd been excluding it because of the compilation error. Do you think it's worthwhile duplicating the errors when the compiler will already be sad?

I'd not thought of an implicit conversion. This is an excellent point. I will pursue.

In the meantime, I've merged the analyzers (and tests, because once we had one analyzer with 2 diagnostics, many of the tests fought), but am lagging on the VB tests. Updates to trickle in over the next few days.

thomaslevesque: > Do you think it's worthwhile duplicating the errors when the compiler will already be sad?

Probably not

> In the meantime, I've merged the analyzers (and tests, because once we had one analyzer with 2 diagnostics, many of the tests fought), but am lagging on the VB tests. Updates to trickle in over the next few days.

👍

blairconrad:
Whoops. Accidentally pushed some changes to this branch. Probably what I would've pushed soon anyhow, but the commit message isn't nice.
Anyhow, I renamed the analyzer and test classes and am handling implicit conversions properly, I think.

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchAnalyzer.cs
```diff
@@ -61,23 +63,62 @@ protected override void AnalyzeArgumentConstraintCore(SyntaxNodeAnalysisContext
...
+                {
+                    return true;
+                }
+
+                currentType = currentType.BaseType;
+            }
+            while (currentType != null);
+
+            return false;
+        }
```
thomaslevesque: Nicely done! Another possible approach would be to use [`ClassifyConversion`](http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/CSharpExtensions.cs,1068)

blairconrad: I gave it a go. It was a lot of guesswork, and I'm not sure the result made it worthwhile, but I'll let you judge.

thomaslevesque: I like it!

blairconrad:
Okay, so all I need now is a code fix. And possibly cleanup. The nesting is getting pretty bad in the analyzer.
I expect there will be much squashing of commits as well.

blairconrad:
Oh! I just realized that I haven't actually tried the code fix provider in the IDE!

blairconrad:
> Oh! I just realized that I haven't actually tried the code fix provider in the IDE!

I'm glad I checked.

```
System.Collections.Generic.KeyNotFoundException : The given key was not present in the dictionary.
   at System.Collections.Immutable.ImmutableDictionary`2.get_Item(TKey key)
   at async FakeItEasy.Analyzer.ArgumentConstraintTypeMismatchCodeFixProvider.ChangeConstraintTypeAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetChangedSolutionAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.ComputeOperationsAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.ComputePreviewOperationsAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.CodeActions.CodeAction.GetPreviewOperationsAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedAction.GetPreviewResultAsync(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.Editor.Implementation.Suggestions.SuggestedActionWithNestedFlavors.<>c__DisplayClass11_0.<GetPreviewAsync>b__0(<Unknown Parameters>)
   at async Microsoft.CodeAnalysis.Extensions.IExtensionManagerExtensions.PerformFunctionAsync[T](<Unknown Parameters>)
```

I think it's coming from 

```c#
var parameterTypeNode = SyntaxFactory.ParseName(diagnostic.Properties["parameterType"]);
```

I'll investigate later.

blairconrad:
Thanks for the help, @thomaslevesque. Pushed the raw changes. If you like, I'll have to squash.

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchCodeFixProvider.cs
```diff
@@ -41,27 +44,38 @@ public override Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                        ct => this.MakeNotNullConstraintAsync(context, diagnostic, ct)),
+                    diagnostic);
+            }
+            else if (diagnostic.Descriptor.Id == DiagnosticDefinitions.ArgumentConstraintTypeMismatch.Id)
+            {
+                context.RegisterCodeFix(
+                    CodeAction.Create(
+                        ChangeConstraintTypeCodeFixTitle,
+                        ct => this.ChangeConstraintTypeAsync(context, diagnostic, ct)),
+                    diagnostic);
```
thomaslevesque: I don't think we should suggest a code fix for constraints other than `A<T>._`/`A<T>.Ignored`, because changing the parameter type can cause the predicate in `A<T>.That.Matches(...)` to become invalid. This is the same reason we don't suggest a code fix (*) for argument constraint nullability mismatch on `A<T>.That.Matches(...)`.

(*) In fact we don't even report a warning in this case, because we can't provide a code fix, and it might be intentional.

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchAnalyzer.cs
```diff
@@ -0,0 +1,172 @@
...
+                        context.ReportDiagnostic(diagnostic);
+                    }
+                }
+                else
+                {
+#if CSHARP
+                    if (!conversionType.IsImplicit || !conversionType.IsReference)
+#elif VISUAL_BASIC
+                    if (!conversionType.IsWidening || !conversionType.IsReference)
+#endif
```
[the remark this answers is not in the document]
thomaslevesque: I do!

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchCodeFixProvider.cs
thomaslevesque: Yes, I mean non-compilable. For instance if you change the type from int to long, the predicate could work on an int but not on a long (e.g. if it calls a method that takes an int)

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchCodeFixProvider.cs
```diff
@@ -137,5 +151,49 @@ private static MemberAccessExpressionSyntax SimpleMemberAccess(ExpressionSyntax
...
+                var parameterType = diagnostic.Properties[ArgumentConstraintTypeMismatchAnalyzer.ParameterTypeKey];
+                var parameterTypeNode = SyntaxFactory.ParseName(parameterType);
+                var newRoot = root.ReplaceNode(constraintType, parameterTypeNode);
+                return context.Document.WithSyntaxRoot(newRoot);
+            }
+
+            return context.Document;
+        }
+
+        private TypeSyntax GetConstraintType(CodeFixContext context, Diagnostic diagnostic, SyntaxNode root)
```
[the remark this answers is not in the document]
thomaslevesque: No need, it's fine

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchCodeFixProvider.cs
```diff
@@ -137,5 +151,49 @@ private static MemberAccessExpressionSyntax SimpleMemberAccess(ExpressionSyntax
...
+            var constraintNode = GetConstraintNode(diagnostic, root);
+            var memberAccessNode = constraintNode as MemberAccessExpressionSyntax;
+            if (memberAccessNode != null)
+            {
+                var expressionNode = memberAccessNode.Expression as GenericNameSyntax;
+                return expressionNode?.TypeArgumentList.Arguments.FirstOrDefault();
+            }
+
+            var invocationNode = constraintNode as InvocationExpressionSyntax;
+            if (invocationNode != null)
```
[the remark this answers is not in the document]
thomaslevesque: The new version looks much better 👍

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchCodeFixProvider.cs
thomaslevesque: Also, it's not just numeric types. You even added a test for an implicit conversion ;)

blairconrad: Right! The implicit types.

Nevertheless, I still think we should report the warning and suggest the code fix for type mismatches (not nullability mismatches) in `A<T>.That.Matches(…)`. We didn't for nullability mismatches because it was possible for the complex `A<T>.That.Matches(…)` constraints to do the right thing. In the case of a type mismatch, nothing that can be put inside `Matches(…)` will ever work, so I think we should definitely report the error. And I'd rather suggest a code fix that left produced uncompiling code in this situation. There's something terribly wrong with the test, and the user should fix it. If the type of the parameter is correct, then the type of the constraint _must_ change, and if that means extensive updates to the test code, so be it.

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchCodeFixProvider.cs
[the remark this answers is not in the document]
blairconrad: I think so too.

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchCodeFixProvider.cs
thomaslevesque: OK, you convinced me ;)

thomaslevesque:
I think you addressed all my comments. I'll merge after you squash.

thomaslevesque:
Thanks @blairconrad, nice work 👍

blairconrad:
Thank you. You were instrumental.

PR #1195, opened 2017-08-31
Better description for IsSameSequenceAs constraint

### LINKED ISSUE #1193: Enhance formatting of received array (or enumerable?) arguments in "unmatched call" output
I'm using FIE 2.3.0

I notice it's a little hard to figure out why my test failed with the current error messages. I am doing an assertion like this:

```
            A.CallTo(() => fake.CreateNewPortRules(
                A<IList<int>>.That.IsSameSequenceAs(new[] { 987, 123 }), 
                true))
            .MustHaveHappened();
```

And the output of the failed test is this:

```
Message: FakeItEasy.ExpectationException : 

    RedisLaunch.PortManager.CreateNewPortRules(<specified sequence>, True)
    1: RedisLaunch.PortManager.SSLPorts
    2: RedisLaunch.PortManager.CreateNewPortRules(ports: System.Int32[], allow: True)
    3: RedisLaunch.PortManager.ClearTextPorts
    4: RedisLaunch.PortManager.CreateNewPortRules(ports: System.Int32[], allow: True)
    5: RedisLaunch.PortManager.CreateStartupFirewallSettings(clearTextOpen: True, sslOpen: True)
```

thomaslevesque:
Fixes #1193 

It seems to work pretty well, but I'm worried about one thing: when formatting argument values in the list of calls, what should happen for nested collections? My current implementation will recursively format the collections, but it will probably hurt the readability of the output...

Also, I noticed a few inconsistencies:
- When we write constraint descriptions, we use `IOutputWriter.WriteArgumentValue` to format values, but when we write the list of calls, we use `ArgumentValueFormatter`.
- Anyway, shouldn't `DefaultOutputWriter.WriteArgumentValue` use `ArgumentValueFormatter`, in order to take custom argument value formatters into account?
- Null values are formatted inconsistently between `DefaultOutputWriter` (`NULL`) and `ArgumentValueFormatter` (`<NULL>`).

### DIFF
--- a/src/FakeItEasy/OutputWriterExtensions.cs
+++ b/src/FakeItEasy/OutputWriterExtensions.cs
@@ -1,6 +1,8 @@
 namespace FakeItEasy
 {
     using System;
+    using System.Collections;
+    using System.Linq;
 
     /// <summary>
     /// Provides extensions for <see cref="IOutputWriter"/>.
@@ -47,6 +49,66 @@ namespace FakeItEasy
             Guard.AgainstNull(value, nameof(value));
 
             writer.Write(value.ToString());
+            return writer;
+        }
+
+        /// <summary>
+        /// Formats the specified argument values as strings and writes them to the output.
+        /// </summary>
+        /// <param name="writer">The writer to write to.</param>
+        /// <param name="values">The values to write to the writer.</param>
+        /// <param name="skipFormatting">Specify that argument values should not be formatted.</param>
+        /// <returns>The writer.</returns>
+        internal static IOutputWriter WriteArgumentValues(this IOutputWriter writer, IEnumerable values, bool skipFormatting = false)
+        {
+            Guard.AgainstNull(writer, nameof(writer));
+            Guard.AgainstNull(values, nameof(values));
+
+            var list = values.AsList();
+            if (list.Count <= 5)
+            {
+                writer.WriteArgumentValuesImpl(list, skipFormatting);
+            }
+            else
+            {
+                writer.WriteArgumentValuesImpl(list.Take(2), skipFormatting);
+                int remainingCount = list.Count - 4;
+                writer.Write($", … ({remainingCount} more elements) …, ");
+                writer.WriteArgumentValuesImpl(list.Skip(list.Count - 2), skipFormatting);
+            }
+
+            return writer;
+        }
+
+        /// <summary>
+        /// Formats the specified argument values as strings and writes them to the output.
+        /// </summary>
+        /// <param name="writer">The writer to write to.</param>
+        /// <param name="values">The values to write to the writer.</param>
+        /// <param name="skipFormatting">Specify that argument values should not be formatted.</param>
+        /// <returns>The writer.</returns>
+        private static IOutputWriter WriteArgumentValuesImpl(this IOutputWriter writer, IEnumerable values, bool skipFormatting = false)
+        {
+            bool first = true;
+            foreach (var value in values)
+            {
+                if (!first)
+                {
+                    writer.Write(", ");
+                }
+
+                if (skipFormatting)
+                {
+                    writer.Write(value);
+                }
+                else
+                {
+                    writer.WriteArgumentValue(value);
+                }
+
+                first = false;
+            }
+
             return writer;
         }
     }
--- a/tests/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
+++ b/tests/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
@@ -5,7 +5,6 @@ namespace FakeItEasy.Tests.Core
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
-    using System.Reflection;
     using FakeItEasy.Core;
     using FakeItEasy.Tests.TestHelpers;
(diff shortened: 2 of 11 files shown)


src/FakeItEasy/OutputWriterExtensions.cs
```diff
@@ -49,5 +51,72 @@ public static IOutputWriter Write(this IOutputWriter writer, object value)
...
+            var list = values.AsList();
+            if (list.Count <= maxNumber)
+            {
+                writer.WriteArgumentValues(list, separator, skipFormatting);
+            }
+            else
+            {
+                writer.WriteArgumentValues(list.Take(maxNumber - 1), separator, skipFormatting);
+                int remainingCount = list.Count - maxNumber + 1;
+                writer.Write($", ... ({remainingCount} more elements)");
```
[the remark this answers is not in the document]
thomaslevesque: > Should `,` not be the passed-in separator?

Yes. Although I doubt we'll ever use anything other than `,`... we could probably remove the parameter and hard-code it instead.

> Perhaps we could print all the elements if less than six, otherwise the first two and last two?

👍

tests/FakeItEasy.Tests/Core/ArgumentValueFormatterTests.cs
```diff
@@ -153,7 +153,7 @@ public void Built_in_formatters_should_have_lower_than_default_priority()
             // Arrange
             var allArgumentValueFormatters = typeof(A).GetTypeInformation().Assembly.GetTypes()
                 .Where(t => t.CanBeInstantiatedAs(typeof(IArgumentValueFormatter)))
-                .Select(Activator.CreateInstance)
+                .Select(t => Activator.CreateInstance(t, GetDefaultArgumentsForConstructor(t)))
```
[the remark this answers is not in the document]
thomaslevesque: Good point. It would certainly make things simpler.

src/FakeItEasy/OutputWriterExtensions.cs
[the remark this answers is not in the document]
blairconrad: I think removing the parameter is a good plan.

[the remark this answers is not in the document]
thomaslevesque: While I was at it, I also removed the `maxNumber` parameter.

thomaslevesque:
I made the requested changes, but I gave up on making fixup commits, because it was too difficult to squash cleanly...

src/FakeItEasy/OutputWriterExtensions.cs
```diff
@@ -49,5 +51,70 @@ public static IOutputWriter Write(this IOutputWriter writer, object value)
             writer.Write(value.ToString());
             return writer;
         }
+
+        /// <summary>
+        /// Formats the specified argument values as strings and writes
+        /// them to the output, separated by the specified separator.
```
blairconrad: always a comma

src/FakeItEasy/OutputWriterExtensions.cs
```diff
@@ -49,5 +51,70 @@ public static IOutputWriter Write(this IOutputWriter writer, object value)
...
+                writer.Write($", … ({remainingCount} more elements) …, ");
+                writer.WriteArgumentValuesImpl(list.Skip(list.Count - 2), skipFormatting);
+            }
+
+            return writer;
+        }
+
+        /// <summary>
+        /// Formats the specified argument values as strings and writes
+        /// them to the output, separated by the specified separator.
```
blairconrad: comma

src/FakeItEasy/OutputWriterExtensions.cs
```diff
@@ -49,5 +51,70 @@ public static IOutputWriter Write(this IOutputWriter writer, object value)
...
+        /// </summary>
+        /// <param name="writer">The writer to write to.</param>
+        /// <param name="values">The values to write to the writer.</param>
+        /// <param name="skipFormatting">Specify that argument values should not be formatted.</param>
+        /// <returns>The writer.</returns>
+        private static IOutputWriter WriteArgumentValuesImpl(this IOutputWriter writer, IEnumerable values, bool skipFormatting = false)
+        {
+            Guard.AgainstNull(writer, nameof(writer));
+            Guard.AgainstNull(values, nameof(values));
```
blairconrad: Are the guards necessary for a private method (I often wonder the same thing about the internal methods as well).

src/FakeItEasy/OutputWriterExtensions.cs
thomaslevesque: oops

src/FakeItEasy/OutputWriterExtensions.cs
thomaslevesque: No, probably not. I'll remove them.
The way I see it, the guards are necessary for public methods to ensure the caller is using the method correctly. Internally, we're supposed to know how to use it, so if we pass a null, it's our own problem, and there's no point in throwing a specific exception. I haven't always applied this rule, though...

thomaslevesque:
I just amended the last commit, since the changes were trivial.

thomaslevesque:
Squashed

blairconrad:
Thanks, @thomaslevesque!

Issue #1144, opened 2017-06-19
Find replacement for FxCop (VS "Code Analysis")


thomaslevesque:
The current Visual Studio "Code Analysis" feature
- doesn't support SDK-style projects
- doesn't support .NET Standard & .NET Core assemblies

Apparently there won't be a short-term fix to make FxCop support these scenarios.

~~So, right now it's only working for the two VB projects, which haven't been migrated to SDK-style projects.~~ (they have now be migrated as well)

Checking and fixing code analysis warnings is part of the release checklist, and it should probably stay this way, so we need a solution.

thomaslevesque:
Possible solution: [Microsoft.CodeAnalysis.FxCopAnalyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/) NuGet package

This package seems to be a mess. It's actually a meta-package that groups several code analysis packages, some of which don't seem to work (e.g. the analyzer from System.Runtime.Analyzers fails to load due to a strong-naming issue). It's unclear which rules are contained in which package.

thomaslevesque:
Possible solution: [Microsoft Code Analysis 2017] VS extension.

From the extension's page:

> - 100+ live diagnostics for API design, performance, security, and best practices for C# and Visual Basic.
> - Quick fixes for applicable diagnostics (using Ctrl+.).
> - Diagnostics appear in the editor, Error List, and scroll bar.
> - All diagnostics are suppressible. 

Pros:
- immediate feedback during development and compilation (no need to explicitly run analysis)

Cons:
- no easy way to run analysis during CI build, since the CI environment won't have the extension
- supports only a subset of FxCop rules

thomaslevesque:
In fact, both the extension and the NuGet package are based on the same analyzers: https://github.com/dotnet/roslyn-analyzers

thomaslevesque:
I installed the extension, but I haven't been able to get it to report a single warning...

thomaslevesque:
> I installed the extension, but I haven't been able to get it to report a single warning...

This is because it requires VS 2017.3, which hasn't been released yet... See https://github.com/dotnet/roslyn-analyzers/issues/1238

thomaslevesque:
Well, it was supposed to work with VS2017 15.3, but it doesn't... I'm still unable to get any warning.

thomaslevesque:
[...]

PR #1198, opened 2017-09-13
Use Directory.Build.props for common build properties

thomaslevesque:
Using `Directory.Build.props` files as explained [here](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build), we can factorize a lot of project properties and package references.

Benefits:
- even smaller project files
- no repetition of common properties and package references
- dependencies versions are specified in one place, which ensures consistency

Drawbacks:
- Visual Studio project properties designer doesn't seem to handle inherited properties very well... it can read them just fine, but editing them doesn't work (no error, but the changes are not persisted). Not really worse than before, since it already didn't support multiple target frameworks...
- Updating a package in NuGet package manager doesn't modify the Directory.Build.props file, but the csproj file.

Marked as WIP because we need to discuss if we want to use this.

### DIFF
--- a/samples/Directory.Build.props
+++ b/samples/Directory.Build.props
@@ -1,0 +1,9 @@
+<Project>
+
+  <Import Project="../Directory.Build.props" />
+
+  <PropertyGroup>
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)FakeItEasy.Examples.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+
+</Project>
--- a/src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.CSharp.csproj
+++ b/src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.CSharp.csproj
@@ -8,16 +8,11 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <FileAlignment>512</FileAlignment>
     <SignAssembly>true</SignAssembly>
-    <AssemblyOriginatorKeyFile>..\..\FakeItEasy.snk</AssemblyOriginatorKeyFile>
-    <CodeAnalysisRuleSet>..\..\FakeItEasy.ruleset</CodeAnalysisRuleSet>
     <DefineConstants>$(DefineConstants);CSHARP</DefineConstants>
     <PackageTargetFallback>portable45-net45+win8</PackageTargetFallback>
     <BaseIntermediateOutputPath>objcs\</BaseIntermediateOutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
-
-  <Import Project="../packages.targets" />
-  <Import Project="packages.analyzer.targets" />
 
   <PropertyGroup Label="Package properties">
     <Title>FakeItEasy.Analyzer.CSharp</Title>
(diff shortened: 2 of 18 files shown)


samples/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+
+  <Import Project="../Directory.Build.props" />
```
thomaslevesque: MSBuild scans directories upwards from the project file, and stop as soon as it encounters a DIrectory.Build.props file. To also import the Directory.Build.props file from the parent directory, we have to do it explicitly. Note that the approach mentioned [here](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build#use-case-multi-level-merging) doesn't seem to work; this one is simpler anyway...

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.CSharp.csproj
```diff
@@ -9,16 +9,12 @@
...
     <AssemblyOriginatorKeyFile>..\..\FakeItEasy.snk</AssemblyOriginatorKeyFile>
-    <CodeAnalysisRuleSet>..\..\FakeItEasy.ruleset</CodeAnalysisRuleSet>
     <DefineConstants>$(DefineConstants);CSHARP</DefineConstants>
     <PackageTargetFallback>portable45-net45+win8</PackageTargetFallback>
     <BaseIntermediateOutputPath>objcs\</BaseIntermediateOutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
 
-  <Import Project="../packages.targets" />
-  <Import Project="packages.analyzer.targets" />
```
thomaslevesque: By renaming these files Directory.Build.props, they're now implicitly imported.

blairconrad:
Thanks, @thomaslevesque. A very shallow think about this doesn't make me concerned. I think the drawbacks you mention are all "livewithable". Sadly, I'm a little busy with work and some home stuff right now (another reason it's taking me weeks to finish my in-flight analyzer/codefix issue). Will take a deeper look as soon as I can.

tests/FakeItEasy.Specs/FakeItEasy.Specs.csproj
```diff
@@ -3,19 +3,8 @@
   <PropertyGroup>
     <TargetFrameworks>net461;netcoreapp1.0</TargetFrameworks>
     <AssemblyName>FakeItEasy.Specs</AssemblyName>
-    <CodeAnalysisRuleSet>FakeItEasy.Specs.ruleset</CodeAnalysisRuleSet>
```
blairconrad: I don't see where this is included in the new files.

thomaslevesque: Oops! It isn't. I'll restore this line, since this ruleset isn't included in any other project.
I'm surprised that I didn't get any warning without the ruleset, though... I'll check again.

thomaslevesque: Ah, of course, it's because of #1144...

thomaslevesque: I amended the commit

blairconrad: Lovely, M. Levesque.

thomaslevesque:
> Did you want to squash, or shall I merge as is?

I just squashed

blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
Thanks for the merge!

blairconrad:
This change has been released as part of [FakeItEasy 4.1.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/4.1.0).

PR #1204, opened 2017-09-24
Fix FakeItEasy0003 false positive

### LINKED ISSUE #1202: False positive for FakeItEasy0003 (argument constraint not valid outside call specification)
This code incorrectly triggers the FakeItEasy0003 diagnostic:

            A.CallTo(() => foo.Foo(A<long>.That.Matches(x => Test(x)))).DoesNothing();

> Argument constraint 'A<long>.That.Matches(x => Test(x))' is not valid outside a call specification.

thomaslevesque:
Fixes #1202

### DIFF
--- a/src/FakeItEasy.Analyzer/SymbolHelpers.cs
+++ b/src/FakeItEasy.Analyzer/SymbolHelpers.cs
@@ -1,5 +1,6 @@
-﻿namespace FakeItEasy.Analyzer
+namespace FakeItEasy.Analyzer
 {
+    using System.Linq;
     using Microsoft.CodeAnalysis;
 #if CSHARP
     using Microsoft.CodeAnalysis.CSharp;
@@ -18,10 +19,18 @@
         /// </summary>
         /// <param name="call">The invocation expression that the symbol is required.</param>
         /// <param name="context">Current context.</param>
+        /// <param name="useFirstCandidateIfNotResolved">Use the first candidate if symbol is not resolved.</param>
         /// <returns>The symbol for the invocation call.</returns>
-        internal static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax call, SyntaxNodeAnalysisContext context)
+        internal static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax call, SyntaxNodeAnalysisContext context, bool useFirstCandidateIfNotResolved = false)
         {
-            var method = context.SemanticModel.GetSymbolInfo(call).Symbol as IMethodSymbol;
+            var symbolInfo = context.SemanticModel.GetSymbolInfo(call);
+            var symbol = symbolInfo.Symbol;
+            if (symbol == null && useFirstCandidateIfNotResolved)
+            {
+                symbol = symbolInfo.CandidateSymbols.FirstOrDefault();
+            }
+
+            var method = symbol as IMethodSymbol;
 
             // In Roslyn for VB.NET, some generic extension methods are "reduced", i.e. WhereConfigurationExtensions.Where<T>
             // becomes WhereConfigurationExtensions.Where, as if it were non-generic. We need the non-reduced version.
--- a/tests/FakeItEasy.Analyzer.CSharp.Tests/UnusedReturnValueAnalyzerTests.cs
+++ b/tests/FakeItEasy.Analyzer.CSharp.Tests/UnusedReturnValueAnalyzerTests.cs
@@ -1,4 +1,4 @@
-﻿namespace FakeItEasy.Analyzer.CSharp.Tests
+namespace FakeItEasy.Analyzer.CSharp.Tests
 {
     using System.Diagnostics.CodeAnalysis;
     using FakeItEasy.Analyzer.Tests.Helpers;
@@ -359,7 +359,7 @@ namespace TheNamespace
         void Test()
         {
             var fake = A.Fake<IFoo>();
-            A.CallTo(fake).Where(call => true, output => new object());
+            A.CallTo(fake).Where(call => true, null);
         }
     }
     interface IFoo { int Bar(); }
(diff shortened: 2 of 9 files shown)


src/FakeItEasy.Analyzer/SymbolHelpers.cs
```diff
@@ -18,10 +19,18 @@ internal static class SymbolHelpers
...
-        internal static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax call, SyntaxNodeAnalysisContext context)
+        internal static IMethodSymbol GetCalledMethodSymbol(InvocationExpressionSyntax call, SyntaxNodeAnalysisContext context, bool useFirstCandidateIfNotResolved = false)
         {
-            var method = context.SemanticModel.GetSymbolInfo(call).Symbol as IMethodSymbol;
+            var symbolInfo = context.SemanticModel.GetSymbolInfo(call);
+            var symbol = symbolInfo.Symbol;
+            if (symbol == null && useFirstCandidateIfNotResolved)
+            {
+                symbol = symbolInfo.CandidateSymbols.FirstOrDefault();
+            }
```
thomaslevesque: `GetSymbolInfo` fails to resolve the method if the parameters are invalid, which prevented an invalid call to `A.CallTo(...)` from being seen as a method supporting argument constraints.
Fortunately `GetSymbolInfo` suggests possible candidates, so we use that instead.

tests/FakeItEasy.Analyzer.CSharp.Tests/UnusedReturnValueAnalyzerTests.cs
```diff
@@ -359,7 +359,7 @@ class TheClass
         void Test()
         {
             var fake = A.Fake<IFoo>();
-            A.CallTo(fake).Where(call => true, output => new object());
+            A.CallTo(fake).Where(call => true, null);
```
thomaslevesque: For consistency with the corresponding VB test that had to be changed after enabling `Option Strict`

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -45,8 +45,9 @@ public abstract partial class DiagnosticVerifier
         /// </summary>
         /// <param name="analyzer">The analyzer to run on the documents.</param>
         /// <param name="documents">The Documents that the analyzer will be run on.</param>
+        /// <param name="allowCompilationErrors">Allow compiler errors.</param>
         /// <returns>An IEnumerable of Diagnostics that surfaced in the source code, sorted by Location.</returns>
-        protected static Diagnostic[] GetSortedDiagnosticsFromDocuments(DiagnosticAnalyzer analyzer, Document[] documents)
+        protected static Diagnostic[] GetSortedDiagnosticsFromDocuments(DiagnosticAnalyzer analyzer, Document[] documents, bool allowCompilationErrors)
```
thomaslevesque: A compilation error in the analyzed code always caused the test to fail, making it impossible to verify diagnostics in invalid code. This lets us skip the assertion of compilation success.

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -175,7 +179,7 @@ private static Project CreateProject(string[] sources, string language = Languag
 #elif VISUAL_BASIC
                     .WithProjectCompilationOptions(
                         projectId,
-                        new VisualBasicCompilationOptions(OutputKind.DynamicallyLinkedLibrary))
+                        new VisualBasicCompilationOptions(OutputKind.DynamicallyLinkedLibrary, optionStrict: OptionStrict.On))
```
thomaslevesque: Should avoid a bit of head scratching in VB analyzer tests... For instance, with `Option Strict` off, VB was perfectly happy to silently convert `Integer` to `Byte` 😕

tests/FakeItEasy.Analyzer.Tests.Shared/Helpers/DiagnosticVerifier.cs
```diff
@@ -24,7 +24,12 @@ public abstract partial class DiagnosticVerifier
         /// <param name="expected"> DiagnosticResults that should appear after the analyzer is run on the source.</param>
         internal void VerifyCSharpDiagnostic(string source, params DiagnosticResult[] expected)
         {
-            this.VerifyDiagnostics(new[] { source }, LanguageNames.CSharp, this.GetCSharpDiagnosticAnalyzer(), expected);
+            this.VerifyDiagnostics(new[] { source }, LanguageNames.CSharp, this.GetCSharpDiagnosticAnalyzer(), expected, false);
+        }
+
+        internal void VerifyCSharpDiagnosticWithCompilationErrors(string source, params DiagnosticResult[] expected)
```
thomaslevesque: A bit ugly, but I couldn't add an optional parameter because of the `params`

tests/FakeItEasy.Analyzer.VisualBasic.Tests/UnusedReturnValueAnalyzerTests.cs
```diff
@@ -351,7 +351,7 @@ Namespace TheNamespace
     Class TheClass
         Sub Test()
             Dim foo = New Fake(Of IFoo)()
-            A.CallTo(foo).Where(Function() True, New Object())
```
thomaslevesque: Had to fix this after enabling `Option Strict`. Yes, VB is perfectly happy to accept `New Object()` where an `Action<IOutputWriter>` is expected... How anyone could think it would be a good idea is beyond me 😕

blairconrad:
Very nice, sir. Thanks.

thomaslevesque:
Thanks for the merge!

PR #1206, opened 2017-09-28
Use MSBuild binary log

### LINKED ISSUE #1205: Use MSBuild binary logger
https://github.com/Microsoft/msbuild/wiki/Binary-Log

The binary log has many advantages:
- smaller log file (a bit smaller than the detailed log we currently have, *much* smaller than the diagnostic log)
- contains even more information than the diagnostic log
- structured
- can include projects and their imports, so we can easily see exactly  what was being built, including .targets files from the build server
- can be opened with [MSBuild Structured Log Viewer](https://github.com/KirillOsenkov/MSBuildStructuredLog)

Downside: the file isn't human-readable. But the current text log is pretty hard to read too, so it's not a big loss... MSBuild Structured Log Viewer is much more convenient IMO.

thomaslevesque:
Fixes #1205

### DIFF
--- a/build.csx
+++ b/build.csx
@@ -245,7 +245,7 @@ public void RunMsBuild(string target)
     var packagesDirectoryOption = string.IsNullOrEmpty(packagesDirectory) ? "" : $"/p:NuGetPackagesDirectory={packagesDirectory}";
     Cmd(
         msBuild,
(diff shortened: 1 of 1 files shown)


blairconrad:
Thanks, @thomaslevesque!

PR #1207, opened 2017-09-28
Make value formatting more consistent

### LINKED ISSUE #1196: Improve consistency of argument value formatting
As mentioned in https://github.com/FakeItEasy/FakeItEasy/pull/1195#issue-254495781

We have (at least) 3 different ways to format argument values:

- In `That` constraints, we use `DefaultOutputWriter.WriteArgumentValue`, which:
    - formats null values as `NULL`
    - formats empty strings as `string.Empty`
    - formats everything else using `ToString()`
(The whole constraint description is enclosed in `< >`.)
- In `EqualityArgumentConstraint`, the formatting for the expected value is as follows:
    - formats null values as `<NULL>`
    - formats empty strings as `""`
    - formats everything else using `ToString()`
- In the list of calls in the error message for failed assertions, we use `DefaultFakeObjectFormatter`, which in turns uses `ArgumentValueFormatter`, which:
    - formats null values as `<NULL>`
    - formats empty strings as `string.Empty`
    - formats everything else using user-defined formatters or `ToString()` if there's no formatter for the type
- In `CastleInvocationCallAdapter.ToString()` (not sure where exactly it is used), we use yet another approach which:
    - formats null as `<NULL>`
    - formats empty strings as `<string.Empty>`
    - formats everything else using `ToString()`

Ideally, we should always use `ArgumentValueFormatter` so that we take user-defined formatters into account.

thomaslevesque:
Fixes #1196

### DIFF
--- a/src/FakeItEasy/Core/TypeCatalogue.cs
+++ b/src/FakeItEasy/Core/TypeCatalogue.cs
@@ -168,7 +168,7 @@ namespace FakeItEasy.Core
 
         private static void WarnFailedToGetSomeTypes(Assembly assembly, ReflectionTypeLoadException ex)
         {
-            var writer = new DefaultOutputWriter(Console.Write);
+            var writer = CreateConsoleWriter();
             Write(
                 writer,
                 ex,
@@ -192,7 +192,7 @@ namespace FakeItEasy.Core
 
         private static void Write(Exception ex, string message)
         {
-            var writer = new DefaultOutputWriter(Console.Write);
+            var writer = CreateConsoleWriter();
             Write(writer, ex, message);
         }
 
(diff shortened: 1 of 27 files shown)


src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -168,7 +168,7 @@ private static void WarnFailedToGetTypes(Assembly assembly, Exception ex)
 
         private static void WarnFailedToGetSomeTypes(Assembly assembly, ReflectionTypeLoadException ex)
         {
-            var writer = new DefaultOutputWriter(Console.Write);
+            var writer = new DefaultOutputWriter(Console.Write, null);
```
thomaslevesque: At this point we can't use `ArgumentValueFormatter` because we haven't loaded the custom formatters yet. But it's safe to pass null because we know that the writer won't be used to format argument values.

thomaslevesque:
I made the requested change (amended the commit), and rebased onto develop

PR #1211, opened 2017-10-02
Pin assembly version to Major.0.0.0

### LINKED ISSUE #1209: Consider changing AssemblyVersion to only include major version
See #1208, https://github.com/castleproject/Core/issues/288 for details

In short, best practice is that FakeItEasy 4.x.y should always have 4.0.0.0 as the `AssemblyVersion`, in order to avoid assembly binding issues. `AssemblyFileVersion ` would still follow the package version (4.x.y.0).

thomaslevesque:
Fixes #1209

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,3 +1,4 @@
 ignore:
   commits-before: 2017-01-01T00:00:00
 legacy-semver-padding: 3
+assembly-versioning-scheme: Major


thomaslevesque:
I was afraid it would be difficult to make this change, since the VersionInfo.cs file is generated by GitVersion, but it turns out there's a setting for this exact use case :)

https://gitversion.readthedocs.io/en/latest/configuration/#assembly-versioning-scheme

blairconrad:
Very nice. Thanks, @thomaslevesque.

PR #1212, opened 2017-10-03
Upgrade to Castle.Core 4.2.0

### LINKED ISSUE #1210: Upgrade to Castle.Core 4.2.0
To avoid future assembly binding issues (see #1208)

thomaslevesque:
Fixes #1210

### DIFF
--- a/src/FakeItEasy/FakeItEasy.csproj
+++ b/src/FakeItEasy/FakeItEasy.csproj
@@ -39,7 +39,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.6'">
-    <PackageReference Include="Castle.Core" Version="4.1.0" />
+    <PackageReference Include="Castle.Core" Version="4.2.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="1.0.0" />
     <PackageReference Include="System.Collections.Concurrent" Version="4.3.0" />
     <PackageReference Include="System.Runtime.Loader" Version="4.0.0" />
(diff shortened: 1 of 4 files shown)


blairconrad:
Does this become a breaking issue, since we'll be publishing the new `IProxyTargetAccessor.DynProxySetTarget` member?
I know _we_ don't expect people to implement the interface, but…

thomaslevesque:
Ah, good question... I'd be inclined to say "no", because even though it is exposed in our API surface, it's not really part of FakeItEasy.

Look at it this way: if we didn't ILMerge Castle.Core into FIE, we would still take a dependency on it. If we changed the dependency version from 4.1.0 to 4.2.0, the end-user would end up with Castle.Core 4.2.0, forcing them to implement the new interface member if they were implementing `IProxyTargetAccessor`. Would you consider it a breaking change for FIE in this case?

(OTOH, it *is* a breaking change for Castle.Core. With the change on `AssemblyVersion`, it makes at least 2 reasons why it should have been a major release)

blairconrad:
> Look at it this way: if we didn't ILMerge Castle.Core into FIE, we would still take a dependency on it.

Which we do for the .NET Standard editions of the library.

> If we changed the dependency version from 4.1.0 to 4.2.0, the end-user would end up with Castle.Core 4.2.0, forcing them to implement the new interface member if they were implementing IProxyTargetAccessor. Would you consider it a breaking change for FIE in this case?

I was also wondering about this. And I was leaning toward "yes". Although it's going to be very tricky to detect every case in which we introduce a breaking change via our dependencies.

thomaslevesque:
> I was also wondering about this. And I was leaning toward "yes".

So, are you saying we should consider this upgrade a breaking change? Technically, it is, but there's no reason for anyone to implement `IProxyTargetAccessor`, so it's unlikely to affect anyone...

> Although it's going to be very tricky to detect every case in which we introduce a breaking change via our dependencies.

Well, it would be easy if everyone strictly followed semantic versioning ;)

blairconrad:
>< > I was also wondering about this. And I was leaning toward "yes".

> So, are you saying we should consider this upgrade a breaking change? Technically, it is, but there's no reason for anyone to implement `IProxyTargetAccessor`, so it's unlikely to affect anyone...

I'm saying we should at least think about it. Was/am torn. I know it's _technically_ breaking, but agree with you that it's unlikely to be an issue. We can probably proceed in 4.1.0.

> > Although it's going to be very tricky to detect every case in which we introduce a breaking change via our dependencies.

> Well, it would be easy if everyone strictly followed semantic versioning ;)

Indeed.

thomaslevesque:
> We can probably proceed in 4.1.0.

Agreed

blairconrad:
Thanks, @thomaslevesque!

PR #1213, opened 2017-10-03
Add params overload to IsSameSequenceAs

### LINKED ISSUE #1194: Friendlier API for A<T>.That.IsSameSequenceAs
From comment: https://github.com/FakeItEasy/FakeItEasy/issues/1193#issuecomment-325925228

Currently `IsSameSequenceAs` takes an non-generic `IEnumerable`. This is less than ideal, because:
- it doesn't ensure type safety, e.g. you can do `IsSameSequenceAs(new[]{ "hello", "world" })` when the parameter is actually a list of ints.
- it's cumbersome to type; in the typical case you'd just want to write `IsSameSequenceAs(1, 2, 3)` rather than `IsSameSequenceAs(new[]{1, 2, 3})`

Proposal:

- add an overload that works on `IEnumerable<T>`
- add overloads which accept the expected elements as a `params` array

New overloads:

```csharp
public static T IsSameSequenceAs<T, TItem>(
    this IArgumentConstraintManager<T> manager,
    IEnumerable<TItem> values) where T : IEnumerable<TItem>;
public static T IsSameSequenceAs<T, TItem>(
    this IArgumentConstraintManager<T> manager,
    params TItem[] values) where T : IEnumerable<TItem>;
public static T IsSameSequenceAs<T>(
    this IArgumentConstraintManager<T> manager,
    params object[] values) where T : IEnumerable;
```

thomaslevesque:
Fixes #1194

### DIFF
--- a/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
+++ b/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
@@ -142,7 +142,7 @@ namespace FakeItEasy
 
         /// <summary>
         /// The tested argument collection should contain the same elements as the
-        /// as the specified collection.
+        /// specified collection, in the same order.
         /// </summary>
         /// <param name="manager">The constraint manager to match the constraint.</param>
         /// <param name="values">The sequence to test against.</param>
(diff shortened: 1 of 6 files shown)


src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
```diff
@@ -157,6 +157,19 @@ public static string IsNullOrEmpty(this IArgumentConstraintManager<string> manag
         }
 
         /// <summary>
+        /// The tested argument collection should contain the same elements as the
+        /// as the specified collection.
```
blairconrad: "as the as the". As I comment, I see that it appears that way above, as well.
While I'm complaining, I wonder if we shouldn't mention that we expect the elements to be in the same order.

thomaslevesque: Ah, yes. I just copy-pasted the doc comment and didn't notice the problem. I'll fix it.

thomaslevesque:
> new test?

I considered it, and thought it didn't add much value, since the new method is just calling the old one. But I can add a test if you prefer.

> In our docs docs, we talk about the argument to `IsSameSequenceAs` as being an "enumerable". I think this meets the spirit of the method, but perhaps we should alert people to the new version?

OK, I'll update the docs.

blairconrad:
Hey, @thomaslevesque. I didn't notice the upload. If you're done, I'm happy to merge after squash. Thanks for accommodating me.

thomaslevesque:
Squashed!

PR #1214, opened 2017-10-04
Document that WhenArgumentsMatch ignores inline argument constraints

### LINKED ISSUE #1200: Improve WhenArgumentsMatch documentation
Recently @drauch had questions about the interaction between `WhenArgumentsMatch` and inline argument constraints. He'd often been tripped up, thinking that _both_ would be applied when matching calls. Instead when `WhenArgumentsMatch` is invoked, nothing that appears inside the `A.CallTo(() => fake.Method(…))` is considered. I think this is the intent, and would not want to change it, but we should augment:

1. the XML-docs, and
2. the readthedocs docs

to make this explicit.

blairconrad:
Fixes #1200.

### DIFF
--- a/docs/argument-constraints.md
+++ b/docs/argument-constraints.md
@@ -135,8 +135,11 @@ Sometimes individually constraining arguments isn't sufficient. In
 such a case, other methods may be used to determine which calls match
 the fake's configuration.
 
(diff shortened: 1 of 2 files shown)


thomaslevesque:
Thanks @blairconrad!

PR #1215, opened 2017-10-04
Change FakeItEasy0002 diagnostic to an error (was a warning)

### LINKED ISSUE #1203: Update severity of analyzer diagnostics
The current severity of out analyzer diagnostics is probably less than ideal:

_(updated after discussion)_

| Id             | Summary                                  | Current severity | Suggested severity |
|----------------|------------------------------------------|------------------|--------------------|
| FakeItEasy0001 | Unused call specification                | Error            | Error        |
| FakeItEasy0002 | Non-virtual member configuration         | Warning          | **Error**          |
| FakeItEasy0003 | Argument constraint outside call spec    | Warning          | Warning          |
| FakeItEasy0004 | Argument constraint nullability mismatch | Warning          | Warning            |
| FakeItEasy0005 | Argument constraint type mismatch        | Error            | Error              |

~~FakeItEasy0001: it's almost certainly a mistake, but it won't cause an error at runtime (it might make the test fail, though).~~
FakeItEasy0002: we know for a fact that it will throw at runtime, since we can't fake a non-virtual method
~~FakeItEasy0003: we know for a fact that it will throw at runtime, since we're not in a call spec~~
~~FakeItEasy0005 (not merged yet): if we follow the same reasoning, it should be a warning.~~

I'm not sure if this strategy is sound, but in any case, we should try to be have consistent rules for choosing the diagnostic severity.

blairconrad:
Fixes #1203.

### DIFF
--- a/src/FakeItEasy.Analyzer/DiagnosticDefinitions.cs
+++ b/src/FakeItEasy.Analyzer/DiagnosticDefinitions.cs
@@ -30,7 +30,7 @@
                 nameof(NonVirtualSetupSpecification),
                 "FakeItEasy0002",
                 "FakeItEasy.Usage",
-                DiagnosticSeverity.Warning,
+                DiagnosticSeverity.Error,
                 true);
 
         public static DiagnosticDescriptor ArgumentConstraintOutsideCallSpec { get; } =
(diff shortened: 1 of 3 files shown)


thomaslevesque:
Thanks @blairconrad!

PR #1217, opened 2017-10-04
Build projects in parallel

### LINKED ISSUE #1216: Have MSBuild build in parallel
Any reason not to?

blairconrad:
Fixes #1216.

### DIFF
--- a/build.csx
+++ b/build.csx
@@ -245,7 +245,7 @@ public void RunMsBuild(string target)
     var packagesDirectoryOption = string.IsNullOrEmpty(packagesDirectory) ? "" : $"/p:NuGetPackagesDirectory={packagesDirectory}";
     Cmd(
         msBuild,
-        $"{solution} /target:{target} /p:configuration=Release /nr:false /verbosity:minimal /nologo /bl:artifacts/logs/{target}.binlog {packagesDirectoryOption}");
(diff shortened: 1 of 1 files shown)


build.csx
```diff
@@ -245,7 +245,7 @@ public void RunMsBuild(string target)
     var packagesDirectoryOption = string.IsNullOrEmpty(packagesDirectory) ? "" : $"/p:NuGetPackagesDirectory={packagesDirectory}";
     Cmd(
         msBuild,
-        $"{solution} /target:{target} /p:configuration=Release /nr:false /verbosity:minimal /nologo /bl:artifacts/logs/{target}.binlog {packagesDirectoryOption}");
+        $"{solution} /target:{target} /p:configuration=Release /maxcpucount /nr:false /verbosity:minimal /nologo /bl:artifacts/logs/{target}.binlog {packagesDirectoryOption}");
```
thomaslevesque: Unrelated to your change, but do you have any idea why we have `/nr:false`? Was node reuse causing issues at some point? Disabling node reuse is probably not great for performance.

thomaslevesque: Just tried with `/nr:true`; the difference isn't significant (1 or 2 second difference for a total time of ~40s)

thomaslevesque:
Thanks @blairconrad! It does seem to improve the build time.

build.csx
blairconrad: I have no idea why we have `/nr:false`, and no opinion about it being on or off. Whatever you like.

PR #1220, opened 2017-10-06
Add script to automate release branch creation, PR

### LINKED ISSUE #1218: Automate release branch creation
From https://github.com/FakeItEasy/FakeItEasy/issues/1170#issuecomment-334441840.

This is one of several automation steps suggested in that comment.

How fancy do we want to get?

1. Just create the release branch from where we are and push upstream?
2. Number 1 and create PRs to master and develop? (assumes that we're not releasing a hot fix after we've already created a release branch, but our workflow typically involves no commits on the release branch, so I think this is safe)
3. Create release branch, merge to master and develop and push both branches?
4. Number 3 and tag master before pushing? This will trigger a full release - deploying to NuGet and what not. But it's not safe for hotfixes. Nor would it actually release the hotfix. Now that I type it, I think we'll need special logic for hotfixes anyhow.

I'm a bit of a maverick and would consider number 4.

blairconrad:
Fixes #1218

### DIFF
--- a/packages.config
+++ b/packages.config
@@ -6,4 +6,5 @@
   <package id="xunit.runner.console" version="2.0.0" />
   <package id="vswhere" version="1.0.62" />
   <package id="Microsoft.Net.Compilers" version="2.2.0" />
+  <package id="Octokit" version="0.26.0" />
 </packages>
--- a/prepare_release.csx
+++ b/prepare_release.csx
@@ -39,23 +39,9 @@ await UpdateRelease(existingRelease, version);
 await CreateNextRelease();
 await UpdateIssue(existingIssue, existingMilestone, version);
 await CreateNextIssue(existingIssue, nextMilestone);
-
-var referenceClient = gitHubClient.Git.Reference;
-
-Console.WriteLine($"Fetching reference to branch {releaseFromBranchName}...");
-var releaseFromBranch = await referenceClient.Get(repoOwner, repoName, $"heads/{releaseFromBranchName}");
-Console.WriteLine($"Fetched reference to branch {releaseFromBranch.Url}");
-
-Console.WriteLine($"Creating branch {releaseBranchName} from {releaseFromBranchName}...");
-var releaseBranch = await referenceClient.CreateBranch(repoOwner, repoName, releaseBranchName, releaseFromBranch);
-Console.WriteLine($"Created branch at {releaseBranch.Url}");
-
-Console.WriteLine($"Creating pull request to merge {releaseBranchName} into {targetBranchName}...");
-var pr = await gitHubClient.PullRequest.Create(
-    repoOwner,
-    repoName,
-    new NewPullRequest($"Release {version}", releaseBranchName, targetBranchName));
-Console.WriteLine($"Created pull request '{pr.Title}' at {pr.HtmlUrl}");
+var releaseFromBranch = await GetReleaseFromBranch();
+await CreateReleaseBranch(releaseBranchName, releaseFromBranch);
+await CreatePullRequest(version, releaseBranchName);
 
 public static GitHubClient GetAuthenticatedGitHubClient()
 {
@@ -168,4 +154,29 @@ public static async Task CreateNextIssue(Issue existingIssue, Milestone nextMile
     Console.WriteLine($"Creating new release issue '{newIssue.Title}'...");
     var nextIssue = await gitHubClient.Issue.Create(repoOwner, repoName, newIssue);
     Console.WriteLine($"Created new release issue #{nextIssue.Number}: '{newIssue.Title}'");
+}
+
+public static async Task<Reference> GetReleaseFromBranch()
+{
+    Console.WriteLine($"Fetching reference to branch {releaseFromBranchName}...");
+    var releaseFromBranch = await gitHubClient.Git.Reference.Get(repoOwner, repoName, $"heads/{releaseFromBranchName}");
+    Console.WriteLine($"Fetched reference to branch {releaseFromBranch.Url}");
+    return releaseFromBranch;
+}
+
+public static async Task CreateReleaseBranch(string releaseBranchName, Reference releaseFromBranch)
+{
+    Console.WriteLine($"Creating branch {releaseBranchName} from {releaseFromBranchName}...");
+    var releaseBranch = await gitHubClient.Git.Reference.CreateBranch(repoOwner, repoName, releaseBranchName, releaseFromBranch);
+    Console.WriteLine($"Created branch at {releaseBranch.Url}");
+}
+
+public static async Task CreatePullRequest(string version, string releaseBranchName)
+{
+    Console.WriteLine($"Creating pull request to merge {releaseBranchName} into {targetBranchName}...");
+    var pr = await gitHubClient.PullRequest.Create(
+        repoOwner,
+        repoName,
+        new NewPullRequest($"Release {version}", releaseBranchName, targetBranchName));
+    Console.WriteLine($"Created pull request '{pr.Title}' at {pr.HtmlUrl}");
 }
--- a/prepare_release.csx
+++ b/prepare_release.csx
@@ -1,0 +1,44 @@
+#r ".\packages\Octokit.0.26.0\lib\net45\Octokit.dll"
+#r ".\packages\system.runtime.interopservices.runtimeinformation\4.0.0\lib\net45\System.Runtime.InteropServices.RuntimeInformation.dll"
+
+using Octokit;
+using Octokit.Helpers;
+
+const string repoOwner = "FakeItEasy";
+const string repoName = "FakeItEasy";
+const string releaseFromBranchName = "develop";
+const string targetBranchName = "master";
+
+var version = Args.FirstOrDefault();
+if (version == null)
+{
+    throw new Exception("No version supplied");
+}
+
+var releaseBranchName = $"release/{version}";
+
+var token = ReadGitHubToken();
+var credentials = new Credentials(token);
+
+var gitHubClient = new GitHubClient(new ProductHeaderValue("FakeItEasy-build-scripts")) { Credentials = credentials };
+var referenceClient = gitHubClient.Git.Reference;
+
+Console.WriteLine($"Fetching reference to branch {releaseFromBranchName}...");
+var releaseFromBranch = await referenceClient.Get(repoOwner, repoName, $"heads/{releaseFromBranchName}");
+Console.WriteLine($"Fetched reference to branch {releaseFromBranch.Url}");
+
+Console.WriteLine($"Creating branch {releaseBranchName} from {releaseFromBranchName}...");
+var releaseBranch = await referenceClient.CreateBranch(repoOwner, repoName, releaseBranchName, releaseFromBranch);
+Console.WriteLine($"Created branch at {releaseBranch.Url}");
+
+Console.WriteLine($"Creating pull request to merge {releaseBranchName} into {targetBranchName}...");
+var pr = await gitHubClient.PullRequest.Create(
+    repoOwner,
+    repoName,
+    new NewPullRequest($"Release {version}", releaseBranchName, targetBranchName));
+Console.WriteLine($"Created pull request '{pr.Title}' at {pr.HtmlUrl}");
+
+public static string ReadGitHubToken()
+{
+    return File.ReadAllText(".githubtoken").Trim();
+}
--- a/prepare_release.csx
+++ b/prepare_release.csx
@@ -1,6 +1,8 @@
 #r ".\packages\Octokit.0.26.0\lib\net45\Octokit.dll"
 #r ".\packages\system.runtime.interopservices.runtimeinformation\4.0.0\lib\net45\System.Runtime.InteropServices.RuntimeInformation.dll"
 
+using System.Linq;
+
 using Octokit;
 using Octokit.Helpers;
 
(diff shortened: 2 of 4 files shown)


packages.config
```diff
@@ -6,4 +6,5 @@
   <package id="xunit.runner.console" version="2.0.0" />
   <package id="vswhere" version="1.0.62" />
   <package id="Microsoft.Net.Compilers" version="2.2.0" />
+  <package id="Octokit" version="0.26.0" />
```
blairconrad: I feel a little bad making everyone pick this up, even non-owners, but it's probably not such a big deal.

release.cmd
```diff
@@ -0,0 +1,32 @@
...
+copy /Y %NUGET_CACHE_DIR%\NuGet.exe .nuget\NuGet.exe > nul
+
+rem restore packages for build script
+rem .nuget\NuGet.exe restore .\packages.config -PackagesDirectory .\packages -Verbosity quiet
+
+rem run build script
+".\packages\Microsoft.Net.Compilers.2.2.0\tools\csi.exe" %~n0.csx %*
+
+rem return to original working directory
+popd
```
blairconrad: This script is nearly identical to build.cmd. (And probably needs a rename anyhow.)

We could consolidate and just pass in the name of the csx to run.
But then maybe we should talk about where to put "helper" scripts. I don't feel great about having them all litter the top level directory.

release.csx
```diff
@@ -0,0 +1,84 @@
...
+if (version == null)
+{
+    throw new Exception("No version supplied");
+}
+
+var releaseBranchName = $"release/{version}";
+
+var token = ReadTokenFromNetrc();
+
+var gitHubClient = new GitHubClient(new ProductHeaderValue("FakeItEasy-build-scripts"));
```
blairconrad: this header value is arbitrary

release.csx
```diff
@@ -0,0 +1,84 @@
...
+gitHubClient.Credentials = new Credentials(token);
```
blairconrad: I'm assuming the ,netrc contains a personal token, not username password, and don't particularly want to encourage username password anyhow.

release.csx
```diff
@@ -0,0 +1,84 @@
...
+Cmd("git", $"checkout -b {releaseBranchName}");
+Cmd("git", $"push {remote} {releaseBranchName}");
```
blairconrad: The output from these commands is dumped to the console, so it's not super pretty. Then again, convenient if something goes wrong.

Note we're assuming that the caller is already using the proper branch.

release.csx
```diff
@@ -0,0 +1,84 @@
...
+
+Console.WriteLine($"Creating pull request to merge {releaseBranchName} into master...");
+var pr = await gitHubClient.PullRequest.Create(
+    repoOwner,
+    repoName,
+    new NewPullRequest($"Release {version}", releaseBranchName, "master"));
+Console.WriteLine($"Created pull request {pr.Number}: '{pr.Title}'");
```
blairconrad: Assuming a merge into master, and we're not (yet?) making a PR to go into develop. (I guess in that case, we could git checkout develop.)
So we're not supporting release of hotfixes. Just "regular" releases.

release.csx
```diff
@@ -0,0 +1,84 @@
...
+
+
+public static string ReadTokenFromNetrc()
+{
```
blairconrad: I think this is "good enough", but not particularly robust.

1. assumes the .netrc file is called {{.netrc}}. Some may call {{_netrc}}.
2. assumes a {{HOME}} environment variable, in which .netrc lives. This is standard, and would've been needed for our ruby implementation
3. only looks for api.github.com. Technically we should use the "default" entry as well, but that just seems silly

release.csx
```diff
@@ -0,0 +1,84 @@
...
+        if (tokens[i] == "machine" || tokens[i] == "default")
+        {
+            break;
+        }
+    }
+
+    throw new Exception($"Can't find password for {githubApiHost} in .netrc file");
+}
+
+public static void Cmd(string fileName, string args)
```
blairconrad: Nearly identical to the version in {{build.csx}}. We could extract and consolidate, with the same "where to put the common stuff?" question from above.

packages.config
thomaslevesque: I'm sure nobody will mind ;)

release.cmd
thomaslevesque: > We could consolidate and just pass in the name of the csx to run.

I like the idea 👍 

> But then maybe we should talk about where to put "helper" scripts. I don't feel great about having them all litter the top level directory.

Yeah, it's already a bit of a mess... We could add a `build` or `scripts` directory.

release.csx
```diff
@@ -0,0 +1,84 @@
+#r ".\packages\Octokit.0.26.0\lib\net45\Octokit.dll"
+#r ".\packages\system.runtime.interopservices.runtimeinformation\4.0.0\lib\net45\System.Runtime.InteropServices.RuntimeInformation.dll"
+
+using Octokit;
+
+const string repoOwner = "FakeItEasy";
+const string repoName = "FakeItEasy";
+const string remote = "upstream";
```
thomaslevesque: Not everyone calls their upstream remote `upstream`. We could use the URL instead.

release.csx
thomaslevesque: > Note we're assuming that the caller is already using the proper branch.

It's a pretty big assumption... I think we should checkout `develop` first (or use `checkout -b {releaseBranchName} develop`).

release.csx
thomaslevesque: 1. Fine by me. If we want to add support for `_netrc` later, it won't be hard anyway.
2. `HOME` isn't a standard environment variable on Windows (I realized this recently at work, when I used it in a script and it didn't work on my colleagues' machines). The equivalent is `USERPROFILE`.
3. Fine by me

release.csx
```diff
@@ -0,0 +1,84 @@
...
+        if (tokens[i] == "password")
+        {
+            return tokens[i + 1];
+        }
+
+        if (tokens[i] == "machine" || tokens[i] == "default")
+        {
+            break;
+        }
+    }
```
thomaslevesque: The format is quite ugly... Is it worth sticking to the `.netrc` format? We could just have a `githubtoken` file in the repo that contains just the token (of course it would have to be added to `.gitignore`).

release.csx
thomaslevesque: Yes, but I think it can wait. No need to worry about that in this PR.

release.csx
blairconrad: We can proceed as you suggest. I only conformed to the `.netrc` format because it's established and tools like Octokit.rb use it and maybe Octokit.NET will eventually, but I'm not married to it. I would not mind seeing this method go away!

release.csx
blairconrad: Can do. Assumes user's local `develop` is up to date.
If that's a concern, we could `checkout -b {releaseBranchName} {remote}/develop`. Although that seems to set tracking to the remote branch. I'm surprised.

release.csx
blairconrad: Yes. Good idea. What do you call yours?

release.csx
blairconrad: Yeah, I know `HOME` is nonstandard, but the .netrc-using tools all use it. I was going to suggest falling back to `USERPROFILE` if you were concerned. But I'm content to take your other suggestion and just look for a file in the repo.

release.csx
blairconrad: Something else. Having one file in the repo means that we'd have to do a similar thing for _every repo_ that we set up like this. Which is probably fine. At least for now.

release.cmd
blairconrad: "build" always makes me wonder if that's where the builds go, but I'm not worried enough to fight either suggestion. If you've a favourite, that's fine.
Like the common `csx` we talk about below, do you want to wait, or would you rather have this now?

blairconrad:
> have you been using JIRA a lot recently? Looks like you forgot how to format inline code in Markdown 

I don't know what you're talking about.

:blush:. Yeah. Too much JIRA. And Confluence.

release.csx
thomaslevesque: True. If it comes to this, we could also look for the file in parent folders, like `Directory.Build.props`

release.csx
thomaslevesque: `upstream`. But it doesn't mean everyone does ;)

thomaslevesque:
> Too much JIRA. And Confluence.

Oh man... all my condolences. I used and disliked both of them in the past ^^

release.cmd
thomaslevesque: Let's wait. We'll do it in a separate PR.

release.csx
thomaslevesque: > Can do. Assumes user's local develop is up to date.
> If that's a concern, we could checkout -b {releaseBranchName} {remote}/develop.

But `{remote}/develop` could also be out of date... to be sure, you'd need to `git fetch` first.

> Although that seems to set tracking to the remote branch. I'm surprised.

Yes, it does. `--no-track` should prevent this.

blairconrad:
Removing [WIP], because we're a little closer now. Not that I don't expect some comments.

release.csx
```diff
@@ -0,0 +1,44 @@
...
+Console.WriteLine($"Creating pull request to merge {releaseBranchName} into {targetBranchName}...");
+var pr = await gitHubClient.PullRequest.Create(
+    repoOwner,
+    repoName,
+    new NewPullRequest($"Release {version}", releaseBranchName, targetBranchName));
+Console.WriteLine($"Created pull request '{pr.Title}' at {pr.HtmlUrl}");
+
+public static string ReadGitHubToken()
+{
+    return File.ReadAllText("githubtoken").Trim();
```
blairconrad: I took your suggested name, @thomaslevesque, but didn't (yet) see a need to walk up the directory tree.

thomaslevesque: No, let's keep it simple for now. I'm open to a different name if you have a better idea (maybe `.githubtoken`?)

blairconrad: I prefer `.githubtoken` as well, even though the `.` doesn't give great benefits on Windows.

blairconrad:
Renamed `githubtoken`, @thomaslevesque.

If/when this PR is merged, I'd replace 

```markdown
- [ ] create release-x.y.z branch on upstream
- [ ] perform any last-minute tasks on release-x.y.z
- [ ] submit a PR to merge the release-x.y.z branch to master
- [ ] submit a PR to merge the release-x.y.z branch to develop, if there are any changes on release-x.y.z
``` 

in #1170 with something like

```markdown
- [ ] run `release.cmd x.y.z` in the root folder, which will create
    - a release branch off `develop` for _x.y.z_ and
    - a PR to merge the release branch into `master`
```

blairconrad:
Oh, while I'm thinking of it, do we like the name of the command? It's not _actually_ releasing. Maybe `prepare_release.cmd`?

thomaslevesque:
> Oh, while I'm thinking of it, do we like the name of the command? It's not actually releasing. Maybe prepare_release.cmd?

Sounds good to me!

In a next step, I think this script could do more, such as:
- rename the vNext issue, release and milestone
- create the new issue, release and milestone

blairconrad:
> > Oh, while I'm thinking of it, do we like the name of the command? It's not actually releasing. Maybe prepare_release.cmd?

> Sounds good to me!

I'll do that. And squash. It'll be harder for you to untangle, but you're up to it!

> In a next step, I think this script could do more, such as:
> - rename the vNext issue, release and milestone
> - create the new issue, release and milestone

Me too. I started on these steps yesterday! Stay tuned.

blairconrad:
> Me too. I started on these steps yesterday! Stay tuned.

So if you want, I can just add to this PR.

thomaslevesque:
> Me too. I started on these steps yesterday! Stay tuned.

You're way ahead of me!

> So if you want, I can just add to this PR.

That works for me. A separate commit would be nice, though ;)

prepare_release.csx
```diff
@@ -38,7 +57,115 @@ var pr = await gitHubClient.PullRequest.Create(
...
 {
     return File.ReadAllText(".githubtoken").Trim();
+}
+
+public static async Task<Milestone> GetExistingMilestone()
+{
+    Console.WriteLine("Fetching milestone 'vNext'...");
+    var milestoneRequest = new MilestoneRequest { State = ItemStateFilter.Open };
+    var existingMilestone = (await gitHubClient.Issue.Milestone.GetAllForRepository(repoOwner, repoName, milestoneRequest))
+        .Single(milestone => milestone.Title == "vNext");
```
thomaslevesque: This seems a bit fragile... but I can't think of a good alternative. Too bad we can't add arbitrary metadata to Github items.

prepare_release.csx
```diff
@@ -38,7 +57,115 @@ var pr = await gitHubClient.PullRequest.Create(
...
+    Console.WriteLine($"Renamed GitHub release '{existingRelease.Name}' to {updatedRelease.Name}");
+}
+
+public static async Task CreateNextRelease()
+{
+    const string newReleaseBody = @"
+### Changed
+
+### New
+* Issue Title (#12345)
```
thomaslevesque: Hopefully this will be high enough ;)

prepare_release.csx
```diff
@@ -0,0 +1,182 @@
...
+    Console.WriteLine($"Fetched GitHub release '{existingRelease.Name}'");
+    return existingRelease;
+}
+
+public static async Task<Issue> GetExistingIssue(Milestone existingMilestone)
+{
+    Console.WriteLine("Fetching release issue 'Release vNext'...");
+    var issueRequest = new RepositoryIssueRequest { Milestone = existingMilestone.Number.ToString() };
+    var existingIssue = (await gitHubClient.Issue.GetAllForRepository(repoOwner, repoName, issueRequest))
+        .Single(issue => issue.Title == "Release vNext");
```
thomaslevesque: Oh, just one thing. Shouldn't you filter by open/closed status? This will avoid causing an error if a closed duplicate exists.

prepare_release.csx
blairconrad: Yeah, I couldn't come up with anything better either.

prepare_release.csx
blairconrad: Default filter is by open status. I can be explicit, if you prefer.

blairconrad: I've added such a filter. Sadly, there is nonesuch for Issues. I could explicitly post-filter, which just occurred to me now.

thomaslevesque:
> I've added such a filter. Sadly, there is nonesuch for Issues.

I don't get it... your code uses `State = ItemStateFilter.Open` for `RepositoryIssueRequest`. Doesn't it work? Why do you also need to post-filter?

blairconrad:
> I don't get it... your code uses State = ItemStateFilter.Open for RepositoryIssueRequest. Doesn't it work? Why do you also need to post-filter?

I can explain. There's no filter for unpublished _releases_. I found this out, and in my early-morning haze, by the time I got to the code, I thought it was _issues_ and added the post-filter.
Changed to a post-filter on releases. :blush:

thomaslevesque:
Looks all good to me! If you're done, I'll merge after you squash, and we can even go ahead with the release 👍

blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
Nice work, thanks @blairconrad!

PR #1225, opened 2017-10-10
Release 4.1.0

blairconrad:

PR #1229, opened 2017-10-11
[DO NOT MERGE] Test AppVeyor build on branch with slash in name

thomaslevesque:
Do not merge.

AppVeyor seems to be having trouble with branches that have slashes in their name (see https://github.com/FakeItEasy/FakeItEasy/pull/1225#issuecomment-335502288). Just trying to see if I can reproduce the problem.

### DIFF
--- a/test.txt
+++ b/test.txt
@@ -1,0 +1,1 @@
+Hello world


thomaslevesque:
OK, so that's not it. AppVeyor is building this branch just fine.
Unfortunately there's no trace of the failed build for #1225 in the history...

According to [this](http://help.appveyor.com/discussions/problems/4830-stuck-on-appveyor-was-unable-to-build-non-mergeable), the "AppVeyor was unable to build non-mergeable pull request" error can be temporary. Maybe we should just have tried again...

PR #1231, opened 2017-10-13
Fix false positive "nested argument constraint" when constraint is of the wrong type

### LINKED ISSUE #1230: False positive "nested constraint" exception when using an argument constraint of the wrong type
*Original title: **Weird issue with nested constraints and simple types***

```csharp
public interface IEmailFactory
    {
        Task<IList<MailMessage>> GetEmailMessages(long idPairGame, EmailTemplateType emailTemplateType);
    }

_emailFactory = new Fake<IEmailFactory>().FakedObject;

A.CallTo(() => _emailFactory.GetEmailMessages(A<int>._, A<EmailTemplateType>._)).Returns(mailMessages);

```
EmailTemplateType is an enum. When I change it to int it is still wrong, so it is not an issue.

FakeIt version 4.1.

It throws the following exception

>FakeItEasy.Expressions.ExpressionArgumentConstraintFactory.ArgumentConstraintExpressionVisitor.VisitMember(MemberExpression node)
   at System.Linq.Expressions.MemberExpression.Accept(ExpressionVisitor visitor)
   at System.Linq.Expressions.ExpressionVisitor.Visit(Expression node)
   at System.Linq.Expressions.ExpressionVisitor.VisitUnary(UnaryExpression node)
   at System.Linq.Expressions.UnaryExpression.Accept(ExpressionVisitor visitor)
   at FakeItEasy.Expressions.ExpressionArgumentConstraintFactory.CheckArgumentExpressionIsValid(Expression expression)
   at FakeItEasy.Expressions.ExpressionArgumentConstraintFactory.GetArgumentConstraintFromExpression(Expression expression, Object& value)
   at FakeItEasy.Expressions.ExpressionArgumentConstraintFactory.GetArgumentConstraint(ParsedArgumentExpression argument)
   at System.Linq.Utilities.<>c__DisplayClass2_0`3.<CombineSelectors>b__0(TSource x)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at FakeItEasy.Expressions.ExpressionCallMatcher..ctor(ParsedCallExpression parsedExpression, ExpressionArgumentConstraintFactory constraintFactory, MethodInfoManager methodInfoManager, Factory outputWriterFactory)
   at FakeItEasy.RootModule.<>c__DisplayClass0_0.<RegisterDependencies>b__4(ParsedCallExpression callSpecification)
   at FakeItEasy.Configuration.FakeConfigurationManager.CallTo[T](Expression`1 callSpecification)

thomaslevesque:
Fixes #1230

### DIFF
--- a/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
+++ b/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
@@ -27,8 +27,7 @@ namespace FakeItEasy.Expressions
 
             var isByRefArgument = IsByRefArgument(argument);
 
-            object argumentValue;
-            var constraint = this.GetArgumentConstraintFromExpression(argument.Expression, out argumentValue);
+            var constraint = this.GetArgumentConstraintFromExpression(argument.Expression, out var argumentValue);
             if (isByRefArgument)
             {
                 if (IsOutArgument(argument))
@@ -90,14 +89,7 @@ namespace FakeItEasy.Expressions
             }
 
             var visitor = new ArgumentConstraintExpressionVisitor();
-            var methodCallExpression = expression as MethodCallExpression;
-            if (methodCallExpression == null)
-            {
-                // An unknown kind of expression - could be a constructor, or almost anything else. Play it safe and
-                // check it out.
-                visitor.Visit(expression);
-            }
-            else
+            if (expression is MethodCallExpression methodCallExpression)
             {
                 // A method call. It might be A<T>.That.Matches, or one of the other extension methods, so don't
                 // check the method node itself. Instead, look at all the arguments (except the first, if it's an
@@ -107,6 +99,12 @@ namespace FakeItEasy.Expressions
                 {
                     visitor.Visit(argument);
                 }
+            }
+            else
+            {
+                // An unknown kind of expression - could be a constructor, or almost anything else. Play it safe and
+                // check it out.
+                visitor.Visit(expression);
             }
         }
 
@@ -147,8 +145,7 @@ namespace FakeItEasy.Expressions
 
             foreach (var argumentExpression in expression.Expressions)
             {
-                object ignored;
-                result.Add(this.GetArgumentConstraintFromExpression(argumentExpression, out ignored));
+                result.Add(this.GetArgumentConstraintFromExpression(argumentExpression, out _));
             }
 
             return new AggregateArgumentConstraint(result);
--- a/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
+++ b/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
@@ -1,4 +1,4 @@
-﻿namespace FakeItEasy.Expressions
+namespace FakeItEasy.Expressions
 {
     using System;
     using System.Collections.Generic;
@@ -81,7 +81,7 @@
 
         private static void CheckArgumentExpressionIsValid(Expression expression)
         {
-            expression = GetExpressionWithoutBoxing(expression);
+            expression = GetExpressionWithoutConversion(expression);
 
             if (expression is MemberExpression)
             {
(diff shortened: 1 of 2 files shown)


blairconrad:
Thanks for the PR, @thomaslevesque. I like the structure. And the change looks nice.
I do have some concerns about it, though.
First, while it's great to not throw a misleading exception, I don't think that masking the error and in effect producing a non-matching constraint is desirable. Ideally we'd throw, but a better exception.

You mentioned in https://github.com/FakeItEasy/FakeItEasy/issues/1230#issuecomment-336433274

> The problem is that there's no way to distinguish between explicit and implicit conversions with the Expression API...
> I think we should always eliminate the conversion, even if it's not a boxing conversion, but I'm a bit worried about unwanted side effects.

Do we need to worry about explicit and implicit conversions? I think all conversion nodes _whose unary member contains an `A<>`-style argument constraint_ are bad argument constraints. Converting `A<>._`  between numeric types or invoking an implicit _or_ explicit conversion on them is an error, no? And since implicit casts up and down the type hierarchy don't' produce a conversion node in the Expression API, I think we might be safe to flag all conversion expressions containing argument constraints as "badly typed argument constraints", no?

thomaslevesque:
> First, while it's great to not throw a misleading exception, I don't think that masking the error and in effect producing a non-matching constraint is desirable. Ideally we'd throw, but a better exception.

Agreed. I thought about this too, but I wanted to focus on fixing the "false positive" nested constraint detection. But I can add a more adequate exception. It will basically do the same thing as the FakeItEasy0005 diagnostic ;)

> Do we need to worry about explicit and implicit conversions? I think all conversion nodes whose unary member contains an A<>-style argument constraint are bad argument constraints. Converting A<>._ between numeric types or invoking an implicit or explicit conversion on them is an error, no? And since implicit casts up and down the type hierarchy don't' produce a conversion node in the Expression API, I think we might be safe to flag all conversion expressions containing argument constraints as "badly typed argument constraints", no?

Sounds good to me 👍

thomaslevesque:
> And since implicit casts up and down the type hierarchy don't' produce a conversion node in the Expression API, I think we might be safe to flag all conversion expressions containing argument constraints as "badly typed argument constraints", no?

In fact, we just need to allow boxing conversions. For instance, `A<int>._` is a valid constraint for an `object` argument.

thomaslevesque:
Looking at the code again, I think the nested constraint detection is flawed in other ways... For instance, take [this](https://github.com/FakeItEasy/FakeItEasy/blob/develop/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs#L86-L90). I don't remember what the exact reasoning was at the time, but I think the comment is wrong. The condition would match something like `Foo.Bar(A<int>._).Value` and wouldn't detect the nested constraint. I'll try to look more closely tonight, but we'll probably need to rethink that part.

thomaslevesque:
> The condition would match something like `Foo.Bar(A<int>._).Value` and wouldn't detect the nested constraint.

In fact, the static method isn't necessary to reproduce the problem; this works too: `new Foo(A<int>._).Value`.

Opened #1232

thomaslevesque:
> Agreed. I thought about this too, but I wanted to focus on fixing the "false positive" nested constraint detection. But I can add a more adequate exception. It will basically do the same thing as the FakeItEasy0005 diagnostic ;)

In fact, I think this PR should just fix the false positive after all. It won't detect constraints of the wrong type, but they were already not detected before... Adding the type check is an enhancement; this PR just attempts to fix a bug. I don't think we should mix the two in the same PR.

thomaslevesque:
I reworded the commits and changed the PR title, to clarify what I was attempting to fix. I can also rename the branch, but it will require closing this PR and making a new one.

blairconrad:
Thanks, @thomaslevesque.

thomaslevesque:
Thanks for the merge!

PR #1233, opened 2017-10-16
Fix missed nested constraint

### LINKED ISSUE #1232: Nested constraint is not detected
This code has a nested constraint, but it isn't detected:

```
public interface IFoo { int Bar(string s); }

public class X
{
    public X(int value) { Value = x.ToString(); }
    public string Value { get; }
}

var fake = A.Fake<IFoo>();
A.CallTo(() => fake.Bar(new X(A<int>._).Value)).Returns(42);
```

thomaslevesque:
Fixes #1232

### DIFF
--- a/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
+++ b/src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
@@ -82,9 +82,9 @@ namespace FakeItEasy.Expressions
         {
             expression = GetExpressionWithoutConversion(expression);
 
-            if (expression is MemberExpression)
+            if (expression is MemberExpression memberExpression && IsMemberOfA(memberExpression.Member))
             {
-                // It's A._, or A.Ignore, or some other property/field, so it's safe.
+                // It's A._, or A.Ignore, so it's safe.
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -82,7 +82,7 @@ private static void CheckArgumentExpressionIsValid(Expression expression)
         {
             expression = GetExpressionWithoutConversion(expression);
 
-            if (expression is MemberExpression)
+            if (expression is MemberExpression memberExpression && IsMemberOfA(memberExpression.Member))
             {
                 // It's A._, or A.Ignore, or some other property/field, so it's safe.
```
blairconrad: I'd consider updating this comment. We originally meant it to refer to `A._` or `A.Ignore` or some field or property on _another type_ (which was an error, as you point out). I'd strike ", or some other property/field".
An alternative would be to say "It's A._, or A.Ignore, or some other property/field on A, so it's safe.".

This isn't a big deal, and if you disagree, I'm content to merge as is.

thomaslevesque: I agree, and I'll change it

thomaslevesque: I amended the commit

blairconrad:
Thanks, @thomaslevesque!

PR #1246, opened 2017-10-29
Detect wrong-typed constraints

### LINKED ISSUE #1237: Detect wrong argument constraint type at runtime
Related to #1177, #1230, #1163 

We already have the FakeItEasy0005 diagnostic to detect this, but we can also detect it at runtime, and we should throw in this case.

A possible approach would be to examine all the constraints collected by `IArgumentConstraintTrapper.TrapConstraints` and check if their type is compatible with the parameter type.

thomaslevesque:
Fixes #1237 

The 3rd commit is a simplification. I realized that only `MatchesConstraint` could be created by the argument constraint trap, so we don't need to introduce a `Type` property for other types of constraint.

### DIFF
--- a/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -4,6 +4,7 @@ namespace FakeItEasy.Specs
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
     using System.Runtime.InteropServices;
+    using FakeItEasy.Configuration;
     using FakeItEasy.Tests.TestHelpers;
     using FluentAssertions;
     using Xbehave;
@@ -49,6 +50,11 @@ namespace FakeItEasy.Specs
         public interface IIHaveACollectionParameter
         {
             int Bar(object[] args);
+        }
+
+        public interface IHaveAnObjectParameter
+        {
+            int Bar(object arg);
         }
 
         [Scenario]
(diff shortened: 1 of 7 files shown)


tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -631,22 +634,86 @@ public static void UnusedArgumentMatcherDescriptionNotUsed(IHaveNoGenericParamet
...
+            "And a delegate that produces a constraint of the wrong type"
+                .x(() => constraintFactory = () => A<byte>.Ignored);
+
+            "When I try to configure a method of the fake with this delegate"
+                .x(() => exception = Record.Exception(() => A.CallTo(() => fake.Bar(constraintFactory()))));
+
+            "Then the call configuration throws a FakeConfigurationException"
+                .x(() => exception.Should().BeAnExceptionOfType<FakeConfigurationException>());
+
+            "And the message should indicate the actual and expected type"
```
blairconrad: nit: previous steps used the imperative mood, not the subjunctive (fun fact: I would probably not even know these moods exist were it not for learning about "le subjontif" in school).
Consider rewording. Maybe something like "And the message indicates the actual and expected types"
(oh, I changed "type" to "types" there as well)

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -631,22 +634,86 @@ public static void UnusedArgumentMatcherDescriptionNotUsed(IHaveNoGenericParamet
...
+                .x(() => exception.Message.Should().Be("Argument constraint is of type System.Byte, but parameter is of type System.Int32. No call can match this constraint."));
+        }
+
+        [Scenario]
+        public static void PassingHiddenConstraintWithWrongTypeToAMethod(
```
blairconrad: I like this test.

src/FakeItEasy/Core/IArgumentConstraint.cs
```diff
@@ -20,4 +22,9 @@ internal interface IArgumentConstraint
         /// <returns>True if the argument is valid.</returns>
         bool IsValid(object argument);
     }
+
+    internal interface ITypedArgumentConstraint : IArgumentConstraint
```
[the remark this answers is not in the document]
thomaslevesque: Sure

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
thomaslevesque: Good point

thomaslevesque:
The rebase was a bit messy, hopefully I didn't break anything...

blairconrad:
Nice, @thomaslevesque! Thanks.

thomaslevesque:
Thanks for the merge!

PR #1248, opened 2017-11-01
Upgrade to .NET Core SDK 2.0.2

### LINKED ISSUE #1197: Warning BC40057 in FakeItEasy.IntegrationTests.VB when TargetFramework is netcoreapp1.0
The full error message is:

> Namespace or type specified in the project-level Imports 'System.Data' doesn't contain any public member or cannot be found

This project-level import comes from `Microsoft.NET.Sdk.VisualBasic.targets` in `%ProgramFiles%\dotnet\sdk\1.0.4\Sdks\Microsoft.NET.Sdk\build`, which unconditionally imports `System.Data`. But in .NET Core 1.0, there's no System.Data... This problem appears to be fixed in the 2.0.0 SDK (we should probably consider switching to that version, btw).

I tried to add `<Import Remove="System.Data" />` in an `<ItemGroup>`, but it has no effect.

### LINKED ISSUE #1247: Upgrade to .NET Core SDK 2.0.2
This should fix #1197, and have better support for VB.NET projects.

thomaslevesque:
Fixes #1247 
Fixes #1197

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,5 @@
 {
-  "projects": [ "src", "tests" ],
   "sdk": {
-    "version": "1.0.3"
+    "version": "2.0.2"
   }
 }
--- a/how_to_build.md
+++ b/how_to_build.md
@@ -13,6 +13,8 @@ At the time of writing the build is only confirmed to work on Windows using the 
 
 1. Ensure you have version 15.3 or later of either Visual Studio 2017 or MSBuild installed.
 
+1. Ensure you have .NET Core SDK 2.0.2 or later installed
+
 ## Building
 
 Using a command prompt, navigate to your clone root folder and execute `build.cmd`.


global.json
```diff
@@ -1,6 +1,5 @@
 {
-  "projects": [ "src", "tests" ],
```
thomaslevesque: Based on [this](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json), the `projects` property no longer exists.

blairconrad:
I checked [how_to_build](https://github.com/FakeItEasy/FakeItEasy/blob/ba200ed4a7b3d36340eb85e927203023d1c65138/how_to_build.md) to see if we needed to make an update to the SDK we list there (since I had only 2.0.0 installed, and the build failed for me when I ran `build.cmd`), and I see that we list no SDK. An oversight? One that we should correct as part of this PR?

thomaslevesque:
Good point @blairconrad. I don't recall installing SDK 2.0.2 myself, so I assume it was installed as part of a VS update... It might be included in 15.4, but I'm not sure. Anyway, I'll update the build instructions.

thomaslevesque:
Amended the commit to update how_to_build.md

blairconrad:
Looking good, @thomaslevesque. Thanks!

thomaslevesque:
Thanks!

PR #1250, opened 2017-11-02
Unclutter the root directory

### LINKED ISSUE #1249: Unclutter the root directory
The root of the repo is currently a bit of a mess. There are 23 files in the root directory, and every new script we add makes it worse (especially for .csx scripts which need a .cmd bootstrapper). We should probably try to move some of these files to a subdirectory.

Files and directories that could be moved to a subdirectory:
- `build.csx`: the .cmd bootstrapper probably needs to stay at the root, but we can move the actual script
- `prepare_release.csx`: same reasoning
- `.githubtoken`*: should probably be next to `prepare_release.csx`
- `deploy.ps1`: only called by AppVeyor, we don't need it at the root
- `packages.config` and `packages/`*: it's only used by csx scripts (all projects now use the package cache and `PackageReference`)
- `.nuget/`*: the projects no longer need it, it's only used by scripts
- `FakeItEasy.Dictionary.xml`, `FakeItEasy.ruleset`, `FakeItEasy.snk`: referenced from `Directory.Build.props`, which can specify a subdirectory
- `mkdocs.yml`: according [to the documentation](http://read-the-docs.readthedocs.io/en/latest/builds.html#mkdocs), we could move it to the `docs` folder

Files that should stay at the root:
- `.editorconfig`, `.gitignore`, `.gitattributes`: apply to the whole repo
- `.mailmap`: Git looks for it at the root of the repo
- `appveyor.yml`: AppVeyor looks for it at the root of the repo
- `build.cmd`: should be easily accessible
- `CONTRIBUTING.md`: it's the standard location on GitHub
- `Directory.Build.props`: applies to projects in all subdirectories
- `FakeItEasy.sln`: should be easily accessible, and the root is the only directory that contains all projects
- `GitVersion.yml`: it has to be in the working directory, which is the root during build. Putting it somewhere else would be awkward
- `global.json`: applies to projects in all subdirectories
- `Licence.txt`: it's the standard location on GitHub
- `README.md`: standard location on GitHub

Unsure:
- `how_to_build.md`: could probably be in a subdirectory, but it needs to be easy to find. If we move it, we could add a link in the README.
- `prepare_release.cmd`: it's more convenient to have it at the root, but on the other hand, it's only used by the maintainers, and we don't need it very often.

(files and folders with a `*` are excluded from source control, so they wouldn't appear in the GitHub repo anyway)

thomaslevesque:
Fixes #1249


tools/build.csx
```diff
@@ -272,3 +272,5 @@ public string GetVSLocation()
 
     return installationPath;
 }
+
+public static string GetCurrentScriptDirectory([CallerFilePath] string path = null) => Path.GetDirectoryName(path);
```
thomaslevesque: There doesn't seem to be a simpler way to get the path of the current script...

tools/build.csx
```diff
@@ -5,13 +5,13 @@
...
 using System.Net.Http;
 using System.Xml.Linq;
+using System.Runtime.CompilerServices;
 using static SimpleTargets;
 
 var solutionName = "FakeItEasy";
 
 var solution = "./" + solutionName + ".sln";
 var versionInfoFile = "./src/VersionInfo.cs";
-var packagesDirectory = Path.GetFullPath("packages");
```
thomaslevesque: We no longer need this; all projects use the package cache in `%USERPROFILE%\.nuget\packages`.

tools/deploy.ps1
```diff
@@ -53,4 +55,6 @@ $artifacts | ForEach-Object {
     Write-Output "Pushed  $($_.Name)"
 }
 
-Write-Output "Finished deploying $releaseName"
\ No newline at end of file
+Write-Output "Finished deploying $releaseName"
+
+Pop-Location
```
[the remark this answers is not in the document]
thomaslevesque: Oops. Not that it really matters, since the build won't do anything after this script, but I'll fix it anyway. A try/finally should do the trick.

thomaslevesque:
> I think we also need to update appveyor.yml to point at the new deploy.ps1 location.

Oh. I was sure I had changed it, but apparently I didn't...

thomaslevesque:
I sent a fixup commit.

tools/deploy.ps1
```diff
@@ -0,0 +1,66 @@
+$ErrorActionPreference = "Stop"
+
+Push-Location (Get-Item $PSScriptRoot).Parent.FullName
+
+try {
```
blairconrad: A moustache on the same line as a `try`? What is this, Java?

blairconrad: Actually, strike that. We seem to have a lot of moustaches on same lines in our powershell. But later on you have

thomaslevesque: Better now?

blairconrad: I still see

thomaslevesque: And now?

thomaslevesque:
Squashed

blairconrad:
Thanks, @thomaslevesque!

PR #1251, opened 2017-11-03
Align Solution items with uncluttered up root directory

blairconrad:
Contributes to #1249.
I should've thought of this when reviewing #1250.

### DIFF
--- a/FakeItEasy.sln
+++ b/FakeItEasy.sln
@@ -10,17 +10,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		.mailmap = .mailmap
 		appveyor.yml = appveyor.yml
 		build.cmd = build.cmd
-		build.csx = build.csx
 		CONTRIBUTING.md = CONTRIBUTING.md
 		Directory.Build.props = Directory.Build.props
-		FakeItEasy.Dictionary.xml = FakeItEasy.Dictionary.xml
-		FakeItEasy.ruleset = FakeItEasy.ruleset
-		FakeItEasy.snk = FakeItEasy.snk
 		GitVersion.yml = GitVersion.yml
 		global.json = global.json
 		how_to_build.md = how_to_build.md
 		License.txt = License.txt
-		mkdocs.yml = mkdocs.yml
 		README.md = README.md
 	EndProjectSection
 EndProject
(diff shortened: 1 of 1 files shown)


thomaslevesque:
Oops. I should have thought of it when I made #1250 :wink:
In fact I didn't even open the solution in VS, I used VS Code.

FakeItEasy.sln
```diff
@@ -10,17 +10,12 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		.mailmap = .mailmap
 		appveyor.yml = appveyor.yml
 		build.cmd = build.cmd
-		build.csx = build.csx
 		CONTRIBUTING.md = CONTRIBUTING.md
 		Directory.Build.props = Directory.Build.props
-		FakeItEasy.Dictionary.xml = FakeItEasy.Dictionary.xml
-		FakeItEasy.ruleset = FakeItEasy.ruleset
-		FakeItEasy.snk = FakeItEasy.snk
```
thomaslevesque: Shouldn't these files be under `src` ?

FakeItEasy.sln
```diff
@@ -199,6 +204,7 @@ Global
 		{DC5EAE30-96E3-4D38-B3C5-33E587C9FB76} = {AD790E7D-D664-4B5F-8BE8-01507A2A9F3B}
 		{5D41D463-D53B-4ADC-90AD-44AA03D6340D} = {AD790E7D-D664-4B5F-8BE8-01507A2A9F3B}
 		{575A3C3B-17BA-4A8A-9F53-05320913D0A6} = {AD790E7D-D664-4B5F-8BE8-01507A2A9F3B}
+		{AD790E7D-D664-4B5F-8BE8-01507A2A9F3B} = {59466C4D-C407-4DD9-B77B-D99162B9FC7C}
```
thomaslevesque: Unless I read this wrong, it means that `tests` has moved under `src`...

blairconrad: You are correct. And I have _no idea_ how I managed that.

FakeItEasy.sln
blairconrad: Er, yes. I forgot about the files you moved to `src`. I was only thinking of the new `tools` dir.

blairconrad:
> Oops. I should have thought of it when I made #1250 :wink:
> In fact I didn't even open the solution in VS, I used VS Code.

I've never been entirely sold on the benefits of including every file in the repo in the sln, myself. Maybe it's time to reconsider. :smile:

adamralph:
> I've never been entirely sold on the benefits of including every file in the repo in the sln, myself.

The general idea is that you can open VS and do "everything". That said, it's about time that VS just did this itself, as it now does with SDK-style csproj.

thomaslevesque:
> I've never been entirely sold on the benefits of including every file in the repo in the sln, myself. Maybe it's time to reconsider. 😄

I like it; it's handy to have everything in the same place. But as @adamralph said, it would be nice if it were automatic...

thomaslevesque:
Thanks @blairconrad, LGTM!
I'll merge after you squash

blairconrad:

thomaslevesque:
Thanks @blairconrad!

PR #1255, opened 2017-11-05
Document ref argument matching

### LINKED ISSUE #1123: Document ref argument matching
We recently had [a question about matching `ref` parameters](https://github.com/FakeItEasy/FakeItEasy/issues/18#issuecomment-308413232), and I couldn't find any docs to point at.

We talk about [matching `out` parameters](http://fakeiteasy.readthedocs.io/en/stable/argument-constraints/#out-parameters), but never `ref`.

We should show how to use the default equality matcher as well as some [fancier matching,](http://fakeiteasy.readthedocs.io/en/stable/argument-constraints/#other-matchers) such as `Ignored`. A pointer to [assigning out and ref parameters](http://fakeiteasy.readthedocs.io/en/stable/assigning-out-and-ref-parameters/) to simulate changing the value is probably not a bad idea.

blairconrad:
Fixes #1123.

### DIFF
--- a/docs/argument-constraints.md
+++ b/docs/argument-constraints.md
@@ -107,17 +107,17 @@ events in the context of an `A.CallTo`.
 So, tempting as it might be to save one of the constraints away in a
 handy variable, don't do it.
 
-# Out parameters
+# `out` parameters
 
-The incoming argument value of out parameters is ignored when matching
-calls. The incoming value of an out parameter can't be seen by the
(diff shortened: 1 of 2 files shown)


blairconrad:
I mixed in some minor formatting/grammar fixes as well. I can try to pull them out, if you like. (But am lazy and would rather not.)

blairconrad:
Thanks, @thomaslevesque.

PR #1268, opened 2017-11-12
Hoist mkdocs.yml to root

### LINKED ISSUE #1267: Docs do not render properly
| |4.1.1|4.2.0|
|-|-----|-----|
|Table of Contents|image|image|
|Code styling|image|image|

I'm sure this is due to the move of `mkdocs.yml` to the `docs` directory as part of the fix for #1249.

blairconrad:
Fixes #1267.


thomaslevesque:
Thanks @blairconrad. I guess we can live with the workaround until this is fixed on ReadTheDocs' side (or is it MkDocs?)

blairconrad:
Thanks, @thomaslevesque. I think it's readthedocs. I've filed an issue (linked to from #1267).
If I move the `mkdocs.yml` into the `docs` directory and lightly edit it to fix up some paths, running `mkdocs` locally works just fine, so this makes me think that it's readthedocs not pointing mkdocs at the YML.
I may check out the code if the readthedocs people don't come back first.

PR #1273, opened 2017-11-24
Implement WithVoidReturnType

### LINKED ISSUE #1272: Add WithVoidReturnType method
As mentioned a few times in the past, we should have a `WithVoidReturnType` method that mirrors `WithNonVoidReturnType`.

thomaslevesque:
Fixes #1272


src/FakeItEasy/Configuration/IAnyCallConfigurationWithNoReturnTypeSpecified.cs
```diff
@@ -21,5 +21,11 @@ public interface IAnyCallConfigurationWithNoReturnTypeSpecified
         /// </summary>
         /// <returns>A configuration object.</returns>
         IAnyCallConfigurationWithReturnTypeSpecified<object> WithNonVoidReturnType();
+
+        /// <summary>
+        /// Matches calls that have void return type.
+        /// </summary>
+        /// <returns>A configuratino object.</returns>
```
blairconrad: "configuration"

docs/specifying-a-call-to-configure.md
```diff
@@ -52,6 +52,9 @@ the fake to `A.CallTo` to refer to any method on the fake:
 ```csharp
 A.CallTo(fakeShop).Throws(new Exception());
 
+// Or limit the calls to void methods
+A.CallTo(fakeShop).WithVoidReturnType().Throws("sugar overflow");
```
blairconrad: 1. nice exception

src/FakeItEasy/Configuration/IAnyCallConfigurationWithVoidReturnType.cs
```diff
@@ -0,0 +1,11 @@
+namespace FakeItEasy.Configuration
+{
+    /// <summary>
+    /// Aggregate of IVoidArgumentValidationConfiguration&lt;T&gt; and IWhereConfiguration&lt;IAnyCallConfigurationWithVoidReturnType&gt;.
```
blairconrad: `IVoidArgumentValidationConfiguration` is not generic - strike the `&lt;T&gt;`.

thomaslevesque:
@blairconrad thanks for the review. What about the breaking/non-breaking question I mentioned in the issue? If we consider this to be breaking, merging it now means the next version will be 5.0.0. (IMO we shouldn't consider this a breaking change; it's true that it changes an interface, but I'm willing to bet that not a single user will be affected)

src/FakeItEasy/Configuration/IAnyCallConfigurationWithNoReturnTypeSpecified.cs
thomaslevesque: amended

blairconrad:
@thomaslevesque, I was stalling with the easy review before addressing the breaking/non-breaking question. I will respond in the issue.

src/FakeItEasy/Configuration/IAnyCallConfigurationWithVoidReturnType.cs
blairconrad: @thomaslevesque, did this comment get lost in the excitement?

thomaslevesque: Yes ! I'll fix it

thomaslevesque: Fixed

blairconrad:
Thanks, @thomaslevesque!

PR #1274, opened 2017-12-04
Fix generation of analyzer packages

### LINKED ISSUE #1263: Updating to 4.2.0 removes the analyzer from the project file
I had 4.0.0 installed of FakeItEasy and FakeItEasy.Analyzer.CSharp. I just tried to update them to 4.2.0 and noticed that the analyzer was removed from the project file. I then tried to reinstall it but that didn't add it back.

I then removed the analyzer entirely and tried installing 4.0.0, then removed it and tried 4.1.0, etc. and the only versions that add the analyzer to the project file are 4.0.0 and 4.1.0.

I've tried restarting VS2017, using the package manager console, and VS2017 Preview but the analyzer still isn't installed from 4.1.1 on.

thomaslevesque:
Fixes #1263

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -9,14 +9,7 @@
     <PackageLicenseUrl>https://github.com/FakeItEasy/FakeItEasy/blob/master/License.txt</PackageLicenseUrl>
     <PackageReleaseNotes>https://github.com/FakeItEasy/FakeItEasy/releases</PackageReleaseNotes>
     <NoPackageAnalysis>true</NoPackageAnalysis>
+    <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);GetPackageFiles</TargetsForTfmSpecificContentInPackage>
   </PropertyGroup>
-
-  <ItemGroup>
-    <!-- Include symbols in the package -->
-    <Content Include="$(OutputPath)/**/$(AssemblyName).pdb">
-      <Pack>true</Pack>
-      <PackagePath>$(BuildOutputTargetFolder)/%(RecursiveDir)$(AssemblyName).pdb</PackagePath>
-    </Content>
-  </ItemGroup>
 
 </Project>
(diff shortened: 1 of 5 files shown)


src/Directory.Build.props
```diff
@@ -9,14 +9,7 @@
...
+    <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);GetPackageFiles</TargetsForTfmSpecificContentInPackage>
   </PropertyGroup>
 
-  <ItemGroup>
-    <!-- Include symbols in the package -->
-    <Content Include="$(OutputPath)/**/$(AssemblyName).pdb">
-      <Pack>true</Pack>
-      <PackagePath>$(BuildOutputTargetFolder)/%(RecursiveDir)$(AssemblyName).pdb</PackagePath>
-    </Content>
-  </ItemGroup>
```
thomaslevesque: No longer necessary; we have to use the new target for analyzer dlls anyway, so we might as well have only one way of doing this.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.CSharp.csproj
```diff
@@ -20,6 +20,7 @@
     <Description>Provides diagnostic analyzers to warn about incorrect usage of FakeItEasy in C#. Works in Visual Studio 2015 Update 1 or later.</Description>
     <PackageTags>$(PackageTags);csharp;c#</PackageTags>
     <BuildOutputTargetFolder>analyzers/dotnet/cs</BuildOutputTargetFolder>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
```
thomaslevesque: So that the dll doesn't also get copied to `analyzers/dotnet/cs/netstandard1.1`

thomaslevesque:
Thanks for the merge!

I was about to make a small change (just a redundant `IncludeBuildOutput` that could have been in Directory.Build.props), but I'll do it in a separate PR. Or not... it doesn't really matter.

PR #1275, opened 2017-12-04
Refactor IncludeBuildOutput property in Directory.Build.props

thomaslevesque:
Relates to #1274


blairconrad:
Looks very good, @thomaslevesque, sorry I merged the other PR early.

I don't suppose you want to use https://github.com/thomaslevesque/Analyzers.Build.Packaging, do you?

thomaslevesque:
> I don't suppose you want to use https://github.com/thomaslevesque/Analyzers.Build.Packaging, do you?

We could use it, but I don't see the point in taking an external dependency when we have it working directly in the project.

PR #1279, opened 2017-12-11
Use the same wording (never, once, twice, n times) when reporting act…

### LINKED ISSUE #1278: Inconsistent use of the # sign in error messages
As you can see in the following example, the error message tries to be very natural in its language.
This should also be applied to the 2nd number in the text `#7`.

Have it be "never", "once", "n times" like the first number.

```
    SomeMethodName()
  Expected to find it exactly 6 times but found it #7 times among the calls:
[...]
```

Mertsch:
…ual amount of found calls

Fixes #1278

### DIFF
--- a/src/FakeItEasy/Core/FakeAsserter.cs
+++ b/src/FakeItEasy/Core/FakeAsserter.cs
@@ -61,7 +61,21 @@ namespace FakeItEasy.Core
 
             if (calls.Any())
             {
-                writer.Write("but found it #{0} times among the calls:", matchedCallCount);
+                switch (matchedCallCount)
+                {
+                    case 0:
+                        writer.Write("but was not found among the calls:");
+                        break;
+                    case 1:
+                        writer.Write("but found it once among the calls:");
+                        break;
+                    case 2:
+                        writer.Write("but found it twice among the calls:");
+                        break;
+                    default:
+                        writer.Write("but found it {0} times among the calls:", matchedCallCount);
+                        break;
+                }
             }
             else
             {
--- a/src/FakeItEasy/Core/FakeAsserter.cs
+++ b/src/FakeItEasy/Core/FakeAsserter.cs
@@ -64,7 +64,7 @@ namespace FakeItEasy.Core
                 switch (matchedCallCount)
                 {
                     case 0:
-                        writer.Write("but was not found among the calls:");
+                        writer.Write("but didn't find it among the calls:");
                         break;
                     case 1:
                         writer.Write("but found it once among the calls:");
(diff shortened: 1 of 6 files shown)


src/FakeItEasy/Core/FakeAsserter.cs
```diff
@@ -61,7 +61,21 @@ private static void AppendExpectation(IEnumerable<IFakeObjectCall> calls, string
 
             if (calls.Any())
             {
-                writer.Write("but found it #{0} times among the calls:", matchedCallCount);
+                switch (matchedCallCount)
+                {
+                    case 0:
+                        writer.Write("but was not found among the calls:");
```
thomaslevesque: Switching to the passive voice for this case is a bit strange... Maybe _"but didn't find it among the calls:"_ instead?

Mertsch: @thomaslevesque I chose "was not" over "didn't" since it's used more often in the whole solution
But I can also use "didn't" :-)
image
image

BTW how do I update a PR from my changes in VS?

Mertsch: I just noticed that a lot of those are "was not matched" and maybe those should be changed to "didn't match", too

thomaslevesque: > But I can also use "didn't" :-)

Let's leave @blairconrad a chance to review as well. He's a native English speaker, which I'm not ;)

> BTW how do I update a PR from my changes in VS?

You just push to the same branch. You can either add a commit, or amend the first one and force push.

Mertsch: @thomaslevesque Since he thumbed up, I have changed it.
Thanks for the hint, didn't expect things to be this comfortable in git :-)

thomaslevesque:
Thanks @Mertsch, looks good.

Just an issue I didn't notice before: you made your changes on your `develop` branch. They should be done on a feature branch instead (e.g. `call-count-wording`). Here's how to fix it:

- create a feature branch from the tip of your `develop` branch and switch to that branch:
    > git checkout -b call-count-wording develop
- push the new branch to GitHub:
    > git push -u origin call-count-wording
- recreate the pull request (PRs on Github are based on the branch name, so you can't change the branch once the PR is created)

While you're at it, would you mind squashing the commits before I merge? Having separate commits makes review easier, but the history is cleaner without the additional commits. Let me know if you need help for this.

thomaslevesque:
@Mertsch you seem to be having trouble with Git... Ignore the rebase for now, I can do it if you allow me to push to your branch (there's a checkbox for this in the PR page)

Mertsch:
@thomaslevesque I will never understand why git has to be such a CLI hell, but I think I finally managed to do it, see #1282

thomaslevesque:
> I will never understand why git has to be such a CLI hell

It's not so bad once you know it well enough, but I understand the feeling...

PR #1282, opened 2017-12-11
Use the same wording (never, once, twice, n times) when reporting act…

### LINKED ISSUE #1278: Inconsistent use of the # sign in error messages

Mertsch:
…ual amount of found calls

Fixes #1278

### DIFF
--- a/docs/formatting-argument-values.md
+++ b/docs/formatting-argument-values.md
@@ -23,7 +23,7 @@ This would help FakeItEasy display this error message:
 <pre>
 Assertion failed for the following call:
   SampleTests.ILibrary.Checkout(<Ignored>)
-Expected to find it never but found it #1 times among the calls:
+Expected to find it never but found it once among the calls:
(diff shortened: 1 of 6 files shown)


thomaslevesque:
Thanks @Mertsch! Looks good to me. Just one little thing: could you remove "was not found -> didn't find it" from the commit message? Just amend your last commit (`git commit --amend`) and force push to the same branch.

blairconrad:
@thomaslevesque, leaving in case you haven't yet had a chance to review the latest upload, but I'm more than content with the results.

thomaslevesque:
Very nice, thanks @Mertsch!

Issue #1284, opened 2018-01-10
Add reference to Microsoft.CSharp in the package


thomaslevesque:
This is required in order to use dynamic features, such as `Raise.FreeForm`.

Thanks to @MrVermond for [reporting the issue in the chat](https://gitter.im/FakeItEasy/FakeItEasy?at=5a55fec4ba39a53f1afb18e8).

thomaslevesque:
Adding `<Reference Include="Microsoft.CSharp" />` to the csproj is correctly reflected in the nuspec as a `<frameworkAssembly>` element, but this only works for the full .NET Framework. To support all target frameworks, we need to reference the `Microsoft.CSharp` NuGet package.

blairconrad:
@thomaslevesque, shall we consider this a bug fix, since we were not properly exposing our dependencies?

thomaslevesque:
> shall we consider this a bug fix

I wasn't sure, so I didn't add a tag... I guess we could consider this a bug. To be fair, FakeItEasy itself doesn't use any dynamic feature; the user does, although it's not obvious from the usage.

> since we were not properly exposing our dependencies?

Strictly speaking, it's not really our dependency, but we should probably make things easy for the user and add the reference when the package is installed.

thomaslevesque:
> Adding `<Reference Include="Microsoft.CSharp" />` to the csproj is correctly reflected in the nuspec as a `<frameworkAssembly>` element, but this only works for the full .NET Framework. To support all target frameworks, we need to reference the Microsoft.CSharp NuGet package.

Apparently the package isn't compatible with net40; it requires net45 or higher. So, we need a `PackageReference` for netstandard1.6, and a `Reference` for net40; for net45 it doesn't matter (the package just adds a reference to the framework assembly).

blairconrad:
Pretend I don't understand things. How are we affected if the client is writing in VB?

thomaslevesque:
> Pretend I don't understand things. How are we affected if the client is writing in VB?

Excellent question. My guess is that it should already work in VB, since VB.NET natively supports late binding. Only C# requires a special assembly for using dynamic features.

But a guess is not enough; I'll experiment a bit to be sure.

thomaslevesque:
My guess was wrong. It doesn't work at all in VB (although it compiles fine). Also, the syntax is ugly; I guess VB wasn't really considered when the `Raise` feature was designed.

[...]

PR #1286, opened 2018-01-12
Document Raise.FreeForm.With limitation and dependency

thomaslevesque:
Relates to #1284

### DIFF
--- a/docs/raising-events.md
+++ b/docs/raising-events.md
@@ -1,7 +1,6 @@
 # Raising events
 
-Let's say - for argument's sake - that we have an interface that has
-an event defined:
+Given a virtual or abstract event, or one defined on an interface:
 
 ```csharp
 public interface IRobot
@@ -56,10 +55,9 @@ fake.CustomEvent += Raise.FreeForm.With(fake, sampleCustomEventArgs);
 fake.FreeformEvent += Raise.FreeForm.With(7);
 ```
 
-Just as when we're trying to
-[override a method's behavior](what-can-be-faked#what-members-can-be-overriden.md),
-_for FakeItEasy to raise an event, the event must be virtual (if
-defined on a class) or defined on an interface_.
+Notes on `Raise.Freeform.With`:
+- it only works in C#, not in VB.NET, due to language limitations;
+- it uses late binding, so you need a reference to the `Microsoft.CSharp` assembly in order to use it.
 
 ## VB.Net
 


blairconrad:
I generally like the text that you added, but have some small concerns about the positioning (some of which are exacerbated by the way the article is already structured) and a little bit the phrasing. I had a number of thoughts when I read the new article. I'd consider addressing all of them, but perhaps not in this PR:

1. The use of the bulleted list does serve to separate the `Raise.Freeform.With` notes from the following general notes on the need for a virtual method, but I wonder if some readers won't lump the latter in under the former. Perhaps we should shift the "Just as when we're trying to…" elsewhere. But I'm not entirely sure where.
2. We have a VB.NET section (which could probably use a few more words of introduction around the code sample). Since we're now explicitly saying that `Raise.Freeform.With` is C#-only, perhaps we should introduce the a C# section as well.
3. Until (unless?) we have a new mechanism for invoking free-form delegates from VB.Net, we should probably restore the `Raise.With(Of TEventHandler )` documentation to the VB.Net section of the page.

If you disagree on any points, or agree, but think they should be done later, I'm happy to merge.
Also, I'm really attacking the article based on some pre-existing problems, so if you'd rather I made the new changes instead of dumping the work on you, I'm happy to do that too.

thomaslevesque:
You raise (haha!) valid points, but I don't really know how to address them...

> Perhaps we should shift the "Just as when we're trying to…" elsewhere. But I'm not entirely sure where.

I'm not sure either. Maybe closer to the top.

> We have a VB.NET section (which could probably use a few more words of introduction around the code sample). Since we're now explicitly saying that Raise.Freeform.With is C#-only, perhaps we should introduce the a C# section as well.

> Until (unless?) we have a new mechanism for invoking free-form delegates from VB.Net, we should probably restore the Raise.With(Of TEventHandler ) documentation to the VB.Net section of the page.

Can we defer those decisions until we know how to address the VB.NET issue?

blairconrad:
> You raise (haha!)

You made my morning.

> > Perhaps we should shift the "Just as when we're trying to…" elsewhere. But I'm not entirely sure where.

> I'm not sure either. Maybe closer to the top.

Probably. I think we didn't originally have the line, but added it after a user had difficulties faking an event.
We could just switch

```
Let's say - for argument's sake - that we have an interface that has an event defined:
```

for

```
Given a virtual or abstract event, or one defined on an interface: 
```

and drop the other note altogether.

thomaslevesque:
Sounds good to me. I made the changes you suggested.

blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
Thanks for the merge!

blairconrad:

PR #1289, opened 2018-01-19
Add Raise.FreeForm<TEventHandler> for Visual Basic

### LINKED ISSUE #1287: Raise.Freeform.With is not available from VB.Net and our only alternative is deprecated
spun off from https://github.com/FakeItEasy/FakeItEasy/issues/1284#issuecomment-356742263

@thomaslevesque suggested either un-deprecating
`Raise.With(Of PropertyChangedEventHandler)(foo, new PropertyChangedEventArgs("Blah"))`
or introducing something like
`Raise.FreeForm(Of PropertyChangedEventHandler).With(foo, new PropertyChangedEventArgs("Blah"))`
for VB.Net.

His rational for the latter is 
> …it's less confusing than having two generic overloads of `Raise.With` that are so similar yet different (the generic type argument means something different for `With<TEventArgs>` and `With<TEventHandler>`, which can possibly cause ambiguities)

blairconrad:
Fixes #1287.

### DIFF
--- a/docs/raising-events.md
+++ b/docs/raising-events.md
@@ -66,13 +66,23 @@ event FreeformEventHandler FreeformEvent;
 event CustomEventHandler CustomEvent;
 ```
 
-To do that, you can use `Raise.FreeForm.With`, which automatically infers the correct delegate type:
+### From C#
+To raise a free-form event from C#, use `Raise.FreeForm.With`, which automatically infers the correct delegate type:
 
 ```csharp
 fake.FreeformEvent += Raise.FreeForm.With(7);
 fake.CustomEvent += Raise.FreeForm.With(fake, sampleCustomEventArgs);
 ```
 
-Notes on `Raise.Freeform.With`:
-- it only works in C#, not in VB.NET, due to language limitations;
-- it uses late binding, so you need a reference to the `Microsoft.CSharp` assembly in order to use it.
+Due to language limitations, `Raise.Freeform.With` does not work in VB.NET, and it uses late binding, so you need a reference to the `Microsoft.CSharp` assembly in order to use it.
+
+### From VB.NET
+To raise a free-form event from VB.NET, you must use `Raise.FreeForm<TEventHandler>.With`:
+
+```csharp
+fake.FreeformEvent += Raise.FreeForm<FreeformEventHandler>.With(7);
+fake.CustomEvent += Raise.FreeForm<CustomEventHandler>.With(fake, sampleCustomEventArgs);
+```
+
+Specifying the type of the event handler gets around the language restrictions in VB.NET.
+This method may also be used from C# if you don't want to rely on late binding.
--- a/docs/raising-events.md
+++ b/docs/raising-events.md
@@ -1,6 +1,10 @@
 # Raising events
 
-Given a virtual or abstract event, or one defined on an interface:
+FakeItEasy can be used to simulate the raising of an event from a Fake object, assuming the event is virtual or abstract, or defined on an interface.
+
+## `EventHandler`-based events
+
+Suppose a standard `EventHandler`-based event such as this one:
 
 ```csharp
 public interface IRobot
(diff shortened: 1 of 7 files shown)


docs/raising-events.md
```diff
@@ -1,6 +1,10 @@
 # Raising events
 
-Given a virtual or abstract event, or one defined on an interface:
+FakeItEasy can be used to simulate the raising of an event from a Fake object, assuming the event is virtual or abstract, or one defined on an interface.
```
thomaslevesque: Maybe remove "one" from "one defined on an interface"? It sounds a bit strange

docs/raising-events.md
```diff
@@ -9,58 +13,76 @@ public interface IRobot
...
 
+```csharp
+fake.FreeformEvent += Raise.FreeForm<FreeformEventHandler>.With(7);
+fake.CustomEvent += Raise.FreeForm<CustomEventHandler>.With(fake, sampleCustomEventArgs);
 ```
-AddHandler robot.FellInLove, Raise.With(EventArgs.Empty)
-```
+
+Specifying the type of the event handler gets around the language restrictions in VB.NET.
+This method may also be used from C# if you want to avoid a reference to `Microsoft.CSharp`.
```
thomaslevesque: I don't think anyone really cares about referencing Microsoft.CSharp, but some people prefer to avoid late binding. I'd suggest changing this to "if you don't want to rely on late binding" or something like that.

tests/FakeItEasy.IntegrationTests.VB/RaisingEventsTests.vb
```diff
@@ -176,4 +176,21 @@ Public Class RaisingEventsTests
...
+        Dim target = A.Fake(Of IHaveEvents)()
+        Dim anObject As Object = New Object()
+
+        AddHandler target.ObjectEvent, AddressOf HandlesObjectEvent
+
+        ' Act
+        AddHandler target.ObjectEvent, Raise.FreeForm(Of IHaveEvents.ObjectEventEventHandler).With(anObject)
+
+        ' Assert
+        ReferenceEquals(capturedObject, anObject).Should().BeTrue()
```
thomaslevesque: Nitpick: could be written as `capturedObject.Should().BeSameAs(anObject)`

tests/FakeItEasy.Specs/EventRaisingSpecs.cs
```diff
@@ -390,6 +390,102 @@ public void ActionEvent()
...
+            "When I raise the event as freeform specifying the arguments"
+                .x(() => Fake.ActionEvent += Raise.FreeForm<Action<int, bool>>.With(19, true));
+
+            "Then the first value is passed as the first event argument"
+                .x(() => CapturedArgs1.Should().Be(19));
+
+            "Then the second value is passed as the second event argument"
+                .x(() => CapturedArgs2.Should().Be(true));
+        }
+
```
thomaslevesque: I didn't look closely, but I assume these scenarios are equivalent to the dynamic free-form ones?

blairconrad: Yup. Copy, paste, and the minimum modification to make them apply to the new method.

tests/FakeItEasy.IntegrationTests.VB/RaisingEventsTests.vb
blairconrad: '''

blairconrad: I think that's why we were using `ReferenceEquals` to begin with.  I can comment up, if you like.

blairconrad:
Quality comments, @thomaslevesque. Docs comments incorporated, and I commented on the `ReferenceEquals`.

tests/FakeItEasy.IntegrationTests.VB/RaisingEventsTests.vb
thomaslevesque: That's weird... there's no late binding involves here, as far as I can tell

thomaslevesque: Apparently VB.NET doesn't support extension methods on `Object`; this is related to the fact that there is no `dynamic` in VB.NET : anything typed as `Object` is late bound.

More details here: https://blogs.msdn.microsoft.com/vbteam/2007/01/24/extension-methods-and-late-binding-extension-methods-part-4/

blairconrad: Wow. Thanks for the note. If you'd like, I can introduce a method to help us document what's going on here. Or change `Object` to something else. It doesn't _need_ to be `Object`.

thomaslevesque: Not sure it's worth the trouble... Maybe changing to something other than `Object` would be enough.

blairconrad: Did so. But then noticed the "sender", which is an object. I pushed a thing. Hate it, if you want!

thomaslevesque: > Did so. But then noticed the "sender", which is an object. I pushed a thing. Hate it, if you want!

Perfect!

thomaslevesque:
I'll merge after you squash

blairconrad:
Squashed! Thanks for the great comments.

thomaslevesque:
Thanks @blairconrad !

PR #1291, opened 2018-01-25
Remove 'repeated' from completed call assertion messages

### LINKED ISSUE #1290: Word "repeated" in assertion failure messages may mislead reader
The presence of the word "repeated" in some of our assertion failure messages may be misleading. Consider this message issued when we expected to see `Bar("")` called at least twice, but instead `Bar()` was called twice:

```
  FakeItEasy.Tests.IFoo.Bar(argument: "")
Expected to find it at least twice but didn't find it among the calls:
  1: FakeItEasy.Tests.IFoo.Bar() repeated 2 times
```

If `Repeated` is confusing when specifying a desired call frequency, as is noted in #808, then it could be misinterpreted here, so the reader may think that `Bar()` was called and then was repeated twice more, for a total of 3 calls.

blairconrad:
Fixes #1290

### DIFF
--- a/tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
@@ -85,12 +85,12 @@ namespace FakeItEasy.Specs
                 .x(() => exception.Should().BeAnExceptionOfType<ExpectationException>().WithMessage(@"
 
   Assertion failed for the following calls:
-    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 1)' repeated exactly twice
-    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 2)' repeated at least once
-    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 3)' repeated at least once
+    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 1)' exactly twice
+    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 2)' at least once
+    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 3)' at least once
   The calls were found but not in the correct order among the calls:
     1: FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 3)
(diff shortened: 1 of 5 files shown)


tests/FakeItEasy.Specs/OrderedCallMatchingSpecs.cs
```diff
@@ -85,12 +85,12 @@ public static void OrderedAssertionsOutOfOrder(IFoo fake, Exception exception)
...
-    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 1)' repeated exactly twice
-    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 2)' repeated at least once
-    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 3)' repeated at least once
+    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 1)' exactly twice
+    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 2)' at least once
+    'FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 3)' at least once
   The calls were found but not in the correct order among the calls:
     1: FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 3)
-    2: FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 1) repeated 2 times
+    2: FakeItEasy.Specs.OrderedCallMatchingSpecs+IFoo.Bar(baz: 1) (2 times)
```
blairconrad: I don't think the difference between how the expected calls are rendered (with single quotes) and the found ones (without) is desirable, but I figure it's a different issue, so I wasn't going to attack at this point.

thomaslevesque: Agreed. Shouldn't we include parenthesis for the expected call count, though?

blairconrad:
```

blairconrad: For some reason the `()`s around the "at least once" don't sit well with m.e. I don't know why.

thomaslevesque: Dropping the parens everywhere sounds good to me 👍

blairconrad:
Wow. Quick merge. Thanks for the content change suggestion!

PR #1293, opened 2018-01-29
New fluent MustHaveHappened* API

### LINKED ISSUE #1292: Provide way to assert completed calls without using Repeated class
As mentioned by @jholovacs in #808, the `Repeated` in  `A.CallTo(() => f.Boo()).MustHaveHappened(Repeated.Exactly.Once)` could be construed as meaning "Boo must have happened, and then it must have been repeated, for a total of _two_ calls". After much discussion, we've resolved to provide alternative methods and eventually deprecate and remove `Repeated`.
[As noted in that issue](https://github.com/FakeItEasy/FakeItEasy/issues/808#issuecomment-359484707), we'll introduce the following API:

```c#
MustHaveHappened()    // as today
MustNotHaveHappened() // as today

MustHaveHappenedOnceExactly()
MustHaveHappenedOnceOrMore()
MustHaveHappenedOnceOrLess()

MustHaveHappenedTwiceExactly()
MustHaveHappenedTwiceOrMore()
MustHaveHappenedTwiceOrLess()

MustHaveHappened(3, Times.Exactly)
MustHaveHappened(3, Times.OrMore)
MustHaveHappened(3, Times.OrLess)

MustHaveHappenedANumberOfTimesMatching(n => n %2 == 0) // the argument is an Expression<Func<int, bool>>
```

`Times` is a new class.

This satisfies the following goals:

1. Eliminates the confusion over the exact semantics of `Repeated`
2. Retains `MustHaveHappened()` and `MustNotHaveHappened()`, which are the most popular methods in use today
3. Improves discoverability, even over the existing `Repeated` mechanism
4. Provides a consistent, natural word ordering (to contrast with  `Repeated.Times(3)`)
5. Does not require an extension method on `int`
6. Avoids forgettable terminal operators (e.g. `MustHaveHappened(3).Times`, where the `Times` could be forgotten)
7. Allows for a soft transition away from the current methods; the `Repeated` variants can live for a major release or two and be deprecated in 5.0.0 and removed in 6.0.0.

blairconrad:
Fixes #1292.


blairconrad:
Some notes on the commits: 
* b22d60a Add Specs for MustHaveHappened, MustNotHaveHappened
    Adds specs for pre-existing methods
* bb6ab9e Implement new fluent MustHaveHappened* methods
    The work of implementing the new API     
* 02f85c6 Prefer never, once, and twice to 0, 1, or 2 times
    An optional "improvement". Makes us output "expected never" or "once" or "twice" even when building an assertion like `MustHaveHappened(n, Times.Exactly)`, similar to what was done in #1278
* 9f9958c Prefer new MustHaveHappened* methods in tests
    Another optional improvement. If the new API is the blessed one, we might as well use it ourselves. When the old methods were used in the "Act" phase of tests, I duplicated, so we wouldn't accidentally break them, but if they were used in our own "Assert"ions, I just replace with the new ones

blairconrad:
Oh, I should've said. I'd want to continue on and invert the new API→`Repeated` dependency, so that `Repeated` could be removed more easily. I can pursue in this PR or in a future one; it's all the same to me.

src/FakeItEasy/Configuration/AssertConfigurationBase.cs
```diff
@@ -0,0 +1,28 @@
+namespace FakeItEasy.Configuration
+{
+    using System;
+    using System.Linq.Expressions;
+
+    /// <inheritdoc />
+    /// <summary>
+    /// Provides extension methods for <see cref="IAssertConfiguration"/>.
+    /// </summary>
+    internal abstract class AssertConfigurationBase : IAssertConfiguration
```
thomaslevesque: Is this class intended to remain after you invert the dependency between `Repeated` and the new API? The fact that the root configuration classes (`RuleBuilder`, `AnyCallConfiguration`, `PropertySetterConfiguration`) inherit from feels awkward, IMO... I think it would be better to duplicate the code in this case.

tests/FakeItEasy.Specs/AssertingCallCountSpecs.cs
```diff
@@ -0,0 +1,1402 @@
...
+
+            "When I assert that the call must not have happened"
+                .x(() => exception = Record.Exception(() => A.CallTo(() => fake.Method()).MustNotHaveHappened()));
+
+            "Then the assertion succeeds"
+                .x(() => exception.Should().BeNull());
+        }
+
+        [Scenario]
+        public static void MustNotHaveHappened1Call(IFoo fake, Exception exception)
```
thomaslevesque: Would it be possible to parameterize the scenarios? There's a lot of repetition, this is going to be a pain to maintain...

thomaslevesque:
> Oh, I should've said. I'd want to continue on and invert the new API→Repeated dependency, so that Repeated could be removed more easily. I can pursue in this PR or in a future one; it's all the same to me.

I think it should be done in the same PR.

(I like your optional commits, btw. Let's keep them!)

tests/FakeItEasy.Specs/AssertingCallCountSpecs.cs
blairconrad: It was a pain to write! And if it's hard to write, it should be harder to maintain.

src/FakeItEasy/Configuration/AssertConfigurationBase.cs
blairconrad: I've no strong feelings about it remaining. I waffled over adding it at all. Happy to push the methods down to the (5) concrete classes.
That raises a question about all the testing, though. The "must be null guarded" tests I added went directly at this base class (a bit of a cheat), so I'm considering duplicating (probably parameterizing) for all the concrete instances. Is okay?

And if that's okay, should I do it to the new Specs I added? Currently they just hit `RuleBuilder`, not `AnyCallConfiguration, `PropertySetterConfiguration`, `PropertySetterAdapter`, or `ReturnValueConfiguration`.

src/FakeItEasy/Configuration/AssertConfigurationBase.cs
```diff
@@ -0,0 +1,28 @@
+namespace FakeItEasy.Configuration
+{
+    using System;
+    using System.Linq.Expressions;
+
+    /// <inheritdoc />
+    /// <summary>
+    /// Provides extension methods for <see cref="IAssertConfiguration"/>.
```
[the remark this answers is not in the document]
blairconrad: Yes. Yes they aren't. Um. They used to be before I moved the code (in steps you didn't see).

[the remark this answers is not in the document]
blairconrad: But nice catch. I really tried to reread all the test and docs changes. Glad you're here.

src/FakeItEasy/Configuration/AssertConfigurationBase.cs
thomaslevesque: In fact, I realize what bothers me is the fact that the base class for all configuration classes is named `AssertConfigurationBase`. It makes it look as if all derived classes have something to do with configuration (which they do, but it's only one of their aspects). If we needed to refactor some other aspect of these classes into a base class, the name would no longer fit, and we can't have multiple base classes.

So maybe just renaming the class to `ConfigurationBase` would be enough.

thomaslevesque: Anyway, this class makes sense now because you're relying on the legacy API to implement the new one, but will it still make sense after you refactor?

thomaslevesque: > That raises a question about all the testing, though. The "must be null guarded" tests I added went directly at this base class (a bit of a cheat), so I'm considering duplicating (probably parameterizing) for all the concrete instances. Is okay?
> 
> And if that's okay, should I do it to the new Specs I added? Currently they just hit RuleBuilder, not AnyCallConfiguration,PropertySetterConfiguration,PropertySetterAdapter, or ReturnValueConfiguration.

Good points.

I just looked at the code coverage, the current `MustHaveHappened` method is covered for `RuleBuilder`, `ReturnValueConfiguration` and `AnyCallConfiguration`, but not for `PropertySetterConfiguration`/`PropertySetterAdapter`. Ideally, the specs should cover all implementations, but I realize it's a lot of work... Maybe it can be postponed for now.

tests/FakeItEasy.Tests/Configuration/AssertConfigurationTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            var unnaturalFake = new Fake<IFoo>();
+            var fake = unnaturalFake.FakedObject;
+            return TestCases.FromObject(
+                A.CallTo(() => fake.VoidMethod()),
+                unnaturalFake.AnyCall(),
+                unnaturalFake.AnyCall().WithNonVoidReturnType(),
+                A.CallToSet(() => fake.WriteableProperty),
+                A.CallToSet(() => fake.WriteableProperty).To(7));
+        }
```
thomaslevesque: Nice! But why use an unnatural fake?

blairconrad: It was the call that I found that got me the assertion object I wanted.

blairconrad:
Specs have been parameterized, including adding coverage for all 5 implementations of `IAssertConfiguration`. I'll wait for comments before pushing commits that invert `Repeated` and `Times`.

blairconrad:
Thanks, @thomaslevesque. Since you seem to be happy so far, I'm going to take the perhaps odd step of squashing before I continue. The commits are getting ugly, and the spec conversion one in particular fights with "f9ada278344f77e5558c751f766239bf40abe1b7 Prefer never, once, and twice to 0, 1, or 2 times" due to the changed phrasing of the call counts in the exception messages.

Pre-squashed version is at https://github.com/blairconrad/FakeItEasy/tree/d893f0b13c2b47f01b3bb0244ebd3bbf3be5f469, and squashed at https://github.com/blairconrad/FakeItEasy/tree/60b5c413b56561970154d802367f578e64a83e2b, in case you want to check 'em to keep me honest.

thomaslevesque:
Yeah, applying the fixups is probably going to be a pain... go ahead and squash!

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -141,11 +142,21 @@ public UnorderedCallAssertion MustHaveHappened(Repeated repeatConstraint)
...
 
-            return new UnorderedCallAssertion(this.manager, this.Matcher, this.RuleBeingBuilt.WriteDescriptionOfValidCall, repeatConstraint);
+            return this.MustHaveHappened(timesOption.ToCallCountConstraint(numberOfTimes));
+        }
+
+        public UnorderedCallAssertion MustHaveHappenedANumberOfTimesMatching(Expression<Func<int, bool>> predicate)
+        {
+            Guard.AgainstNull(predicate, nameof(predicate));
+
+            return this.MustHaveHappened(new CallCountConstraint(predicate.Compile(), $"the number of times specified by the predicate '{predicate}'"));
```
thomaslevesque: Maybe _"a number of times matching the predicate"_ instead of _"the number of times specified by the predicate"_? It would be more accurate, and would fit the API language better.

blairconrad: Oh, fantastic observation, @thomaslevesque. Tell you what. I'll bundle [a slightly modified version of] it in with another commit that I was planning on sending anyhow. I look forward to your reaction.

thomaslevesque: Very nice! 👍

blairconrad:
Oh, I have one more coming! As for squashing, I had been content about the separate commits (although if you prefer one commit, I can do that, or if you prefer selected squashing).

blairconrad:
I thought also to remove our less-obvious use of "repeat", at least as much as I could without triggering a change in the API Approval docs.
I can remove the commit if you prefer. And I'll squash as much as you want.

thomaslevesque:
> As for squashing, I had been content about the separate commits (although if you prefer one commit, I can do that, or if you prefer selected squashing).

It's fine as it is!

> I thought also to remove our less-obvious use of "repeat", at least as much as I could without triggering a change in the API Approval docs.

Looks good to me.

Besides the to-be-deprecated `Repeated`/`IRepeatSpecification`, the only remaining ambiguous API is `IRepeatConfiguration` (used for call behavior specifications, not assertions), and the user doesn't manipulate it explicitly; we should probably rename it in the next major release anyway.

blairconrad:
Thanks, @thomaslevesque. Issue created. And if you're still happy with the number of commits, the PR is ripe for merging, as I've remove "WIP". Thanks for your feedback!

thomaslevesque:
Thanks @blairconrad! It was a much bigger PR than I expected, but I like the end result ;)

src/FakeItEasy/Configuration/IRepeatConfiguration.cs
```diff
@@ -9,7 +9,7 @@ public interface IRepeatConfiguration<out TInterface> : IHideObjectMembers
         /// <summary>
         /// Specifies the number of times for the configured event.
         /// </summary>
-        /// <param name="numberOfTimesToRepeat">The number of times to repeat.</param>
```
thomaslevesque: I missed that when I reviewed... You changed the parameter name in the doc comments, but not in the method signature itself ;)
I'd like to change the name, but it's a public API, so technically it would be a breaking change... Not a big one, though; I doubt anyone writes `.NumberOfTimes(numberOfTimesToRepeat: 3)`.
Anyway, I'd like to change the method description as well. It's weird that it mentions an "event". Something like "the number of times the configured behavior should be applied" would be better, I think.

PR #1300, opened 2018-02-03
Change documentation comments for IRepeatConfiguration

thomaslevesque:
As mentioned in https://github.com/FakeItEasy/FakeItEasy/pull/1293/files#r165813185

The first commit
- fixes the inconsistency in the parameter name between doc comments and actual signature
- changes the wording of the interface and method summaries

The second commits actually renames the parameter to `numberOfTimes`. I did it as a separate commit, because it's a breaking change. I think we could still do it, because the chances that it will actually impact anyone are extremely low. If you disagree, I'll remove the commit.

Contributes to #1292.

### DIFF
--- a/src/FakeItEasy/Configuration/IRepeatConfiguration.cs
+++ b/src/FakeItEasy/Configuration/IRepeatConfiguration.cs
@@ -9,8 +9,8 @@ namespace FakeItEasy.Configuration
         /// <summary>
         /// Specifies the number of times the configured behavior should be applied.
         /// </summary>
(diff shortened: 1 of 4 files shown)


blairconrad:
Oh, nice catch, @thomaslevesque. I ruined this when I was trying to rename the parameter to `numberOfTimes` and was scared off by the need to change the approval docs. Then I partially reverted.
I think you're right, though. We can keep the commit.

blairconrad:

PR #1301, opened 2018-02-03
Fix setter description in assertion

### LINKED ISSUE #1298: Call description is wrong when asserting a call to a property setter
Consider this code:

```c#
public interface IFoo {  int Bar { get; set; } }
public void Test()
{
    var foo = A.Fake<IFoo>();
    foo.Bar = 0;
    A.CallToSet(() => foo.Bar).To(42).MustHaveHappened();
}
```
As expected, this throws an `ExpectationException`, but the call description in the exception message is wrong:

> Assertion failed for the following call:
>     Castle.Proxies.ObjectProxy_12.set_Bar(value: 42)
>   Expected to find it once or more but didn't find it among the calls:
>     1: UserQuery+IFoo.Bar = 0

It should say `UserQuery+IFoo` instead of `Castle.Proxies.ObjectProxy_12`. It works as expected for method calls and property getters, the issue seems to exist only for property setters.

thomaslevesque:
Fixes #1298 

I didn't find a good place to put those assertions, so I created a new class.

### DIFF
--- a/src/FakeItEasy/Configuration/PropertyExpressionHelper.cs
+++ b/src/FakeItEasy/Configuration/PropertyExpressionHelper.cs
@@ -26,8 +26,8 @@ namespace FakeItEasy.Configuration
                 .Concat(new[] { parsedCallExpression.CalledMethod.ReturnType })
                 .ToArray();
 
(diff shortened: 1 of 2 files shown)


blairconrad:
Nice, @thomaslevesque!

PR #1303, opened 2018-02-03
Make description of asserted call consistent with actual calls for property accessors

### LINKED ISSUE #1299: Make descriptions of calls to property accessors more consistent
When we assert calls on property accessors, the descriptions are inconsistent between the asserted calls and the list of actual calls made to the fake.

Currently we have this for the asserted calls:
- getter: `IFoo.get_Bar()`
- setter: `IFoo.set_Bar(value: 42)` (*)

But in the list of actual calls, we have this: 
- getter: `IFoo.Bar`
- setter: `IFoo.Bar = 42`

Which I think is better. Anyway, we should probably make the descriptions consistent with each other.

(*) That's a lie. In fact we have `Castle.Proxies.ObjectProxy_12` instead of `IFoo`, but it's another issue (#1298)

thomaslevesque:
Fixes #1299

Note: Don't merge yet. This PR is built on top of #1301, so it includes the commits from that branch. I'll rebase it on develop after #1301 is merged.

### DIFF
--- a/src/FakeItEasy/Expressions/ExpressionCallMatcher.cs
+++ b/src/FakeItEasy/Expressions/ExpressionCallMatcher.cs
@@ -71,8 +71,7 @@ namespace FakeItEasy.Expressions
 
             result.Append(this.Method.DeclaringType);
             result.Append(".");
-            result.Append(this.Method.Name);
-            result.Append(this.Method.GetGenericArgumentsString());
+            this.AppendMethodName(result);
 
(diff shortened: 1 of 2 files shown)


blairconrad:
Thanks, @thomaslevesque. #1301 is merged. Rebase away, and I think we're good to merge!

thomaslevesque:
Done!

thomaslevesque:
Thanks for the merge!

Issue #1297, opened 2018-02-02
Add code fix provider to convert assertions to new API


blairconrad:
#1292 added a new assertion API, superseding the old version that had methods using the `Repeated` class.
The analyzer could help people convert the old uses into new ones.

One danger of doing this is that someone may use a FakeItEasy analyzer from the future, that is, a version with the code fix provider, while targeting a FakeItEasy version that doesn't contain the new API.
It's probably not very likely, so we should consider a way to reduce the risk. Perhaps (and I'm just thinking out loud here) we could interrogate the `IAssertConfiguration` instance to see if it has the new methods before offering to convert to them.

thomaslevesque:
> Perhaps (and I'm just thinking out loud here) we could interrogate the `IAssertConfiguration` instance to see if it has the new methods before offering to convert to them.

Yes, I think we can do this.

thomaslevesque:
> blairconrad added the in-progress label 11 hours ago

Wow, already?

blairconrad:
> Wow, already?

Oh, :blush:. I figured I could always un-progress it if someone disagreed.

blairconrad:
I'd like to offer the "nicest" options possible for code fixes, that is, I'd like to offer these fixes, where the user could choose the format they wanted when there are multiple conversions from a given source assertion:

```c#
MustHaveHappened(Repeated.Never) → MustNotHaveHappened()

MustHaveHappened(Repeated.Exactly.Once)    → MustHaveHappenedOnceExactly()
MustHaveHappened(Repeated.AtLeast.Once)    → MustHaveHappenedOnceOrMore()
MustHaveHappened(Repeated.NoMoreThan.Once) → MustHaveHappenedOnceOrLess()

MustHaveHappened(Repeated.Exactly.Twice)    → MustHaveHappenedTwiceExactly()
MustHaveHappened(Repeated.AtLeast.Twice)    → MustHaveHappenedTwiceOrMore()
MustHaveHappened(Repeated.NoMoreThan.Twice) → MustHaveHappenedTwiceOrLess()

MustHaveHappened(Repeated.Exactly.Times(n))    → MustHaveHappened(n, Times.Exactly)
MustHaveHappened(Repeated.AtLeast.Times(n))    → MustHaveHappened(n, Times.OrMore)
MustHaveHappened(Repeated.NoMoreThan.Times(n)) → MustHaveHappened(n, Times.OrLess)

[...]

PR #1304, opened 2018-02-05
Use StyleCop.Analyzers 1.1.0-beta006

blairconrad:
I wanted to use a `!` with pattern matching as part of my work on #1297.
StyleCop.Analyzers couldn't handle it, thinking that 
```c#
if (!(context.Node is InvocationExpressionSyntax call))
```

had too many parentheses. I noticed the bug was fixed in 1.1.0-beta006, so thought, "why not upgrade?".

### DIFF
--- a/samples/FakeItEasy.Examples.ruleset
+++ b/samples/FakeItEasy.Examples.ruleset
@@ -13,5 +13,6 @@
     <Rule Id="CA1824" Action="None" />
     <Rule Id="CA2201" Action="None" />
     <Rule Id="CA2210" Action="None" />
+    <Rule Id="SA0001" Action="None" />
   </Rules>
 </RuleSet>
--- a/src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchAnalyzer.cs
+++ b/src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchAnalyzer.cs
@@ -158,17 +158,17 @@ namespace FakeItEasy.Analyzer
                     }
 #if CSHARP
                 case ElementAccessExpressionSyntax elementAccess:
-                {
-                    var indexer = SymbolHelpers.GetAccessedIndexerSymbol(elementAccess, context);
-                    if (indexer == null)
                     {
-                        return false;
-                    }
+                        var indexer = SymbolHelpers.GetAccessedIndexerSymbol(elementAccess, context);
+                        if (indexer == null)
+                        {
+                            return false;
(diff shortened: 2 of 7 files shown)


samples/FakeItEasy.Examples.ruleset
```diff
@@ -13,5 +13,6 @@
     <Rule Id="CA1824" Action="None" />
     <Rule Id="CA2201" Action="None" />
     <Rule Id="CA2210" Action="None" />
+    <Rule Id="SA0001" Action="None" />
```
blairconrad: 1.1.0-beta006 doesn't like it when a project is set to not generate XML docs. We don't really need them for the Examples or Tests.

src/FakeItEasy.Analyzer/ArgumentConstraintTypeMismatchAnalyzer.cs
```diff
@@ -158,17 +158,17 @@ protected override void AnalyzeArgumentConstraintCore(string propertyFullName, S
...
+                        }
 
-                    methodOrIndexerName = indexer.Name;
-                    parameters = indexer.Parameters;
-                    return true;
-                }
+                        methodOrIndexerName = indexer.Name;
+                        parameters = indexer.Parameters;
+                        return true;
+                    }
```
blairconrad: The analyzer noticed that a few files had inconsistent indentation.

src/FakeItEasy.ruleset
```diff
@@ -11,6 +11,7 @@
     <Rule Id="SA1124" Action="Warning" />
     <Rule Id="SA1127" Action="None" />
     <Rule Id="SA1128" Action="None" />
+    <Rule Id="SA1413" Action="None" />
```
blairconrad: The analyzer [prefers a trailing comma in multiline initializers](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1413.md). I do too, but there are many instances where we don't use one, and I wasn't sure of the prevailing sentiment. I'm happy to remove this and add the commas…

thomaslevesque: I usually don't add a trailing comma, but I don't care if there is. Suppressing the warning is fine, as far as I'm concerned.

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thanks, @thomaslevesque!

blairconrad:

PR #1305, opened 2018-02-05
Add Code Fix to replace MustHaveHappened(Repeated)

### LINKED ISSUE #1297: Add code fix provider to convert assertions to new API

blairconrad:
Fixes #1297


blairconrad:
A first stab at the analyzer we'll need to support the code fix.

src/FakeItEasy.Analyzer/RepeatedAssertionAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+            if (methodSymbol?.GetFullName() != "FakeItEasy.Configuration.IAssertConfiguration.MustHaveHappened")
+            {
+                return;
+            }
+
+            // call is an InvocationExpression, with two children - a SimpleMemberAccessExpression
+            // (like A.CallTo(() => fake.Method()).MustHaveHappened), and an argument list.
+            // The SimpleMemberAccessExpression's first child is the IAssertConfiguration (for example,
+            // A.CallTo(() => fake.Method())), and the second is the IdentifierName MustHaveHappened.
+            var memberAccessNode = call.ChildNodes().FirstOrDefault();
```
thomaslevesque: [`call.Expression`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntax.invocationexpressionsyntax.expression?view=roslyn-dotnet) would achieve the same result, and doesn't rely on the order of child nodes.

src/FakeItEasy.Analyzer/RepeatedAssertionAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+            var mustHaveHappenedNode = memberAccessNode?.ChildNodes().Skip(1).FirstOrDefault();
```
thomaslevesque: [`MemberAccessExpressionSyntax.Name`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntax.memberaccessexpressionsyntax.name)

src/FakeItEasy.Analyzer/RepeatedAssertionAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            if (!(context.Node is InvocationExpressionSyntax call))
+            {
+                return;
+            }
+
+            if (call.ArgumentList.Arguments.Count != 1)
+            {
+                return;
+            }
```
thomaslevesque: With the current API, it's enough to check the number of arguments, but if we add a new overload of `MustHaveHappened` with a single argument, this analyzer will break. Why not check if `Repeated` is used in the argument list instead?

src/FakeItEasy.Analyzer/RepeatedAssertionAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+            // call is an InvocationExpression, with two children - a SimpleMemberAccessExpression
+            // (like A.CallTo(() => fake.Method()).MustHaveHappened), and an argument list.
+            // The SimpleMemberAccessExpression's first child is the IAssertConfiguration (for example,
+            // A.CallTo(() => fake.Method())), and the second is the IdentifierName MustHaveHappened.
+            var memberAccessNode = call.ChildNodes().FirstOrDefault();
+            var mustHaveHappenedNode = memberAccessNode?.ChildNodes().Skip(1).FirstOrDefault();
+            if (mustHaveHappenedNode == null)
+            {
+                return;
+            }
```
thomaslevesque: You already checked that the call was to the `MustHaveHappened` method, so I don't think this can be null.

src/FakeItEasy.Analyzer/RepeatedAssertionAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            var descriptor = DiagnosticDefinitions.RepeatedAssertion;
+            var diagnostic = Diagnostic.Create(descriptor, mustHaveHappenedNode.GetLocation(), call.ToString());
```
thomaslevesque: This will put the squiggles only on the `MustHaveHappened` token. Ideally the argument list should also be included. This can be achieved by creating a location that starts at the start of  `mustHaveHappened` and ends at the end of `call.ArgumentList`

blairconrad:
> I have a few remarks, mostly regarding the way to find specific nodes.

I welcome them!

> BTW, the Roslyn Syntax Visualizer is very helpful to understand the syntax tree for a given expression.

Yes. Without it, I woudn't've gotten as far as I did. I am terrible at this.

src/FakeItEasy.Analyzer/RepeatedAssertionAnalyzer.cs
blairconrad: Ha. I'd figured this sort of thing out when working on the code fix and hadn't brung it back. Thanks!

src/FakeItEasy.Analyzer/RepeatedAssertionAnalyzer.cs
blairconrad: I wonder if my problem before was that I had "Info" level squiggles, because

src/FakeItEasy.Analyzer/RepeatedAssertionAnalyzer.cs
```diff
@@ -0,0 +1,76 @@
...
+            context.ReportDiagnostic(Diagnostic.Create(descriptor, location));
+        }
+
+        private static string GetTypeName(SyntaxNodeAnalysisContext context, ArgumentSyntax argument)
+        {
+#if CSHARP
+            return context.SemanticModel.GetTypeInfo(argument.Expression).ConvertedType.ToString();
+#elif VISUAL_BASIC
+            return context.SemanticModel.GetTypeInfo(argument.GetExpression()).ConvertedType.ToString();
+#endif
```
blairconrad: I look forward to learning how I could've done this better!

thomaslevesque: I don't think you could have done it better.

blairconrad: :blush:

blairconrad:
I think I got all your earlier comments (good ones, BTW).
I'd add my code fix provider, but am for some reason having problems with the "suppress the diagnostics if FakeItEasy is too old" portion. I don't have unit tests, and they may be too cumbersome, but I took a stab, using the presence of `MustHaveHappenedANumberOfTimesMatching` to indicate that we have the new API. And... nothing. For some reason, I'm falling through the check every time - it's as if the member is always there, regardless of the version of FakeItEasy the host project references.
I'll work on it a little later at home, but anyone want to take a gander in the meantime?
https://github.com/blairconrad/FakeItEasy/blob/add-code-fix/src/FakeItEasy.Analyzer/RepeatedAssertionAnalyzer.cs#L59.

blairconrad:
>  am for some reason having problems with the "suppress the diagnostics if FakeItEasy is too old" portion.

working now. I'm not sure what's going on, but I seem to have to remove the analyzer, shut down my test solution, start it up, and add a new build in order to get things picked up. Anyhow, I see the right behaviour now.

I've added the changes for the code fix provider. My approach _may_ be unconventional. I do a little extra work in the analyzer when there's a finding, and then shove bonus information over to the code fix provider in the diagnostic's properties. These help build the code fix provider title with little work, and contribute to running the fix as well. I can try to rebalance the work if desired.

thomaslevesque:
> I'm not sure what's going on, but I seem to have to remove the analyzer, shut down my test solution, start it up, and add a new build in order to get things picked up.

Yeah, testing analyzers is a pain...

blairconrad:
> Very nice, @blairconrad!
>
> In addition to my other remarks, have you considered adding support for "fix all occurrences"? I think it's as simple as this:
> 
>    `public override FixAllProvider GetFixAllProvider() => WellKnownFixAllProviders.BatchFixer;`

It _was_ as simple as that! Unfortunately, with my very specific code fix titles, we then have something like this:

image

where the suggestion is that we'll change all assertions like this to  `MustHaveHappenedANumberOfTimesMatching`, when really we'll change all assertions  using `Repeated` to the preferred call. I wonder if I shouldn't go with the generic title if we're going to have the bulk update option (which we totally should!).

thomaslevesque:
> I wonder if I shouldn't go with the generic title if we're going to have the bulk update option (which we totally should!).

That's what I was going to suggest. Maybe "Change assertion to use the new API"? The preview would still show the result, so we don't need it in the title. It also makes the implementation simpler.

thomaslevesque:
> I'll merge after you squash.

And after you un-WIP, of course. Is there anything else you want to add?

blairconrad:
Thanks for noticing the whitespace. I fixed up while squashing.
I agree that the solution is better now. Thanks for your suggestions!
And no, nothing else. I just forgot to un-WIP *blush*. Done.

thomaslevesque:
Thanks @blairconrad! This analyzer was a great idea, it will make it very easy to migrate to the new API.

thomaslevesque:
BTW, we should probably mention the code fix [in the docs](http://fakeiteasy.readthedocs.io/en/stable/assertion/#using-the-legacy-musthavehappenedrepeated-overload).

PR #1312, opened 2018-02-11
Enable code analysis

### LINKED ISSUE #1144: Find replacement for FxCop (VS "Code Analysis")

thomaslevesque:
Fixes #1144

The `Microsoft.CodeAnalysis.FxCopAnalyzers` package now seems sufficiently stable.

There are many warnings that weren't previously reported by FxCop, so I had to fix or suppress many of them. I tried to make commits easy to follow, hopefully it won't be to painful to review...

### DIFF
--- a/src/FakeItEasy.ruleset
+++ b/src/FakeItEasy.ruleset
@@ -1,8 +1,9 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <RuleSet Name="FakeItEasy" ToolsVersion="11.0">
   <Include Path="allrules.ruleset" Action="Default" />
   <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
     <Rule Id="CA1020" Action="None" />
+    <Rule Id="CA1200" Action="None" />
     <Rule Id="CA1305" Action="None" />
   </Rules>
   <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
--- a/src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
@@ -1,4 +1,4 @@
-﻿namespace FakeItEasy.Core
+namespace FakeItEasy.Core
 {
     using System;
     using System.Collections.Generic;
@@ -14,12 +14,14 @@
         private class ObjectMemberRule
             : IFakeObjectCallRule
         {
+#pragma warning disable CA2235 // Mark all non-serializable fields
             private static readonly List<MethodInfo> ObjectMethods =
+#pragma warning restore CA2235 // Mark all non-serializable fields
                 new List<MethodInfo>
                     {
                         typeof(object).GetMethod("Equals", new[] { typeof(object) }),
-                        typeof(object).GetMethod("ToString", new Type[] { }),
-                        typeof(object).GetMethod("GetHashCode", new Type[] { })
+                        typeof(object).GetMethod("ToString", Type.EmptyTypes),
+                        typeof(object).GetMethod("GetHashCode", Type.EmptyTypes)
                     };
 
             private readonly FakeManager fakeManager;
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -1,4 +1,4 @@
-﻿namespace FakeItEasy.Core
+namespace FakeItEasy.Core
 {
     using System;
     using System.Collections.Concurrent;
@@ -18,10 +18,12 @@
     {
         private readonly CallRuleMetadata[] postUserRules;
         private readonly CallRuleMetadata[] preUserRules;
+#pragma warning disable CA2235 // Mark all non-serializable fields
         private readonly LinkedList<IInterceptionListener> interceptionListeners;
         private readonly WeakReference objectReference;
 
         private ConcurrentQueue<ICompletedFakeObjectCall> recordedCalls;
+#pragma warning restore CA2235 // Mark all non-serializable fields
 
         /// <summary>
         /// Initializes a new instance of the <see cref="FakeManager"/> class.
@@ -66,13 +68,16 @@
         /// <summary>
         /// Gets the faked proxy object.
         /// </summary>
-        [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "Object", Justification = "Renaming this property would be a breaking change.")]
+#pragma warning disable CA1716, CA1720 // Identifier contains keyword, Identifier contains type name
         public virtual object Object => this.objectReference.Target;
+#pragma warning restore CA1716, CA1720 // Identifier contains keyword, Identifier contains type name
 
         /// <summary>
         /// Gets the faked type.
         /// </summary>
+#pragma warning disable CA2235 // Mark all non-serializable fields
         public virtual Type FakeObjectType { get; }
+#pragma warning restore CA2235 // Mark all non-serializable fields
 
         /// <summary>
         /// Gets the interceptions that are currently registered with the fake object.
@@ -82,7 +87,9 @@
             get { return this.AllUserRules.Select(x => x.Rule); }
         }
 
+#pragma warning disable CA2235 // Mark all non-serializable fields
         internal LinkedList<CallRuleMetadata> AllUserRules { get; }
+#pragma warning restore CA2235 // Mark all non-serializable fields
 
         private IEnumerable<CallRuleMetadata> AllRules =>
             this.preUserRules.Concat(this.AllUserRules.Concat(this.postUserRules));
--- a/src/FakeItEasy/Core/MethodInfoManager.cs
+++ b/src/FakeItEasy/Core/MethodInfoManager.cs
@@ -160,7 +160,7 @@ namespace FakeItEasy.Core
             return type.GetInterfaces().Any(x => x.Equals(interfaceType));
         }
 
-        private struct TypeMethodInfoPair
+        private struct TypeMethodInfoPair : IEquatable<TypeMethodInfoPair>
         {
             public TypeMethodInfoPair(Type type, MethodInfo methodInfo)
                 : this()
@@ -186,6 +186,11 @@ namespace FakeItEasy.Core
             {
                 var other = (TypeMethodInfoPair)obj;
 
+                return this.Equals(other);
+            }
+
+            public bool Equals(TypeMethodInfoPair other)
+            {
                 return this.Type == other.Type && this.MethodInfo == other.MethodInfo;
             }
         }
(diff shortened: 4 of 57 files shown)


src/FakeItEasy.ruleset
```diff
@@ -1,9 +1,11 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <RuleSet Name="FakeItEasy" ToolsVersion="11.0">
   <Include Path="allrules.ruleset" Action="Default" />
   <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
     <Rule Id="CA1020" Action="None" />
+    <Rule Id="CA1200" Action="None" />
```
thomaslevesque: "Avoid using cref tags with a prefix". Apparently it doesn't like the `<see cref="!:http://..." />` syntax. This isn't a standard syntax, so I think what it actually doesn't like is prefixes such as "M:" (method), "T:" (type) etc; the "!:" prefix was just caught accidentally.

src/FakeItEasy.ruleset
```diff
@@ -1,9 +1,11 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <RuleSet Name="FakeItEasy" ToolsVersion="11.0">
   <Include Path="allrules.ruleset" Action="Default" />
   <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
     <Rule Id="CA1020" Action="None" />
+    <Rule Id="CA1200" Action="None" />
     <Rule Id="CA1305" Action="None" />
+    <Rule Id="CA2235" Action="None" />
```
thomaslevesque: "Mark all non-serializable fields". I think occurrences of this warning are false positive, because they only appears in the netstandard1.6 build, where serialization is disabled... It seems to be ignoring the `#if FEATURE_BINARY_SERIALIZATION`.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -66,8 +66,9 @@ internal FakeManager(Type fakeObjectType, object proxy, IFakeManagerAccessor fak
         /// <summary>
         /// Gets the faked proxy object.
         /// </summary>
-        [SuppressMessage("Microsoft.Naming", "CA1716:IdentifiersShouldNotMatchKeywords", MessageId = "Object", Justification = "Renaming this property would be a breaking change.")]
```
thomaslevesque: I think we could remove all suppression attributes and use `pragma warning disable` instead.

src/FakeItEasy/Core/MethodInfoManager.cs
```diff
@@ -160,7 +160,7 @@ private static bool TypeImplementsInterface(Type type, Type interfaceType)
             return type.GetInterfaces().Any(x => x.Equals(interfaceType));
         }
 
-        private struct TypeMethodInfoPair
+        private struct TypeMethodInfoPair : IEquatable<TypeMethodInfoPair>
```
thomaslevesque: I don't remember the warning number, but apparently you should implement `IEquatable<T>` when you override `Equals`.

tests/FakeItEasy.IntegrationTests/DummyTests.cs
```diff
@@ -55,8 +55,8 @@ public void Same_dummy_type_can_be_created_concurrently_on_different_threads()
             // the other is still being created.
             // Guards against the concurrent creations being detected as a
             // circular dependency and throwing an exception.
-            var makeOne = Task.Factory.StartNew(A.Dummy<ConcurrentCreationDummy>);
-            var makeTwo = Task.Factory.StartNew(() =>
+            var makeOne = Task.Run(() => A.Dummy<ConcurrentCreationDummy>());
+            var makeTwo = Task.Run(() =>
```
thomaslevesque: I don't remember the warning number, but basically, when using `Task.Factory.StartNew`, a `TaskScheduler` should be specified. In most cases, it's better to use `Task.Run`, unless you want some very specific behavior.

tests/FakeItEasy.Tests.ruleset
```diff
@@ -12,7 +12,9 @@
     <Rule Id="CA1025" Action="None" />
     <Rule Id="CA1034" Action="None" />
     <Rule Id="CA1045" Action="None" />
+    <Rule Id="CA1051" Action="None" />
```
thomaslevesque: "Do not declare visible instance fields". Doesn't matter in unit tests.

tests/FakeItEasy.Tests.ruleset
```diff
@@ -12,7 +12,9 @@
     <Rule Id="CA1025" Action="None" />
     <Rule Id="CA1034" Action="None" />
     <Rule Id="CA1045" Action="None" />
+    <Rule Id="CA1051" Action="None" />
     <Rule Id="CA1062" Action="None" />
+    <Rule Id="CA1068" Action="None" />
```
thomaslevesque: "CancellationToken parameters must come last". Doesn't matter in tests; violated by several scenarios in `CancellationSpecs`.

tests/FakeItEasy.Tests.ruleset
```diff
@@ -23,6 +25,7 @@
     <Rule Id="CA1709" Action="None" />
     <Rule Id="CA1716" Action="None" />
     <Rule Id="CA1720" Action="None" />
+    <Rule Id="CA1724" Action="None" />
```
thomaslevesque: "Type names should not match namespaces". Doesn't matter in tests. Violated by `Generic`, which matches `System.Collection.Generic`

tests/FakeItEasy.Tests/Core/MethodInfoManagerTests.cs
```diff
@@ -211,16 +211,14 @@ public void DoSomething(string text)
...
+            public T GetDefault<T>() => default(T);
 
             [SuppressMessage("Microsoft.Usage", "CA1801:ReviewUnusedParameters", MessageId = "other", Justification = "Required for testing.")]
-            public bool Equals(Base other)
-            {
-                return true;
-            }
+            public bool Equals(Base other) => true;
+
+            public override bool Equals(object other) => true;
```
thomaslevesque: Should override `Equals(object)` when implementing `IEquatable<T>`

tests/FakeItEasy.Tests/Expressions/CallExpressionParserTests.cs
```diff
@@ -176,7 +176,7 @@ private static MethodInfo GetPropertyGetter<T>(string propertyName)
         {
             return typeof(T)
                 .GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
-                .Where(x => x.Name.Equals(propertyName))
+                .Where(x => x.Name == propertyName)
```
thomaslevesque: More readable IMHO, and doesn't incur a warning (equivalent to `Equals` with `StringComparison.Ordinal`)

src/FakeItEasy.ruleset
blairconrad: In that case, should we consider disabling the warning on a case-by-case basis? And/or report the missed `#if`?

src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
```diff
@@ -18,8 +18,8 @@ private class ObjectMemberRule
                 new List<MethodInfo>
                     {
                         typeof(object).GetMethod("Equals", new[] { typeof(object) }),
-                        typeof(object).GetMethod("ToString", new Type[] { }),
-                        typeof(object).GetMethod("GetHashCode", new Type[] { })
+                        typeof(object).GetMethod("ToString", Type.EmptyTypes),
+                        typeof(object).GetMethod("GetHashCode", Type.EmptyTypes)
```
blairconrad: I didn't know about `Type.EmptyTypes`. I like!

src/FakeItEasy/Raise.cs
```diff
@@ -77,7 +77,9 @@ public static TEventHandler With<TEventHandler>(params object[] arguments)
         /// with all CLR languages (for example Visual Basic).
         /// To raise non-standard events from other languages, use <see cref="Raise.FreeForm{TEventHandler}"/>.
         /// </summary>
+#pragma warning disable CA1034
```
blairconrad: I've always enjoyed the justifications we've had on the suppressed warnings, or failing that, an indication of what we're suppressing, not unlike what you've done in FakeManager.cs. Would you consider giving a bit of a hint for these pragmas?

src/FakeItEasy.ruleset
thomaslevesque: How would you do this? By putting `#pragma warning disable` between `#if`s? It would look a bit ugly, but it should work.
What I don't understand is that for netstandard1.6, the `[Serializable]` attribute is suppressed, so this warning shouldn't kick in... I'll open an issue in the roslyn-analyzers repo.

src/FakeItEasy/Raise.cs
thomaslevesque: Sure, I'll add a comment

src/FakeItEasy.ruleset
thomaslevesque: OK, I investigated a bit more. In fact, the warning is *not* triggered when building netstandard1.6, but in the VS error list, the reported project is "FakeItEasy (netstandard1.6)":

image

Intellisense and command line build report it only in net40/net45. So it looks like a bug in VS.

So, maybe we shouldn't disable this warning entirely, but just suppress it where appropriate. Anyway, I think this warning is a good indication that the serialization of fakes is flaky at best...

thomaslevesque: > So, maybe we shouldn't disable this warning entirely, but just suppress it where appropriate. Anyway, I think this warning is a good indication that the serialization of fakes is flaky at best...

Or maybe not... there's clearly something wrong with the analyzer: the types it reports as not serializable *are* actually serializable (e.g. object, MethodInfo...).

thomaslevesque: Ah! https://github.com/dotnet/roslyn-analyzers/issues/1510

src/FakeItEasy/Core/FakeManager.cs
blairconrad: I glossed over this comment the first time around. You mean _all_ `SuppressMessage`s!

src/FakeItEasy.ruleset
blairconrad: Thanks for the pragmas and pointer to the roslyn-analyzer issue.

src/FakeItEasy/Raise.cs
blairconrad: Thanks.

blairconrad:
Thanks, @thomaslevesque. Looks really nice. Squash as much or little as you want (well, it'd be nice if the fixup! got squashed at least) and I'll merge.

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: It was just a remark in passing. I don't intend to change all of them now, but we could. It's more concise, and just as readable (as long as we include the warning description)

blairconrad:
Very nice, @thomaslevesque . Thank you.

thomaslevesque:
Thanks!

Issue #684, opened 2016-04-28
find a better way to describe a call than making an ExpressionCallMatcher


blairconrad:
From https://github.com/FakeItEasy/FakeItEasy/pull/673/files#r60991134

blairconrad:
Of course, if #673 doesn't get merged, the point may be moot.

thomaslevesque:
This change has been released as part of [FakeItEasy 4.5.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/4.5.0).

PR #1313, opened 2018-02-12
Write call constraint to writer instead of converting to string

blairconrad:
When fixing #684, I noticed that we use output writers to make strings that we pass to output writers that we use to make strings…

This will take out at least one place where we go writer → string → writer.
It's just a peeve of mine, and if we don't want to bother, I understand.


blairconrad:
Thank you!

PR #1314, opened 2018-02-12
Make CallConstraintDescriber static

blairconrad:
This is embarrassing. After #1313, CAllConstraintDescriber no longer needed to have instances.
Why not make it static?

### DIFF
--- a/src/FakeItEasy/Configuration/PropertyExpressionHelper.cs
+++ b/src/FakeItEasy/Configuration/PropertyExpressionHelper.cs
@@ -58,9 +58,8 @@ namespace FakeItEasy.Configuration
         {
             var writer = ServiceLocator.Current.Resolve<StringBuilderOutputWriter>();
             var constraintFactory = ServiceLocator.Current.Resolve<ExpressionArgumentConstraintFactory>();
-            var describer = ServiceLocator.Current.Resolve<CallConstraintDescriber>();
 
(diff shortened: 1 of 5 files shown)


thomaslevesque:
> Why not make it static?

It's true that it makes the code a bit simpler, but I'm not sure it's worth losing the ability to fake it for unit testing...

blairconrad:
1. [WDNIY](http://wiki.c2.com/?WdNiy), and we can always make unstatic again, if we need to fake.
2. Personally, I'd rather test this sort of thing from the outside, by verifying the actual usage, as we do now. There are no real moving parts.
3. I'm not wed to the idea of making it static, if you still prefer non-static, I can do, and I'll submit another PR that clean up the warnings we have now, but stopping short of making the class static.

thomaslevesque:
OK, I'm sold. I just wanted to make sure we thought this through.

blairconrad:
Thanks, @thomaslevesque. It's good to check these things. Me, I'm happier with static and sealed classes than most people, I think, and I do worry that I'd go overboard.

thomaslevesque:
This change has been released as part of [FakeItEasy 4.5.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/4.5.0).

thomaslevesque:
(oops, we don't do this on PRs...)

blairconrad:
> (oops, we don't do this on PRs...)

sometimes we do, if the PR isn't specifically to fix an issue!

PR #1317, opened 2018-02-17
Support configuration of property setter defined on base interface

### LINKED ISSUE #1316: CallToSet fails when setting base interface property
[Bug.zip](https://github.com/FakeItEasy/FakeItEasy/files/1731953/Bug.zip)

Run attached, the CallToSet method fails when setting a base interface property.

blairconrad:
Fixes #1316.


src/FakeItEasy/Configuration/PropertyExpressionHelper.cs
```diff
@@ -27,7 +27,11 @@ internal class PropertyExpressionHelper
                 .ToArray();
 
             var callTargetType = Fake.GetFakeManager(parsedCallExpression.CallTarget).FakeObjectType;
-            var indexerSetterInfo = callTargetType.GetMethod("set_" + propertyName, parameterTypes);
+            var setPropertyName = "set_" + propertyName;
+            var indexerSetterInfo = callTargetType.GetMethod(setPropertyName, parameterTypes)
+                                    ?? callTargetType.GetInterfaces()
+                                        .Select(i => i.GetMethod(setPropertyName, parameterTypes))
+                                        .FirstOrDefault();
```
thomaslevesque: I must have broken this in #1301... `GetMethod` looks in parent types for a class, but not for an interface.

I think there might be an issue with this fix, though...

Suppose I have this:

```
interface IFoo
{
    int X { get; set; }
}

interface IBar
{
    int X { get; set; }
}

interface IBaz : IFoo, IBar
{
}

var baz = A.Fake<IBaz>();
A.CallToSet(() => baz.X).DoesNothing();
```
This wouldn't compile, because there would be an ambiguity between `IFoo.X` and `IBar.X`. I have to specify which `X` I want to configure, like this:

```
var baz = A.Fake<IBaz>();
A.CallToSet(() => ((IFoo)baz).X).DoesNothing();
```

But with the proposed fix, `indexerSetterInfo` would be `IBar.set_X` (at least it could; I'm not sure if it's deterministic, but in my test it picked `IBar` rather than `IFoo`). So we would be configuring the wrong property.

Maybe we should use `parsedCallExpression.CalledMethod.DeclaringType` as the `callTargetType`.

tests/FakeItEasy.Specs/ConfiguringPropertySetterSpecs.cs
```diff
@@ -428,6 +432,102 @@ public interface IHaveInterestingProperties
...
 
+        [Scenario]
+        public static void ConfiguringSetterFromBaseInterfaceOnFakedInterface(
+            IHaveInterestingDerivedProperties subject,
+            bool wasConfiguredBehaviorUsed)
+        {
+            "Given an interface with a read/write property"
+                .See<IHaveInterestingProperties>();
+
+            "And an interface that implements it"
```
thomaslevesque: An interface doesn't implement anything :wink: ([yet!](https://github.com/dotnet/csharplang/issues/52)). It just inherits from another interface.

src/FakeItEasy/Configuration/PropertyExpressionHelper.cs
thomaslevesque: > (at least it could; I'm not sure if it's deterministic, but in my test it picked `IBar` rather than `IFoo`)

I just tried, and it seems to be non-deterministic... I got a different result when I tried in a xBehave scenario, the first interface was picked instead of the second.

Anyway, here's the scenario that fails for me:

```
        public interface IBase1
        {
            int X { get; set; }
        }

        public interface IBase2
        {
            int X { get; set; }
        }

        public interface IAmbiguousDerived : IBase1, IBase2
        {
        }

        [Scenario]
        public static void ConfiguringSetterFromAmbiguousBaseInterface(
            IAmbiguousDerived subject,
            bool wasConfiguredBehaviorUsed)
        {
            "Given a base interface with a read/write property"
                .See<IBase1>();

            "And another base interface with a read/write property with the same name"
                .See<IBase2>();

            "And an interface that inherits from both base interfaces"
                .x(() => subject = A.Fake<IAmbiguousDerived>());

            "And assignment of the property of the second base interface is configured for any value"
                .x(() => A.CallToSet(() => ((IBase2)subject).X).Invokes(call => wasConfiguredBehaviorUsed = true));

            "When I assign a value to the property"
                .x(() => ((IBase2)subject).X = 0);

            "Then the configured behavior is used"
                .x(() => wasConfiguredBehaviorUsed.Should().BeTrue());
        }
```

thomaslevesque: And here's the code that fixes it:

```
            var callTargetType = parsedCallExpression.CalledMethod.DeclaringType
                                 ?? Fake.GetFakeManager(parsedCallExpression.CallTarget).FakeObjectType;
            var setPropertyName = "set_" + propertyName;
            var indexerSetterInfo = callTargetType.GetMethod(setPropertyName, parameterTypes);
```

The `?? Fake.GetFakeManager(parsedCallExpression.CallTarget).FakeObjectType` part is because R# complains that `DeclaringType` could be null. It's technically possible (e.g. dynamic method created with LCG), but I don't think it could really happen unless someone really tries very hard to do it, e.g. by passing a specially crafted expression to `A.CallToSet`. Feel free to drop it ;)

blairconrad: I'm not fussy about whether the `??` bit is there, but if it stays, probably best to include your explanation in a comment.

tests/FakeItEasy.Specs/ConfiguringPropertySetterSpecs.cs
blairconrad: Good point. Will push a fixup.

blairconrad:
I just pushed your code. The glory will be mine!

thomaslevesque:
LGTM, go ahead and squash!

blairconrad:
Thanks for the comments, additional test, replacement fix, and merge, @thomaslevesque!

PR #1339, opened 2018-03-15
Document limitations around in parameters and ref returns

### LINKED ISSUE #1337: Document limitations of working with ref returns, in parameters
As revealed by:

 https://gist.github.com/thomaslevesque/76a8b2677ab37f9f7ff04e87f80b9b74.

The `in` parameter weakness is a result of https://github.com/castleproject/Core/issues/339.

blairconrad:
Fixes #1337.


tests/FakeItEasy.Specs/FakeItEasy.Specs.csproj
```diff
@@ -11,7 +11,11 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net461'">
-    <DefineConstants>$(DefineConstants)</DefineConstants>
+    <DefineConstants>$(DefineConstants);FEATURE_IN_PARAM</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <LangVersion>latest</LangVersion>
```
blairconrad: If you prefer, could go to 7.2.

thomaslevesque: `latest` is fine, but maybe it should be set in the root Directory.Build.props, in order to affect all projects.

docs/what-can-be-faked.md
```diff
@@ -35,3 +43,7 @@ overridden if they are:
 
 Note that this means that static members, including extension methods,
 **cannot** be overridden.
+
+### Methods with reference return values
+
+Methods with [reference return values](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/ref-returns#what-is-a-reference-return-value) can not be invoked on a Fake. Any attempt to do so will result in a `NullReferenceException` being thrown.
```
thomaslevesque: Even though the official documentation uses it, I find the term "reference return value" confusing... It could easily be understood as "return value of reference type". I don't have a good alternative besides "ref return", though... Any better idea?

tests/FakeItEasy.Specs/RefReturnSpecs.cs
```diff
@@ -0,0 +1,53 @@
+namespace FakeItEasy.Specs
+{
+    using System;
+    using FakeItEasy.Tests.TestHelpers;
+    using FluentAssertions;
+    using Xbehave;
+    using Xunit;
+
+    /// <summary>
+    /// Characterization tests, representing the current capabilities of the code, not the desired state.
```
thomaslevesque: I don't think it makes sense to add characterization tests for ref return. It makes sense for `in` params, because it's a scenario that should work, but currently doesn't. But faking ref return methods is not something that could ever work, because of fundamental limitations of the platform.

docs/what-can-be-faked.md
blairconrad: "methods that return values by reference"?

blairconrad: Still confusing?

tests/FakeItEasy.Specs/RefReturnSpecs.cs
blairconrad: I'll drop.

docs/what-can-be-faked.md
thomaslevesque: Maybe just "Methods that return by reference"?

blairconrad: Is okay with me. keep "(officially…)", though?

thomaslevesque: Sorry, I posted my suggestion before I saw your answer. Either "methods that return values by reference" or "Methods that return a reference to a variable" are fine by me. Although the reference might be to an array location, rather than a variable...
Mentioning the official term is a good idea.

thomaslevesque:
Very nice. Thanks @blairconrad!

blairconrad:
"Cool. I'll squash it up." Is what I'd say under other circumstances.

blairconrad:
Thanks for the improvements and merge!

PR #1344, opened 2018-05-01
Use ArgumentValueFormatter in EqualityArgumentConstraint

### LINKED ISSUE #1343: EqualityArgumentConstraint doesn't use argument value formatters
As mentioned in #1196, `EqualityArgumentConstraint` (used to represent "constant" argument values in call specifications) should have been updated to use `ArgumentValueFormatter`, but was somehow forgotten during the implementation.

thomaslevesque:
Fixes #1343

### DIFF
--- a/src/FakeItEasy/Core/ArgumentValueFormatter.cs
+++ b/src/FakeItEasy/Core/ArgumentValueFormatter.cs
@@ -74,40 +74,11 @@ namespace FakeItEasy.Core
         private IArgumentValueFormatter ResolveTypeFormatter(Type forType)
         {
             return
-                (from formatter in this.typeFormatters
-                 where formatter.ForType.IsAssignableFrom(forType)
-                 select formatter)
-                .Min(f => new RangedFormatter(f, GetDistanceFromKnownType(f.ForType, forType)))
-                .Formatter;
-        }
-
-        /// <summary>
-        /// Holds a formatter as well as the distance between a type to be formatted
-        /// and the type for which the formatted is registered.
-        /// </summary>
-        private class RangedFormatter : IComparable<RangedFormatter>
-        {
-            private readonly int distanceFromKnownType;
-
-            public RangedFormatter(IArgumentValueFormatter formatter, int distanceFromKnownType)
-            {
-                this.Formatter = formatter;
-                this.distanceFromKnownType = distanceFromKnownType;
-            }
-
-            public IArgumentValueFormatter Formatter { get; }
-
-            public int CompareTo(RangedFormatter other)
-            {
-                Guard.AgainstNull(other, nameof(other));
-
-                if (other.distanceFromKnownType == this.distanceFromKnownType)
-                {
-                    return other.Formatter.Priority.CompareTo(this.Formatter.Priority);
-                }
-
-                return this.distanceFromKnownType.CompareTo(other.distanceFromKnownType);
-            }
+                this.typeFormatters
+                    .Where(formatter => formatter.ForType.IsAssignableFrom(forType))
+                    .OrderBy(formatter => GetDistanceFromKnownType(formatter.ForType, forType))
+                    .ThenByDescending(formatter => formatter.Priority)
+                    .First();
         }
 
         private class DefaultFormatter
--- a/src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
+++ b/src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
@@ -24,20 +24,11 @@ namespace FakeItEasy.Expressions.ArgumentConstraints
         [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Any type of exception may be encountered.")]
         public override string ToString()
         {
-            if (this.ExpectedValue == null)
-            {
-                return "<NULL>";
-            }
-
-            var stringValue = this.ExpectedValue as string;
-            if (stringValue != null)
-            {
-                return $@"""{stringValue}""";
-            }
-
             try
             {
-                return this.ExpectedValue.ToString();
+                var writer = ServiceLocator.Current.Resolve<StringBuilderOutputWriter>();
+                writer.WriteArgumentValue(this.ExpectedValue);
+                return writer.Builder.ToString();
             }
             catch (Exception)
             {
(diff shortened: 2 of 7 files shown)


tests/FakeItEasy.Specs/StringExtensions.cs
```diff
@@ -53,6 +53,20 @@ public static IStepBuilder See<T>(this string text, Func<T, Action<Type, IFakeOp
             return text.x(() => { });
         }
 
+        /// <summary>
+        /// Creates a step builder that can be used to refer to a type's method.
+        /// Useful when we want to establish a condition with actually performing an action.
```
blairconrad: "without actually", I think

src/FakeItEasy/Core/ArgumentValueFormatter.cs
```diff
@@ -74,40 +74,11 @@ private static int GetDistanceFromKnownType(Type comparedType, Type knownType)
...
-                    return other.Formatter.Priority.CompareTo(this.Formatter.Priority);
-                }
-
-                return this.distanceFromKnownType.CompareTo(other.distanceFromKnownType);
-            }
+                this.typeFormatters
+                    .Where(formatter => formatter.ForType.IsAssignableFrom(forType))
+                    .OrderBy(formatter => GetDistanceFromKnownType(formatter.ForType, forType))
+                    .ThenByDescending(formatter => formatter.Priority)
+                    .First();
```
blairconrad: I like it!

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -732,23 +732,18 @@ public static void UnusedArgumentMatcherDescriptionNotUsed(IHaveNoGenericParamet
         public static IEnumerable<object[]> Fakes()
         {
             yield return new object[] { A.Fake<object>(), "Faked " + typeof(object) };
-            yield return new object[] { A.Fake<List<int>>(), "Faked " + typeof(List<int>) };
-            yield return new object[] { A.Fake<IList<int>>(), "Faked " + typeof(IList<int>) };
```
blairconrad: Why remove `List<int>`?

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -732,23 +732,18 @@ public static void UnusedArgumentMatcherDescriptionNotUsed(IHaveNoGenericParamet
         public static IEnumerable<object[]> Fakes()
         {
             yield return new object[] { A.Fake<object>(), "Faked " + typeof(object) };
-            yield return new object[] { A.Fake<List<int>>(), "Faked " + typeof(List<int>) };
-            yield return new object[] { A.Fake<IList<int>>(), "Faked " + typeof(IList<int>) };
+            yield return new object[] { A.Fake<IList<int>>(), "[]" };
```
blairconrad: I'm not entirely sure about rendering as `[]`. We _could_ add a formatter that detected fakes and render as `Faked Blah`. I'm not sure which is better. Honestly, I'd hope that there aren't too many Fakes being passed around, but continue to note that it's a Fake could still be helpful. Thoughts?

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
```diff
@@ -37,7 +37,9 @@ public override string ToString()
```
blairconrad: `ArgumentValueFormatter` already handles `null` (rendering it as `NULL` all the time, sadly not `<NULL>`) and strings (by default rendering either as `string.Empty` or `"thestring"`, but a custom formatter would override). I think we should defer to `ArgumentValueFormatter` all the time. Of course if we don't like `<NULL>` or `string.Empty`, we can adjust…

src/FakeItEasy/Core/ArgumentValueFormatter.cs
thomaslevesque: To be honest, it's slightly less efficient, since it requires ordering the sequence, so it's `O(n log n)` instead of `O(n)`. What we really needs here is a `MinBy` method, which IMO should have been in Linq since the beginning. Anyway, I don't think it's a problem, since there are typically very few formatters.

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
thomaslevesque: Oops, I should have removed this. Will do.

tests/FakeItEasy.Specs/StringExtensions.cs
thomaslevesque: Actually, I copied it from the above method, which I believe you wrote... so it's your fault 😛
I'll fix it, and the others as well.

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
thomaslevesque: > We could add a formatter that detected fakes and render as Faked Blah

Actually we can't. Formatters are selected based on their `ForType` property, and this doesn't let us detect if an object is a fake. We would need formatters to have `CanFormat` method, but that would be a breaking change.

I think there's little value in knowing that a given collection is actually a fake. But if we really want to do it, we could modify the `DefaultEnumerableValueFormatter` to check if it's a fake, and add a `Faked ` prefix to the collection description.

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
thomaslevesque: Because from xUnit's perspective, both test cases are the same (both have an empty collection and the "[]" string), and it only runs one of them.

src/FakeItEasy/Core/ArgumentValueFormatter.cs
blairconrad: Yeah, I spotted the performance change, but like you I figured it didn't matter much due to the presumably small list and the hopefully few times the code is used during a test run.

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
blairconrad: > Actually we can't.

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
blairconrad: Darn you to heck, xUnit! Thanks.

tests/FakeItEasy.Specs/StringExtensions.cs
blairconrad: That does look like the sort of error I would make. Your story checks out. Thanks.

thomaslevesque:
> Want to appease the draconian "DontMergeMeYet" bot?

Done

Issue #1345, opened 2018-05-09
Not compatible with dotnetcore 2.1-rc1-final


ryanwischkaemper:
Dotnetcore 2.1-rc1 was [released the other day](https://blogs.msdn.microsoft.com/dotnet/2018/05/07/announcing-net-core-2-1-rc-1/). I updated a class library and it's accompanying unit test project to 2.1-rc1 (it was previously targeting 2.0). 

Now I get errors in my tests like this:

`System.NotSupportedException: 'Cannot create boxed ByRef-like values.'` (stack trace below)

A lot of other tests are breaking with the same error, but this is a simpler one. 

`dotnet --version` is **2.1.300-rc1-008673**

### Test Method
```csharp
[TestMethod]
        public async Task GetValue_WhenRequestedByMoreThanOneTask_ShouldOnlyCallValueFactoryOnce()
        {
            var fakeValueFactory = A.Fake<Func<string, Task<string>>>();

            var sut = new ThreadSafeAsyncCache<string, string>(fakeValueFactory);

	        var callingTasks = Enumerable.Range(0,3).Select(x => sut.GetValue(A.Dummy<string>())).ToArray();

            await Task.WhenAll(callingTasks);

            A.CallTo(() => fakeValueFactory.Invoke(A.Dummy<string>())).MustHaveHappened(Repeated.Exactly.Once);
        }
```

### Class Being Tested
```csharp
    public class ThreadSafeAsyncCache<TKey, TValue>
    {
        private readonly Func<TKey, Task<TValue>> _valueFactory;
        private readonly ConcurrentDictionary<TKey, Lazy<Task<TValue>>> _map;

        public ThreadSafeAsyncCache(Func<TKey, Task<TValue>> valueFactory)
        {
            if(valueFactory == null) throw new ArgumentNullException(nameof(valueFactory));
            _valueFactory = valueFactory;
            _map = new ConcurrentDictionary<TKey, Lazy<Task<TValue>>>();
        }

        public Task<TValue> GetValue(TKey key)
        {
            if(key == null) throw new ArgumentNullException(nameof(key));
            return _map.GetOrAdd(key,
                x => new Lazy<Task<TValue>>(() => _valueFactory(x)))?.Value;
        }
    }
```

Stack Trace:

```
System.NotSupportedException: Cannot create boxed ByRef-like values.
[...]

PR #1349, opened 2018-05-10
Protect argument resolution failure

### LINKED ISSUE #1348: Protect against exceptions when resolving constructor arguments for Fake/Dummy creation
As mentioned in #1347, if a `ResolveStrategy` throws, it could derail the entire Fake or Dummy creation process. We should protect against this.

On the other hand, catching `ResolveStrategy` errors may just end up concealing the reason for the failure, so taking the simple path of putting a try/catch or two in `TryResolveDummyValueWithAllAvailableStrategies`, may not be the best bet.

For example, if we try to make a Dummy `Lazy`, `ResolveByCreatingLazyStrategy` is really the best bet. If it fails, maybe we'd like the exception to fly.

However, for the more speculative resolvers, such as `ResolveByCreatingFakeStrategy` and `ResolveByInstantiatingClassUsingDummyValuesAsConstructorArgumentsStrategy`, which could each make several attempts to create a Dummy, it might be best to catch exceptions raised within them and carry on with the attempts. For example, as noted in #1345, we cannot make a Dummy `string` under dotnet core 2.1 rc1 because we fail to create a `ReadOnlySpan<T>`, which can be used as a string constructor argument.

If we just accepted that we can't make a `ReadOnlySpan<T>`, we could move on to another constructor and still create a `string`.

blairconrad:
Fixes #1348.
Works around the particular failure reported in #1345.

### DIFF
--- a/src/FakeItEasy/Creation/DummyValueResolver.cs
+++ b/src/FakeItEasy/Creation/DummyValueResolver.cs
@@ -221,18 +221,18 @@ namespace FakeItEasy.Creation
                 foreach (var constructor in GetConstructorsInOrder(typeOfDummy))
                 {
                     var parameterTypes = constructor.GetParameters().Select(x => x.ParameterType);
-                    var resolvedArguments = this.ResolveAllTypes(session, parameterTypes);
-
-                    if (resolvedArguments != null)
+                    try
                     {
-                        try
+                        var resolvedArguments = this.ResolveAllTypes(session, parameterTypes);
+
+                        if (resolvedArguments != null)
                         {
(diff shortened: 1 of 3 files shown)


blairconrad:
@ryanwischkaemper, I believe this will solve your problem. If you've time and interest, want to grab and try it? If it does, and one of my colleagues approves the changes, I'd be inclined to publish a release containing it quite soon.

ryanwischkaemper:
@blairconrad I tried out your PR and it solves our problems - all our tests pass again

blairconrad:
@ryanwischkaemper, thanks. It's nice to have outside confirmation. I'm glad you'll be able to proceed until we release a version containing some kind of fix (I hope this one).

PR #1351, opened 2018-05-10
Resolve constructor arguments lazily

### LINKED ISSUE #1350: Resolve Fakes' constructor arguments lazily
From #1347: `TryCreateFakeWithDummyArgumentsForConstructor` eagerly determines which constructors' arguments can be resolved, by resolving the arguments. That is, every possible argument for every possible constructor is attempted, even if the first constructor succeeds.

For example, if we construct a Dummy `string` (under net461), we end up creating 1 `Char *`, 3 `Char[]`s, 12 `Int32`s, and 1 `SByte *`—much more than we need, since the first constructor found (taking a `Char[]` and 2 `Int32`s) would succeed. Contrast `ResolveByInstantiatingClassUsingDummyValuesAsConstructorArgumentsStrategy`, which only resolves constructor arguments as needed.

blairconrad:
Fixes #1350

### DIFF
--- a/src/FakeItEasy/Creation/FakeObjectCreator.cs
+++ b/src/FakeItEasy/Creation/FakeObjectCreator.cs
@@ -36,11 +36,10 @@ namespace FakeItEasy.Creation
             return result != null ? result.GeneratedProxy : null;
         }
 
-        private static ResolvedConstructor[] ResolveConstructors(Type typeOfFake, DummyCreationSession session, IDummyValueResolver resolver)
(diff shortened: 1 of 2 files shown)


blairconrad:
Marking as WIP so we don't accidentally merge before next release.

blairconrad:
Oh, we could. It's just that I figured you'd not be around and I'd planned the release branch process. UnWIPing.

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thanks for the merge.

Issue #1347, opened 2018-05-10
Dummy/Fake creation process improvements


blairconrad:
The Dummy/Fake creation process has a number of weaknesses, some of which I found during an investigation into #1345. I'll compile a few here so we can create issues for them and discuss and optionally address them. Once we've made all the issues we want for now, we can close this issue.

- [x] (#1348) many of the `ResolveStrategy`s may throw if downstream code (for example constructors, value activation, etc.) does, and this will stop `TryResolveDummyValueWithAllAvailableStrategies` in its tracks, so no Dummy is created. And since Dummies are used to create Fakes, this can impact Fake creation, as in #1345.
- [x] (#1350) `TryCreateFakeWithDummyArgumentsForConstructor` eagerly determines which constructors' arguments can be resolved, by resolving the arguments. This is inefficient at best, and hazardous at worst, in case some constructors take unresolvable arguments.
- [x] (#1355) the proxy generator selector attempts to use generators to make Fakes for which they are unsuited. For example, we try to use `CastleDynamicProxyGenerator` to create a `string`, even though `string`s are sealed (the `CastleDynamicProxyGenerator` exits early, but not before we create all the constructor arguments mentioned above. Oh yeah, none of the constructor arguments are even used in this case)
- [x] (#1359) even though we cache the `ResolveStrategy` used to create a Dummy, we still may be doing significant work on subsequent calls. If the `ResolveByCreatingFakeStrategy` or `ResolveByInstantiatingClassUsingDummyValuesAsConstructorArgumentsStrategy` are cached, invoking them will still require enumerating the constructors and possibly creating many extra constructor arguments. We could consider caching more targeted resolvers that will go right to the proper constructor/set of fake creation arguments

blairconrad:
I'd created this issue mostly as a note for myself so I wouldn't forget the ideas that came to me while working on #1345. Now that I've created issues from all the nuggets, I think this can be closed.

PR #1352, opened 2018-05-11
Remove GenerateProxy that doesn't take attribute list

blairconrad:
Found while looking at #1347. The method was not used in production code.


blairconrad:
Thanks for the merge, @thomaslevesque. Your point about the extra attributes is a good one. I created #1353.

blairconrad:
This change has been released as part of [FakeItEasy 4.6.0](https://github.com/FakeItEasy/FakeItEasy/releases/4.6.0).

PR #1357, opened 2018-05-13
Release 4.5.2

blairconrad:


blairconrad:
This PR will be blocked by DontMergeMeYet, since there's a "squash!" commit. Also "fixup!". However, this is how those commits were merged into develop, so we should merge anyhow.

PR #1360, opened 2018-05-15
Avoid asking proxy generators to create Fakes that they can't

### LINKED ISSUE #1355: Avoid asking proxy generators to create Fakes that they can't
Currently we try to invoke the proxy generators, often once per available constructor, when making a Fake, even if the proxy generator can't possibly succeed.

For example, we try to use CastleDynamicProxyGenerator to create a string, even though strings are sealed (the CastleDynamicProxyGenerator exits early, but not before we create constructor arguments).

(From #1347.)

blairconrad:
Fixes #1355.


blairconrad:
00764387cd5dfe5c6dcb6b2820d5ac0ebec7a702 is optional, and does introduce temporal coupling, but avoids some double checking about whether we can fake certain types.

I inlined the private `GenerateProxy` at first because it kept it from jumping around when it changed from static to non-static, but in the end, I liked how it let us do more basic checks up front. I'm not married to it, though, if nobody else likes it.

thomaslevesque:
I know I already approved the PR, but...

> 0076438 is optional, and does introduce temporal coupling, but avoids some double checking about whether we can fake certain types.

It's not an expensive check; I think I'd rather keep the double checking and avoid the temporal coupling.

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you for the merge, and vetoing the temporal coupling!

PR #1371, opened 2018-06-10
Add missing null checks in AnyCallConfiguration

### LINKED ISSUE #1370: Missing argument validation in AnyCallConfiguration.WhenArgumentsMatch
This code throws `NullReferenceException` when the fake method is called. It should throw `ArgumentNullException` when `WhenArgumentsMatch` instead.

```c#
var foo = A.Fake<IFoo>();
A.CallTo(foo).WhenArgumentsMatch(null).DoesNothing();
foo.Bar();
```

thomaslevesque:
Fixes #1370


thomaslevesque:
> No tests!?

For a null check?
I thought about it, but it didn't seem to fit in specs, and I thought we didn't write unit tests any more.
I can add some if you like, though they'll be of little value.

blairconrad:
> we didn't write unit tests any more

Oh, I wouldn't say that, but we try not to. Or I try not to. I'm sure there will come a situation where a brace of unit or integration tests will come in handy. But you've convinced me; we're probably fine here.

thomaslevesque:
Thanks!

PR #1374, opened 2018-06-22
Output () for failed parameterless constructor signature instead of '…

### LINKED ISSUE #1373: Change wording when referring to parameterless constuctors during failed fake creation
From @thomaslevesque's comment in https://github.com/FakeItEasy/FakeItEasy/pull/1372#discussion_r196890730

```c#
[Fact]
public void Test()
{
    A.Fake<ClassWithFailingParameterlessConstructor>();
}

public class ClassWithFailingParameterlessConstructor
{
    public ClassWithFailingParameterlessConstructor()
    {
        throw new NotImplementedException();
    }
}
```

yields

```
 Failed to create fake of type FakeItEasyPlay.FailingParameterlessConstructor+ClassWithFailingParameterlessConstructor.

  Below is a list of reasons for failure per attempted constructor:
    No constructor arguments failed:
      No usable default constructor was found on the type FakeItEasyPlay.FailingParameterlessConstructor+ClassWithFailingParameterlessConstructor.
      An exception of type System.NotImplementedException was caught during this call. Its message was:
      The method or operation is not implemented.
         at FakeItEasyPlay.FailingParameterlessConstructor.ClassWithFailingParameterlessConstructor..ctor() in D:\Users\amidu\Documents\Visual Studio 2017\Projects\FakeItEasyPlay\FakeItEasyPlay\HiddenMembers.cs:line 19
         at Castle.Proxies.ClassWithFailingParameterlessConstructorProxy..ctor(IInterceptor[] )
```

"No constructor arguments failed" is confusing. A parameterful constructor would yield a message like

`Constructor with signature (System.Int32) failed:`

Perhaps we should remove the special-casing and just emit

`Constructor with signature () failed:`

for the failing parameterless constructor.

Alternatively:

`Parameterless constructor failed:`

blairconrad:
…No constructor arguments'

Fixes #1373.

### DIFF
--- a/src/FakeItEasy/Core/DefaultExceptionThrower.cs
+++ b/src/FakeItEasy/Core/DefaultExceptionThrower.cs
@@ -56,20 +56,11 @@ namespace FakeItEasy.Core
             {
                 foreach (var constructor in resolvedConstructors.Where(x => x.WasSuccessfullyResolved))
                 {
-                    if (constructor.Arguments == null)
-                    {
-                        message
-                            .AppendIndented("    ", "No constructor arguments failed:");
-                    }
-                    else
-                    {
-                        message
-                            .AppendIndented("    ", "Constructor with signature (")
-                            .Append(constructor.Arguments.ToCollectionString(x => x.ArgumentType.ToString(), ", "))
-                            .AppendLine(") failed:");
-                    }
-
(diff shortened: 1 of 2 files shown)


blairconrad:
We take a step back so we can take a bigger one forward!

Thanks for merge!

PR #1375, opened 2018-06-24
switch to Bullseye and SimpleExec

adamralph:
- [Bullseye](https://github.com/adamralph/bullseye) replaces simple-targets-csx (it's a regular binary package instead of scripts, so can optionally be used in a [regular .NET project](https://github.com/xbehave/xbehave.net/blob/6d798877b4b891ce95ea8f2c5f9bee3d56a26bae/targets/Program.cs#L8)).
- [SimpleExec](https://github.com/adamralph/simple-exec/) wraps up command invocation, so removes the need to include local helpers

### DIFF
--- a/tools/build.csx
+++ b/tools/build.csx
@@ -1,8 +1,10 @@
 #r "packages/Bullseye.1.0.0-rc.3/lib/netstandard2.0/Bullseye.dll"
+#r "packages/SimpleExec.2.1.0/lib/netstandard1.3/SimpleExec.dll"
 
 using System.Runtime.CompilerServices;
 using Bullseye;
 using static Bullseye.Targets;
+using static SimpleExec.Command;
 
 var solutionName = "FakeItEasy";
 
@@ -75,13 +77,13 @@ Targets.Add("testsDirectory", () => Directory.CreateDirectory(testsDirectory));
 
 Targets.Add("build", DependsOn("clean", "restore", "versionInfoFile"), () => RunMsBuild("Build"));
 
-Add("versionInfoFile", () => Cmd(gitversion, $"/updateAssemblyInfo {versionInfoFile} /ensureAssemblyInfo"));
+Add("versionInfoFile", () => Run(gitversion, $"/updateAssemblyInfo {versionInfoFile} /ensureAssemblyInfo"));
 
 Targets.Add("clean", DependsOn("logsDirectory"), () => RunMsBuild("Clean"));
 
 Targets.Add(
     "restore",
-    () => Cmd("dotnet", $"restore"));
+    () => Run("dotnet", $"restore"));
 
 Targets.Add(
     "unit",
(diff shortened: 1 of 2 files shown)


tools/build.csx
```diff
@@ -64,129 +66,75 @@ var outputDirectory = Path.GetFullPath("./artifacts/output");
 static var testsDirectory = "./artifacts/tests";
 
 // targets
-var targets = new TargetDictionary();
+Add("default", DependsOn("unit", "integ", "spec", "approve", "pack"));
```
blairconrad: I appreciate the brevity, but slightly prefer the explicitness of `Targets.Add` (and `.Run`). It's not a huge deal for me though; if I'm alone in this, I'll go with the flow.

thomaslevesque: Agreed; `Add` is a little too generic for my taste. Otherwise, I love the changes!

adamralph: Thanks for the feedback! I can see why this may look a little odd in the context of this script. Elsewhere, I've moved away from scripting entirely and [I use Bullseye in a regular .NET project named `Targets`](https://github.com/xbehave/xbehave.net/blob/6d798877b4b891ce95ea8f2c5f9bee3d56a26bae/targets/Program.cs). Hence the context of "targets" is already set and I think the plain `Add()` makes more sense. But in the context of a script named `build.csx` it does indeed look a little odd.

What about `DependOn()`? Would you prefer that qualified with the type name or not? If not, then we have to have both of these `using` statements at the top of the file:

```c#
using Bullseye;
using static Bullseye.Targets;
```

I'm also now thinking whether the API design is optimal. Maybe this would make more sense:

```c#
using Bullseye.Targets;

AddTarget(...)
RunTargets(...)
```

adamralph:
@blairconrad perhaps we should change Bullseye so that anything less than 1ms is rendered as `(<1 ms)`?

tools/build.csx
thomaslevesque: > I'm also now thinking whether the API design is optimal. Maybe this would make more sense:
>
> ```csharp
> using Bullseye.Targets;
> 
> AddTarget(...)
> RunTargets(...)
> ```

It would be my favorite approach 👍

blairconrad: I like `AddTarget` and `RunTargets`. In this case, `DependsOn` would stay as is? I think in context, that makes sense. You'd still have a `using static`, yes?

blairconrad:
> perhaps we should change Bullseye so that anything less than 1ms is rendered as (<1 ms)?

That works for me!

adamralph:
@FakeItEasy/owners I've pushed another commit which qualifies `Add()` and `Run()` with `Targets`, but leave `DependsOn()` unqualified. How does that look? Bear in mind it requires the two `using` statements.

I think I'd rather not change the API of Bullseye at this stage since it's already in RC. I'm happy to consider changing things for 2.0.

I'll push another RC with the change to the timings output ASAP and update the PR to that.

blairconrad:
Thanks, @adamralph. I do prefer the new syntax. The two `using`s don't bother me a bit. As far as I'm concerned (@thomaslevesque, do speak up if you disagree), you could fixup 2aa7d90e575fdb278e7072c197846ed0aa05264c with the latest commit at any time. Including When we get the new Bullseye RC.

adamralph:
@blairconrad I've made the relevant fixups, including an upgrade to Bullseye RC3:

image

blairconrad:
I Love it. Thanks, @adamralph!

blairconrad:

PR #1378, opened 2018-06-26
fix GetVSLocation

adamralph:
If vswhere returns an empty line, the method returns an empty string instead of throwing the exception.


adamralph:
I have a better idea.

blairconrad:
Thanks, @adamralph!

blairconrad:

PR #1379, opened 2018-06-26
drop VS finding

adamralph:
Replaces #1378

### DIFF
--- a/tools/build.csx
+++ b/tools/build.csx
@@ -54,9 +54,7 @@ var testSuites = new Dictionary<string, string[]>
 // tool locations
 
 static var toolsPackagesDirectory = Path.Combine(GetCurrentScriptDirectory(), "packages");
-var vswhere = $"{toolsPackagesDirectory}/vswhere.1.0.62/tools/vswhere.exe";
 var gitversion = $"{toolsPackagesDirectory}/GitVersion.CommandLine.4.0.0-beta0012/tools/GitVersion.exe";
-var msBuild = $"{GetVSLocation()}/MSBuild/15.0/Bin/MSBuild.exe";
 var nuget = $"{GetCurrentScriptDirectory()}/.nuget/NuGet.exe";
 var pdbGit = $"{toolsPackagesDirectory}/pdbGit.3.0.41/tools/PdbGit.exe";
 static var xunit = $"{toolsPackagesDirectory}/xunit.runner.console.2.0.0/tools/xunit.console.exe";
@@ -75,11 +73,11 @@ Targets.Add("logsDirectory", () => Directory.CreateDirectory(logsDirectory));
 
 Targets.Add("testsDirectory", () => Directory.CreateDirectory(testsDirectory));
 
-Targets.Add("build", DependsOn("clean", "restore", "versionInfoFile"), () => RunMsBuild("Build"));
+Targets.Add("build", DependsOn("clean", "restore", "versionInfoFile"), () => RunDotNet("build"));
 
 Add("versionInfoFile", () => Run(gitversion, $"/updateAssemblyInfo {versionInfoFile} /ensureAssemblyInfo"));
 
-Targets.Add("clean", DependsOn("logsDirectory"), () => RunMsBuild("Clean"));
+Targets.Add("clean",() => RunDotNet("clean"));
 
 Targets.Add(
     "restore",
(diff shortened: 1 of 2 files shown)


blairconrad:
That does seem better. To verify that I understand, `dotnet` now includes MSBuild, and since we specify the SDK we like, we don't need to worry about anything outside that, yes?

adamralph:
@blairconrad correct.

blairconrad:
Worse than it not building, we reported success!

adamralph:
I think I know why the failure is not propagating. `run-csx.cmd` is calling csi to run `build.csx` but there is no fail-fast on that line in case of error. It needs some error handling like in https://github.com/adamralph/simple-exec/blob/master/build.cmd

blairconrad:
Well spotted, @adamralph. The nodeReuse problem appears to be https://github.com/Microsoft/msbuild/issues/1135, which is fixed but not released yet, AFAICT.

blairconrad:
Yeah, there's no `/nodeReuse` either. I like the expanded `binaryLogger`, though!

adamralph:
@blairconrad I amended the commit so it now uses `dotnet` `build` and `clean` (this is how I do all my builds now). This means you lose the binary log file, but is that something you need?

adamralph:
Now something weird is happening with loading analysis libraries during the build...

adamralph:
It looks like some more work is required to get the solution buildable with pure `dotnet` commands. I recommend it, as it does simplify things.

For this PR, I see two options:

a) withdraw it
b) switch back to `dotnet msbuild` but drop the `nodeReuse` switch. I guess that means you may have stray msbuild processes hanging around after you run the command line build locally.

blairconrad:
@adamralph, we've enjoyed the binary log file in the past (especially @thomaslevesque).
The new Analysis libraries problem is consistent with something when I tried to use the 2.1.300. I didn't spend much time on it.
There's more, though: we all of a sudden don't understand the `in` keyword:

```
InParametersSpecs.cs(14,21): error CS1041: Identifier expected; 'in' is a keyword [C:\projects\fakeiteasy\tests\FakeItEasy.Specs\FakeItEasy.Specs.csproj]
```

I wonder if we shouldn't take a step back and (for now) go with your original PR and just ignore all whitespace in the output of finding the VS location (either the way you pushed it, or by optimistically trimming the result before checking, since we have to trim it anyhow).

blairconrad:
> > I think I know why the failure is not propagating. run-csx.cmd is calling csi to run build.csx but there is no fail-fast on that line in case of error. It needs some error handling like in adamralph/simple-exec:build.cmd@master

> Well spotted, @adamralph. The nodeReuse problem appears to be Microsoft/msbuild#1135, which is fixed but not released yet, AFAICT.

I'm backing up on that. `run-csx.cmd` works fine if I introduce a bad msbuild parameter and run it with via simple-targets: https://ci.appveyor.com/project/blairconrad/fakeiteasy/build/862

My new guess is that the failure to fail is related to Bullseye.

thomaslevesque:
I'd rather keep the binary log, it's really useful for diagnostic when something goes wrong. Isn't it possible to pass MSBuild arguments via `dotnet build`?

thomaslevesque:
> Isn't it possible to pass MSBuild arguments via dotnet build?

It is: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-msbuild#examples

I just tried, `dotnet msbuild /bl` generates a binary log.

adamralph:
@thomaslevesque if you use `dotnet msbuild` then yes, you can pass the usual MSBuild params (with the exception of `nodeReuse` as above). The problem with that is that is that the lack of support for `nodeReuse` could mean that phantom msbuild.exe processes are left over after running the build locally (not a concern in Appveyor since the machines are destroyed after use). If you want to keep using msbuild explicitly then I suggest waiting until https://github.com/Microsoft/msbuild/issues/1135 has been released.

So I discovered the problem in Bullseye. I originally wrote it for use in regular .NET projects, as I'm doing in SimpleExec, LiteGuard, xBehave.net, etc. To that end, I made it return an exit code from `Run()` which can be propagated to the exit code of `Program.Main`. However, in the context of a script, that doesn't work, since the returned value from a script is just a returned value, not an exit code. I'm going to change Bullseye to throw an exception in RC4. I had a feeling that switching from exceptions to error codes was going to bit me eventually, and hey presto. Lesson learned. Even when you think "yeah, but in this case..." libraries should _always_ respect the exception model, and not attempt to switch over to error codes. I'll send another PR to fix that problem as soon as I release Bullseye RC4.

I agree that we should drop this PR and go back to https://github.com/FakeItEasy/FakeItEasy/pull/1378. When https://github.com/Microsoft/msbuild/issues/1135 is released, the intent of this PR can be revisited.

thomaslevesque:
> if you use dotnet msbuild then yes, you can pass the usual MSBuild params

In fact, I meant to say that I tried `dotnet build /bl`, not `dotnet msbuild /bl`. So it works even if you don't explicitly invoke MSBuild.

adamralph:
Ah, interesting!

PR #1380, opened 2018-06-26
inline variable in build script

adamralph:


blairconrad:

PR #1383, opened 2018-06-27
Support in parameters in non-generic classes under .NET Core

### LINKED ISSUE #1338: Allow Faking of non-generic types with methods that have in parameters
As mentioned in #1337, and due to castleproject/Core#339, types containing a method with an `in` parameter cannot be faked by FakeItEasy .NET Standard, and if the type is a generic type, it cannot be faked by either the FakeItEasy .NET Standard or .NET Framework editions. 

Once Castle.Core supports these scenarios, we should do so as well.

blairconrad:
Fixes #1338.

### DIFF
--- a/docs/what-can-be-faked.md
+++ b/docs/what-can-be-faked.md
@@ -20,11 +20,8 @@ Note that special steps will need to be taken to
 
 ### Types whose methods have `in` parameters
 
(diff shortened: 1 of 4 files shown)


blairconrad:
Oh, that was fast! Thanks.

PR #1384, opened 2018-06-28
update to Bullseye 1.0.0-rc.5

adamralph:
Contains a minor fix concerning console output for targets with a `null` action (an unusual thing to do). You shouldn't notice any difference.


blairconrad:

PR #1389, opened 2018-07-04
Reject attempts to assign values to in parameters

### LINKED ISSUE #1388: Faked delegates with in parameters don't honor the read-only semantics
Reported by @zvirja https://github.com/FakeItEasy/FakeItEasy/issues/1338#issuecomment-402251902

Faked delegates with `in` parameters don't honor the read-only semantics of `in`, i.e. it's possible to configure the faked delegate to mutate the parameter:

```csharp
delegate void Foo(in int x);
...
var foo = A.Fake<Foo>();
A.CallTo(() => foo(A<int>.Ignored)).AssignsOutAndRefParameters(123);
int z = 0;
foo(z);
Console.WriteLine(z); // Should print 0, but prints 123
```

### LINKED ISSUE #1390: Faked types with in parameters don't honor the read-only semantics under .NET Framework
From https://github.com/FakeItEasy/FakeItEasy/issues/1388#issuecomment-402392358.

```c#
public interface IFoo
{
    void Bar(in int x);
}

var foo = A.Fake<IFoo>();
A.CallTo(() => foo.Bar(A<int>.Ignored)).AssignsOutAndRefParameters(123);
int z = 0;
foo.Bar(z);
Console.WriteLine(z); // Should print 0, but prints 123
```

### LINKED ISSUE #1391: Reject attempts to assign values to in parameters
Found while investigating #1388.

blairconrad:
Fixes #1388.
Fixes #1390.
Fixes #1391.

### DIFF
--- a/src/FakeItEasy/ParameterInfoExtensions.cs
+++ b/src/FakeItEasy/ParameterInfoExtensions.cs
@@ -1,0 +1,31 @@
+namespace FakeItEasy
+{
+    using System.Linq;
+    using System.Reflection;
+
+    /// <summary>
+    /// Provides extension methods for <see cref="ParameterInfo"/>.
+    /// </summary>
+    internal static class ParameterInfoExtensions
+    {
+        private const string IsReadOnlyAttributeFullName = "System.Runtime.CompilerServices.IsReadOnlyAttribute";
+
+        public static bool IsOutOrRef(this ParameterInfo parameterInfo)
+        {
+            if (!parameterInfo.ParameterType.IsByRef)
+            {
+                return false;
+            }
+
+#if FEATURE_PARAMETERINFO_CUSTOMATTRIBUTES_PROPERTY
+            var parameterAttributes = parameterInfo.CustomAttributes;
+            return parameterAttributes == null ||
+                   parameterAttributes.All(customAttributeData => customAttributeData.AttributeType.FullName != IsReadOnlyAttributeFullName);
+#else
+            var parameterAttributes = parameterInfo.GetCustomAttributesData();
+            return parameterAttributes == null ||
+                   parameterAttributes.All(customAttributeData => customAttributeData.Constructor.DeclaringType?.FullName != IsReadOnlyAttributeFullName);
+#endif
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


blairconrad:
I realized that we should be disallowing attempts to set `in` parameters, not just ignoring the attempt. So I tried a thing. It's a little rough, doesn't account for net40 (does it need to?), and relies on the name of the custom attribute, but it seems to work.

src/FakeItEasy/ParameterInfoExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            if (!parameterInfo.ParameterType.IsByRef)
+            {
+                return false;
+            }
+
+#if FEATURE_READONLY_IN_PARAMETERS
+            var parameterAttributes = parameterInfo.CustomAttributes;
+            return parameterAttributes == null ||
+                   parameterAttributes.All(customAttributeData => customAttributeData.AttributeType.FullName != "System.Runtime.CompilerServices.IsReadOnlyAttribute");
```
thomaslevesque: If I understand correctly, the `#if` is just an optimization; platforms that don't have `IsReadOnlyAttribute` will just never see the attribute. Right?
I wonder if it's worth keeping the `#if`. Projects that target net471+ should reference the net45 version of FakeItEasy, but if for some reason they reference the net40 version (e.g. because the assembly was referenced manually instead of via nuget, or because of a transitive reference), they won't get the correct behavior.

blairconrad: Honestly, I wasn't sure. Even though `ReadOnlyAttribute` is only available in .NET Framework 4.7.1+ and .NET Core 2.0+, it's clearly present on the parameter when we build to target .NET Framework 4.5 and .NET Standard 1.0, and run under .NET Framework 4.6.1 and netcoreapp 1.0. Of course it was put there by a super-modern compiler.
Regardless, the real reason I put the `#if` in was because in net40, the custom attributes look all different. I couldn't compile without it. When I combined that with the uncertainty over the value in trying to support `in` on net40, I figured I'd throw the `#if` in just to see how well it worked on the more modern platforms.

blairconrad:
Fixupped to work with net40. Sadly, the automated tests don't cover this, but locally I removed the net45 target from FakeItEasy (and `RaisingEventSpecs.cs`) and manually ran the tests, and all passed.

blairconrad:
Wow. Thanks, @thomaslevesque. Squashed.

PR #1396, opened 2018-07-24
Improved failed Dummy creation exception type and message

### LINKED ISSUE #1367: Misleading exception message when failing to create a Dummy
```c#
public class Unconstructable
{
    private Unconstructable() { }
}

[Test]
public void Test()
{
    A.Dummy<Unconstructable>();
}
```

Fails like so:

```
Test 'FakeItEasyQuestions2015.PrivateConstructor.Test' failed: FakeItEasy.Core.FakeCreationException : Unable to create fake object.
	at FakeItEasy.Creation.DefaultFakeAndDummyManager.CreateDummy(Type typeOfDummy)
	at FakeItEasy.A.Dummy[T]()
	PrivateConstructor.cs(17,0): at FakeItEasyQuestions2015.PrivateConstructor.Test()
```

"Unable to create fake object." is misleading. We were trying to make a Dummy, which need not be a Fake.

### LINKED ISSUE #1397: Dummy failure exception should include reasons for failure
From https://github.com/FakeItEasy/FakeItEasy/issues/1367#issuecomment-401353711.

Because a Dummy can be made in one of about 6 different ways, I think we should indicate each of the things that went wrong along the chain. Just reporting which public constructors failed when we tried to instantiate a type would be misleading when the class (in theory) could've been made into a Dummy had it been fakeable or if it had a Dummy Factory.

blairconrad:
Fixes #1367.
Fixes #1397.


tests/FakeItEasy.Specs/DummyCreationSpecs.cs
```diff
@@ -194,6 +196,179 @@ public abstract class DummyCreationSpecsBase
...
+                .x(() => exception.Message.Should().Be(@"
+  Failed to create dummy of type FakeItEasy.Specs.DummyCreationSpecsBase+PrivateAbstractClass:
+    It is abstract.
+"));
+        }
+
+        [Scenario]
+        public void ClassWhoseConstructorArgumentsCannotBeResolvedCreation(Exception exception)
+        {
+            "Given a type that can't be made into a Dummy because the arguements for its constructor cannot be resolved"
```
thomaslevesque: "arguements" => "arguments"

src/FakeItEasy/Creation/CreationResult.cs
```diff
@@ -60,13 +47,33 @@ public static CreationResult FailedToCreate(Type type, List<ResolvedConstructor>
             return new CreationResult(type, null, consideredConstructors);
         }
 
-        private string GetFailedToCreateResultMessage()
+        public object GetResultAsFake()
```
thomaslevesque: I'm not sure about this approach. The caller could easily call `GetResultAsFake` instead of `GetResultAsDummy`, or vice versa (of course, the caller is us, since it's an internal class, but still...)
Could the `CreationResult` record whether it's for a fake or dummy creation instead? I think it's not very straightforward, since some dummies are actually fakes, but it would probably be better. Forget about it if it's too complex

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
blairconrad: Thanks! Speling is hard.

src/FakeItEasy/Creation/CreationResult.cs
blairconrad: I had some of the same concerns, and even debated creating the `CreationResult` to know whether it was for a Fake or a Dummy, just as you suggest. I didn't mostly for the reason you mention: sometimes we start creating a Fake but actually intend to use it as a Dummy. Hmmm. I will quick try something to see how you like it. Stay tuned.

blairconrad: Okay. I put one rough attempt at https://github.com/blairconrad/FakeItEasy/commit/30b9ec09a994277806e879d7fcbc0284d824f8f5.

thomaslevesque: Not bad! But don't spend too much time on this, it's probably not worth it...

blairconrad: I massaged that rough attempt a little bit and added it here as a fixup. Tell me what you think.

src/FakeItEasy/Creation/CreationResult.cs
```diff
@@ -21,73 +23,83 @@ private CreationResult(object result)
...
-        public object GetResultAsDummy()
-        {
-            if (this.WasSuccessful)
-            {
-                return this.result;
-            }
-
-            throw new DummyCreationException(this.GetFailedToCreateResultMessage("dummy"));
+            original.creationMode = CreationMode.Dummy;
+            return original;
```
thomaslevesque: I would expect such a method to create a new object rather than modify the original. Maybe it's overkill, but immutability does help reducing bugs.

src/FakeItEasy/Creation/CreationResult.cs
```diff
@@ -231,5 +243,22 @@ public int GetHashCode(ResolvedConstructor obj)
...
+                this.ExceptionFactory = exceptionFactory;
+            }
+
+            public static CreationMode Fake { get; } = new CreationMode("fake", message => new FakeCreationException(message));
+
+            public static CreationMode Dummy { get; } = new CreationMode("dummy", message => new DummyCreationException(message));
+
+            public string Name { get; }
+
+            public Func<string, Exception> ExceptionFactory { get; }
```
thomaslevesque: nitpicking: maybe expose this as a method rather than a delegate?

blairconrad: Oh, yes. How foolish of me!

src/FakeItEasy/Creation/CreationResult.cs
blairconrad: Will do.

src/FakeItEasy/Creation/CreationResult.cs
blairconrad: Wait. A method that calls the delegate, or do you prefer me to subclass in order to have a method all the way down?

blairconrad: Taking the easy route for now. Pushed.

thomaslevesque: > A method that calls the delegate

That's what I had in mind

blairconrad:
Thanks, @thomaslevesque. I enjoyed your comments. Squashed and pushed.

PR #1398, opened 2018-07-28
Stop trying parameterless constructor

### LINKED ISSUE #1395: Parameterless constructor is always tried when creating fake classes, even if there isn't one
Found while working on #1367. We always ask DynamicProxy to create a Fake class using its parameterless constructor, even if the class hasn't got one.
This is wasteful and can clutter up the error messages.

blairconrad:
Fixes #1395.


blairconrad:
I tweaked the messages as we discussed, plus a little more.

PR #1399, opened 2018-07-30
Include action in CallTo documentation

### LINKED ISSUE #1342: Ensure documentation always shows an action after a call specification
[Specifying a method call or property get using an Expression](http://fakeiteasy.readthedocs.io/en/stable/specifying-a-call-to-configure/#specifying-a-method-call-or-property-get-using-an-expression) has a sample block of code that does nothing with the return value of `A.CallTo`. The next paragraphs explain that you should use the return value, but a hasty reading may lead people to copy the example and not supply an action (behaviour or call assertion).

I think we should rework the documentation to include a better example. I've not yet searched, but the bad examples may appear elsewhere.

blairconrad:
Fixes #1342.

### DIFF
--- a/docs/specifying-a-call-to-configure.md
+++ b/docs/specifying-a-call-to-configure.md
@@ -7,21 +7,16 @@ done using a method on the `A` class: `A.CallTo`.
 ## Specifying a method call or property `get` using an Expression
 
 ```csharp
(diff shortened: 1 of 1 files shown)


blairconrad:
I didn't find any other examples of "unfinished A.CallTos".

thomaslevesque:
Thanks @blairconrad!

PR #1400, opened 2018-08-02
Set new NuGet API key

thomaslevesque:


blairconrad:
Thanks, @thomaslevesque

blairconrad:
This change has been released as part of [FakeItEasy 4.8.0](https://github.com/FakeItEasy/FakeItEasy/releases/4.8.0).

Mertsch:
Is this a public key? Looks like this should be a build server secret?!

Mertsch:
@thomaslevesque Ah OK. Thanks for the info.

PR #1403, opened 2018-08-04
Fail if unsupported options are used when faking delegates

### LINKED ISSUE #1353: Custom attributes, additional interfaces, and constructor parameters unused when we create a Fake delegate
Mentioned by @thomaslevesque in https://github.com/FakeItEasy/FakeItEasy/pull/1352#discussion_r187769671

We should

1. document this, as [Creating Fakes](http://fakeiteasy.readthedocs.io/en/stable/creating-fakes/#explicit-creation-options) does not mention it, and
2. make sure we throw a nice exception if attributes are supplied during creation

blairconrad:
Fixes #1353.

I may have gotten carried away. I could do less. Or something entirely different.

| commit  | notes |
|-----------------|--------|
| 44e7563 :green_heart: Add test for creating delegates |  |
| 8628e71 :lipstick: Use CreationStrategy in FakeObjectCreator |  |
| 2457f8e :lipstick: Add DelegateCreationStrategy |  |
| 48fb8fd :lipstick: Introduce ICreationStrategy.IsResponsibleForCreating |  |
| 385d4e7 :x: Require fast failure when using illegal options to fake delegates |  |
| 6ed92fc :green_heart: Fail when using illegal options to fake delegates | Fixes the issue. We could stop here. |
| 71fd797 :lipstick: Remove ProxyGeneratorSelector | ProxyGeneratorSelector seemed thin now, with most of the work handled by the creation strategies |
| 9600be4 :lipstick: Remove IProxyGenerator | there was no longer any need to use a generator without knowing what it was |
| 7da6b87 :lipstick: Make a few things static | … to shut up some warnings |
| 9432f67 :lipstick: Remove ResolverFakeObjectCreator | it wasn't really doing much, and it let us reclaim `IFakeObjectCreator` |


src/FakeItEasy/Creation/IMethodInterceptionTester.cs
```diff
@@ -0,0 +1,16 @@
+namespace FakeItEasy.Creation
+{
+    using System.Reflection;
+
+    internal interface IMethodInterceptionTester
+    {
+        /// <summary>
+        /// Gets a value indicating whether the specified method can be intercanepted by  instance.
```
thomaslevesque: "intercanepted" ?

src/FakeItEasy/Creation/IMethodInterceptionTester.cs
```diff
@@ -0,0 +1,16 @@
...
+    internal interface IMethodInterceptionTester
+    {
+        /// <summary>
+        /// Gets a value indicating whether the specified method can be intercanepted by  instance.
+        /// </summary>
+        /// <param name="method">The member to test.</param>
+        /// <param name="callTarget">The instance the method will be called on.</param>
+        /// <param name="failReason">The reason the method can not be intercepted.</param>
+        /// <returns>True if the member can be intercepted.</returns>
+        bool MethodCanBeInterceptedOnInstance(MethodInfo method, object callTarget, out string failReason);
```
thomaslevesque: Looks like this interface has the same method as `CastleDynamicProxyInterceptionValidator`. Maybe `CastleDynamicProxyInterceptionValidator` could implement `IMethodInterceptionTester` (which could be renamed to `IMethodInterceptionValidator`), and be called directly without going through the `ProxyGenerator`?

blairconrad:
About dropping `IProxyGenerator`: first, I'd be content to not do that, if you prefer, but here was my motivation:

The proxy generators do different things and require different inputs to perform their work. It seemed better to me not to force them into all the same shape. In the following commit, I end up making the method signatures different (truth be told, I probably should've made those two commits as one. Here's what happened to `DelegateProxyGenerator.GenerateProxy`:

```diff
diff --git a/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs b/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
index 78e18ec..d55c0f4 100644
--- a/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
+++ b/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
@@ -1,4 +1,4 @@
-        public virtual ProxyGeneratorResult GenerateProxy(
+        public static ProxyGeneratorResult GenerateProxy(
             Type typeOfProxy,
-            IEnumerable<Type> additionalInterfacesToImplement,
-            IEnumerable<object> argumentsForConstructor,
-            IEnumerable<Expression<Func<Attribute>>> attributes,
             IFakeCallProcessorProvider fakeCallProcessorProvider)
```

Had we had this signature to begin with, we'd've known off the bat the 3 options weren't used. We now have the `CreationStrategy`s to provide a common interface on top of our proxy generators.

Still, if you'd like it rolled back, I've no problem doing so.

src/FakeItEasy/Creation/IMethodInterceptionTester.cs
blairconrad: Yes. That's not a word. I don't know how I managed that.

docs/creating-fakes.md
```diff
@@ -28,15 +28,16 @@ IList<object> fakes = Create.CollectionOfFake(type, 10);
...
+When creating fakes you can, through a fluent interface, specify options for how the fake should be created, depending on the type of fake being made:
+
+| Option                                                                                            | Applies to    |
+|---------------------------------------------------------------------------------------------------|---------------|
+| Specify arguments for the constructor of the faked type                                           | non-delegates |
+| Specify additional interfaces that the fake should implemen                                       | non-delegates |
+| Assign additional custom attributes to the faked type                                             | non-delegates |
+| Cause a fake to have [strict mocking semantics](strict-fakes.md)                                  | any fake      |
+| Configure all of a fake's methods to [use their original implementation](calling-base-methods.md) | classes       |
+| Create a fake that wraps another object, and optinally specify a recorder                         | any fake      |
```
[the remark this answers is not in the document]
blairconrad: Glad you like it. I was unsure about the change. Note that I just pushed a fixup that removes some additional lines that I'd accidentally added in the "pass" commit. They are redundant, now that the table is here.

blairconrad:
I can pursue the `IMethodInterceptionValidator` change, but I think the shape will depend on whether we do get rid of `IProxyGenerator`. For example, if we make the `IMethodInterceptionValidator` change and drop `IProxyGenerator`, then `CastleDynamicProxyGenerator` can become static…

thomaslevesque:
Ah, I see. OK to drop `IProxyGenerator` then.

> For example, if we make the IMethodInterceptionValidator change and drop IProxyGenerator, then CastleDynamicProxyGenerator can become static…

Indeed! That would be better, I think. It was bothering me that `DelegateProxyGenerator` was static but `CastleDynamicProxyGenerator` was not.

blairconrad:
Thanks for the comments, @thomaslevesque. I think I've addressed everything, and I've repushed. The commits are the same up to the `fail` commit, but then deviate. I left the original branch at [original-attributes-unused](https://github.com/blairconrad/FakeItEasy/tree/773021ada31baee548671c2033e7c2c5a83bdc1d) in case you want to compare or whatever.

blairconrad:
Thanks for improvements and merges!

PR #1409, opened 2018-08-13
upgrade Bullseye and SimpleExec

adamralph:


tools/build.csx
```diff
@@ -117,18 +116,13 @@ Targets.Add(
...
     DependsOn("build"),
-    () =>
-    {
-        foreach (var pdb in pdbs)
-        {
-            Run(pdbGit, $"-u {repoUrl} -s {pdb}");
-        }
-    });
+    pdbs,
+    pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}"));
```
blairconrad: This is an interesting syntax. I'm uneasy about the magic connection between `pdbs` and the `pdb` variable.

adamralph: Would it help if the parameter name(s) were included?

```c#
forEach: pdbs,
action: pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}"));
```

adamralph: 👍 thanks! This is the very reason I raise these pre-release PR's. The feedback is invaluable! (In retrospect perhaps it should have beta rather than RC.)

blairconrad: So I asked myself what I'd do if I wanted a more concise version of the original code if I didn't have access to the new Bullseye feature, and ended up with

```diff
 Target(
     "pdbgit",
     DependsOn("build"),
+    () => Array.ForEach(
         pdbs,
-    pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}"));
+        pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}")));
 
 RunTargets(Args);
```

(relative your original change). It's got a little more noise, but I think it makes the relationship between `pdbs` and `pdb` clear. Or if `pdbs` were a list:

```diff
 Target(
     "pdbgit",
     DependsOn("build"),
-    pdbs,
-    pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}"));
+    () => pdbs.ForEach(
+        pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}")));
 
 RunTargets(Args);
``` 
 
Either one has a little more noise than your version (although perhaps not if the parameter names were included), and is arguably not as pretty, but I do enjoy the clarity and knowing that we'll have a single task argument after the `DependsOn`.

blairconrad: Ah. I did not realize that Bullseye gained additional powers beyond syntactic sugar. I didn't walk through all the issues, and I didn't see anything about the feature in the docs (did I miss it, or is there an opportunity to augment the docs?).
I do like this enhanced output.

adamralph: Yeah, the only docs that exist are in the README.md. Which of course has the problem that anyone reading them on the default branch will be reading those for the current pre-release rather than the current stable (assuming they are updated together with the feature addition)

The only docs that exist for this feature currently are in the issue: https://github.com/adamralph/bullseye/pull/36

blairconrad: Thanks for that. I see that originally, you'd imagined a lightweight wrapper over the enumerable, not unlike `DependsOn`.  I think it helped, but I can understand if you want to limit number of such things, or even message chaining. The parameter names do help. I wonder if a small twist wouldn't read better, though:

```c#
 Target(
     "pdbgit",
     DependsOn("build"),
     forEach: pdbs,
     do: pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}"));
```

alternatively, if you did want to get fancier with a more fluent API:

```c#
 Target(
     "pdbgit",
     DependsOn("build"),
     pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}"))
    .ForEachElementOf(pdbs);
```

There's no way to force people to complete the statement with the `ForEachElement`, but you could fail quickly up front as soon as `RunTargets` is called if they forgot.

Oh, I suppose that could evolve to

```c#
 Target(
     "pdbgit",
     DependsOn("build"))
     .Does(pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}")))
    .ForEachElementOf(pdbs);
```

But maybe I've gotten carried away.

Bottom line, now that I understand what the new call offers, I do like it, and I think your original proposal plus the explicit parameter names is clear enough, if you don't see any gain in further altering the syntax.

adamralph: I'd rather avoid introducing a fluent API. I considered it during initial development but decided against it.

I quite like the `do` parameter name. I'm not sure if I can do that in a minor version though. I guess it's unlikely that anyone is stating the parameter name currently. However, this is almost like building a little "fluent-ish" API in the parameter names, i.e. we currently have:

```c#
Target(
    name: ...,
    dependsOn: ...,
    forEach: ...,
    action: ...,
```

 which is a bit of a mix.

I guess fully "fluentish" would be something like:

```c#
Target(
    named: ...,
    dependsOn: ...,
    forEachItemIn: ...,
    does: ...,
```

and "non-fluent" or "conventional" naming would be:

```c#
Target(
    name: ...,
    dependencies: ...,
    inputs: ...,
    action: ...,
```

Perhaps it's time to stop trying to build a DSL within a method signature and switch to conventional naming?

Although, at second glance, your proposal does look quite nice:

```c#
Target(
    name: ...,
    dependsOn: ...,
    forEach: ...,
    do: ...,
```

I love the smell of bike shedding in the morning... 😆 

BTW - `ForEach()` does exist, but only for the same reason as `DependsOn()`. If we ever get terse array declaration in C# then I'll remove both.

adamralph: @blairconrad what colour shall we paint this shed? I'm happy to rename the parameter from `action` to `do` if you're keen on it.

blairconrad: Sorry, for the delay. I got sidetracked.
Given the DSL-ish names of the other parameters, I'd say go with `do`.

adamralph: @blairconrad no problem. I have bad news though. I tried to make the change and immediately hit a wrinkle: `do` is a C# keyword, so the parameter would have to be named `@do`. 😏 

Unless we can think of a compelling alternative, I'd say we leave it as `action`. I'm not very keen on `does`.

blairconrad: Now that you say it out loud, I've heard of this `do`!

I think 

```c#
Target(
     "pdbgit",
     DependsOn("build"),
     forEach: pdbs,
     action: pdb => Run(pdbGit, $"-u {repoUrl} -s {pdb}"));
```

is livewithable. Unless @thomaslevesque absolutely hates it (note: I'm talking about how FIE would use the call, not trying to "dictate" what you can do in Bullseye, of course!).

blairconrad: once you have `forEach`, do you not _need_ `action`?

adamralph:
Oh, I think you're right. IIRC what I was suggesting will be allowed in C# 8.

tools/build.csx
thomaslevesque: I don't hate it. `forEach`/`do` would have been nicer, but `forEach`/`action` is OK.

adamralph:
I've amended the commit to include `forEach: ` and `action: `.

blairconrad:

Thanks, @adamralph. Look for your name in the release notes. :trophy:

PR #1412, opened 2018-08-18
Serialize dummy/fake creation in specs that count constructor calls

blairconrad:
As noted in https://github.com/FakeItEasy/FakeItEasy/pull/1410#issuecomment-413938393, there is a race condition in a couple of tests that count the number of times failing constructors are called. This should ensure we run those tests serially rather than concurrently.


blairconrad:
Mind like a sieve. I meant to assert that the waits succeed. I can amend on Sunday. And/or respond to other comments.

blairconrad:
Instead of checking the `Wait`, now releasing the lock in a `Teardown`, which I had forgotten existed.

blairconrad:
Thank you!

blairconrad:

PR #1414, opened 2018-08-28
Convert csx scripts to csproj

### LINKED ISSUE #1413: Consider replacing the csx scripts by full-fledged csprojs
As discussed in https://github.com/FakeItEasy/FakeItEasy/pull/1410#discussion_r211909659

@adamralph said:

> You might find it simpler to do away with scripting completely, as I've done with xbehave and @blairconrad has done with SelfInitializingFakes. It allows you to drop nuget.exe, csi.exe, packages.config, etc. All you need is a csproj for package references and a Program.Main for your build "script". You can then invoke simply with dotnet run.

It would make things easier and IDE support would be better. Although it feels kind of strange to run a C# project to build a C# project :wink:

### LINKED ISSUE #1415: Source stepping is not available for net45 library
Somehow we're missing it!
- https://github.com/FakeItEasy/FakeItEasy/pull/1414/files#r214055579

thomaslevesque:
Fixes #1413
Fixes #1415 

Note: this is a direct translation of the existing scripts, I didn't attempt to make things cleaner at this point. I just made the code look more like a real program (e.g. converted "global variables" to static fields).


tools/FakeItEasy.Build/FakeItEasy.Build.csproj
```diff
@@ -0,0 +1,41 @@
...
+    <PackageReference Include="Bullseye" Version="1.1.0-rc.2" />
+    <PackageReference Include="SimpleExec" Version="2.2.0" />
+    <PackageReference Include="GitVersion.CommandLine" Version="4.0.0-beta0012" ToolName="GitVersion" ToolExe="GitVersion.exe" />
+    <PackageReference Include="PdbGit" Version="3.0.41" ToolName="PdbGit" ToolExe="PdbGit.exe" />
+    <PackageReference Include="xunit.runner.console" Version="2.0.0" ToolName="XUnit" ToolExe="xunit.console.exe" />
+    <PackageReference Include="vswhere" Version="1.0.62" ToolName="VSWhere" ToolExe="vswhere.exe" />
+    <PackageReference Include="NuGet.CommandLine" Version="4.1.0" ToolName="NuGet" ToolExe="nuget.exe" />
+  </ItemGroup>
+
+  <Target Name="GenerateToolPaths" BeforeTargets="BeforeBuild">
```
thomaslevesque: This is used to generate a code file with constants for tool paths.
It's slightly ugly, but hey, it works :wink:

tools/FakeItEasy.Build/FakeItEasy.Build.csproj
```diff
@@ -0,0 +1,41 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Bullseye" Version="1.1.0-rc.2" />
+    <PackageReference Include="SimpleExec" Version="2.2.0" />
+    <PackageReference Include="GitVersion.CommandLine" Version="4.0.0-beta0012" ToolName="GitVersion" ToolExe="GitVersion.exe" />
```
thomaslevesque: The extra properties `ToolName` and `ToolExe` are used in the target below.

tools/FakeItEasy.Build/FakeItEasy.Build.csproj
```diff
@@ -0,0 +1,41 @@
...
+      <ClassName>ToolPaths</ClassName>
+      <OutputFile>$(IntermediateOutputPath)\$(ClassName).cs</OutputFile>
+    </PropertyGroup>
+    <ItemGroup>
+      <Tool Include="@(PackageReference)" Condition="'%(PackageReference.ToolName)' != ''" />
+      <Compile Include="$(OutputFile)" />
+    </ItemGroup>
+    <PropertyGroup>
+      <Semicolon>%3B</Semicolon>
+      <Indent>%20%20%20%20%20%20%20%20</Indent>
```
thomaslevesque: `;` has a special meaning for MSBuild, so it needs to be escaped. And for some reason, indenting with spaces works on the first line, but the spaces are stripped on the following lines, so I escape the spaces as well.

tools/FakeItEasy.PrepareRelease/FakeItEasy.PrepareRelease.csproj
```diff
@@ -0,0 +1,12 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Octokit" Version="0.26.0" />
+  </ItemGroup>
+
+</Project>
```
thomaslevesque: Note: I built this project but didn't actually try to run it, since there's no "dry run" option.

tools/FakeItEasy.PrepareRelease/Program.cs
```diff
@@ -0,0 +1,280 @@
...
+        {
+            var version = args.FirstOrDefault();
+            if (version == null)
+            {
+                throw new Exception("No version supplied");
+            }
+
+            var releaseBranchName = $"release/{version}";
+
+            var gitHubClient = GetAuthenticatedGitHubClient();
```
thomaslevesque: This was a `static var` in the script so that it could be used by other methods. The equivalent in a real C# program would have been to make it a static field, but it felt ugly, so instead I kept it as a local variable and converted the static methods to extension methods with `gitHubClient` as the `this` parameter.

tools/prepare_release.cmd
```diff
@@ -1 +1 @@
-@powershell -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "%~dp0\tools\run-csx.ps1 %~dp0\tools\prepare-release.csx %*"
```
thomaslevesque: This was never tested and probably didn't work, because the path was wrong... I broke it in https://github.com/FakeItEasy/FakeItEasy/pull/1410

tools/FakeItEasy.Build/FakeItEasy.Build.csproj
```diff
@@ -0,0 +1,41 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Bullseye" Version="1.1.0-rc.2" />
+    <PackageReference Include="SimpleExec" Version="2.2.0" />
+    <PackageReference Include="GitVersion.CommandLine" Version="4.0.0-beta0012" ToolName="GitVersion" ToolExe="GitVersion.exe" />
+    <PackageReference Include="PdbGit" Version="3.0.41" ToolName="PdbGit" ToolExe="PdbGit.exe" />
+    <PackageReference Include="xunit.runner.console" Version="2.0.0" ToolName="XUnit" ToolExe="xunit.console.exe" />
+    <PackageReference Include="vswhere" Version="1.0.62" ToolName="VSWhere" ToolExe="vswhere.exe" />
+    <PackageReference Include="NuGet.CommandLine" Version="4.1.0" ToolName="NuGet" ToolExe="nuget.exe" />
```
thomaslevesque: This rids us of downloading nuget.exe manually :wink:
(we should probably consider upgrading the version too, but I didn't want to make unrelated changes)

thomaslevesque:
Oops. For some reason it doesn't work on AppVeyor, although it works on my machine. Let me try something.

tools/FakeItEasy.Build/FakeItEasy.Build.csproj
```diff
@@ -18,10 +18,11 @@
   <Target Name="GenerateToolPaths" BeforeTargets="BeforeBuild">
     <PropertyGroup>
       <ClassName>ToolPaths</ClassName>
-      <OutputFile>$(IntermediateOutputPath)\$(ClassName).cs</OutputFile>
+      <OutputFile>$(ClassName).cs</OutputFile>
     </PropertyGroup>
     <ItemGroup>
       <Tool Include="@(PackageReference)" Condition="'%(PackageReference.ToolName)' != ''" />
+      <Compile Remove="$(OutputFile)" />
```
thomaslevesque: I have to remove the file before including it, because if it already existed, it would already be implicitly included. There might be a way to use a `Condition` for this (i.e. "include if not already included"), but it was easier this way.

tools/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+<Project>
+    <!-- To suppress inherited Directory.Build.props -->
+
+    <PropertyGroup>
+        <LangVersion>latest</LangVersion>
+    </PropertyGroup>
+
+</Project>
```
blairconrad: Why are we suppressing the Directory.Build.props?

I could check out and remove this file and find out myself, I suppose.

blairconrad: Hmm. Mostly things we could easily fix (make a few fields private and const) plus 

> CSC : warning CS8032: An instance of analyzer Microsoft.CodeQuality.Analyzers.ApiDesignGuidelines.ParameterNamesShouldMatchBaseDeclarationAnalyzer cannot be created from D:\Users\amidu\.nuget\packages\microsoft.codequality.analyzers\2.6.0\analyzers\dotnet\cs\Microsoft.CodeQuality.Analyzers.dll : Could not load file or assembly 'Microsoft.CodeAnalysis, Version=2.6.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. The system cannot find the file specified.. [D:\Sandbox\FakeItEasy\tools\FakeItEasy.Build\FakeItEasy.Build.csproj]

This seems familiar, but I can't remember how to get around it. One option is to push 
```
<PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.6.0" PrivateAssets="all" />
```

down into src and tests' Directory.Build.props

tools/FakeItEasy.Build/.gitignore
```diff
@@ -0,0 +1 @@
+ToolPaths.cs
```
blairconrad: we don't just add the appropriate line to the top level .gitignore?

tools/FakeItEasy.Build/FakeItEasy.Build.csproj
blairconrad: I don't mind it so much.

tools/FakeItEasy.Build/FakeItEasy.Build.csproj
```diff
@@ -0,0 +1,42 @@
...
+      <Compile Include="$(OutputFile)" />
+    </ItemGroup>
+    <PropertyGroup>
+      <Semicolon>%3B</Semicolon>
+      <Indent>%20%20%20%20%20%20%20%20</Indent>
+      <FileContent><![CDATA[namespace FakeItEasy.Build
+{
+    internal static class $(ClassName)
+    {
+@(Tool -> '$(Indent)public static string %(ToolName) = @"$(NuGetPackageRoot)%(Identity)\%(Version)\tools\%(ToolExe)"$(Semicolon)')
```
blairconrad: const instead of static? Silences some warnings if you end up running StyleCop on the generated output, like I'm pushing for above.

build.cmd
```diff
@@ -1 +1 @@
-@powershell -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "%~dp0\tools\run-csx.ps1 %~dp0\tools\build.csx %*"
+@dotnet run --project "%~dp0\tools\FakeItEasy.Build\FakeItEasy.Build.csproj"
```
blairconrad: add a `-- %*` on the end of this so we don't always execute the default build.

tools/FakeItEasy.Build/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                "tests/FakeItEasy.Specs"
+            },
+            ["approve"] = new[]
+            {
+                "tests/FakeItEasy.Tests.Approval"
+            }
+        };
+
+        private static string LogsDirectory = "./artifacts/logs";
```
blairconrad: nit: `const` and up near `TestsDirectory`?

tools/FakeItEasy.Build/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+            "./src/FakeItEasy/FakeItEasy.csproj",
+            "./src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.CSharp.csproj",
+            "./src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.VisualBasic.csproj"
+        };
+
+        private const string AnalyzerMetaPackageNuspecPath = "./src/FakeItEasy.Analyzer.nuspec";
+
+        private static readonly string[] Pdbs =
+        {
+            "src/FakeItEasy/bin/Release/net40/FakeItEasy.pdb",
```
blairconrad: any idea why we don't do net45? I think this is pre-existing.

tools/FakeItEasy.Build/Program.cs
```diff
@@ -0,0 +1,157 @@
...
+            Target("clean", DependsOn("logsDirectory"), () => RunMsBuild("Clean"));
+
+            Target(
+                "restore",
+                () => Run("dotnet", $"restore"));
+
+            Target(
+                "unit",
+                DependsOn("build", "testsDirectory"),
+                () => RunTests("unit"));
```
blairconrad: A bonus change, for another review, but I think we could use the fancy new "target that iterates" feature on these tests.

tools/FakeItEasy.PrepareRelease/FakeItEasy.PrepareRelease.csproj
blairconrad: I usually just modify it to return early before we start modifying things. Right after the new check for release notes/milestone agreement.

tools/prepare_release.cmd
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+dotnet run --project "%~dp0\FakeItEasy.PrepareRelease\FakeItEasy.PrepareRelease.csproj"
```
blairconrad: we'll need a `-- %*` here too

blairconrad:
Generally nice, @thomaslevesque, and a fair bit of work. I noticed some things that are definite errors, some things I think we should change, and made an annoying comment about turning off our stylecop analysis in the new csprojs!

.gitignore
```diff
@@ -1,15 +1,12 @@
...
 .vs
 [Oo]bj
 [Bb]in
 [Oo]bjcs
 [Oo]bjvb
 artifacts
-packages
 src/VersionInfo.cs
 src/FakeItEasy.nuspec
-tools/packages.config.sha1
```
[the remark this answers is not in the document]
thomaslevesque: I didn't think of that. I guess I could leave it there, as it does no harm.

tools/Directory.Build.props
thomaslevesque: Well, the root Directory.Build.props has plenty of things we don't necessarily want in the build project: analyzers, rule sets, strong name key... So I thought it would be simpler to start from a blank slate.
And also the warnings you noticed. I don't know what's causing them. I can look into it if you like.

tools/FakeItEasy.Build/.gitignore
thomaslevesque: We could. I thought it would be cleaner this way, but I can change it.

tools/FakeItEasy.Build/FakeItEasy.Build.csproj
thomaslevesque: Oh! Yes, of course. For some reason I considered making it a method at some point, and forgot to change the static. I'll fix it.

build.cmd
thomaslevesque: Oops! Yes, of course.

tools/FakeItEasy.Build/Program.cs
thomaslevesque: Sure

tools/FakeItEasy.Build/Program.cs
thomaslevesque: No idea. We certainly should.

tools/FakeItEasy.PrepareRelease/FakeItEasy.PrepareRelease.csproj
thomaslevesque: 👍 
I'll do that.

tools/prepare_release.cmd
thomaslevesque: Will do

tools/Directory.Build.props
blairconrad: I'm not against analyzers and rule sets being in the build project (admittedly they are less important).  Chasing the Microsoft.CodeQuality.Analyzers warnings is probably not worth it, so however you want to avoid, I'm content. Either as you have the directory.build.props as you do or maybe include the StyleCop analyzer (but I don't feel strongly enough to insist even on that; it's just that I'd default to attempting to preserver the style from the rest of the project).

tools/FakeItEasy.Build/.gitignore
blairconrad: I know every time I go back to svn (where every ignore is in the folder) I'm annoyed at having to repeat myself, so perhaps I like the top-level .gitignore too much. We do already ignore some files in lower level directories from that file, so I thought that was our intention going forward. I think I do prefer one ignore file, but if you do want to have a second, shouldn't we have one in src as well?

tools/FakeItEasy.Build/Program.cs
blairconrad: Created #1415.

tools/FakeItEasy.Build/Program.cs
blairconrad: Of course. I created #1416.

tools/FakeItEasy.Build/.gitignore
thomaslevesque: > I think I do prefer one ignore file, but if you do want to have a second, shouldn't we have one in src as well?

And one int tests, for the `.received.txt` files... OK, you got me there. I don't want to go down that road at this point. I'll put everything in the same .gitignore.

tools/FakeItEasy.PrepareRelease/FakeItEasy.PrepareRelease.csproj
thomaslevesque: Well, good thing I tried, because it was broken. The .githubtoken is not in the same directory, but in the parent directory. We could keep the code as is and just move the file, but if we make other tools that use it, it would be better to have it in a common location. What would you prefer?

tools/Directory.Build.props
thomaslevesque: Apparently the issue is caused by the SDK version we use (2.0.3). I don't get those warnings with the latest (2.1.400). We should probably upgrade, but I get build errors in the FIE analyzer projects with the latest SDK, not sure why... For now, I guess we could just remove the `Microsoft.CodeAnalysis.FxCopAnalyzers` package from the build project, and solve the issue separately.

tools/FakeItEasy.PrepareRelease/FakeItEasy.PrepareRelease.csproj
thomaslevesque: For now, I just manually moved the file; you'll have to do the same. I can change it if you like.

thomaslevesque: Yes

blairconrad:
Thanks, @thomaslevesque!

adamralph:
Nice work! The `ToolPaths.cs` generation is... fascinating! 😆

thomaslevesque:
> The ToolPaths.cs generation is... fascinating!

lol, it's a kinder word than I'd have used 😉

PR #1418, opened 2018-09-01
Build tweaks

adamralph:


build.cmd
```diff
@@ -1 +1,2 @@
-@dotnet run --project "%~dp0\tools\FakeItEasy.Build\FakeItEasy.Build.csproj" -- %*
+@cd %~dp0
```
adamralph: Relative paths are used inside `Program.cs`. If you don't change working directory, then the targets will fail when run from a subdirectory, e.g. `../build.cmd`.

tools/FakeItEasy.Build/Program.cs
```diff
@@ -104,7 +104,7 @@ public static void Main(string[] args)
                 DependsOn("build", "outputDirectory", "pdbgit"),
                 () =>
                 {
-                    var version = Read(ToolPaths.GitVersion, "/showvariable NuGetVersionV2", ".");
+                    var version = Read(ToolPaths.GitVersion, "/showvariable NuGetVersionV2");
```
adamralph: When `SimpleExec` is passed a working directory, it prints an extra line like so:

`Working directory: .`

Both this argument and this line are redundant, since `"."` simple refers to the current directory.

The other changes to remove `./` are pure refactoring (no observable effect).

adamralph:
BTW, you could choose to prefix _all_ paths, with `./` for explicitness, but it should be consistent. The paths used for `Pdbs` and `TestSuites` are already missing the prefix, so I removed them in all other places for consistency.

build.cmd
adamralph: @blairconrad it only changes it for the scope of the script:

image

thomaslevesque: > Or I might just swap build.cmd for build.ps1 to force people to use PowerShell. 😉

I'd consider doing that for FakeItEasy too... I never use cmd anymore.

adamralph:
@thomaslevesque I swapped the `cd` for `pushd` and `popd`. Since the change was so trivial I amended the first commit directly, rather than pushing a fixup.

thomaslevesque:
> @thomaslevesque I swapped the cd for pushd and popd. Since the change was so trivial I amended the first commit directly, rather than pushing a fixup.

👍 
Thanks!

thomaslevesque:
@blairconrad since your concerns were addressed, I'm taking the liberty of dismissing your review

thomaslevesque:
Thanks @adamralph!

blairconrad:

Thanks, @adamralph. Look for your name in the release notes. :trophy:

PR #1419, opened 2018-09-01
Split analyzers into separate directories

thomaslevesque:
This gets rid of two [warnings](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/637-bqrwurem#L13) that had been annoying me for a while:

> C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(818,5): warning MSB3539: The value of the property "BaseIntermediateOutputPath" was modified after it was used by MSBuild which can lead to unexpected build results. Tools such as NuGet will write outputs to the path specified by the "MSBuildProjectExtensionsPath" instead. To set this property, you must do so before Microsoft.Common.props is imported, for example by using Directory.Build.props.  For more information, please visit https://go.microsoft.com/fwlink/?linkid=869650 [C:\projects\fakeiteasy\src\FakeItEasy.Analyzer\FakeItEasy.Analyzer.VisualBasic.csproj]

And it's cleaner anyway. Having two projects in the same directory forced us to jump through hoops, like specifying a different intermediate output path (objcs/objvb instead of obj), and explicitly excluding files from the other project's intermediate output path.


src/FakeItEasy.Analyzer.CSharp/FakeItEasy.Analyzer.CSharp.csproj
```diff
@@ -1,16 +1,16 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
+  <Import Project="../FakeItEasy.Analyzer.Shared/Directory.Build.props" />
+
   <PropertyGroup>
     <TargetFramework>netstandard1.1</TargetFramework>
-    <AssemblyName>FakeItEasy.Analyzer.CSharp</AssemblyName>
```
thomaslevesque: Implicit

src/FakeItEasy.Analyzer.CSharp/FakeItEasy.Analyzer.CSharp.csproj
```diff
@@ -1,16 +1,16 @@
...
     <TargetFramework>netstandard1.1</TargetFramework>
-    <AssemblyName>FakeItEasy.Analyzer.CSharp</AssemblyName>
     <RootNamespace>FakeItEasy.Analyzer</RootNamespace>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <FileAlignment>512</FileAlignment>
     <SignAssembly>true</SignAssembly>
     <DefineConstants>$(DefineConstants);CSHARP</DefineConstants>
     <PackageTargetFallback>portable45-net45+win8</PackageTargetFallback>
-    <BaseIntermediateOutputPath>objcs\</BaseIntermediateOutputPath>
```
thomaslevesque: No longer necessary

src/FakeItEasy.Analyzer.CSharp/FakeItEasy.Analyzer.CSharp.csproj
```diff
@@ -21,25 +21,6 @@
     <PackageTags>$(PackageTags);csharp;c#</PackageTags>
   </PropertyGroup>
 
-  <ItemGroup>
-    <EmbeddedResource Remove="VisualBasic/Resources.resx" />
-    <Compile Remove="objvb\**" />
-    <EmbeddedResource Remove="objvb\**" />
-    <None Remove="objvb\**" />
-  </ItemGroup>
```
thomaslevesque: No longer necessary

src/FakeItEasy.Analyzer.CSharp/FakeItEasy.Analyzer.CSharp.csproj
```diff
@@ -21,25 +21,6 @@
...
-    <Compile Include="../*.cs" />
-  </ItemGroup>
-
-  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-    <DebugType>full</DebugType>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
-    <DebugType>pdbonly</DebugType>
-  </PropertyGroup>
```
thomaslevesque: Moved to ../FakeItEasy.Analyzer.Shared/Directory.Build.props

src/FakeItEasy.Analyzer.Shared/DiagnosticDefinitions.cs
```diff
@@ -11,11 +11,7 @@ internal static class DiagnosticDefinitions
                 ResourceBaseName,
                 typeof(DiagnosticDefinitions).GetTypeInfo().Assembly);
 
-#if CSHARP
-        private const string ResourceBaseName = "FakeItEasy.Analyzer.CSharp.Resources";
-#elif VISUAL_BASIC
-        private const string ResourceBaseName = "FakeItEasy.Analyzer.VisualBasic.Resources";
-#endif
```
thomaslevesque: The CSharp/VisualBasic subfolders are no longer necessary

src/FakeItEasy.Analyzer.Shared/Directory.Build.props
```diff
@@ -0,0 +1,25 @@
...
+
+  <ItemGroup>
+    <!-- Shared source files -->
+    <Compile Include="$(MSBuildThisFileDirectory)../*.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)**/*.cs" />
+    <!-- Analyzer installation scripts -->
+    <Content Include="$(MSBuildThisFileDirectory)tools/*.ps1" Pack="true" PackagePath="tools/%(Filename)%(Extension)" />
+  </ItemGroup>
+
+</Project>
```
thomaslevesque: While I was at it, I extracted some common bits to this file.
The file name might be misleading, though, since it's not a "real" Directory.Build.props (in the sense that it's not implicitly imported). Maybe we should rename it to something else, e.g. `FakeItEasy.Analyzers.Shared.proj`.

thomaslevesque: I'd like to try something. Currently it's difficult to add new files to the projects, because there's no place to do it in the solution explorer. Introducing a shared project might make things easier (I'm not sure it works with the .NET Core SDK though, I need to try it)

thomaslevesque: OK, it works with a shared project, but it's not perfect. Shared projects don't have all the nice features of SDK-style projects, so the files are listed explicitly in the `.projitems` file (we can manually add glob patterns, but VS doesn't really support it so new files are added explicitly, which causes them to be included twice in the build). But at least we can now add files to the right place from Visual Studio.

Let me know what you think.

thomaslevesque:
rebased, and removed [WIP]

thomaslevesque:
Thanks @blairconrad!

blairconrad:

PR #1421, opened 2018-09-02
switch targets to using enumerable inputs

### LINKED ISSUE #1416: Run tests using Bullseye's fancy "enumerable input" feature
From https://github.com/FakeItEasy/FakeItEasy/issues/1415

We loop over test directories manually and run the tests. Bullseye may be able to do this for us, supplying more granular output.

adamralph:
Fixes #1416

### DIFF
--- a/tools/FakeItEasy.Build/Program.cs
+++ b/tools/FakeItEasy.Build/Program.cs
@@ -81,29 +81,14 @@
                 "restore",
                 () => Run("dotnet", $"restore"));
 
-            Target(
-                "unit",
-                DependsOn("build", "testsDirectory"),
-                forEach: TestSuites["unit"],
-                action: testDirectory => RunTests(testDirectory));
-
-            Target(
-                "integ",
-                DependsOn("build", "testsDirectory"),
-                forEach: TestSuites["integ"],
-                action: testDirectory => RunTests(testDirectory));
-
-            Target(
-                "spec",
-                DependsOn("build", "testsDirectory"),
-                forEach: TestSuites["spec"],
-                action: testDirectory => RunTests(testDirectory));
-
-            Target(
-                "approve",
-                DependsOn("build", "testsDirectory"),
-                forEach: TestSuites["approve"],
-                action: testDirectory => RunTests(testDirectory));
+            foreach (var testSuite in TestSuites)
+            {
+                Target(
+                    testSuite.Key,
+                    DependsOn("build", "testsDirectory"),
+                    forEach: testSuite.Value,
+                    action: testDirectory => RunTests(testDirectory));
+            }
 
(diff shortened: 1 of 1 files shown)


tools/FakeItEasy.Build/Program.cs
```diff
@@ -83,36 +84,43 @@ public static void Main(string[] args)
...
+
+            Target(
+                "pack-projects",
+                DependsOn("build", "outputDirectory", "pdbgit", "get-version"),
+                forEach: ProjectsToPack,
+                action: project => Run("dotnet", $"pack {project} --configuration Release --no-build --output {OutputDirectory} /p:Version={version}"));
+
+            Target(
+                "pack-nuspecs",
+                DependsOn("build", "outputDirectory", "pdbgit", "get-version"),
```
thomaslevesque: This one doesn't actually depend on `pdbgit`. It just creates an empty package with dependencies on other packages. In fact, it doesn't even depend on `build` ;)

blairconrad:
Dude, you're stealing my work! (#1416).

tools/FakeItEasy.Build/Program.cs
adamralph: Ah, ok. I'l remove the deps on `pdbgit` and `build`.

adamralph:
@blairconrad oh, sorry! So much for checking current issues before doing work... 😏 

I don't mind withdrawing this. Just say the word. (Or just close it.)

FWIW, I amended the commit to address @thomaslevesque's review comment.

blairconrad:
@adamralph, no worries!

I'm content for you to continue. I had 2 substantive differences in my sandbox, though. First, I was creating the test targets from `TestSuites`:

```c#
foreach (var testSuite in TestSuites)
{
    Target(
        testSuite.Key,
        DependsOn("build", "testsDirectory"),
        forEach: testSuite.Value,
        action: testDirectory => RunTestsInDirectory(testDirectory));
}
```

We have general programming language available to us. Why not use it?
Of course this is very much a "bonus change".

The other I'll add to an actual review response.

adamralph:
I've added a commit to generate test targets from `TestSuites`.

blairconrad:
Thanks, @adamralph! I like it. I'll just merge before @thomaslevesque  can even see the new addition!

adamralph:
BTW - I can't remember if we ever considered this when I was an owner, but it may be worth switching on this in the branch protection settings: 

image

I'm quite surprised it isn't switch on by default. If further changes are made after an approval then it follows that the current state of the PR has yet to be approved.

blairconrad:
I think this is a good idea. Thanks for mentioning it, @adamralph.

PR #1422, opened 2018-09-02
upgrade to SDK 2.1.202

adamralph:
I'm endlessly frustrated when working with FIE locally because I don't have SDK 2.0.2. That means I keep having to temporarily delete or hack `global.json` before anything will build for me. Worse, is that 2.0.2 isn't even listed for download under https://www.microsoft.com/net/download/dotnet-core/2.0

I'm not sure if you already want to move to an SDK which is paired with the 2.1 runtime, so I went for 2.1.202, which is the latest SDK paired with the 2.0 runtime (and I have it installed 😉).


thomaslevesque:
@adamralph the build failed... Looks like the problem is related to the objcs/objvb folders in FakeItEasy.Analyzer.

>  Generating MSBuild file C:\projects\fakeiteasy\src\FakeItEasy.Analyzer\**obj**\FakeItEasy.Analyzer.CSharp.csproj.nuget.g.props.
>  Generating MSBuild file C:\projects\fakeiteasy\src\FakeItEasy.Analyzer\**obj**\FakeItEasy.Analyzer.CSharp.csproj.nuget.g.targets.
> 
> ...
> ... Assets file 'C:\projects\fakeiteasy\src\FakeItEasy.Analyzer\**objvb**\project.assets.json'

(probably related to the MSB3539 warning earlier)

I'm trying to get rid of those folders in #1419

thomaslevesque:
Also, I was unable to build the analyzer projects with the latest SDK (2.1.400), but on my #1419 branch, it works fine. So we'll probably be able to upgrade the SDK after it's merged.

adamralph:
@thomaslevesque yep, and when I run locally (with SDK 2.1.201), I get yet a different error, which is weird:

image

I've tried with SDK 2.1.2 (the oldest on my machine) and 2.1.401 (the latest available) and I get the same error.

thomaslevesque:
> Worse, is that 2.0.2 isn't even listed for download under https://www.microsoft.com/net/download/dotnet-core/2.0

Yes, that's weird... I use 2.0.3 instead, but there's no installer, just a zip (even though there's an installer for macOS...)

adamralph:
Yeah, it looks like the oldest Windows installer available is for SDK 2.1.4.

thomaslevesque:
The problem is that the `BaseIntermediateOutputPath` is modified after it has already been used for restore. Restore uses the original value (`obj`), but build uses the modified value (`objcs`/`objvb`). The problem will go away when #1419 is merged.

adamralph:
FYI I "installed" SDK 2.0.3 locally by dumping the appropriate folder from the zip into `C:\Program Files\dotnet\sdk` but I still get the same error as in the screenshot above.

thomaslevesque:
> FYI I "installed" SDK 2.0.3 locally by dumping the appropriate folder from the zip into C:\Program Files\dotnet\sdk but I still get the same error as in the screenshot above.

I did the same, and it's working for me.

adamralph:
@thomaslevesque it worked for me after `git clean -xdf`. :smirk:

BTW - that reminds me (/cc @blairconrad), I would consider getting rid of the `clean` build target. An msbuild clean only does so much, and clearly it wasn't doing enough here. `git clean -xdf` (nuke it from orbit) is the only way to be sure. FWIW, I no longer define a `clean` target in any of my repos.

thomaslevesque:
> I would consider getting rid of the clean build target. An msbuild clean only does so much, and clearly it wasn't doing enough here. git clean -xdf (nuke it from orbit) is the only way to be sure. FWIW, I no longer define a clean target in any of my repos.

👍 
It's not doing anything useful, it just makes each build take longer because everything has to be rebuilt...
In fact, even the `restore` target is probably unnecessary, since build automatically restores packages.

blairconrad:
> I'm not sure if you already want to move to an SDK which is paired with the 2.1 runtime

I've no objection. Is there a reason not to try to move to 2.1.401?

adamralph:
I think it's worth a try. Appveyor has it, and anyone else can install it easily enough. It may be even worth unpinning the SDK completely, to get early warning of incompatibility with new versions as they are released. It shouldn't happen too often.

blairconrad:
Unpinning is worth a try, but for sure we should make sure [how to build](https://github.com/FakeItEasy/FakeItEasy/blob/develop/how_to_build.md) spells out the proper requirements. Which it should anyhow, but I think we've been lazy. At least the `global.json` gave a hint of what might be going wrong if contributors had build problems.

adamralph:
In any case, it seems either changing or removing `global.json` is dependent on https://github.com/FakeItEasy/FakeItEasy/pull/1419. This PR is a step in the right direction, but feel free to amend/close it as and when required.

adamralph:
I rebased to force a rebuild, and it looks like this is good to go!

thomaslevesque:
👍 for unpinning

adamralph:
This PR, as is stands, is an improvement. But up to you if you want to close and unpin instead.

thomaslevesque:
Superseded by #1432. Thanks for raising this, @adamralph!

PR #1431, opened 2018-09-04
Add FakeItEasy.Build and FakeItEasy.PrepareRelease projects to solution

thomaslevesque:
So that we can edit them in Visual Studio and VSCode (technically, it was already possible to edit them in VSCode, but OmniSharp doesn't take the projects into account if they're not part of the solution, so no intellisense or analysis)

I disabled build for these projects in the solution configuration so that running `FakeItEasy.Build` doesn't attempt to rebuild `FakeItEasy.Build` 😉


blairconrad:

PR #1432, opened 2018-09-04
Unpin the .NET Core SDK version

blairconrad:
Supersedes #1422


how_to_build.md
```diff
@@ -9,17 +9,22 @@ At the time of writing the build is only confirmed to work on Windows using the
...
 
-1. Ensure you have version 15.3 or later of either Visual Studio 2017 or MSBuild installed.
+Ensure that the following are installed:
 
-1. Ensure you have .NET Core SDK 2.0.2 or later installed
+1. .NET framework 4.6.1 or later
+
+1. a very recent version of Visual Studio 2017 or MSBuild (currently this means 15.7 or later)
+
+1. a very recent version of the .NET Core SDK (currently this means 2.1.401 or later)
```
thomaslevesque: I don't think we need a version this high. It works fine with 2.1.4 for instance.

thomaslevesque: Also, it might seem strange, but I think we also need .NET 3.5 installed. For some reason, ILMerge doesn't work without it.

how_to_build.md
```diff
@@ -9,17 +9,22 @@ At the time of writing the build is only confirmed to work on Windows using the
...
 
-1. Ensure you have .NET framework 4.6.1 or later installed.
+The build requires that a few pieces of software be installed on the host computer. We're somewhat aggressive about adoptiong new language features and the like, so rather than specifying exactly which versions are required, we'll tend toward
+"latest" or "at least" forms of guidance. If it seems you have an incompatible version of the software, prefer to upgrade rather than downgrade.
 
-1. Ensure you have version 15.3 or later of either Visual Studio 2017 or MSBuild installed.
+Ensure that the following are installed:
 
-1. Ensure you have .NET Core SDK 2.0.2 or later installed
+1. .NET framework 4.6.1 or later
```
thomaslevesque: I think we also need the 4.6.1 targeting pack, not just the runtime. (but it's installed with VS anyway, so maybe not worth mentioning)

blairconrad: Is it installed with MSBuild?

thomaslevesque: No idea. I don't even know how to install MSBuild without installing VS...

blairconrad: Build from source? ;)

how_to_build.md
blairconrad: You've reproduced? I heard that was the case for < 2.14.1208, but that 2.1.4.1208 didn't require 3.5.

how_to_build.md
thomaslevesque: Actually, there's that thing called "Build tools for Visual Studio", intended for build servers which don't need the IDE itself. I think it includes the targeting packs.
Anyway, it's probably not an issue. Forget I ever mentioned it :wink:

how_to_build.md
thomaslevesque: Not recently, but every time I installed or reinstalled a machine, I was unable to build FakeItEasy without installing .NET 3.5 installed.

blairconrad: Wow. I'd no idea. I downloaded one of MS's fancy developer evaluation OVAs yesterday and tried to build and failed miserably. Thanks for the note!

blairconrad:
Thanks, @thomaslevesque. I amended the commit. Only the ordered list in `how_to _build.md` changed

thomaslevesque:
Thanks @blairconrad!

blairconrad:

PR #1436, opened 2018-09-05
Test under netcoreapp2.1

### LINKED ISSUE #1361: Run tests under .NET Core 2.1
We currently run tests under .NET Framework 4.6.1 and .NET Core 1.0.
Sometimes newer platforms introduce changes in behaviour, as noted in #1345. Even though that one bug was resolved, we should test under the newer .NET Core to ensure that everything else works (and continues to work).

blairconrad:
Fixes #1361

### DIFF
--- a/tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
+++ b/tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
@@ -1,9 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


blairconrad:
@thomaslevesque, I added the framework, rather than replacing netcoreapp1.0, since I think there's still value in testing against netcoreapp1.0, but I'm happy to hear dissenting opinions, as I know this will increase build times.

thomaslevesque:
> @thomaslevesque, I added the framework, rather than replacing netcoreapp1.0, since I think there's still value in testing against netcoreapp1.0, but I'm happy to hear dissenting opinions, as I know this will increase build times.

I think we should keep netcoreapp1.0.

thomaslevesque:
Thanks @blairconrad

blairconrad:
Thank you!

PR #1439, opened 2018-09-07
Optimize dummy creation

### LINKED ISSUE #1354: Optimize creation of string and value-type Dummies
As discussed in https://github.com/FakeItEasy/FakeItEasy/issues/1348#issuecomment-388546371

(@blairconrad is the first speaker, @thomaslevesque the second)
> > But what if somebody has made a Dummy factory for Char[]? Then all of a sudden we've changed the string Dummies out from under them. 😄

> I don't remember the priorities when creating a dummy... Don't we use user-provided dummy factories over built-in resolution strategies?

> > What if something else uses a ReadOnlySpan?

> Maybe we could have a special case for this too, e.g. create an empty `ReadOnlySpan`. We'd have to do it via reflection, though, since we don't want to take a dependency on `System.Memory`

Ties to #1345.

blairconrad:
Fixes #1354.

### DIFF
--- a/src/FakeItEasy/Creation/DummyValueResolver.cs
+++ b/src/FakeItEasy/Creation/DummyValueResolver.cs
@@ -28,8 +28,8 @@ namespace FakeItEasy.Creation
                     new ResolveFromDummyFactoryStrategy(dummyFactory),
                     new ResolveByCreatingTaskStrategy(this),
                     new ResolveByCreatingLazyStrategy(this),
-                    new ResolveByCreatingFakeStrategy(fakeObjectCreator, this),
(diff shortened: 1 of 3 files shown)


blairconrad:
Sadly, the new specs pass even before I make any changes, and I could not think of a way to verify that we create a value type without trying to fake it.
Am happy to remove useless spec and/or add some other test if you like.

blairconrad:
> If you want to leave it as is, I'm happy to merge.

Yeah. I think every once in a while we have to trust that the code is doing the right thing when it appears to be doing the right thing. Thanks.

thomaslevesque:
Thanks @blairconrad!

PR #1440, opened 2018-09-09
Fix CA1307 warning

thomaslevesque:
Introduced when `netcoreapp2.1` targets were added to tests.

### DIFF
--- a/tests/FakeItEasy.IntegrationTests/StringArgumentConstraints.cs
+++ b/tests/FakeItEasy.IntegrationTests/StringArgumentConstraints.cs
@@ -11,7 +11,11 @@ namespace FakeItEasy.IntegrationTests
 
         public static string Contains(this IArgumentConstraintManager<string> scope, string value)
         {
+#if FEATURE_STRING_CONTAINS_COMPARISONTYPE
+            return scope.Matches(x => x.Contains(value, StringComparison.CurrentCulture), $@"Contains ""{value}""");
+#else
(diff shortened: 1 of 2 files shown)


tests/FakeItEasy.IntegrationTests/StringArgumentConstraints.cs
```diff
@@ -11,7 +11,11 @@ public static string StartsWith(this IArgumentConstraintManager<string> scope, s
 
         public static string Contains(this IArgumentConstraintManager<string> scope, string value)
         {
+#if FEATURE_STRING_CONTAINS_COMPARISONTYPE
+            return scope.Matches(x => x.Contains(value, StringComparison.CurrentCulture), $@"Contains ""{value}""");
```
thomaslevesque: `String.Contains` now has multiple overloads. Not specifying the comparison type means current culture is used implicitly, which could be an issue and triggers CA1307. I specified current culture explicitly to keep the same behavior.

thomaslevesque:
@blairconrad the warning was in tests, so I could just suppress it if you prefer.

blairconrad:
Thanks, @thomaslevesque. I noticed this on Friday and hadn't yet gotten back to it.

thomaslevesque:
Wow, you're fast @blairconrad!

thomaslevesque:
Thanks @blairconrad!

blairconrad:
> Wow, you're fast @blairconrad!

Luck. I just happened to see the e-mail for the PR come in.

blairconrad:

PR #1441, opened 2018-09-09
Drop MSBuild, use dotnet build instead

thomaslevesque:
As discussed with @adamralph, not sure where/when it was.
We no longer need to call MSBuild explicitly, so we don't need VSWhere anymore.

### DIFF
--- a/how_to_build.md
+++ b/how_to_build.md
@@ -16,9 +16,7 @@ Ensure that the following are installed:
 
 1. .NET 3.5, required by ILMerge
 
-1. .NET framework 4.6.1 or later
-
-1. a very recent version of Visual Studio 2017 or MSBuild (currently this means 15.7 or later)
+1. a very recent version of Visual Studio 2017 (currently this means 15.7 or later) or the Build Tools for Visual Studio 2017
 
 1. a very recent version of the .NET Core SDK (currently this means 2.1.3 or later)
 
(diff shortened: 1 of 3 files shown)


blairconrad:
It's beautiful, @thomaslevesque. Thanks.

Confession time: I didn't completely follow when last we talked about the build prereqs. It _seems_ that we now don't need VS or MSBuild, but you were concerned about some other extensions or plugins or something that I don't know what they are. Are you happy with `how_to_build.md` as it is?

thomaslevesque:
> It seems that we now don't need VS or MSBuild

I think we don't. But I'd like to check, first. I'll create a clean VM in Azure to make sure.

> but you were concerned about some other extensions or plugins or something that I don't know what they are.

I don't remember what it might be.

> Are you happy with how_to_build.md as it is?

No, I should update it. We might want to state clearly that it only works on Windows (because of ILMerge). It was implied by the VS requirement, but if that requirement is gone we should be more explicit. Currently it's a little understated:

> At the time of writing the build is only confirmed to work on Windows using the Microsoft .NET framework.

how_to_build.md
```diff
@@ -16,9 +16,7 @@ Ensure that the following are installed:
 
 1. .NET 3.5, required by ILMerge
 
-1. .NET framework 4.6.1 or later
```
thomaslevesque: I think this is implied by VS2017 (you can't have VS2017 without having net461 or later)

blairconrad:
I do love me some deleted lines. Thanks, @thomaslevesque!

thomaslevesque:
Thanks for the merge!

blairconrad:

PR #1442, opened 2018-09-10
split test inputs by framework

adamralph:
For some reason (yet to be discovered), I get repeatable test failures when building locally. Looking at the output, I found it very difficult to tell exactly where it's happening. With this change, I can see it's when running under `netcoreapp1.0`:

image

(As soon as I know why it's failing, I'll report the problem in another issue/PR.)

PR #1445, opened 2018-09-11
Upgrade xunit to 2.4.0, switch to dotnet test

### LINKED ISSUE #1444: Update xunit, drop dotnet xunit in favor of dotnet test
As discussed in https://github.com/FakeItEasy/FakeItEasy/pull/1442#issuecomment-420230719

@adamralph said:

> BTW - you may want to think about switching to `dotnet test` soon, since `dotnet xunit` is no longer a thing.

Also, we're using a beta version of xunit 2.3.0, but xunit 2.4.0 RTM has been released and seems to work fine. We should consider upgrading. `dotnet xunit` is no longer supported in 2.4.0, which means we have to drop it anyway if we upgrade. This might fix an issue where the name of the failed test is not clearly visible in the build output.

thomaslevesque:
Fixes #1444

### DIFF
--- a/tests/Directory.Build.props
+++ b/tests/Directory.Build.props
@@ -14,14 +14,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="xunit.core" Version="2.3.0-beta2-build3683" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.0-beta2-build1317" />
-    <PackageReference Include="Microsoft.TestPlatform.TestHost" Version="15.0.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
+    <PackageReference Include="xunit.core" Version="2.4.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageReference Include="XunitXml.TestLogger" Version="2.0.0" />
     <PackageReference Include="FluentAssertions" Version="4.19.2" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <DotNetCliToolReference Include="dotnet-xunit" Version="2.3.0-beta2-build3683" />
   </ItemGroup>
 
 </Project>
--- a/tools/FakeItEasy.Build/Program.cs
+++ b/tools/FakeItEasy.Build/Program.cs
@@ -1,6 +1,5 @@
 ﻿namespace FakeItEasy.Build
 {
-    using System;
     using System.Collections.Generic;
     using System.IO;
     using static Bullseye.Targets;
@@ -8,7 +7,6 @@
 
     public class Program
     {
-        private const string TestsDirectory = "artifacts/tests";
         private const string LogsDirectory = "artifacts/logs";
         private const string Solution = "FakeItEasy.sln";
         private const string VersionInfoFile = "src/VersionInfo.cs";
@@ -66,8 +64,6 @@
 
             Target("logsDirectory", () => Directory.CreateDirectory(LogsDirectory));
 
-            Target("testsDirectory", () => Directory.CreateDirectory(TestsDirectory));
-
             Target(
                 "build",
                 DependsOn("versionInfoFile"),
@@ -79,7 +75,7 @@
             {
                 Target(
                     testSuite.Key,
-                    DependsOn("build", "testsDirectory"),
+                    DependsOn("build"),
                     forEach: testSuite.Value,
                     action: testDirectory => RunTests(testDirectory));
             }
@@ -110,10 +106,7 @@
             RunTargets(args);
         }
 
-        private static void RunTests(string testDirectory)
-        {
-            var xml = Path.GetFullPath(Path.Combine(TestsDirectory, Path.GetFileName(testDirectory) + ".TestResults.xml"));
-            Run("dotnet", $"xunit -configuration Release -nologo -nobuild -noautoreporters -notrait \"explicit=yes\" -xml {xml}", testDirectory);
-        }
+        private static void RunTests(string testDirectory) =>
+            Run("dotnet", "test --configuration Release --no-build -- RunConfiguration.NoAutoReporters=true", testDirectory);
     }
 }


tests/Directory.Build.props
```diff
@@ -14,14 +14,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="xunit.core" Version="2.3.0-beta2-build3683" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.0-beta2-build1317" />
-    <PackageReference Include="Microsoft.TestPlatform.TestHost" Version="15.0.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
```
thomaslevesque: I think we should have been using this instead of `Microsoft.TestPlatform.TestHost`. Anyway, xUnit's "getting started" section uses this package.

tests/Directory.Build.props
```diff
@@ -14,14 +14,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="xunit.core" Version="2.3.0-beta2-build3683" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.0-beta2-build1317" />
-    <PackageReference Include="Microsoft.TestPlatform.TestHost" Version="15.0.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
+    <PackageReference Include="xunit.core" Version="2.4.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageReference Include="XunitXml.TestLogger" Version="2.0.0" />
```
thomaslevesque: There's no built-in XML logger. This is the one recommended on [this page](https://github.com/Microsoft/vstest-docs/blob/master/docs/report.md#available-test-loggers). However I see 2 issues with it:
- it's not an official implementation (https://github.com/spekt/xunit.testlogger)
- it doesn't take the framework into account, so there's only one result file, which seems to contain results only for the last framework run.

Alternative options:
- Do we really need those files? Me, I never looked at them. Maybe we could just disable reporting.
- Do we really need them to be in the xUnit format? We could use the built-in trx logger

tools/FakeItEasy.Build/Program.cs
```diff
@@ -8,7 +8,7 @@
 
     public class Program
     {
-        private const string TestsDirectory = "artifacts/tests";
```
thomaslevesque: I didn't really like that name, it was misleading

tools/FakeItEasy.Build/Program.cs
```diff
@@ -112,8 +112,8 @@ public static void Main(string[] args)
 
         private static void RunTests(string testDirectory)
         {
-            var xml = Path.GetFullPath(Path.Combine(TestsDirectory, Path.GetFileName(testDirectory) + ".TestResults.xml"));
-            Run("dotnet", $"xunit -configuration Release -nologo -nobuild -noautoreporters -notrait \"explicit=yes\" -xml {xml}", testDirectory);
+            var xml = Path.GetFullPath(Path.Combine(TestResultsDirectory, Path.GetFileName(testDirectory) + ".TestResults.xml"));
+            Run("dotnet", $"test -c Release --no-build --logger:\"xunit;LogFilePath={xml}\"", testDirectory);
```
thomaslevesque: There are many `dotnet xunit` options that don't have equivalent in `dotnet test`
- `-nologo`: I spent a bit of time trying to find it, but apparently there's no way to suppress the tool output header. I also didn't succeed in reducing the verbosity, even with `--verbosity quiet` (the default seems to be `minimal`, which produces the same output `quiet`)
- `-noautoreporters`: no equivalent, as far as I can tell. The tests are reported to AppVeyor again, and I don't know how to disable it.
- `-notrait`: there is a `--filter` option, but `--filter "explicit!=yes` doesn't work, even though the [docs mention it's possible](https://github.com/Microsoft/vstest-docs/blob/master/docs/filter.md). We're not using this at the moment anyway.

blairconrad:
You probably noticed, but we're only getting XML output for the latest-run framework.

tools/FakeItEasy.Build/Program.cs
blairconrad: For the first problem, I'd found https://github.com/Microsoft/vstest/issues/1701.
Ah, I forgot to check the auto reporters. Good find.

thomaslevesque: > For the first problem, I'd found Microsoft/vstest#1701.

Yes, but it doesn't provide a solution...

tests/Directory.Build.props
thomaslevesque: > it doesn't take the framework into account, so there's only one result file, which seems to contain results only for the last framework run.

The trx logger has the same problem. So if we want per-framework result files, we have to run the tests per framework, as @adamralph did in #1442...

blairconrad: It's true. On the other hand, I'm not against submitting an issue/PR to spekt/xunit.testlogger.

tools/FakeItEasy.Build/Program.cs
blairconrad: No, no. But it's nice to know that someone's aware of it.

tests/Directory.Build.props
blairconrad: It's true. And the project is an absolute mess.

blairconrad: I mostly enjoy the results files when the console output is too hard to read. Seems like we're making that easier, so I'm game to try without them.

blairconrad:
@thomaslevesque, I generally like the way this is going (except for `-configuration` → `-c`, but I'm picking nits now).

I agree we could drop the logger.

For me the only remaining issue of concern would be the inability to disable autoreporters. That was a real pain point for the AppVeyor builds.

adamralph:
> For me the only remaining issue of concern would be the inability to disable autoreporters. That was a real pain point for the AppVeyor builds.

Out of interest, what is this about? I don't do anything special around "autoreporters" in other projects and the Appveyor builds are fine.

thomaslevesque:
> (except for -configuration → -c, but I'm picking nits now).

Well, I tend to prefer the short form, but I can change it to `--configuration` if you prefer.

> Out of interest, what is this about? I don't do anything special around "autoreporters" in other projects and the Appveyor builds are fine.

Tests are reported to AppVeyor and appear in the "Tests" tab of the build. This looks like a good idea, but it has several issues:
- Test result collection by AppVeyor takes a very significant time; disabling this made the build much faster (I don't remember the exact numbers, but I think it was about twice as fast)
- We have a total of almost 20000 reported tests. Just loading them in the page takes a long time and tends to freeze the browser (at least when using Firefox)
- It shows very little useful information: test name, status (failed/passed), and duration.
- The tested are listed in seemingly random order

blairconrad:
> Well, I tend to prefer the short form, but I can change it to `--configuration` if you prefer.

I usually like the explicit version in scripts, but I'm not too bent out of shape. Vote with your :heart:.

blairconrad:
@adamralph, @thomaslevesque, for interest, I checked the [latest build](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/659-xrrbtgqv) for this PR, and the autotest reporting added at least 2:12 to the build time (measured from time of last line of the console output to the reported build time near the top of the page).

tools/FakeItEasy.Build/Program.cs
```diff
@@ -110,10 +106,7 @@ public static void Main(string[] args)
             RunTargets(args);
         }
 
-        private static void RunTests(string testDirectory)
-        {
-            var xml = Path.GetFullPath(Path.Combine(TestResultsDirectory, Path.GetFileName(testDirectory) + ".TestResults.xml"));
-            Run("dotnet", $"test -c Release --no-build --logger:\"xunit;LogFilePath={xml}\"", testDirectory);
-        }
+        private static void RunTests(string testDirectory) =>
+            Run("dotnet", $"test --configuration Release --no-build -- RunConfiguration.NoAutoReporters=true", testDirectory);
```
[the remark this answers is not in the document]
thomaslevesque: > The `$` is bonus now.

Oops. I'll remove it. Usually R# tells me about these things, but I did everything in VS Code...

blairconrad:
@thomaslevesque, looks great. Is there a reason we're still \[WIP\]?

thomaslevesque:
No, I just forgot to remove it.
I'm on my phone now, so I can't do it easily. Feel free to rename!

Le mar. 11 sept. 2018 à 17:48, Blair Conrad <notifications@github.com> a
écrit :

adamralph:
> * Test result collection by AppVeyor takes a very significant time; disabling this made the build much faster (I don't remember the exact numbers, but I think it was about twice as fast)
> 
> * We have a total of almost 20000 reported tests. Just loading them in the page takes a long time and tends to freeze the browser (at least when using Firefox)

> @adamralph, @thomaslevesque, for interest, I checked the [latest build](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/build/659-xrrbtgqv) for this PR, and the autotest reporting added at least 2:12 to the build time (measured from time of last line of the console output to the reported build time near the top of the page).

Ouch! I can understand the reason for suppressing it! 
 
> * The tested are listed in seemingly random order

Yes, this is very frustrating. I've been chiming in on https://github.com/appveyor/ci/issues/239 about that.

blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
Thanks for the merge!

Le mar. 11 sept. 2018 à 19:18, Blair Conrad <notifications@github.com> a
écrit :

PR #1446, opened 2018-09-11
remove xunit console

adamralph:
So this is actually the thing I was trying to fix which, when checking it didn't break anything, led me to raise https://github.com/FakeItEasy/FakeItEasy/pull/1442.

I guess this package reference became redundant some time ago.


blairconrad:

Thanks, @adamralph. Look for you name in the release notes! 🎋

PR #1448, opened 2018-09-12
Use Monitor to ensure tests don't run concurrently

### LINKED ISSUE #1443: Intermittent failure for NonGenericCreationSpecs.CacheSuccessfulConstructor
I've seen this several times in the past few days:

> Failed! The process exited with code 1: FakeItEasy.Specs.NonGenericCreationSpecs.CacheSuccessfulConstructor() [09] And the parameterless constructor was only called once [FAIL]

So I guess the concurrency issues are not completely fixed yet...

blairconrad:
Fixes #1443.

### DIFF
--- a/tests/FakeItEasy.Specs/CreationSpecs.cs
+++ b/tests/FakeItEasy.Specs/CreationSpecs.cs
@@ -169,8 +169,8 @@ namespace FakeItEasy.Specs
             // unsuccessful constructors may be called more than once, so serialize fake
             // creation for this test.
             "And nobody else is trying to fake the class right now"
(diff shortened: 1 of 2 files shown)


blairconrad:
I really don't understand why the original fix didn't work, but I made some local changes to encourage failure and was getting about 1 failure in 3 runs of the specs. Then I changed to Monitor at @thomaslevesque's behest and ran 300 times in a row without failure.
Other tests (not using FIE code) with the `ManualResetEventSlim` and `Monitor` showed it to be more effective as well.

blairconrad:
Thanks, @thomaslevesque

PR #1467, opened 2018-10-03
Cache FakeOption constructor

### LINKED ISSUE #1466: Cache FakeOptions constructor to improve performance
For a lark, I profiled creating 100,000 instances of a small interface (after warming up FakeItEasy by faking a different interface). Using dotTrace, I found that `DefaultFakeAndDummyManager.BuildProxyOptions` was a hotspot. So I started caching the constructor by type of Fake and found this difference:

image

`BuildProxyOptions`  drops from 19.10% of the total time to 9.95%. Or if you prefer the diff view:

image

Of course we're trading away some storage space. If we end up faking many different types once, this is a bad tradeoff, but if types are faked more than once, it may be a benefit.

Also, in many cases we don't even _need_ a `FakeOptions` object. I can explore dropping the construction step as well.

blairconrad:
Fixes #1466.

### DIFF
--- a/src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
+++ b/src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
@@ -1,9 +1,8 @@
 namespace FakeItEasy.Creation
 {
     using System;
-#if FEATURE_NETCORE_REFLECTION
+    using System.Collections.Concurrent;
     using System.Reflection;
-#endif
     using FakeItEasy.Core;
 
     /// <summary>
@@ -12,6 +11,8 @@ namespace FakeItEasy.Creation
     internal class DefaultFakeAndDummyManager
         : IFakeAndDummyManager
     {
+        private static readonly ConcurrentDictionary<Type, Func<ProxyOptions, IFakeOptions>> FakeOptionsFactoryCache = new ConcurrentDictionary<Type, Func<ProxyOptions, IFakeOptions>>();
+
         private readonly IFakeObjectCreator fakeCreator;
         private readonly DynamicOptionsBuilder dynamicOptionsBuilder;
         private readonly IDummyValueResolver dummyValueResolver;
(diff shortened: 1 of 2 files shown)


src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
```diff
@@ -48,19 +52,18 @@ public bool TryCreateDummy(Type typeOfDummy, out object result)
...
-
-            return (IFakeOptions)optionsConstructor.Invoke(new object[] { proxyOptions });
         }
 
         private IProxyOptions BuildProxyOptions(Type typeOfFake, Action<IFakeOptions> optionsBuilder)
         {
             var proxyOptions = new ProxyOptions();
-            var options = CreateFakeOptions(typeOfFake, proxyOptions);
+            var constructor = FakeOptionsConstructorCache.GetOrAdd(typeOfFake, GetFakeOptionsConstructor);
+            var options = (IFakeOptions) constructor.Invoke(new object[] { proxyOptions });
```
thomaslevesque: I think we can do better than that. Instead of caching and invoking the constructor, create a delegate, cache it and invoke it. It would probably be faster.

thomaslevesque: Just did a quick test, the delegate is about 8 times faster. And it doesn't require an array allocation.

src/FakeItEasy/Creation/DefaultFakeAndDummyManager.cs
```diff
@@ -48,19 +49,20 @@ public bool TryCreateDummy(Type typeOfDummy, out object result)
...
+
+        private static Func<ProxyOptions, IFakeOptions> GetFakeOptionsFactory(Type typeOfFake)
         {
-            var optionsConstructor = typeof(FakeOptions<>)
-                .MakeGenericType(typeOfFake)
-                .GetConstructor(new[] { typeof(ProxyOptions) });
+            var method = typeof(DefaultFakeAndDummyManager).GetMethod(nameof(CreateFakeOptions), BindingFlags.NonPublic | BindingFlags.Static).MakeGenericMethod(typeOfFake);
 
-            return (IFakeOptions)optionsConstructor.Invoke(new object[] { proxyOptions });
+            return (Func<ProxyOptions, IFakeOptions>)method.CreateDelegate(typeof(Func<ProxyOptions, IFakeOptions>));
```
thomaslevesque: Nice! It's annoying that you can't create a delegate directly from a constructor... In my test I had created the delegate from a Linq expression, but your approach is more efficient.

blairconrad: Oh. I had absolutely no idea what I was doing. I figured I'd done the worst possible thing and you'd fix it for me!

blairconrad:
> I'll let you look into avoiding the creation completely if you still want to

Yeah, it's a little more complicated than I initially thought. I may pursue on the back burner, but as you say, diminishing returns at this point. I see no reason to hold up this PR for that work.

thomaslevesque:
@blairconrad do you want to squash before I merge? I see no reason to keep the initial commit.

blairconrad:
Bother! I forget. I absolutely do. Sorry/thanks.

blairconrad:
Squashed.

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you. For the merge and the dramatic improvement in the code!

PR #1469, opened 2018-10-06
Create ProxyOptions only when required

blairconrad:
As discussed earlier. There may be a small improvement in the time required to create fakes. Certainly we'll make fewer objects in what I believe to be the most common case - where there's no fake options builder used to create a fake.
I profiled:

image

Tell me what you think.

I made two commits because I didn't want the rename to confuse the diffs.


blairconrad:
Thank you, @thomaslevesque. Squashed.

thomaslevesque:

PR #1471, opened 2018-10-07
switch to Bullseye 2.1.0 beta 1

adamralph:
Looking for feedback on this version. The main change is that a specific colour palette is chosen depending on which CI system is detected. Currently Appveyor, Travis CI, and TeamCity are supported.

For reference, this is how the output looks in a local console:

And these are the console colours available on Appveyor:

### DIFF
--- a/tools/FakeItEasy.Build/FakeItEasy.Build.csproj
+++ b/tools/FakeItEasy.Build/FakeItEasy.Build.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Bullseye" Version="2.0.0-rc.3" />
+    <PackageReference Include="Bullseye" Version="2.1.0-beta.1" />
     <PackageReference Include="SimpleExec" Version="$(SimpleExecVersion)" />
     <PackageReference Include="GitVersion.CommandLine" Version="4.0.0-beta0012" ToolName="GitVersion" ToolExe="GitVersion.exe" />
     <PackageReference Include="PdbGit" Version="3.0.41" ToolName="PdbGit" ToolExe="PdbGit.exe" />


thomaslevesque:
I'm colorblind, so it doesn't really make a big difference to me... @blairconrad might have a more useful opinion :wink:

adamralph:
@thomaslevesque your feedback could still be very useful. E.g. do you find there is enough contrast between the background and the "Bullseye/{target}" labels in the "After" snippet for you to read them? And how does that compare with the "Before" snippet?

I'm not wild about the colour choice for those labels, but the Appveyor palette didn't give me any better option, other than sticking with the old choice, and losing the distinction between the target names and the input values.

thomaslevesque:
> E.g. do you find there is enough contrast between the background and the "Bullseye/{target}" labels in the "After" snippet for you to read them? And how does that compare with the "Before" snippet?

There's enough contrast for me to be able to read everything, both before and after. But a closer look made me realize that some things appeared to be the same color before (e.g. targets and their "sub-targets") and are now different. So I guess it's a change for the better.

blairconrad:
> > I'm not wild about the colour choice for those labels, but the Appveyor palette didn't give me any better option, other than sticking with the old choice, and losing the distinction between the target names and the input values.

I agree. Well, mostly I don't like how the blue background limits your options for attractive, high-contrast labels.
I had never considered that there was a need for a distinction between the target names and the input values, but I can see the value.

> There's enough contrast for me to be able to read everything, both before and after. But a closer look made me realize that some things appeared to be the same color before (e.g. targets and their "sub-targets") and are now different. So I guess it's a change for the better.

Yup. Me too.

adamralph:
👍 thanks for the feedback.

thomaslevesque:
Thanks @adamralph!

Issue #1470, opened 2018-10-06
Improve performance when configuring fake method


blairconrad:
`AssemblyExtesions.Name` [creates a new `AssemblyName`](https://github.com/FakeItEasy/FakeItEasy/blob/develop/src/FakeItEasy/Core/AssemblyExtensions.cs#L27) when I don't think it has to.
If I stop:

![image](https://user-images.githubusercontent.com/3275797/46570560-f96d3900-c933-11e8-86dd-9916f7dc6cd8.png)

And garbage collection time drops from 16% to 6%.

Or if you like the diff format:

![image](https://user-images.githubusercontent.com/3275797/46570592-7698ae00-c934-11e8-8863-9c8fe4e7bda0.png)

thomaslevesque:
Wow. I didn't expect the `AssemblyName` constructor to be so expensive.

thomaslevesque:
Ah, I see. `ProxyUtil.IsProxyType`!

blairconrad:
I marked #1472 as connecting to this issue in case we want to pile on a few more PRs. If you'd rather, I don't mind closing this issue with the one PR and opening more issues (or just sending raw PRs) for other performance work in the area.

thomaslevesque:
No, I think it's OK to keep this one open for now.

blairconrad:
This could be a never-ending process, but we've had good results. I think I'll focus my efforts somewhere else for a while.
Closed in #1472, #1476, #1483, #1485, and #1486.

thomaslevesque:
This change has been released in FakeItEasy [5.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/5.0.0).

PR #1476, opened 2018-10-19
Stop using Linq in GetOutAndRefParametersValueProducer

blairconrad:
Connects to #1470.
Most of the time, we have no out or ref parameters to fetch values from, but we're LINQing and creating a list all the time. Even when we do have out and ref parameters, it's not the cheapest way to construct the list.


blairconrad:
When configuring a method with no parameters:

image

When configuring a method with a single parameter, an `out int`:

image

thomaslevesque:

Issue #1478, opened 2018-10-22
Simplify/replace/remove ServiceLocator


blairconrad:
As @thomaslevesque and I noted in private chat, the `ServiceLocator` we're using provides little (if any) benefit, complicates debugging and _may_ decrease performance (it's hard to tell, mostly because the ServiceLocator injects itself into so many stack traces in weird places that it's hard to tell). Let's simplify it a bit.

blairconrad:
@thomaslevesque, I'm happy with the changes we've made to the `ServiceLocator` so far, and could switch to other tasks.

Do you want us to continue simplifying and/or removing, or shall we leave it as is for now?

blairconrad:
Closed by #1479 and #1481.

thomaslevesque:
This change has been released in FakeItEasy [5.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/5.0.0).

PR #1479, opened 2018-10-22
Only register singletons in the service locator

blairconrad:
… and resolve them eagerly.

Connects to #1478.


tests/FakeItEasy.Tests/FakeTests.cs
```diff
@@ -1,9 +1,71 @@
...
+            // Arrange
+
+            // Act
+
+            // Assert
+#pragma warning disable CS0618 // Type or member is obsolete
+                Expression<Action> call = () => Fake.InitializeFixture(new object());
+#pragma warning restore CS0618 // Type or member is obsolete
+                call.Should().BeNullGuarded();
+        }
```
blairconrad: All the new tests here were sucked over from the `FakeFacadeTests`. Some went to the specs.

blairconrad:
There are other potential improvements as well. If nothing else, we register a lot of services only so we can find them inside the *Modules, and they have no external consumers. While I was chasing that path, I had difficulties with some of the differences between the singletons and the transient "services", and thought this change would be relatively self-contained.

src/FakeItEasy/FakeFacade.cs
```diff
@@ -1,65 +0,0 @@
...
-#pragma warning restore CA1822 // Mark members as static
-        {
-            Guard.AgainstNull(fixture, nameof(fixture));
-
-#pragma warning disable CS0618 // Type or member is obsolete
-            FixtureInitializer.InitializeFakes(fixture);
-#pragma warning restore CS0618 // Type or member is obsolete
-        }
-    }
-}
```
thomaslevesque: Good riddance! I never understood the need for this class

tests/FakeItEasy.Tests/Core/DefaultFakeManagerAccessorTests.cs
```diff
@@ -10,12 +10,7 @@
 
     public class DefaultFakeManagerAccessorTests
     {
-        private readonly DefaultFakeManagerAccessor accessor;
-
-        public DefaultFakeManagerAccessorTests()
-        {
-            this.accessor = new DefaultFakeManagerAccessor();
-        }
+        private readonly IFakeManagerAccessor accessor = ServiceLocator.Current.Resolve<IFakeManagerAccessor>();
```
thomaslevesque: Are you sure about this change? This test fixture is `DefaultFakeManagerAccessorTests`, but it never explicitly creates a `DefaultFakeManagerAccessor`; instead it relies on the service locator to return a `DefaultFakeManagerAccessor`. This seems wrong...

src/FakeItEasy/RootModule.cs
```diff
@@ -106,7 +106,7 @@ public static void RegisterDependencies(DictionaryContainer container)
 
             container.RegisterSingleton(c => new EventHandlerArgumentProviderMap());
 
-            container.Register(c => new SequentialCallContext(c.Resolve<CallWriter>(), c.Resolve<StringBuilderOutputWriter.Factory>()));
+            container.RegisterSingleton<SequentialCallContext.Factory>(c => () => new SequentialCallContext(c.Resolve<CallWriter>(), c.Resolve<StringBuilderOutputWriter.Factory>()));
```
thomaslevesque: Small optimization: resolve the `CallWriter` and `StringBuilderOutputWriter.Factory` outside of the inner delegate, i.e.

```csharp
container.RegisterSingleton<SequentialCallContext.Factory>(c =>
{
    var callWriter = c.Resolve<CallWriter>();
    var outputWriterFactory = c.Resolve<StringBuilderOutputWriter.Factory>();
    return () => new SequentialCallContext(callWriter, outputWriterFactory);
});
```

This could probably be done in other places, too. Not sure how big the performance impact would be, but I guess it can't hurt.

src/FakeItEasy/IoC/DictionaryContainer.cs
```diff
@@ -8,120 +8,69 @@ namespace FakeItEasy.IoC
...
-                        return this.resolver.Resolve(container);
-                    }
-                }
-
-                private void SignalThatSingletonHasBeenCreated()
-                {
-                    this.resolveFunction = null;
-                }
-            }
+            throw new KeyNotFoundException($"The specified service {serviceType} was not registered in the container.");
```
thomaslevesque: Shouldn't there be some locking in this method? If two threads attempt to modified `cachedServices` at the same time, it could become corrupted. Or maybe use a `ConcurrentDictionary`?

src/FakeItEasy/ServiceLocator.cs
```diff
@@ -1,30 +1,45 @@
...
-        }
+            if (this.registeredServices.TryGetValue(typeof(T), out object service))
+            {
+                return (T)service;
+            }
 
-        [DebuggerStepThrough]
-        internal abstract object Resolve(Type componentType);
+            throw new KeyNotFoundException($"The specified service {typeof(T)} was not registered.");
+        }
```
thomaslevesque: Nice! I like how you removed the inheritance between `DictionaryContainer` and `ServiceLocator`.

src/FakeItEasy/RootModule.cs
thomaslevesque: In fact, ignore this comment. I had forgotten that it was a singleton, so it will only be constructed once anyway.

src/FakeItEasy/IoC/DictionaryContainer.cs
thomaslevesque: And... ignore this too. All services are constructed eagerly from a single thread, and the dictionary won't be modified afterward.

src/FakeItEasy/RootModule.cs
blairconrad: Also, those changes will come in time!

src/FakeItEasy/IoC/DictionaryContainer.cs
blairconrad: It's currently safe, since we're only accessing the `DictionaryContainer` from the static constructor of `ServiceLocator`, which is guaranteed to run once. If you want futureproofing, I can add.

tests/FakeItEasy.Tests/Core/DefaultFakeManagerAccessorTests.cs
blairconrad: I am not sure at all. I have no idea why I did that. (Nor any memory of it, but that's not unusual.) Thank you!

blairconrad:
Thanks for the insightful comments, @thomaslevesque. I think I addressed (reverted) the one change you were unhappy with. Also, I rebased and took the liberty of pushing a bonus commit from the future. It stops us using `ServiceLocator` from within `RootModule` code, and is a preview of how we can stop registering many types used only inside the *Modules.

blairconrad:
Sorry, I just amended 72b9a1bf574b55f6afdbd176b41830f92ef35379 instead of including the fixup commit (dumb/lazy), but the only change I made was to revert the area you called out, and am now (once again) creating a `DefaultFakeManagerAccessor` in its test class.

blairconrad:
Thank you, @thomaslevesque! More changes to come in another PR (not today). Unfortunately, the next phase is going to be a bit of a slog, while I switch from registering and resolving everything to eagerly instantiating objects and then only registering if we need to.

PR #1485, opened 2018-11-02
Optimize IsParamArrayExpression

blairconrad:
Connects to #1470.

Checking to see if a parameter is a params array is expensive. Let's at least not do it unless the parameter is an array. Also, let's check smarter.

image


blairconrad:
Playing with a matplotlib to format the BenchmarkDotNet's output:

image

The "alpha" was the last build of develop on AppVeyor.

blairconrad:
Thanks, sir.

Issue #1489, opened 2018-11-07
Improve performance of Fake creation


blairconrad:
It could be faster! We should probably focus on optimizing faking interfaces, since I hope that's the most common use case, but it's worth looking at faking classes and delegates as well.

blairconrad:
While there are always performance improvements to be made, we're probably done for this release.
Closed in #1490, #1491, #1497, and #1504.

thomaslevesque:
This change has been released in FakeItEasy [5.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/5.0.0).

PR #1491, opened 2018-11-08
Share stateless fake call rules

blairconrad:
Connects to #1489.

Avoids creation of a number of rule objects with every Fake, at the cost of a slightly more complicated implementation.

|                                            Method |                  Job |     Mean |     Error |    StdDev | Ratio | RatioSD |
|-------------------------------------------------- |--------------------- |---------:|----------:|----------:|------:|--------:|
| &#39;A.Fake&lt;SimpleClass&gt;(o =&gt; o.Implements&lt;IFake&gt;())&#39; |      4.10.0-alpha070 | 5.741 us | 0.6874 us | 0.4546 us |  1.00 |    0.00 |
| &#39;A.Fake&lt;SimpleClass&gt;(o =&gt; o.Implements&lt;IFake&gt;())&#39; | 4.9.0-share-rules001 | 5.208 us | 0.3416 us | 0.2260 us |  0.91 |    0.09 |
|                             A.Fake&lt;SimpleClass&gt;() |      4.10.0-alpha070 | 4.833 us | 0.3176 us | 0.1890 us |  1.00 |    0.00 |
|                             A.Fake&lt;SimpleClass&gt;() | 4.9.0-share-rules001 | 4.190 us | 0.1527 us | 0.0909 us |  0.87 |    0.05 |
|                         A.Fake&lt;LessSimpleClass&gt;() |      4.10.0-alpha070 | 9.777 us | 0.0985 us | 0.0586 us |  1.00 |    0.00 |
|                         A.Fake&lt;LessSimpleClass&gt;() | 4.9.0-share-rules001 | 9.733 us | 0.0836 us | 0.0437 us |  1.00 |    0.01 |
|                                   A.Fake&lt;IFake&gt;() |      4.10.0-alpha070 | 5.158 us | 0.5939 us | 0.3928 us |  1.00 |    0.00 |
|                                   A.Fake&lt;IFake&gt;() | 4.9.0-share-rules001 | 4.725 us | 0.1679 us | 0.1111 us |  0.92 |    0.07 |

image


blairconrad:
The first couple of commits were separated out to make the diffs easier. If you prefer, we could squash later. I'm not fussy.

blairconrad:
Um. Hold tight. I think I accidentally reordered some methods, making the diff hard.

blairconrad:
Okay. Unreordered the `ObjectMemberRule` methods. Review away!

tests/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
```diff
@@ -141,6 +141,35 @@ public interface IFoo
...
+        }
+
+        [Scenario]
+        public static void EqualsDifferentInstanceOfSameFakedType(FakedClass fake1, FakedClass fake2, bool equalsResult)
+        {
+            "Given a faked class instance"
+                .x(() => fake1 = A.Fake<FakedClass>());
+
+            "And a second faked class instance of the same type"
+                .x(() => fake1 = A.Fake<FakedClass>());
```
blairconrad: copy/paste error! Should be `fake2`. Will update sometime this evening.

tests/FakeItEasy.Tests/Core/FakeManagerTests.cs
```diff
@@ -342,62 +339,14 @@ public void Should_return_hash_code_of_fake_manager_when_GetHashCode_has_not_bee
...
-            A.CallTo(() => interceptedCall.FakedObject).Returns(proxyPassedToEquals);
-            A.CallTo(() => interceptedCall.Arguments).Returns(new ArgumentCollection(new object[] { proxyPassedToEquals }, interceptedCall.Method));
-
-            // Act
-            ProcessFakeObjectCall(interceptingProxyManager, interceptedCall);
-
-            // Assert
-            A.CallTo(() => interceptedCall.SetReturnValue(true)).MustHaveHappened();
-        }
-
```
blairconrad: These started failing, because we don't use the faked `FakeManagerAccessor`. Moved to specs (which you saw above), as near as I was able.

blairconrad:
I added a late-breaking commit that reduces the overhead of the `EventRule`. Would eventually squash.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -241,5 +267,22 @@ private void UpdateLastSequenceNumber(int sequenceNumber)
             while (sequenceNumber > last &&
                    sequenceNumber != Interlocked.CompareExchange(ref this.lastSequenceNumber, sequenceNumber, last));
         }
+
+        /// <summary>
+        /// A hack to allow shared object call rules. Some of these rules need access to a FakeManager when executing, so
+        /// it'll be supplied in the <see cref="ApplyBestRule(IInterceptedFakeObjectCall, out IFakeObjectCallRule)"/> method.
+        /// The only reason these rules extend <see cref="IFakeObjectCallRule"/> is that they are emitted through
+        /// <see cref="IInterceptionListener.OnAfterCallIntercepted"/>.
+        /// </summary>
```
blairconrad: As I hint at, this hack wouldn't be necessary if `IFakeObjectCallRule` didn't expose `Apply`. But it needs to in order for people to create their own rules.

But I don't think that `IInterceptionListener.OnAfterCallIntercepted` needs to be passed an object that has an `Apply` method; there's no need for `OnAfterCallIntercepted` to call `Apply`.
I don't even know why it would need to call `IsApplicable` or `NumberOfTimesToCall`. Shall we explore  changing this in a future major version? We could supply `OnAfterCallIntercepted` with a different interface that's a restricted version of `IFakeObjectCallRule`, or even remove the `IFakeObjectCallRule` parameter.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -143,14 +141,10 @@ void IFakeCallProcessor.Process(IInterceptedFakeObjectCall fakeObjectCall)
...
-            var ruleToUse =
-                (from rule in this.AllRules
-                 where rule.Rule.IsApplicableTo(fakeObjectCall) && rule.HasNotBeenCalledSpecifiedNumberOfTimes()
-                 select rule).First();
-
+            IFakeObjectCallRule ruleUsed = null;
             try
             {
-                ApplyRule(ruleToUse, fakeObjectCall);
+                this.ApplyBestRule(fakeObjectCall, out ruleUsed);
```
thomaslevesque: Is there a reason to return `ruleUsed` as an out parameter here? Can't it just be returned directly, since there's no other return value?

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: Seems like a good idea. I don't see any value in having the call rule available in `OnAfterCallIntercepted`, except for information purposes. Maybe we could just pass a string describing the rule, or just the type of the rule.

src/FakeItEasy/Core/FakeManager.cs
blairconrad: It was a way to set the `ruleUsed` (and ensure that the caller knew what it was) before attempting the application of the rule. If the `.Apply` throws, we won't get an opportunity to return the rule. I could wrap each `Apply` call in `ApplyBestRule` with try/catches and return the rule regardless.

blairconrad:
Thanks, @thomaslevesque. Good comments! Some of the requested changes break unit tests, so there will be a slight delay as I address that, but everything you ask about is achievable.

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: Ah, I see. You can leave it as is, then.

blairconrad: Sorry, I was unclear. In progress, I'm replacing `ApplyBestRule` with 

```
private IFakeObjectCallRule ChooseBestRule(IInterceptedFakeObjectCall fakeObjectCall)
{
    foreach (var preUserRule in this.preUserRules)
    {
        if (preUserRule.IsApplicableTo(fakeObjectCall))
        {
            return preUserRule;
        }
    }

    foreach (var rule in this.AllUserRules)
    {
        if (rule.Rule.IsApplicableTo(fakeObjectCall) && rule.HasNotBeenCalledSpecifiedNumberOfTimes())
        {
            rule.CalledNumberOfTimes++;
            return rule.Rule;
        }
    }

    foreach (var postUserRule in PostUserRules)
    {
        if (postUserRule.IsApplicableTo(fakeObjectCall))
        {
            return postUserRule;
        }
    }

    // The last rule in the postUserRules is applicable to all calls, so this line is here to satisfy
    // the compiler. We will always have returned by now.
    return null;
}
```

And then we can just call `Apply` right after. I think it's better than the `out`. But if you prefer the `out`, I can leave.

thomaslevesque: Ah, I see. What bothers me with this approach is that `CalledNumberOfTimes` is incremented in the method that chooses the rule, rather than in the method that applies it...

blairconrad: A fair point. Back to `out` with an explanatory comment?

thomaslevesque: Sure, a comment wouldn't hurt!

blairconrad:
Thanks, @thomaslevesque. Updated. I added two fixup commits, but the second one sort of spans two original commits. If ever you're happy with the changes, I'd be content to squash down to 1, unless you prefer some separation.

tests/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
```diff
@@ -166,10 +179,26 @@ public static void EqualsDifferentInstanceOfSameFakedType(FakedClass fake1, Fake
...
+
+            "When I get its hash code"
+                .x(() => hashCode = fake.GetHashCode());
+
+            "And I get its manager's hash code"
+                .x(() => managerHashCode = Fake.GetFakeManager(fake).GetHashCode());
+
+            "Then the hash codes are the same"
+                .x(() => hashCode.Should().Be(managerHashCode));
+        }
```
thomaslevesque: Isn't the fact that the fake and its manager have the same hashcode an implementation detail? i.e. do we really need a spec for it?

blairconrad: You're right. I always struggle with how to test `GetHashCode`. We can make sure that fakes that are equal have the same hash code, but since we've a 1:1 map between fakes and their managers, the implemented behaviour devolves to "use reference semantics for `Equals` and `GetHashCode`". Then the only meaningful test I can think of is "a Fake generates the same hash code twice". Which isn't bad, but it is pretty weak.
I'd considered checking that two fakes give different hash codes, but that's not a requirement and the test could fail even when everything is working properly.

I can certainly add a test to ensure that the same fake's `GetHashCode` returns the same value more than once. And I can move the test back (and update it so it passes) if you like,  but the fact that we return the manager's hash code is meaningless enough that I'd almost rather remove the test altogether.

Have you a preference?

blairconrad: Sorry, "remove the test that compares against the FakeManager's `GetHashCode` value".

blairconrad: For now, "improved" the spec at least.

thomaslevesque: > Have you a preference?

I think I'd remove this spec altogether.

thomaslevesque:
> If ever you're happy with the changes, I'd be content to squash down to 1, unless you prefer some separation.

At this point it's probably simpler to just squash everything.

blairconrad:
Thanks, @thomaslevesque. I appreciate the comments. The `GetHashCode` and "idempotent" ones were gold.
Squashed and pushed. I threw https://github.com/blairconrad/FakeItEasy/tree/share-rules-unsquashed up in case you wanted to see that the only change was to remove the spec(s).

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Ha. I gave you so much misery over the bad changes, I'm not sure I deserve thanks. :smile:

Issue #1493, opened 2018-11-14
Improve performance of faked method invocation


blairconrad:
It could be faster!

blairconrad:
While more performance improvements are always possible, we're probably done for this release anyhow.
Closed in #1498 and #1501.

thomaslevesque:
This change has been released in FakeItEasy [5.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/5.0.0).

PR #691, opened 2016-04-30
Remove SequenceNumber from ICompletedFakeObjectCall

thomaslevesque:
Closes #651 
Connects to #637

PR #1498, opened 2018-11-18
[WIP] Move sequencenumber to completed call

blairconrad:
Connects to #1493.

I did in two stages. First, the naive way where I added the `SequenceNumber` to the `ICompletedFakeObjectCall`. It's a new property exposed. I'd argue not breaking since there's no profit in anyone every implementing this. But maybe it's unsightly?
Still, if you like, we could just keep it. 
The second commit restores the public API at the cost of an additional interface and class (both internal).
If you like having that one, I'd squash before merge.

|          Method |                         Job |                        NuGetReferences |      Mean |     Error |    StdDev | Ratio | RatioSD |
|---------------- |---------------------------- |--------------------------------------- |----------:|----------:|----------:|------:|--------:|
|        Callback |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 | 27.185 us | 0.8842 us | 0.8271 us |  1.00 |    0.00 |
|        Callback | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 | 24.824 us | 0.5752 us | 0.5381 us |  0.91 |    0.04 |
|    CallbackOnly |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 |  6.254 us | 0.5566 us | 0.5206 us |  1.00 |    0.00 |
|    CallbackOnly | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 |  3.900 us | 0.1710 us | 0.1599 us |  0.63 |    0.07 |
|    Construction |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 |  7.950 us | 0.1628 us | 0.1523 us |  1.00 |    0.00 |
|    Construction | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 |  7.967 us | 0.2734 us | 0.2557 us |  1.00 |    0.03 |
|     EmptyMethod |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 | 17.067 us | 2.2661 us | 2.1198 us |  1.00 |    0.00 |
|     EmptyMethod | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 | 15.411 us | 1.9658 us | 1.8388 us |  0.91 |    0.14 |
| EmptyMethodOnly |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 |  7.861 us | 0.8080 us | 0.7558 us |  1.00 |    0.00 |
| EmptyMethodOnly | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 |  5.177 us | 0.1575 us | 0.1473 us |  0.67 |    0.08 |
|    EmptyReturns |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 | 17.160 us | 0.2899 us | 0.2711 us |  1.00 |    0.00 |
|    EmptyReturns | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 | 15.324 us | 0.3449 us | 0.3226 us |  0.89 |    0.02 |
| EmptyReturnOnly |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 |  8.280 us | 0.8279 us | 0.7744 us |  1.00 |    0.00 |
| EmptyReturnOnly | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 |  5.508 us | 0.2151 us | 0.2012 us |  0.67 |    0.08 |
|          Return |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 | 26.024 us | 1.1747 us | 1.0988 us |  1.00 |    0.00 |
|          Return | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 | 23.792 us | 0.2633 us | 0.2463 us |  0.92 |    0.03 |
|      ReturnOnly |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 |  8.420 us | 0.7777 us | 0.7275 us |  1.00 |    0.00 |
|      ReturnOnly | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 |  5.666 us | 0.3052 us | 0.2855 us |  0.68 |    0.07 |
|          Verify |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 | 30.875 us | 0.8732 us | 0.8168 us |  1.00 |    0.00 |
|          Verify | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 | 28.324 us | 0.4445 us | 0.4158 us |  0.92 |    0.03 |
|      VerifyOnly |             4.10.0-alpha078 |             FakeItEasy 4.10.0-alpha078 |  6.705 us | 0.3930 us | 0.3676 us |  1.00 |    0.00 |
|      VerifyOnly | 4.10.0-move-sequencenumb001 | FakeItEasy 4.10.0-move-sequencenumb001 |  6.261 us | 0.1246 us | 0.1165 us |  0.94 |    0.05 |

image


src/FakeItEasy/Core/InterceptedFakeObjectCall.cs
```diff
@@ -0,0 +1,54 @@
+namespace FakeItEasy.Core
+{
+    using System.Reflection;
+    using FakeItEasy.Configuration;
+
+    internal abstract class InterceptedFakeObjectCall : IInterceptedFakeObjectCall
```
thomaslevesque: Maybe I'm missing something, but why is this necessary?

src/FakeItEasy/Core/SequenceNumberManager.cs
```diff
@@ -1,25 +1,11 @@
...
-            int sequenceNumber = GetNextSequenceNumber();
-            var box = SequenceNumbers.GetOrCreateValue(call);
-            box.Value = sequenceNumber;
-            return sequenceNumber;
-        }
-
-        public static int GetSequenceNumber(ICompletedFakeObjectCall call) => SequenceNumbers.GetOrCreateValue(call).Value;
-
-        private static int GetNextSequenceNumber() => Interlocked.Increment(ref lastSequenceNumber);
+        public static int GetNextSequenceNumber() => Interlocked.Increment(ref lastSequenceNumber);
```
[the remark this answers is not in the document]
thomaslevesque: (in fact, the `Interlocked.Increment` had always been there, but the overall approach for handling sequence number is what we already had before #691)

src/FakeItEasy/Core/InterceptedFakeObjectCall.cs
blairconrad: It allowed me to return `IOrderedCompletedFakeObjectCall` from `AsReadOnly` when called internally, and only `ICompletedFakeObjectCall` when `AsReadOnly` is invoked through the `IInterceptedFakeObjectCall` interface (that is, by clients).

blairconrad:
> Thanks @blairconrad. Interestingly, we actually did the opposite thing a few years ago (#651).

So we did. I'd forgotten.  And for essentially the same reason as I made the second commit. I agree with your original comment that it's an implementation detail the user doesn't need to know about. The only real complexity added is that there's an abstract class from which our concrete intercepted calls derive. I think we should more often pass around concrete or abstract-typed results internally and expose the restricted interfaces to the outside world (I expect this is how we'll end up taking `AsReadOnly` off the public's view of the intercepted call, which I'd thought we'd talked about recently, but I can't find a reference).

> I'm a bit torn about the second commit. One one hand, it avoids a breaking change (which probably isn't really breaking to anyone, as you mentioned). On the other hand, it adds some complexity, and I'd argue that having the SequenceNumber exposed doesn't hurt anyone and could actually be useful.

blairconrad:
Oh, and I forgot to say earlier: the second commit is reversible if we change our minds later. I'd rather add a member to `ICompletedFakeObjectCall` in a minor release than remove one.

thomaslevesque:
> I think we should more often pass around concrete or abstract-typed results internally and expose the restricted interfaces to the outside world

👍 

> (I expect this is how we'll end up taking `AsReadOnly` off the public's view of the intercepted call, which I'd thought we'd talked about recently, but I can't find a reference).

Not recently, I think. We did discuss it in #651.

OK, let's merge as is. We can probably simplify things a bit when we decide to remove `AsReadOnly`.

thomaslevesque:
Oh, it's still WIP

blairconrad:
I left WIP because I'd thought better to squash before merge. It's really only one change. I just split in case you _hated_ the `InterceptedFakeObjectCall` workaround. Squashing actually makes the change look (a little) less daunting. If nothing else, it removes the reversal in the acceptance tests.
Oh! I can remove `IOrderedCompletedFakeObjectCall` if I make `CompletedFakeObjectCall` serializable. Pushing that as an amendment and heading to work. Will alter or squash or whatever you want later.

thomaslevesque:
> Oh! I can remove `IOrderedCompletedFakeObjectCall` if I make `CompletedFakeObjectCall` serializable.

Nice! But shouldn't it be a fixup for the second commit?

blairconrad:
Oh, I suppose. But if we squash all into one, it works out the same. You
prefer two commits?

thomaslevesque:
> You prefer two commits?

No, the full diff is clear enough as it is. Squash it 👍

blairconrad:
I added an unsquashed commit for you to assess. Takes away the build warnings while still having the tests pass. And makes it even more explicit that we won't preserve calls across serialization, which isn't new behaviour.

blairconrad:

blairconrad:
Ha! Turns out this breaks one of the characterization tests I added in #1499. We gain slightly more serialization powers after this change.

thomaslevesque:
> Turns out this breaks one of the characterization tests I added in #1499.

So what do you want to do? Merge anyway?

blairconrad:
Oh, please do merge. I've already updated the test and repushed to that branch. I'll have to rebase before we merge #1499, but that was always so.
Or if you prefer, I could restore #1499 to original, we could merge it, then I update this PR to contain the change to the characterization test.
Me, I think easier and just as good to merge this PR first.

thomaslevesque:
Let's go, then! Thanks @blairconrad

blairconrad:
Thank you, @thomaslevesque!

PR #1499, opened 2018-11-20
Figure out serialization support

blairconrad:
As an outgrowth of recent discussions in #1498, I thought it would be nice to figure out what support we have for serialization, so I consolidated some tests and added a few (and removed one). I think this will help us reason about how to proceed with serialization support.


tests/FakeItEasy.Specs/SerializationSpecs.cs
```diff
@@ -0,0 +1,136 @@
...
+                .x(() => fake = A.Fake<IInterface>());
+
+            "And I call a method"
+                .x(() => fake.ParseAString("a string"));
+
+            "When I serialize and deserialize the fake"
+                .x(() => deserializedFake = SerializeAndDeserialize(fake));
+
+            "Then the record of the completed call is lost"
+                .x(() => Fake.GetCalls(deserializedFake).Should().BeEmpty());
```
blairconrad: @thomaslevesque, this is the test that I had to update. Now that I don't serialize the recorded calls on the FakeManager, serialization passes. Previously it would've thrown on serialization.
However, we still don't have access to the previously-recorded calls upon deserialization.

thomaslevesque:

PR #1506, opened 2018-11-26
[WIP] switch to MinVer

adamralph:
Another alpha test. Hope you don't mind.

This time I've added the project elements to required to replace the removed assembly attributes. As with last time, for the purpose of the alpha test, I've removed the building of the analyser meta-package.

FYI alpha 21 is the last planned alpha.

### DIFF
--- a/src/CommonAssemblyInfo.cs
+++ b/src/CommonAssemblyInfo.cs
@@ -1,10 +1,7 @@
-using System;
 using System.Reflection;
 using System.Resources;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyCompany("Patrik Hägne")]
-[assembly: AssemblyCopyright("Copyright (c) FakeItEasy contributors. (fakeiteasyfx@gmail.com)")]
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
 [assembly: NeutralResourcesLanguage("en-US")]
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@
 [Oo]bjcs
 [Oo]bjvb
 artifacts
-src/VersionInfo.cs
 src/FakeItEasy.nuspec
 tools/packages.config.sha1
 tools/*/ToolPaths.cs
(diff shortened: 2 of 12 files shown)


adamralph:
All good! https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/builds/20563744#L135

blairconrad:
Hey, @adamralph. We don't mind at all. It's nice getting a preview of what would change. I do like the shape of the change. It'd be _nice_ if there was a way to retain the metapackage, but maybe we should relookat dropping it.

adamralph:
@blairconrad there could be a way to keep it. After I release the first beta (probably tomorrow) I'll send another PR and see if I can work out a way to do it.

blairconrad:
Oh, very cool. If nothing else, the minver-cli should allow this functionality, no?

adamralph:
I guess it could, but I'm not really sure about mixing `MinVer` and `minver-cli` in the same project. The inputs to each would have to kept in sync somehow, and that might not be so easy.

src/CommonAssemblyInfo.cs
```diff
@@ -1,10 +1,7 @@
-using System;
 using System.Reflection;
 using System.Resources;
 using System.Runtime.InteropServices;
 
-[assembly: AssemblyCompany("Patrik Hägne")]
-[assembly: AssemblyCopyright("Copyright (c) FakeItEasy contributors. (fakeiteasyfx@gmail.com)")]
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
 [assembly: NeutralResourcesLanguage("en-US")]
```
thomaslevesque: While we're at it, we should probably consider moving everything from AssemblyInfo.cs to csproj or Directory.Build.props (except `ComVisibleAttribute`, which is required by [CA1017](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1017-mark-assemblies-with-comvisibleattribute?view=vs-2017) and can't be set in csproj as far as I know)

adamralph: I think that's a good move. Probably best done now, regardless of MinVer.

PR #1510, opened 2018-11-29
Remove analyzer meta-package

### LINKED ISSUE #1509: Remove Analyzer meta-package in 5.0.0
There's no need to subscribe to an analyzer that examines both C# and VB code.

thomaslevesque:
Fixes #1509 

As mentioned previously, this PR replaces the analyzer meta-package with a package that only contains a MSBuild target that emits a warning.
The package will have to be built and deployed manually (no point in including it in the build process, since it's a one-shot operation)


src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.targets
```diff
@@ -0,0 +1,10 @@
+<Project>
+
+    <Target Name="WarnAboutWrongAnalyzerPackage" AfterTargets="AfterBuild">
+        <Warning Text="The FakeItEasy.Analyzer package is no longer maintained and does nothing. Use the C# specific analyzer package instead: FakeItEasy.Analyzer.CSharp."
+                 Condition=" '$(Language)' == 'C#' " />
+        <Warning Text="The FakeItEasy.Analyzer package is no longer maintained and does nothing. Use the VB.NET specific analyzer package instead: FakeItEasy.Analyzer.VisualBasic."
+                 Condition=" '$(Language)' == 'VB' " />
```
thomaslevesque: I'm not 100% sure about this. I'm not certain the `Language` property is always defined. Maybe there should be a generic message in case we can't determine the project language.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.nuspec
```diff
@@ -2,21 +2,19 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>FakeItEasy.Analyzer</id>
-    <version>$version$</version>
+    <version>5.0.0</version>
     <title>FakeItEasy.Analyzer</title>
     <authors>Thomas Levesque, FakeItEasy contributors</authors>
-    <description>A convenience package that installs the FakeItEasy Analyzer for both C# and Visual Basic. Clients using FakeItEasy from only one of the two languages should prefer FakeItEasy.Analyzer.CSharp or FakeItEasy.Analyzer.VisualBasic. Provides diagnostic analyzers to warn about incorrect usage of FakeItEasy. Works in Visual Studio 2015 Update 1 or later.</description>
+    <description>DO NO INSTALL THIS PACKAGE. It's no longer maintained and does nothing. Use a language specific analyzer package instead: FakeItEasy.Analyzer.CSharp or FakeItEasy.Analyzer.VisualBasic.</description>
```
blairconrad: "language specific" → "language-specific"

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.targets
blairconrad: So leave the two that are here and add a third with no condition? Sounds nice and safe.

thomaslevesque:
> I think we should make the tarted-up package manually and deploy, as you say, and then the merged-in PR would just remove the meta-package and all dependencies.

Works for me. I thought we would want to keep a trace of the nuspec/targets, but since they'll be in the package anyway, I guess we don't need to.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.targets
thomaslevesque: Well, I still need a condition, otherwise there will be two warnings when `Language` is defined... Or maybe I could use [`Choose`](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-conditional-constructs?view=vs-2017).

thomaslevesque: `Choose` doesn't seem to work in a `Target`...

thomaslevesque:
The last commit is an attempt to make the target look a bit nicer.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.nuspec
```diff
@@ -2,21 +2,19 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>FakeItEasy.Analyzer</id>
-    <version>$version$</version>
+    <version>5.0.0</version>
     <title>FakeItEasy.Analyzer</title>
     <authors>Thomas Levesque, FakeItEasy contributors</authors>
-    <description>A convenience package that installs the FakeItEasy Analyzer for both C# and Visual Basic. Clients using FakeItEasy from only one of the two languages should prefer FakeItEasy.Analyzer.CSharp or FakeItEasy.Analyzer.VisualBasic. Provides diagnostic analyzers to warn about incorrect usage of FakeItEasy. Works in Visual Studio 2015 Update 1 or later.</description>
+    <description>DO NO INSTALL THIS PACKAGE. It's no longer maintained and does nothing. Use one of the language-specific analyzer packages instead: FakeItEasy.Analyzer.CSharp or FakeItEasy.Analyzer.VisualBasic.</description>
```
blairconrad: I wish I'd noticed that "DO NO" should be "DO NOT" the first time around!

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.targets
blairconrad: I'd wondered. It may be possible to phrase things so the two warnings complement one another. For example,

"The FakeItEasy.Analyzer package is no longer maintained and does nothing. Language-specific analyzer packages are provided instead." could always be emitted.

And then one of 

"Use the VB.NET-specific analyzer package instead: FakeItEasy.Analyzer.VisualBasic."

or 

"Use the C#-specific analyzer package instead: FakeItEasy.Analyzer.CSharp."

if the language condition is met.

If the warning order is preserved, it's not terrible, I think.

blairconrad:
I'd commented without noticing your last commit. That's a very nice way to handle the problem of undetected languages.

src/FakeItEasy.Analyzer/FakeItEasy.Analyzer.nuspec
thomaslevesque: oops

thomaslevesque: Fixed

tools/FakeItEasy.Build/Program.cs
```diff
@@ -89,14 +88,9 @@ public static void Main(string[] args)
...
 
-            Target(
-                "pack-nuspecs",
-                DependsOn("outputDirectory", "get-version"),
-                () => Run(ToolPaths.NuGet, $"pack {AnalyzerMetaPackageNuspecPath} -Version {version} -OutputDirectory {OutputDirectory} -NoPackageAnalysis"));
-
             Target(
                 "pack",
-                DependsOn("pack-projects", "pack-nuspecs"));
+                DependsOn("pack-projects"));
```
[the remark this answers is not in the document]
thomaslevesque: Yes, I intended to do it, but since this PR won't be merged anyway, I didn't bother

blairconrad:
@thomaslevesque, I like the new content. I've not tested a package built from it, but it looks like it'll do just the right thing. I think we should consider publishing the new package in conjunction with FakeItEasy 5.0.0. Technically it could be done earlier, I suppose, but then we'd force people from FakeItEasy.Analyzer 4.9.1 to FakeItEasy.Analyzer 5.0.0 to FakeItEasy.Analyzer.CSharp 4.9.1 to (eventually) FakeItEasy.Analyzer.CSharp 5.0.0.

thomaslevesque:
> I've not tested a package built from it

I have (with C#, VB, and F# projects)

> I think we should consider publishing the new package in conjunction with FakeItEasy 5.0.0.

I think so too

thomaslevesque:
I removed the nuspec and targets from the PR, they can be found here: https://gist.github.com/thomaslevesque/289b9f917a129890bed04dd0320e10b1

blairconrad:
We should update the `analyzer.md` file. Actually we should've earlier, but now even more.
And remove `FakeItEasy.Analyzer.nuspec` from `FakeItEasy.sln`

I believe I am otherwise very happy!

FakeItEasy.sln
```diff
@@ -24,9 +24,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{59466C4D-C40
 	ProjectSection(SolutionItems) = preProject
 		src\CommonAssemblyInfo.cs = src\CommonAssemblyInfo.cs
 		src\Directory.Build.props = src\Directory.Build.props
-		src\FakeItEasy.Analyzer.nuspec = src\FakeItEasy.Analyzer.nuspec
 		src\FakeItEasy.Dictionary.xml = src\FakeItEasy.Dictionary.xml
-		src\FakeItEasy.nuspec.template = src\FakeItEasy.nuspec.template
```
thomaslevesque: This one should have been removed a long time ago!

docs/analyzer.md
```diff
@@ -6,13 +6,10 @@ bugs at runtime.
 
 ## Installation
 
-The analyzer works in Visual Studio 2015 Update 1 or later. It can be installed
-[from NuGet](https://www.nuget.org/packages/FakeItEasy.Analyzer) for each project
-that needs it:
-
-```
-PM> Install-Package FakeItEasy.Analyzer
-```
```
thomaslevesque: I removed the part showing how to install the package. By now people should be used to installing NuGet packages, they don't need to be told. And anyway, I doubt many people still use the package manager console (the only place where this command would work).

thomaslevesque:
I squashed the fixup that updated the sln file

blairconrad:
Nice, @thomaslevesque. I'm enjoying the lack of NuGet command line tool. And other things.

thomaslevesque:
Thanks!

PR #1513, opened 2018-12-01
Remove automatic fake injection into system under test (InitializeFixture)

### LINKED ISSUE #992: Remove automatic fake injection into system under test (InitializeFixture) in 5.0.0
As suggested by @adamralph in https://github.com/FakeItEasy/FakeItEasy/issues/964#issuecomment-280834895. I'm creating this issue because if we go through with it, the decision deserves to be visible.

He suggested that the feature is "half-baked bloat". 

@thomaslevesque says

> I liked the feature when I discovered it, but I soon encountered annoying limitations. It's better to use an auto-mocking container such as [this](http://www.thomaslevesque.com/2015/06/14/create-an-auto-mocking-container-with-unity-and-fakeiteasy/).

and that

> Well, it's trying to mimick what an IoC container would do (injecting dependencies automatically into classes), but it's much more limited than a real IoC container. You would be better off with an auto-mocking container (see the link in my previous message, or an IoC integration package such as AutoFac.Extras.FakeItEasy). AutoFixture.FakeItEasy would work too, but it would mean that you have to migrate your tests to use AutoFixture, which is a non-trivial amount of work.

@adamralph agreed. @blairconrad (that's me) said nothing at the time, but agrees. An external mechanism (and several exist) seems like a better choice.

@apacha [says](https://github.com/FakeItEasy/FakeItEasy/issues/964#issuecomment-280915464)

> No, I'd really come to love this feature. Please don't remove it. What exactly do you mean, that it is half-baked? I only wanted that it gets documented a little. 
Switching to AutoFixture.AutoFakeItEasy might be an option, but its even less documented. At least if you deprecate it, please provide a link to an article about AutoFixture.AutoFakeItEasy that explains how to do the same thing. That would be great.

blairconrad:
Fixes #992.


blairconrad:
Thanks, @thomaslevesque!

PR #1514, opened 2018-12-01
Stop passing rule to interception listener

### LINKED ISSUE #1494: Stop passing IFakeObjectCallRule to OnAfterCallIntercepted in 5.0.0
In https://github.com/FakeItEasy/FakeItEasy/pull/1491#discussion_r233072656, @blairconrad said:

> As I hint at, this hack wouldn't be necessary if `IFakeObjectCallRule` didn't expose `Apply`. But it needs to in order for people to create their own rules.
> 
> But I don't think that `IInterceptionListener.OnAfterCallIntercepted` needs to be passed an object that has an `Apply` method; there's no need for `OnAfterCallIntercepted` to call `Apply`.
> I don't even know why it would need to call `IsApplicable` or `NumberOfTimesToCall`. Shall we explore changing this in a future major version? We could supply `OnAfterCallIntercepted` with a different interface that's a restricted version of `IFakeObjectCallRule`, or even remove the `IFakeObjectCallRule` parameter.

@thomaslevesque replied:

> Seems like a good idea. I don't see any value in having the call rule available in `OnAfterCallIntercepted`, except for information purposes. Maybe we could just pass a string describing the rule, or just the type of the rule.

thomaslevesque:
Fixes #1494

### DIFF
--- a/tests/FakeItEasy.Tests/Core/FakeManagerTests.cs
+++ b/tests/FakeItEasy.Tests/Core/FakeManagerTests.cs
@@ -384,17 +384,13 @@ namespace FakeItEasy.Tests.Core
             var listener = A.Fake<IInterceptionListener>();
             var manager = new FakeManager(typeof(int), 0);
 
-            var selectedRule = A.Fake<IFakeObjectCallRule>();
-            A.CallTo(() => selectedRule.IsApplicableTo(interceptedCall)).Returns(true);
-
-            manager.AddRuleFirst(selectedRule);
             manager.AddInterceptionListener(listener);
 
             // Act
             ProcessFakeObjectCall(manager, interceptedCall);
 
             // Assert
-            A.CallTo(() => listener.OnAfterCallIntercepted(completedCall, selectedRule)).MustHaveHappened();
+            A.CallTo(() => listener.OnAfterCallIntercepted(completedCall)).MustHaveHappened();
         }
 
         [Fact]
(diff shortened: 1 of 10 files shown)


thomaslevesque:
Oops, looks like I messed up my branches...

thomaslevesque:
Fixed

The first commit passes the type of the rule. The second commit removes the parameter completely.

blairconrad:
> Oops, looks like I messed up my branches...

Ha. This morning I accidentally pushed to origin/develop, then reset local to upstream/develop, pushed, realized I'd lost all my work, and restored from reflog. 
On the brighter side, a colleague just learning git had had some questions about whether commits disappear when they're no longer reachable, so I was able to copy my fiasco to use as a teaching moment on Monday.

tests/FakeItEasy.Tests/Core/FakeManagerTests.cs
```diff
@@ -394,7 +394,7 @@ public void Should_invoke_listener_after_call_has_been_intercepted()
             ProcessFakeObjectCall(manager, interceptedCall);
 
             // Assert
-            A.CallTo(() => listener.OnAfterCallIntercepted(completedCall, selectedRule)).MustHaveHappened();
+            A.CallTo(() => listener.OnAfterCallIntercepted(completedCall)).MustHaveHappened();
```
blairconrad: We could remove `selectedRule` altogether. We don't _have to_, but we could.

blairconrad: I'm not trying to be pushy, but usually you'd comply or comment after I say something like this, and you did neither. I suspect you missed my late-breaking comment.

thomaslevesque: I'm not sure I understand... I did remove `selectedRule`!

thomaslevesque: Ah, you mean the variable is still there! Sorry, will fix.

blairconrad:
Whoops. I didn't notice your latest commit. Great. Thanks!

thomaslevesque:
Squashed

blairconrad:
Thanks, @thomaslevesque.

PR #1520, opened 2018-12-02
refactor build script

adamralph:
I spotted a few things which were redundant and/or could be simplified.

### DIFF
--- a/tools/FakeItEasy.Build/Program.cs
+++ b/tools/FakeItEasy.Build/Program.cs
@@ -1,16 +1,12 @@
 ﻿namespace FakeItEasy.Build
 {
     using System.Collections.Generic;
-    using System.IO;
+    using System.Linq;
     using static Bullseye.Targets;
     using static SimpleExec.Command;
 
     public class Program
     {
-        private const string LogsDirectory = "artifacts/logs";
-        private const string Solution = "FakeItEasy.sln";
-        private const string RepoUrl = "https://github.com/FakeItEasy/FakeItEasy";
-
         private static readonly string[] ProjectsToPack =
         {
             "src/FakeItEasy/FakeItEasy.csproj",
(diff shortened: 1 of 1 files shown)


adamralph:
BTW - if you can ever figure out a way to get something like SourceLink working, then you can get rid of the `pdggit` target and the explicit `pack` target (by adding `<GeneratePackageOnBuild>true</GeneratePackageOnBuild>` to `src/Directory.Build.props`). That would make the build script incredibly trivial.

tools/FakeItEasy.Build/Program.cs
```diff
@@ -51,45 +47,36 @@ public class Program
...
-        private static readonly string OutputDirectory = Path.GetFullPath("artifacts/output");
-
         public static void Main(string[] args)
         {
-            Target("default", DependsOn("unit", "integ", "spec", "approve", "pack"));
-
-            Target("outputDirectory", () => Directory.CreateDirectory(OutputDirectory));
-
-            Target("logsDirectory", () => Directory.CreateDirectory(LogsDirectory));
+            Target("default", dependsOn: TestSuites.Keys.Concat(new[] { "pack" }));
```
blairconrad: I was worried about this when I saw it. I was worried that the test execution order would be some crazy thing, and I like the current order (well, I like `approve` being last; I'd almost prefer `spec` first, since we're moving so many tests there). I see they ran in the same orer, and wonder if this was just luck or what. Any idea?

adamralph: Good catch. Apparently, [the order is non-deterministic], so I'll revert this change.

adamralph:
@blairconrad I amended the commit to revert the change to the `default` target dependency specification.

thomaslevesque:
> BTW - if you ever figure out a way to get something like SourceLink working, then you can get rid of the `pdbgit` target

Yeah, I know... I really want to get SourceLink working, but ILMerge is getting in the way (again).

blairconrad:
Sorry, @adamralph. Another change got merged underneath you. Would you mind rebasing?

adamralph:
@blairconrad done.

thomaslevesque:
Thanks @adamralph!

PR #1522, opened 2018-12-03
Stop faking delegates not accessible to DynamicProxy

### LINKED ISSUE #1508: Stop faking delegates that are not visible to DynamicProxy in 5.0.0
Connects to #1437, in that if DynamicProxy gives us a way to fake delegates, we would likely have to drop this support anyhow. And it would make our capabilities more consistent, since we cannot fake classes or interfaces that aren't visible to DynamicProxy.

thomaslevesque:
Fixes #1508

### DIFF
--- a/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
+++ b/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
@@ -1,6 +1,7 @@
 namespace FakeItEasy.Creation.DelegateProxies
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Linq;
     using System.Linq.Expressions;
@@ -12,6 +13,7 @@ namespace FakeItEasy.Creation.DelegateProxies
     internal static class DelegateProxyGenerator
     {
         private static readonly ProxyGenerator ProxyGenerator = new ProxyGenerator();
+        private static readonly ConcurrentDictionary<Type, bool> AccessibleToDynamicProxyCache = new ConcurrentDictionary<Type, bool>();
 
         public static ProxyGeneratorResult GenerateProxy(
             Type typeOfProxy,
@@ -21,7 +23,7 @@ namespace FakeItEasy.Creation.DelegateProxies
 
             var invokeMethod = typeOfProxy.GetMethod("Invoke");
 
-            if (!ProxyUtil.IsAccessible(typeOfProxy))
+            if (!IsAccessibleToDynamicProxy(typeOfProxy))
             {
                 try
                 {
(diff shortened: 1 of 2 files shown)


blairconrad:
I don't mind it so much. I'd probably add a comment explaining why we're doing this craziness. Also, does it work well with generic delegates that act on inaccessible types, such as `Func<PrivateClass>`?

thomaslevesque:
> I don't mind it so much. I'd probably add a comment explaining why we're doing this craziness.

Good point, will do.

> Also, does it work well with generic delegates that act on inaccessible types, such as `Func<PrivateClass>`?

I have no idea. I'll try.

thomaslevesque:
> Also, does it work well with generic delegates that act on inaccessible types, such as `Func<PrivateClass>`?

Nope. Well, it works *too* well... it doesn't throw. Apparently `ProxyUtil.IsAccessible` doesn't look at the type parameters...

thomaslevesque:
OK, it should work now

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -1,28 +1,70 @@
...
+    using Castle.DynamicProxy;
     using FakeItEasy.Configuration;
     using FakeItEasy.Core;
 
     internal static class DelegateProxyGenerator
     {
+        private static readonly ProxyGenerator ProxyGenerator = new ProxyGenerator();
+        private static readonly ConcurrentDictionary<Type, bool> AccessibleToDynamicProxyCache = new ConcurrentDictionary<Type, bool>();
+
+
```
blairconrad: the thing that warns about extra blank lines warns about this extra blank line

tests/FakeItEasy.Specs/CreationSpecs.cs
```diff
@@ -325,6 +325,44 @@ public void PrivateClassCannotBeFaked(Exception exception)
...
+
+            "When I create a fake of the delegate"
+                .x(() => exception = Record.Exception(this.CreateFake<Func<PrivateClass>>));
+
+            "Then it throws a fake creation exception"
+                .x(() => exception.Should().BeOfType<FakeCreationException>());
+
+            "And the exception message indicates the reason for failure"
+                .x(() => exception.Message.Should().Match(@"
+  Failed to create fake of type System.Func`1[FakeItEasy.Specs.CreationSpecsBase+PrivateClass]:
```
blairconrad: Ideally we'd be warning that we can't fake `Func<PrivateClass>`, as we're not actually trying to fake `PrivateClass`.
Sorry.

thomaslevesque: Isn't it what we're doing?

blairconrad:
@thomaslevesque, webhooks are not reporting statuses or something, but you might as well squash it up, if you like. :+1:

thomaslevesque:
Squashed!

blairconrad:
Thanks, @thomaslevesque!

PR #1523, opened 2018-12-04
Restore nupkgs to artifacts/output

blairconrad:
Connects to #1520.

### DIFF
--- a/tools/FakeItEasy.Build/Program.cs
+++ b/tools/FakeItEasy.Build/Program.cs
@@ -1,7 +1,7 @@
 ﻿namespace FakeItEasy.Build
 {
     using System.Collections.Generic;
-    using System.Linq;
+    using System.IO;
     using static Bullseye.Targets;
     using static SimpleExec.Command;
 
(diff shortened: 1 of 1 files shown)


blairconrad:
Currently there being placed under their project build areas and are not published to the AppVeyor feed.

adamralph:
Oops, sorry about that.

Personally I would remove that option altogether and let the packages go to the default location. They can be swept up into artifacts on Appveyor like so: https://github.com/adamralph/bullseye/blob/5b1c3d84be5c7daf8cbddb2ee8c164d629b67947/appveyor.yml#L19

thomaslevesque:
> Personally I would remove that option altogether and let the packages go to the default location

I think I prefer to have them all in the same place. @blairconrad, what do you think?

blairconrad:
I prefer them in one place. They're easier for computers and humans to find. While AppVeyor can be configured to sweep them up, and so can FakeItEasy's "sweep up artifacts on build failure" script, having them all appear under `artifacts/outputs` makes another behaviour better: I only have to configure _one_ custom NuGet source directory and I can consumer all my locally-built nupkgs.

blairconrad:
> Oops, sorry about that.

It's all good. No harm done, and all three of us failed to notice the change in behaviour.

thomaslevesque:
Thanks @blairconrad

thomaslevesque:

PR #1524, opened 2018-12-05
update Bullseye to 2.3.0-beta.3

adamralph:
Main features:

### Before

Returns after a two second delay while three entries are written to the OS event logs.

### After

Returns immediately with nothing written to the OS event logs.


thomaslevesque:
Thanks @adamralph!

PR #1525, opened 2018-12-05
Remove binary serialization

### LINKED ISSUE #1500: Drop fake serialization support in 5.0.0
#1499 revealed that we cannot serialize fakes once they've been configured.
Also, if any call has been made to them and we serialize and deserialize the fake, the previously-made calls are lost. (Also note that some calls, such as a property set or get, implicitly configure the fake, so even that last statement is optimistic.)

I don't think there's any value in making the fakes binary serializable and can't imagine that anyone's using the ability in any meaningful way.

blairconrad:
Fixes #1500.


blairconrad:
I split into two commits because I wasn't sure we wanted to drop serialization for the exceptions. I'm inclined to, unless someone thinks it's a bad idea. And if we go for it, I'm happy to squash the commits. Or whatever.

tests/FakeItEasy.Specs/ThrowsAsyncSpecs.cs
```diff
@@ -304,7 +304,6 @@ public static void WithResultPassExceptionFactoryWithFourArgs(IFoo fake, Task<in
                 .x(() => task.Exception?.InnerException.Should().BeAnExceptionOfType<MyException>());
         }
 
-        [SuppressMessage("Microsoft.Usage", "CA2237:MarkISerializableTypesWithSerializable", Justification = "Irrelevant for test purposes, and not compatible with .NET Core")]
```
thomaslevesque: Any idea why we no longer need this warning?

blairconrad: Great question. It's currently superfluous in `develop`.

My guess is that it's because the rule wasn't ported to the Roslyn-based FxCop: https://github.com/dotnet/roslyn-analyzers/issues/641.
If we make the other exceptions non-serializable, we'll get hit by the error all over the place once they port it, I expect. I can leave it if you prefer, or just address it if

1. we decide to make exceptions non-serializable, and
2. the rule is eventually ported and we pick up the new version of the analyzer

blairconrad:
> Thanks @blairconrad. I'm a bit torn about whether or not to keep exceptions serializable, […] but I can't think of any scenario where FakeItEasy exceptions would need to be marshalled across AppDomains...

Yup. Me too. The reason you cited was a big part of why I did this in two commits. I don't think removing the serializability of the exceptions will cause problems, but if you think it's too risky, I can drop the commit.

OR we commit, and then when someone screams, we revert it as a "bug fix".

blairconrad:
Can do. FWIW, Moq's exceptions aren't. Nor NSubstitute's, so if we drop it, at least we won't be alone!

blairconrad:
Heck, [BullseyeException](https://github.com/adamralph/bullseye/blob/master/Bullseye/Internal/BullseyeException.cs) isn't even serializable! :astonished:

thomaslevesque:
> FWIW, Moq's exceptions aren't. Nor NSubstitute's


Good enough for me!
I'd rather keep the commits separate, though. If you agree, I'll merge as is.

blairconrad:
I love it. Thank you.

thomaslevesque:
Thanks @blairconrad!

PR #1527, opened 2018-12-07
Warn that delegates can't call base method at configuration time

### LINKED ISSUE #1492: Warn that delegates can not be configured to call base method at configuration time, not invocation
This passes, but it'd be better to fail on ~~"When I invoke it"~~ "And I configure it to call its base method".

```c#
[Scenario]
public static void CannotBeConfiguredToCallBaseMethod(Action fake, Exception exception)
{
    "Given a fake delegate"
        .x(() => fake = A.Fake<Action>());

    "And I configure it to call its base method"
        .x(() => A.CallTo(() => fake.Invoke()).CallsBaseMethod());

    "When I invoke it"
        .x(() => exception = Record.Exception(() => fake.Invoke()));

    "Then it throws"
        .x(() => exception.Should()
            .BeAnExceptionOfType<NotSupportedException>()
            .WithMessage("Can not configure a delegate proxy to call base method."));
}
```

thomaslevesque:
Fixes #1492

### DIFF
--- a/src/FakeItEasy/Configuration/RuleBuilder.cs
+++ b/src/FakeItEasy/Configuration/RuleBuilder.cs
@@ -3,6 +3,9 @@ namespace FakeItEasy.Configuration
     using System;
     using System.Collections.Generic;
     using System.Linq.Expressions;
+#if FEATURE_NETCORE_REFLECTION
+    using System.Reflection;
+#endif
     using FakeItEasy.Core;
 
     internal class RuleBuilder
@@ -123,6 +126,11 @@ namespace FakeItEasy.Configuration
 
         public virtual IAfterCallConfiguredConfiguration<IVoidConfiguration> CallsBaseMethod()
         {
+            if (this.manager.FakeObjectType.GetTypeInfo().IsSubclassOf(typeof(Delegate)))
+            {
+                throw new FakeConfigurationException(ExceptionMessages.DelegateCannotCallBaseMethod);
+            }
+
             this.AddRuleIfNeeded();
             this.RuleBeingBuilt.UseApplicator(x => { });
             this.RuleBeingBuilt.CallBaseMethod = true;
--- a/src/FakeItEasy/ExceptionMessages.cs
+++ b/src/FakeItEasy/ExceptionMessages.cs
@@ -28,6 +28,8 @@ namespace FakeItEasy
 
         public static string ArgumentConstraintCannotBeNestedInArgument => "An argument constraint, such as That, Ignored, or _, cannot be nested in an argument.";
 
+        public static string DelegateCannotCallBaseMethod => "Can not configure a delegate proxy to call a base method.";
+
         public static string WrongConstructorExpressionType(Type actualConstructorType, Type expectedConstructorType) =>
             $"Supplied constructor is for type {actualConstructorType}, but must be for {expectedConstructorType}.";
 
--- a/tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
+++ b/tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
@@ -23,7 +23,7 @@ namespace FakeItEasy.Tests.Configuration
             this.asserter = A.Fake<IFakeAsserter>();
             this.ruleProducedByFactory = A.Fake<BuildableCallRule>();
 
-            this.fakeManager = A.Fake<FakeManager>(o => o.CallsBaseMethods());
+            this.fakeManager = new FakeManager(typeof(object), new object());
 
             this.builder = this.CreateBuilder();
         }
--- a/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
+++ b/src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
@@ -219,7 +219,7 @@ namespace FakeItEasy.Creation.DelegateProxies
 
             public override void CallBaseMethod()
             {
-                throw new NotSupportedException("Can not configure a delegate proxy to call base method.");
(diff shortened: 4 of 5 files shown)


src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -123,6 +126,11 @@ public virtual IVoidConfiguration Invokes(Action<IFakeObjectCall> action)
 
         public virtual IAfterCallConfiguredConfiguration<IVoidConfiguration> CallsBaseMethod()
         {
+            if (this.manager.FakeObjectType.GetTypeInfo().IsSubclassOf(typeof(Delegate)))
+            {
+                throw new FakeConfigurationException(ExceptionMessages.DelegateCannotCallBaseMethod);
+            }
```
blairconrad: Is there a reason you chose this location for the check? I would've picked the `BuildableCallRule.CallBaseMethod` setter. Both classes are internal, and the property isn't implementing any publicly visible member, so I think it makes no difference, but I'm curious.

src/FakeItEasy/ExceptionMessages.cs
```diff
@@ -28,6 +28,8 @@ internal static class ExceptionMessages
 
         public static string ArgumentConstraintCannotBeNestedInArgument => "An argument constraint, such as That, Ignored, or _, cannot be nested in an argument.";
 
+        public static string DelegateCannotCallBaseMethod => "Can not configure a delegate proxy to call base method.";
```
blairconrad: I know this is the wording we've had up to now, but now that I'm paying attention to it, it hits the ear funny. How do you feel about "Can not configure a delegate proxy to call base method**s**."?

blairconrad:
Thanks, @thomaslevesque. Generally I like it. I've made one question and one hint at a change that I _disguised_ as a question. Also, a late breaking series of comments and questions.

`DelegateProxyCallBaseMethod` is probably technically useless now. I was going to suggest removing it, but I guess we still need some kind of implementation. However, I wonder

1. should it too throw a `FakeConfigurationException`? I know we won't be encountering the problem during configuration, but it's definitely due to an illegal configuration. I am not hugely worried about this, but it popped into my mind.
2. regardless, I think it should use the `ExceptionMessages.DelegateCannotCallBaseMethod` constant for the message

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: It seemed the most natural place at the time. Why do you think the property setter would be better?

src/FakeItEasy/ExceptionMessages.cs
thomaslevesque: Why plural? It's used on a specific call, so it would only call one base method (if it made sense for a delegate). Perhaps you're confusing with `FakeOptionsExtensions.CallsBaseMethods`?

src/FakeItEasy/Configuration/RuleBuilder.cs
blairconrad: I'm not sure I do think it! I'm genuinely just asking a question. I started from the existing exception and walked up the stack. I suspect you started at the test's call site and walked down, and we just arrived at different places. I guess it's possible we later add a new rule that allows base-calling, and it bypasses your check, but I really doubt it's going to happen. The current location is fine. It was the wording of the exception that I thought would have a bigger impact.

thomaslevesque:
> `DelegateProxyCallBaseMethod` is probably technically useless now. I was going to suggest removing it, but I guess we still need some kind of implementation.

Yes, it'd be pretty hard to remove it

>     1. should it too throw a `FakeConfigurationException`? I know we won't be encountering the problem during configuration, but it's definitely due to an illegal configuration. I am not hugely worried about this, but it popped into my mind.

For consistency, yes, it would probably make sense. But it's not supposed to be called anyway.

>     2. regardless, I think it should use the `ExceptionMessages.DelegateCannotCallBaseMethod` constant for the message

Sure, will do

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: Also, in `BuildableCallRule` we don't know the type of the fake...

src/FakeItEasy/ExceptionMessages.cs
blairconrad: I wasn't confusing the calls. I'm going to say "grammar". I would think it would sound funny in French too. In English, I'd say either "it can't call **a** base method" or (equivalently) "it can't call base method**s**". I'd be content with either phrasing, unless you are stuck on the current one.

thomaslevesque: > "it can't call **a** base method"

👍

tests/FakeItEasy.Tests/Configuration/RuleBuilderTests.cs
```diff
@@ -23,7 +23,7 @@ public RuleBuilderTests()
             this.asserter = A.Fake<IFakeAsserter>();
             this.ruleProducedByFactory = A.Fake<BuildableCallRule>();
 
-            this.fakeManager = A.Fake<FakeManager>(o => o.CallsBaseMethods());
+            this.fakeManager = new FakeManager(typeof(object), new object());
```
thomaslevesque: In case you were wondering about this: this is because if we fake the `FakeManager`, `FakeObjectType` returns a fake type on which `GetTypeInfo()` throws an exception. We don't actually need it to be a fake, since we don't configure any call on it.

blairconrad: I was mildly curious, but you had me at "one less fake".

thomaslevesque:
Rebased

PR #1529, opened 2018-12-07
show messages only for build exceptions

adamralph:
Since each target is a one-liner which calls SimpleExec, the stack traces are useless. This change this:

image

to this:

image

### DIFF
--- a/tools/Directory.Build.props
+++ b/tools/Directory.Build.props
@@ -4,7 +4,7 @@
 
     <PropertyGroup>
         <NoWarn>$(NoWarn),SA0001</NoWarn>
-        <SimpleExecVersion>4.0.0-beta.1</SimpleExecVersion>
+        <SimpleExecVersion>4.1.0</SimpleExecVersion>
         <NuGetVersion>4.1.0</NuGetVersion>
         <OctokitVersion>0.31.0</OctokitVersion>
     </PropertyGroup>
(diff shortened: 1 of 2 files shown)


blairconrad:
I very much like the effect, @adamralph. I'm glad _you_ brought the PR though, since I find the new syntax to be a little opaque. But then again, where was I to comment when you were developing it?

adamralph:
Rebased.

It's only a beta feature so I'm still open to suggestions. 😃 

BTW - there is another overload which takes a `IEnumerable<Type>` for specifying an arbitrary number of exception types:

```C#
RunTargetsAndExit(args, exceptionMessageOnly: new[] { typeof(CommandException) });
```

It has the advantage of being able to write `exceptionMessageOnly:` which, come to think of it, could be reduced to `messageOnly:`. But as you can see it's quite verbose because of `typeof` and again, symbol-heavy because of the lack of a nice array initializer in C#. The `<T>`, `<T1, T2>`, `<T1, T2, T3>`, and `<T1, T2, T3, T4>` methods are just syntactic sugar over this method.

thomaslevesque:
Maybe a predicate parameter would make things clearer

adamralph:
@thomaslevesque I assume you mean like this?

```c#
RunTargetsAndExit(args, messageOnly: ex => ex is CommandException);
```

I really like that! I'll sleep on it, but I think that may be a much better way to do it.

adamralph:
WIP PR: https://github.com/adamralph/bullseye/pull/182

thomaslevesque:
> @thomaslevesque I assume you mean like this?

Yes!

thomaslevesque:
Thanks @adamralph!

Issue #1487, opened 2018-11-05
Drop GitFlow


blairconrad:
@thomaslevesque said it best in https://github.com/FakeItEasy/FakeItEasy/pull/1484#issuecomment-435693538:

> Well, I'm not married to GitFlow, so why not.

We don't make use of its power, and it comes with some overhead. Also, if we want to use a simpler versioner (specifically [MinVer](https://github.com/adamralph/min-ver)), it's standing in the way a bit.

I'm thinking that this work is best done immediately following a release. Off the top of my head, we may need to 

- [x] update FakeItEasy.PrepareRelease
- [x] update [CONTRIBUTING.md](https://github.com/FakeItEasy/FakeItEasy/blob/develop/CONTRIBUTING.md)
- [x] update [README.md](https://github.com/FakeItEasy/FakeItEasy/blob/develop/README.md)
- [x] update protected branch configuration
- [x] [merge develop into master one last time](https://github.com/FakeItEasy/FakeItEasy/pull/1532)
- set the default branch to master in 
    - [x] [project settings](https://github.com/FakeItEasy/FakeItEasy/settings/branches)
    - [x] [Appveyor settings](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/settings)
    - [x] [Read The Docs advanced configuration](https://readthedocs.org/dashboard/fakeiteasy/advanced/)
- [x] update [managing pull requests](https://github.com/FakeItEasy/FakeItEasy/wiki/Managing-pull-requests)
- ~~update FakeItEasy.Deploy~~
- [x] fix up the [next release checklist issue](https://github.com/FakeItEasy/FakeItEasy/issues/1462)
- [x] delete the `develop` branch
- ~~release an alpha version~~
- [ ] remove MinVerMajorMinor from `src/Directory.Build.props`
- [x] delete the `develop` branch

blairconrad:
> I'm thinking that this work is best done immediately following a release

I'm rethinking it. Why not do mid-release? We could just fast-forward master to develop and then do all the things in the checklist, no?

blairconrad:
I was holding off while we had a flurry of competing commits, but I'm keen to work on this now. Any objection? Testing will probably mean releasing an alpha.

blairconrad:
Actually, we wouldn't need to release. We could stop the process after the GitHub bookkeeping is done.

thomaslevesque:
No objection from me.
[...]

PR #1530, opened 2018-12-07
Merge develop into master so we can switch to GitHub flow

blairconrad:
Connects to #1487.  I thought we could fast-forward master, but it has a lot of merge commits that develop doesn't, so instead let's merge develop in and then decommission it.


blairconrad:
In retrospect, I didn't need to close this, since the list of commits auto-updated. I didn't see it at first, but I might'be been a little impatient. Oh well. Might as well keep it closed.

PR #1532, opened 2018-12-08
Merge develop into master so we can switch to GitHub flow

blairconrad:
Connects to #1487.  I thought we could fast-forward master, but it has a lot of merge commits that develop doesn't, so instead let's merge develop in and then decommission it.
Supersedes #1530.


thomaslevesque:
Thanks @blairconrad. Farewell, `develop`!

PR #1541, opened 2018-12-11
Compare issues in all releases against milestone

### LINKED ISSUE #1538: prepare_release should compare milestone issues against all pre-release release issues
Since we keep a milestone open until a proper release is done, but we create new pre-release GitHub releases, the list of issues in the one will drift relative the other. In particular, the second pre-release GitHub release for a milestone (say 5.0.0-alpha.**2**) will reference a few issues, and when we call `prepare_release`, it will think we're missing all kinds of issues.

If possible, we should look for all the other pre-release GitHub releases with the same root and include those issues in the comparison against the milestone.

blairconrad:
Fixes #1538.


blairconrad:
> it only checks other pre-releases when preparing a pre-release, right?

That's the intention! Line 67 should take care of it. Let me quick test again.

Also, it appears I need to rebase anyhow!

blairconrad:
Confirmed. With `.\tools\prepare_release.cmd 5.0.0-beta.1`, I'm told there is one issue not in the release, and with `.\tools\prepare_release.cmd 5.0.0`, I'm told there are about a million.

blairconrad:
Thank you.

Issue #1543, opened 2018-12-13
UserCallbackException is thrown when another Exception is expected


waldfee:
I upgraded FakeItEasy from 2.3.1 to 4.9.1 and I have the following issue.

**Scenario:**
1. A Fake is configured to throw
2. Another Fake is configured to indirectly call the throwing fake
3. The exception from the first Fake is catched
4. Another exception is thrown in the catch block

**Expected behaviour:**
The exception thrown in the catch block is raised to the caller

**Observed behaviour:**
An UserCallbackException is raised to the caller

**Workarounds:**
- Unpack the UserCallbackException and assert it's InnerException
- Use a manually written Fake for the throwing interface

**This behaviour seems to have been introduced with #1364
Is this behaviour intended, and is there a way to configure it, or do I have to use one of the workarounds?**

**FakeItEasy version**: 4.9.1, also happens in 5.0.0-alpha.1
.NET Framework version: 4.7.2

Minimal repro:
```c#
[TestClass]
public class UserCallbackExceptionTest
{
	public class CatchedException : Exception
	{
	}

	public class ThrowedException : Exception
	{
	}

	public interface IThrows
	{
		void ThrowIt();
	}

	public interface IDoStuff
	{
		void Execute(Action operation);
	}

	public class UnderTest
	{
		private readonly IThrows _throws;
		private readonly IDoStuff _dostuff;

		public UnderTest(IThrows throws, IDoStuff dostuff)
		{
			_throws = throws;
			_dostuff = dostuff;
		}

		public void DoStuff()
		{
			_dostuff.Execute(DoStuffPrivately);
		}

		private void DoStuffPrivately()
		{
			try
			{
				_throws.ThrowIt();
			}
			catch (CatchedException)
			{
				throw new ThrowedException();
			}
		}
	}

	public class HandWrittenFakeDoStuff : IDoStuff
	{
		public void Execute(Action operation)
		{
			operation();
		}
	}

	[TestMethod]
	public void WorksWhenUsingHandWrittenFake()
	{
		IThrows throws = A.Fake<IThrows>();
		A.CallTo(() => throws.ThrowIt())
			.Throws(new CatchedException());

		UnderTest underTest = new UnderTest(throws, new HandWrittenFakeDoStuff());

		Assert.ThrowsException<ThrowedException>(() => underTest.DoStuff());
	}

	[TestMethod]
	public void DoesNotWorkWhenUsingFakeItEasyFake()
	{
[...]

PR #1546, opened 2018-12-13
Merge hotfix 4.9.2

blairconrad:
Specifically to include

* UserCallbackException is thrown when another Exception is expected (#1543)


blairconrad:
Hmm. We're out of date. This is probably not the best approach.

Issue #1496, opened 2018-11-16
Very explicitly state how to invoke custom calls for methods that take more than 4 arguments


blairconrad:
Inspired by [Stack Overflow question How to pass more than 4 arguments to FakeItEasy Invokes()?](https://stackoverflow.com/q/53341672/1199).

I'd thought the "more advanced scenarios" line in the [docs](https://fakeiteasy.readthedocs.io/en/stable/invoking-custom-code/) would've hinted at this, but it could be more clear.
The particular question involved `Invokes`, but we should take a look at related articles, such as [Specifying Return Values](https://fakeiteasy.readthedocs.io/en/stable/specifying-return-values/#return-values-calculated-at-call-time), which does explicitly call out the 4-argument limitation and workaround.

thomaslevesque:
This change has been released in FakeItEasy [5.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/5.0.0).

PR #1552, opened 2018-12-24
Correct documentation on return value when Invokes is used

blairconrad:
While I was looking at #1496, I noticed that the documentation for `Invokes` said that the return value would be `null` or `default` unless `Returns` was used to change it, but this is not so.


blairconrad:
Thank you!

thomaslevesque:

PR #1553, opened 2018-12-27
Remove bad slash in link to default behaviour

blairconrad:
Fixes up an error I made in #1552.
I'd previewed that change in VSCode, which wasn't bothered by the extra slash, but [on the Read The Docs site](https://fakeiteasy.readthedocs.io/en/latest/invoking-custom-code/), it's a broken link.


blairconrad:
Thanks for the merge.
In case it comes up again, the site did not automatically rebuild (maybe it would've in time), so I had to force it to and then hard-refresh the page.

thomaslevesque:

PR #1554, opened 2018-12-30
Explain how to deal with calls of more than 4 parameters

### LINKED ISSUE #1496: Very explicitly state how to invoke custom calls for methods that take more than 4 arguments

blairconrad:
Fixes #1496.


docs/invoking-custom-code.md
```diff
@@ -33,7 +33,8 @@ A.CallTo(()=>fakeShop.NumberOfSweetsSoldOn(A<DateTime>._))
  .Invokes((DateTime when) => System.Console.Out.WriteLine("showing sweet sales for " + when))
  .Returns(17);
 
-// Pass an IFakeObjectCall into the creation method for more advanced scenarios.
+// Pass an IFakeObjectCall into the creation method for more advanced scenarios,
+// including configuring methods that have more than 4 parameters.
```
thomaslevesque: Why "creation"? I think it was copied from throwing-exceptions.md. It's a callback method here.

blairconrad: Sigh. I eyeballed that on the "throwing" docs and eventually became happy and didn't even notice here! Thanks. Will just amend.

blairconrad:
I did not end up amending, as I noticed a small problem just above the one you reported.

thomaslevesque:
LGTM!

blairconrad:
Squashed, thanks.

blairconrad:
thank you

PR #1561, opened 2019-01-12
Raise internal events

### LINKED ISSUE #1560: Raising internal events silently fails
I have followed the [FakeItEasy documentation](https://bfakeiteasy.readthedocs.io/en/read-the-docs-2.0/how-to-fake-internal-types/) to be able to fake internal types, which works for most of my tests. However, when registering a callback to a faked internal event, the callback registration is silently failing, so when the event is later raised then the expected callback is mysteriously not invoked.

Here is a minimal repro:
```csharp
internal class PublicClass
{
	internal virtual event EventHandler<EventArgs> InternalEvent;

	internal virtual void InternalCallback(object sender, EventArgs e) { }
}

[TestMethod]
public void InternalEventTest()
{
	PublicClass publicClass = A.Fake<PublicClass>();
	publicClass.InternalEvent += publicClass.InternalCallback;
	publicClass.InternalEvent += Raise.With(new EventArgs());

	A.CallTo(() => publicClass.InternalCallback(A<object>._, A<EventArgs>._)).MustHaveHappened();
}
```

As written, this test fails the `MustHaveHappened()` check. However, if the internal event is changed from internal to public, then the test passes (even while leaving the class itself as internal!).

I have determined the code responsible is in FakeManager.EventRule.cs. In that file, if the call to `Type.GetEvents()` is changed to specify `BindingFlags.NonPublic`, and if the calls to `EventInfo.GetAddMethod/GetRemoveMethod` are changed to pass true for the `nonpublic` parameter, then the above test passes.

thomaslevesque:
Fixes #1560

### DIFF
--- a/tests/FakeItEasy.Specs/EventRaisingSpecs.cs
+++ b/tests/FakeItEasy.Specs/EventRaisingSpecs.cs
@@ -267,7 +267,7 @@ namespace FakeItEasy.Specs
         }
 
         [Scenario]
-        public void EventArgumentsThatDoNotExtenedEventArg()
+        public void EventArgumentsThatDoNotExtendEventArg()
         {
             "Given an event of type EventHandler with custom event arguments type that does not extend EventArgs"
                 .See(() => nameof(Fake.EventHandlerOfNonEventArgsEvent));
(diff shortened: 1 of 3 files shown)


tests/FakeItEasy.Specs/EventRaisingSpecs.cs
```diff
@@ -504,5 +504,24 @@ public void DynamicRaiseWithWrongArguments(Exception exception)
                 .x(() => exception.Message.Should().Be(
                     "The event has the signature (System.Object, FakeItEasy.Specs.CustomEventArgs), but the provided arguments have types (System.String, System.Int32, System.Boolean)."));
         }
+
+        [Scenario]
+        public void RaisingInternalEvent(ClassWithInternalEventVisibleToDynamicProxy fake, EventHandler handler)
+        {
+            "Given a fake of a class with an internal even"
```
blairconrad: even → event

thomaslevesque:
> Amend that commit, and I'll merge

Done

blairconrad:
Thanks, @thomaslevesque.

PR #1562, opened 2019-01-12
Look for pre-release checklist issues even if we're not releasing a pre-release

### LINKED ISSUE #1559: Filter out all release issues from milestone when preparing release
When releasing 5.0.0, @thomaslevesque saw that the `tools\prepare_release.cmd` command complained that the 5.0.0-alpha.1 and 5.0.0-beta.1 release checklist issues were marked as "part of the milestone, but not the release notes".

In `FakeItEasy.PrepareRelease`, we're doing an exact match on issue titles against `$"Release {release.Name}"`. Instead we should probably match against the non "-pre-release" portion of the release name.

blairconrad:
Fixes #1559.


blairconrad:
Thank you.

PR #1563, opened 2019-01-12
Suppress CS0067 warning for unused event

blairconrad:
Connects to #1561.


thomaslevesque:
Thanks @blairconrad

blairconrad:
> The event is internal, and there's an InternalsVisibleTo attribute, so the compiler shouldn't assume it isn't used.

Agreed. We have methods with a similar setup, and no warning was issued for them.

Thanks for the merge!

PR #1566, opened 2019-01-21
Add support for named fakes

### LINKED ISSUE #1488: Add support for named fakes
I just found myself in a situation where I have a collection of fakes, and an assertion fails on one of the fakes, but I don't know which. It would be handy to be able to name fakes, e.g. something like this:

```csharp
var foo1 = A.Fake<IFoo>(options => options.Named("foo1"));
var foo2 = A.Fake<IFoo>(options => options.Named("foo2"));
```

Then we could have an assertion failure message like this:

```
Assertion failed for the following call on <foo1>:
  Blah.IFoo.Bar()
Expected to find it once or more but no calls were made to the fake object.
```

The name could also be used for arguments in calls list. (e.g. `Blah.IFoo.Bar(<baz>)`).

For `A.CollectionOfFake<T>`, we might need to introduce an additional overload that takes the index into account:

```csharp
var foos = A.CollectionOfFake<IFoo>(10, (options, index) => options.Named("foo" + index));
```

thomaslevesque:
Fixes #1488

### DIFF
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -84,6 +84,8 @@ namespace FakeItEasy.Core
 
         internal LinkedList<CallRuleMetadata> AllUserRules { get; }
 
+        internal string FakeObjectDisplayName => "Faked " + this.FakeObjectType;
+
         /// <summary>
         /// Adds a call rule to the fake object.
         /// </summary>
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -36,13 +36,15 @@ namespace FakeItEasy.Core
         /// </summary>
         /// <param name="fakeObjectType">The faked type.</param>
         /// <param name="proxy">The faked proxy object.</param>
-        internal FakeManager(Type fakeObjectType, object proxy)
+        /// <param name="fakeObjectName">The name of the fake object.</param>
+        internal FakeManager(Type fakeObjectType, object proxy, string fakeObjectName)
         {
             Guard.AgainstNull(fakeObjectType, nameof(fakeObjectType));
             Guard.AgainstNull(proxy, nameof(proxy));
 
             this.objectReference = new WeakReference(proxy);
             this.FakeObjectType = fakeObjectType;
+            this.FakeObjectName = fakeObjectName;
 
             this.preUserRules = new IFakeObjectCallRule[]
             {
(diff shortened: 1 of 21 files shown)


src/FakeItEasy/Creation/IProxyOptions.cs
```diff
@@ -14,5 +14,7 @@ internal interface IProxyOptions
         IEnumerable<Action<object>> ProxyConfigurationActions { get; }
 
         IEnumerable<Expression<Func<Attribute>>> Attributes { get; }
+
+        string Name { get; }
```
thomaslevesque: I'm not too happy with this, because `Name` isn't really an option of the proxy... However the alternative would have been quite painful, requiring the addition of new parameters in zillions of places. And strictly speaking, `ProxyConfigurationActions` is also not really an option of the proxy (i.e. it's not used for creating the proxy with Castle Core). Maybe we should just rename `IProxyOptions` to `IFakeCreationOptions`?

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -74,6 +77,11 @@ internal FakeManager(Type fakeObjectType, object proxy)
         /// </summary>
         public virtual Type FakeObjectType { get; }
 
+        /// <summary>
+        /// Gets the name of the fake.
+        /// </summary>
+        public string FakeObjectName { get; }
```
thomaslevesque: This isn't really necessary, but I see no harm in it, and some people might find it useful.

thomaslevesque:
> If you'd like to rename `IProxyOptions` in this PR, I'm open to it. If you prefer to leave for later (never?), that's fine too.

I think I'll do it in another PR

> One thing I do care about, though:

Doh! I *always* forget the docs. Will do.

thomaslevesque:
Oh, wait. I just realized it doesn't work for faked delegates. Obviously we can't do anything about `ToString()` for delegates, but we should use the name in call descriptions.

thomaslevesque:
I had to make a small functional changes to accommodate faked delegates: they will now be displayed in call descriptions as `Faked System.Action` (or the fake name if it's specified) instead of `System.Action`. This is more consistent with non-delegate fakes, but unfortunately it doesn't apply to `ToString()`, because we can't configure `ToString()` for a delegate... so `fakedDelegate.ToString()` will continue to return `System.Action`.

blairconrad:
Thanks, @thomaslevesque. I like all the changes. The argument description/`ToString` disparity for delegates is disappointing, but you've done the best that could be hoped for. I especially enjoyed the docs around that foible.
Squash it up!

thomaslevesque:
Thanks, squashed!

blairconrad:
Thanks, @thomaslevesque!

PR #1568, opened 2019-01-23
Upgrade ILMerge to 3.0.21

### LINKED ISSUE #1567: Upgrade to ILMerge 3.0.21
I don't think it fixes any problem we currently have (still loses SourceLink information), but:
- it no longer requires .NET 3.5
- it provides a MSBuild property with the tool path
- I'm unable to build with 2.14.1208 on my new machine, but it works with the newer version

thomaslevesque:
Fixes #1567

### DIFF
--- a/how_to_build.md
+++ b/how_to_build.md
@@ -14,8 +14,6 @@ The build requires that a few pieces of software be installed on the host comput
 
 Ensure that the following are installed:
 
-1. .NET 3.5, required by ILMerge
-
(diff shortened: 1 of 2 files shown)


thomaslevesque:
> I like the `ILMergeConsolePath`.

Yeah. A real MSBuild target would be even better!

> This means we can remove a prerequisite from [how_to_build.md](https://github.com/FakeItEasy/FakeItEasy/blob/master/how_to_build.md), no?

Ah, yes. I'll update it.

thomaslevesque:
I amended the commit

PR #1584, opened 2019-04-14
Record call before applying the best matching rule

### LINKED ISSUE #1583: Calls are recorded after applying the best rule, not when received
From StackOverflow question [FakeItEasy assert that method call to method](https://stackoverflow.com/questions/55675942/fakeiteasy-assert-that-method-call-to-method/55677429#55677429).

@baruchiro had a test that passed in 2.2.0 and failed in 5.1.0:

```c#
public class Foo
{
    public virtual void MethodA()
    {
        MethodB();
    }

    public virtual void MethodB() { }
}

…

var foo_fake = A.Fake<Foo>(options => options.CallsBaseMethods());

foo_fake.MethodA();

A.CallTo(() => foo_fake.MethodA()).MustHaveHappened()
    .Then(A.CallTo(() => foo_fake.MethodB()).MustHaveHappened());
```

The failure indicated that both calls happened, but in the wrong order.

We inadvertently changed the behaviour in #1124, released in 3.4.2, when we started recording the call in the FakeManager after applying the best-matching rule, rather than in the `CastleInvocationCallAdapter`.

We could record the call before looking for the best rule to apply. I think this is a good fix anyhow, as it more accurately reflects when the call was made, irrespective of any time FakeItEasy takes to figure out what to do with it.

blairconrad:
Fixes #1583.

### DIFF
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -159,16 +159,15 @@ namespace FakeItEasy.Core
                 listenerNode.Value.OnBeforeCallIntercepted(fakeObjectCall);
             }
 
+            var readonlyCall = fakeObjectCall.AsReadOnly();
+            this.RecordCall(readonlyCall);
+
             try
             {
                 this.ApplyBestRule(fakeObjectCall);
             }
             finally
             {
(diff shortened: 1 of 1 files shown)


blairconrad:
Thanks, @thomaslevesque!

PR #1585, opened 2019-04-14
Test that calls are recorded before applying best matching rule

blairconrad:
Tests that #1583 stays fixed.
I forgot this before.

box of shame


blairconrad:
Thank you!

PR #1588, opened 2019-04-18
Replace the term framework with library

### LINKED ISSUE #1587: Replace term framework with library in docs
Inspired by our good friends at https://github.com/nsubstitute/NSubstitute/pull/545

The reasoning is that FakeItEasy is a library because client code calls it. Frameworks call clients' code.

Rmkrs:
Fixes #1587.

Also replaced one mention of framework in the .csproj description.

### DIFF
--- a/docs/why-was-fakeiteasy-created.md
+++ b/docs/why-was-fakeiteasy-created.md
@@ -3,7 +3,7 @@
 #Introduction
 
 There was a good question on Stack Overflow that asks what
-distinguishes FakeItEasy from other frameworks. Creator of FakeItEasy,
+distinguishes FakeItEasy from other libraries. Creator of FakeItEasy,
 Patrik H&auml;gne, answered the question there but we reproduce the
 answer here. Note that the text has been preserved, and particular
 constructs referenced (such as `DummyDefinitions`) have changed or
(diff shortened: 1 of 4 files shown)


docs/why-was-fakeiteasy-created.md
```diff
@@ -15,7 +15,7 @@ The question on Stack Overflow:
 #Patrik H&auml;gne's answer
 
 To be clear, I created FakeItEasy so I'll definitely not say whether
-one framework is better than the other, what I can do is point out
```
blairconrad: I hesitate to change this word, and the one on line 48, since this text actually quotes M. Hägne's answer on StackOverflow.

blairconrad:
Thanks, @Rmkrs. We appreciate the contribution!

I will do a proper review later, but I did make one quick comment (and @thomaslevesque, do chime in if you think I'm off base) about one site where I think we might leave the original wording.

docs/why-was-fakeiteasy-created.md
Rmkrs: Ah right, I can imagine. Can change those back if you like.

thomaslevesque: I agree. Better to keep Patrik's original text unchanged.

blairconrad:
@thomaslevesque said

> LGTM, except for the changes in why-was-fakeiteasy-created.md, which should be reverted.

I think the first change in that file can be kept, as it's not part of the quote.

blairconrad:
Thanks for the update, @Rmkrs. I'm taking the liberty of squashing the commits into one, since the second just partially undoes the first. I should've asked for this when I commented earlier, but I am forgetful. If I do it right, the commits will still be listed as your change, so you get the credit due you.

blairconrad:
Thanks, @Rmkrs!

PR #1607, opened 2019-08-29
Detect fake loops

### LINKED ISSUE #1603: Stack overflow exception when creating second Fake of type that takes a parameter of its own type
We are facing an issue while running the following tests:

```c#
using System;
using FakeItEasy;
using NUnit.Framework;

namespace ConsoleApp1
{
    [TestFixture]
    public class ClassTest
    {
        [SetUp]
        public void SetUp()
        {
        }

        [Test]
        public void Test1()
        {
            var fake = A.Fake<Class>();
        }

        [Test]
        public void Test2()
        {
            var fake = A.Fake<Class>();
        }
    }
    
    public class Class
    {
        public Class(Type type)
        {
        }

        public Class(Type type, Class @class)
        {
        }
    }
}
```

Only when they are executed together, the second test (Test2) fails with a Stack Overflow Exception. The first test instead has no problem. Until version 4.6.0 of FakeItEasy this sample works correctly.

blairconrad:
Fixes #1603.

### DIFF
--- a/src/FakeItEasy/Creation/DummyValueResolver.cs
+++ b/src/FakeItEasy/Creation/DummyValueResolver.cs
@@ -49,13 +49,13 @@ namespace FakeItEasy.Creation
         {
             private readonly ResolveStrategy[] strategies;
             private readonly ConcurrentDictionary<Type, ResolveStrategy> strategyCache;
-            private readonly HashSet<Type> typesCurrentlyBeingResolved;
+            private readonly LoopDetectingResolutionContext resolutionContext;
 
             public LoopDetectingResolver(ResolveStrategy[] strategies, ConcurrentDictionary<Type, ResolveStrategy> strategyCache)
             {
                 this.strategies = strategies;
                 this.strategyCache = strategyCache;
-                this.typesCurrentlyBeingResolved = new HashSet<Type>();
+                this.resolutionContext = new LoopDetectingResolutionContext();
             }
 
             public CreationResult TryResolveDummyValue(Type typeOfDummy)
@@ -65,7 +65,7 @@ namespace FakeItEasy.Creation
                     return cachedStrategy.TryCreateDummyValue(typeOfDummy, this);
                 }
 
-                if (!this.TryBeginToResolveType(typeOfDummy))
+                if (!this.resolutionContext.TryBeginToResolve(typeOfDummy))
                 {
                     return CreationResult.FailedToCreateDummy(typeOfDummy, "Recursive dependency detected. Already resolving " + typeOfDummy + '.');
                 }
@@ -73,7 +73,7 @@ namespace FakeItEasy.Creation
                 var creationResult = this.TryResolveDummyValueWithAllAvailableStrategies(typeOfDummy);
                 if (creationResult.WasSuccessful)
                 {
-                    this.OnSuccessfulResolve(typeOfDummy);
+                    this.resolutionContext.OnSuccessfulResolve(typeOfDummy);
                 }
 
                 return creationResult;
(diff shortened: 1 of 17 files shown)


blairconrad:
It probably makes sense to squash all the commits before merge, but I left them separate for now in case they help the diffing.
General approach is pretty much what you'd expect: carry a context object everywhere, holding the current set of types being resolved.
As discussed, where the `DummyValueResolver` records that we're making a Dummy of a given type, it calls a special method on the `FakeObjectCreator` that will not immediately check that we're not resolving that same type.

src/FakeItEasy/Creation/LoopDetectingResolutionContext.cs
```diff
@@ -0,0 +1,29 @@
+namespace FakeItEasy.Creation
+{
+    using System;
+    using System.Collections.Generic;
+
+    internal class LoopDetectingResolutionContext
```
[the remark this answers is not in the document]
thomaslevesque: I think it's fine

src/FakeItEasy/Fake.of.T.cs
```diff
@@ -21,7 +21,7 @@ public class Fake<T> : IHideObjectMembers where T : class
         /// </summary>
         public Fake()
         {
-            this.FakedObject = (T)FakeAndDummyManager.CreateFake(typeof(T), null);
+            this.FakedObject = (T)FakeAndDummyManager.CreateFake(typeof(T), null, new LoopDetectingResolutionContext());
```
thomaslevesque: Not related to your change, but is there a reason why we don't use the `CreateFake` method as in the other constructor?

src/FakeItEasy/Creation/FakeObjectCreator.cs
```diff
@@ -29,6 +29,30 @@ internal class FakeObjectCreator : IFakeObjectCreator, IMethodInterceptionValida
...
+                resolutionContext);
+            if (result.WasSuccessful)
+            {
+                resolutionContext.OnSuccessfulResolve(typeOfFake);
+            }
+
+            return result;
+        }
+
+        public CreationResult CreateFakeWithoutCheckingIfWeAlreadyAre(
```
thomaslevesque: The name is a bit of a mouthful... Maybe `CreateFakeWithoutLoopDetection`?

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -102,10 +102,16 @@ public ResolveByCreatingFakeStrategy(IFakeObjectCreator fakeCreator)
...
-                this.FakeCreator.CreateFake(typeOfDummy, new ProxyOptions(), resolver, resolutionContext);
+                Type typeOfDummy,
+                IDummyValueResolver resolver,
+                LoopDetectingResolutionContext resolutionContext)
+            {
+                return this.FakeCreator.CreateFakeWithoutCheckingIfWeAlreadyAre(
+                    typeOfDummy,
+                    new ProxyOptions(),
+                    resolver,
+                    resolutionContext);
```
thomaslevesque: I think I might be missing the whole point of this method. Why is it safe in this case to not check if we're already resolving `typeOfDummy`?

src/FakeItEasy/Fake.of.T.cs
thomaslevesque: We could just pass an options builder that does nothing... Anyway, it's off topic for this PR

src/FakeItEasy/Creation/FakeObjectCreator.cs
thomaslevesque: I think the doc comments are clear enough. Since it's not public anyway, I think it's OK to rely on them.

src/FakeItEasy/Creation/DummyValueResolver.cs
thomaslevesque: Ah, I see. Thanks for the explanation.

blairconrad:
Oh, thanks for the review and merge, @thomaslevesque! I'd meant to monitor the build and hoot you when it was done.

PR #1611, opened 2019-08-31
Add overloads of Invokes, ReturnsLazily, etc for methods with more than 4 parameters

### LINKED ISSUE #1606: Add overloads of Invokes, ReturnsLazily, etc for methods with more than 4 parameters
As discussed [in chat](https://gitter.im/FakeItEasy?at=5cda88550f381d0a76b314b2) a while ago.

There are currently strongly-typed helper methods for `Invokes` for up to 4 parameters:

```csharp
.Invokes((T1 a) => ...)
.Invokes((T1 a, T2 b) => ...)
.Invokes((T1 a, T2 b, T3 c) => ...)
.Invokes((T1 a, T2 b, T3 c, T4 d) => ...)
```

The same is true for a few other configuration methods:
- `WhenArgumentsMatch`
- `ReturnsLazily`
- `AssignsOutAndRefParametersLazily`
- `Throws`
- `ThrowsAsync`

4 parameters is enough is most cases, but if you add an extra parameter, the build fails with a message that's not easily understood:

> CS1593 Delegate 'Action<IFakeObjectCall>' does not take 5 arguments

Even though it's bad practice to have methods too many parameters, it's not uncommon to have 5 or 6, so I think we should consider adding more overloads to support more parameter (maybe 8?).

These helper methods are a bit tedious to write, so we should probably consider using design-time code generation (T4 templates or another appropriate tool). If possible, the code generation could happen during build; if not, the code generation will be run when needed and the generated files will be committed.

thomaslevesque:
Fixes #1606 

- [x] `Invokes`
- [x] `WhenArgumentsMatch`
- [x] `ReturnsLazily`
- [x] `AssignsOutAndRefParametersLazily`
- [x] `Throws`
- [x] `ThrowsAsync`


src/FakeItEasy/CallbackConfigurationExtensions.cs
```diff
@@ -10,112 +10,22 @@ namespace FakeItEasy
...
-    public static class CallbackConfigurationExtensions
+    public static partial class CallbackConfigurationExtensions
     {
         private const string NameOfInvokesFeature = "invokes";
 
         /// <summary>
         /// Executes the specified action when a matching call is being made. This overload can also be used to fake calls with arguments when they don't need to be accessed.
         /// </summary>
-        /// <typeparam name="TFake">The type of fake object.</typeparam>
+        /// <typeparam name="TInterface">The type of configuration interface to return.</typeparam>
```
thomaslevesque: The type parameter name and description was incorrect and misleading. I changed it to be the same as for similar methods.

blairconrad: Nice catch. I think the `returns` value is still a lie. And in the `.tt` too.

src/FakeItEasy/CallbackConfigurationExtensions.Generated.tt
```diff
@@ -0,0 +1,70 @@
...
+
+<#
+    }
+#>
+    }
+}
+<#+
+private static readonly string[] Ordinals =
+{
+    "", // zero
```
blairconrad: could be `"zeroth"`. Avoids the comment and could be helpful if it accidentally leaks out.
Overall, I really like the use of the `Ordinals` array.

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -95,5 +95,21 @@
   <ItemGroup>
     <None Include="$(OutputPath)/**/$(AssemblyName).pdb" Pack="true" PackagePath="lib" Visible="false" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="CallbackConfigurationExtensions.Generated.tt">
```
blairconrad: Having the `Generated` as part of the filename seems odd to me, given that this _isn't_ a generated file, but I trust your knowledge of the conventions.

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -95,5 +95,21 @@
   <ItemGroup>
     <None Include="$(OutputPath)/**/$(AssemblyName).pdb" Pack="true" PackagePath="lib" Visible="false" />
   </ItemGroup>
+
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
```
blairconrad: I figure this enables the template expansion. Is it worth commenting?

src/FakeItEasy/CallbackConfigurationExtensions.Generated.tt
thomaslevesque: I thought about it, but thought it wouldn't really make sense as we're not supposed to use it... I can change it anyway if you prefer.

src/FakeItEasy/CallbackConfigurationExtensions.cs
thomaslevesque: Oops. I knew I was forgetting something

src/FakeItEasy/FakeItEasy.csproj
thomaslevesque: I think it does. Visual Studio added it automatically. I don't think it's strictly necessary, but VS will probably keep adding it if we remove it. I'll add a comment.

src/FakeItEasy/FakeItEasy.csproj
thomaslevesque: I know, I'm not too happy about it either. But there's no easy way to control the output file name, it's always based on the template file name. Maybe I could name it CallbackConfigurationExtensions.StronglyTyped.tt or something similar ?

src/FakeItEasy/CallbackConfigurationExtensions.Generated.tt
blairconrad: I think I'd feel better. If it won't kill you, maybe do?
We prefer `string.Empty` to `""` anyhow, so it's not like you won't be touching this line regardless. :wink:

src/FakeItEasy/CallbackConfigurationExtensions.StronglyTyped.tt
```diff
@@ -23,7 +31,7 @@ namespace FakeItEasy
         /// Executes the specified action when a matching call is being made.
         /// </summary>
         /// <param name="configuration">The configuration that is extended.</param>
-        /// <param name="actionToInvoke">The <see cref="Action{T1}"/> to invoke.</param>
+        /// <param name="actionToInvoke">The <see cref="Action{<#= typeParamList #>}"/> to invoke.</param>
```
blairconrad: nice

src/FakeItEasy/CallbackConfigurationExtensions.StronglyTyped.tt
```diff
@@ -56,7 +64,7 @@ namespace FakeItEasy
 <#+
 private static readonly string[] Ordinals =
 {
-    "", // zero
+    "zeroth",
```
blairconrad: thanks

src/FakeItEasy/CallbackConfigurationExtensions.StronglyTyped.tt
```diff
@@ -2,6 +2,14 @@
...
 <#@ assembly name="System.Core" #>
 <#@ import namespace="System.Linq" #>
+//------------------------------------------------------------------------------
+// <auto-generated>
+//     This code was generated by a T4 template.
+//
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
```
thomaslevesque: Had to add this so that it's ignored by StyleCop. Previously the ".generated" in the file name was enough.
(I guess I *could* address the warnings, but it's a pain with T4)

src/FakeItEasy/Configuration/ExceptionThrower.StronglyTyped.tt
```diff
@@ -0,0 +1,133 @@
...
+    {
+        private const string NameOfThrowsFeature = "throws";
+
+<# GenerateMethodImplementations("IPropertySetterConfiguration"); #>
+        private partial class PropertySetterAdapter
+        {
+<# GenerateMethodImplementations("IPropertySetterConfiguration"); #>
+        }
+    }
+}
```
thomaslevesque: I know we normally put each class in its own file, but it was much easier to do it like this, and anyway it's generated code.

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
```diff
@@ -6,7 +6,9 @@ namespace FakeItEasy.Configuration
     using FakeItEasy.Core;
     using FakeItEasy.Expressions;
 
-    internal class PropertySetterConfiguration<TValue>
+#pragma warning disable SA1601 // Partial elements should be documented
+    internal partial class PropertySetterConfiguration<TValue>
+#pragma warning restore SA1601 // Partial elements should be documented
```
thomaslevesque: See https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3009. If you prefer, we can globally disable SA1601, or add a StyleCop settings file to ignore it on non public members.

src/FakeItEasy/Configuration/IExceptionThrowerConfiguration.cs
```diff
@@ -19,52 +19,6 @@ public interface IExceptionThrowerConfiguration<out TInterface> : IHideObjectMem
...
-        /// </summary>
-        /// <param name="exceptionFactory">A function that returns the exception to throw when invoked.</param>
-        /// <typeparam name="T1">Type of the first argument of the faked method call.</typeparam>
-        /// <typeparam name="T2">Type of the second argument of the faked method call.</typeparam>
-        /// <typeparam name="T3">Type of the third argument of the faked method call.</typeparam>
-        /// <typeparam name="T4">Type of the fourth argument of the faked method call.</typeparam>
-        /// <returns>Configuration object.</returns>
-        /// <exception cref="FakeConfigurationException">The signatures of the faked method and the <paramref name="exceptionFactory"/> do not match.</exception>
-        IAfterCallConfiguredConfiguration<TInterface> Throws<T1, T2, T3, T4>(Func<T1, T2, T3, T4, Exception> exceptionFactory);
-
```
thomaslevesque: These are now in the generated part of the interface

thomaslevesque:
I think I'm done, except for the tests. There are already tests for the overloads up to 4 parameters. Do you think there's value in adding tests for the new overloads? They're all generated by the same template, so if the 1-to-4 parameter overloads work, there's little reason for the others not to work.

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
blairconrad: Nice new issue. I'm happy with any of the solutions you showed or mentioned.

thomaslevesque:
> How do you feel about updating the docs?

Oops! Forgot that, as usual...

src/FakeItEasy/Configuration/PropertySetterConfiguration.cs
thomaslevesque: We already disabled [SA1600](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1600.md) globally, so I guess it would make sense to do the same for SA1601. Less clutter in the code that way.

docs/assigning-out-and-ref-parameters.md
```diff
@@ -36,15 +36,15 @@ A.CallTo(() => aFake.AMethod(anInt, ref aRef, out anOut))
 ```
 
 As shown above, the inputs to the method may be used to calculate the
-values to assign. Convenient overloads exist for methods of up to four
```
blairconrad: I would've missed this, likely. :+1:

thomaslevesque:
> [ExceptionThrowerConfigurationExtensions](https://github.com/FakeItEasy/FakeItEasy/blob/02e6f4ef3e0ae49fdd3e712dd2e523a025a94d69/src/FakeItEasy/ExceptionThrowerConfigurationExtensions.cs) has some unused parameterized methods (and `NameOfThrowsFeature`) that can be removed, if I read things aright.

I intended to remove them, I forgot... I'll add a fixup commit.

thomaslevesque:
I amended the commit

Issue #1613, opened 2019-09-02
Expose nullability annotations (and enable nullability warnings internally in production code)


blairconrad:
We'll need to support .NET Core 3.0 and C# 8.0.
I fear there may be challenges with the `PublicApiGenerator`. At least there were a few months ago when I started looking at this and eventually dropped it.

blairconrad:
I thought I'd start out by submitting a few PRs that clean up some our internal code. Nothing that requires a language or framework change at this point, but that would reduce the pain we have in the future.

thomaslevesque:
What exactly do you have in mind for nullable ref types?

blairconrad:
I wasn't anticipating a question of this type! Now I fear I do not understand the feature. I guess technically I'm more interested in non-nullable reference types.

I'd hoped to enable the nullable annotation context as well as the nullable warning context. Clients would know which arguments must not be null (and which return values won't be) and we would know the same about our own methods.

Do you think this is a bad idea? Not worthwhile?

thomaslevesque:
> Do you think this is a bad idea?

Not at all. I was just wondering if you just wanted to enable it for internal use, or actually expose the nullability annotations to package consumers (both are useful, of course).

blairconrad:
Ah, sorry. If I had to choose, the second. However.

 ![both](https://media.giphy.com/media/3o7aCRloybJlXpNjSU/giphy.gif)

thomaslevesque:
Both, yes. Anyway, I don't think we could do the first without the second (for public types)

blairconrad:
I think I've gone about as far as I can without taking advantage of the new language features. Or at least without a discussion. Next steps, as I see them, not necessarily in this order:

- [ ] enable nullable warnings and annotations. This is a breaking change, but we're already making breaking changes
- [ ] conduct a public API review. We will need to use some nullable reference types in public members 
- [ ] address warnings
- [ ] potentially remove `Guard.Against` calls for internal methods where we know we'll not have incoming `null`s

Addressing warnings will be interesting. As I type, there are about 90 warnings, although #1680 will take care of about 20 of those. There are at least 4 major classes of remaining warnings:

- legitimate places where we need to support `null`. We'll likely have to `?` those
[...]

PR #1614, opened 2019-09-02
Reduce number of nullable reference warnings in Specs

blairconrad:
Supports #1613.

Reduces nullability-related warnings in the Specs from 227 to 32. The remaining ones are probably better served via changes that involve new language features, like non-nullable typeparams or the new `?` operator.

### DIFF
--- a/tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
@@ -85,7 +85,7 @@
                 .x(() => fake = A.Fake<IFoo>());
 
             "And a fake method with 2 parameters is configured to return a value when the arguments match a predicate"
-                .x(() => A.CallTo(() => fake.Bar(0, null))
+                .x(() => A.CallTo(() => fake.Bar(0, string.Empty))
                             .WhenArgumentsMatch(args => args.Get<int>(0) % 2 == 0 && args.Get<string>(1).Length < 3)
                             .Returns(42));
 
@@ -103,7 +103,7 @@
                 .x(() => fake = A.Fake<IFoo>());
 
             "And a fake method with 2 parameters is configured to return a value when the arguments match a predicate"
-                .x(() => A.CallTo(() => fake.Bar(0, null))
+                .x(() => A.CallTo(() => fake.Bar(0, string.Empty))
                             .WhenArgumentsMatch(args => args.Get<int>(0) % 2 == 0 && args.Get<string>(1).Length < 3)
                             .Returns(42));
 
@@ -177,7 +177,7 @@
                 .x(() => fake = A.Fake<IFoo>());
 
             "And a fake method with 2 parameters is configured to return a value when the arguments match a predicate"
-                .x(() => A.CallTo(() => fake.Bar(0, null))
+                .x(() => A.CallTo(() => fake.Bar(0, string.Empty))
                             .WhenArgumentsMatch((int a, string b) => a % 2 == 0 && b.Length < 3)
                             .Returns(42));
 
@@ -195,7 +195,7 @@
                 .x(() => fake = A.Fake<IFoo>());
 
             "And a fake method with 2 parameters is configured to return a value when the arguments match a predicate"
-                .x(() => A.CallTo(() => fake.Bar(0, null))
+                .x(() => A.CallTo(() => fake.Bar(0, string.Empty))
                             .WhenArgumentsMatch((int a, string b) => a % 2 == 0 && b.Length < 3)
                             .Returns(42));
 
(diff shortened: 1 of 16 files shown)


tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
```diff
@@ -85,7 +85,7 @@ public static void WhenArgumentsMatchWith2ArgsSuccess(IFoo fake, int result)
                 .x(() => fake = A.Fake<IFoo>());
 
             "And a fake method with 2 parameters is configured to return a value when the arguments match a predicate"
-                .x(() => A.CallTo(() => fake.Bar(0, null))
+                .x(() => A.CallTo(() => fake.Bar(0, string.Empty))
```
blairconrad: I suppose technically we could use a `A.Dummy<string>()`.

tests/FakeItEasy.Specs/DomainEvents.cs
```diff
@@ -15,8 +15,6 @@ protected DomainEvent()
 
         public int ID { get; set; }
 
-        public string Name { get; set; }
```
blairconrad: Wasn't used.

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
thomaslevesque: Inside a call configuration? Not sure what that would do... `string.Empty` is fine, I think.

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
```diff
@@ -326,7 +326,7 @@ public static void StronglyTypedWhenArgumentsMatchWith4ArgsWrongSignature(IFoo f
 
             "And a fake method with 4 parameters is configured with an arguments predicate with an incompatible signature"
                 .x(() => exception = Record.Exception(
-                        () => A.CallTo(() => fake.Bar(0, null, false, null))
+                        () => A.CallTo(() => fake.Bar(0, string.Empty, false, new object()))
                             .WhenArgumentsMatch((long a, DateTime b, Type c, char d) => true)
                             .Returns(42)));
```
thomaslevesque: I'm not sure there's much value in those changes. Why not mark the parameters as nullable in the interface? In this situation we don't actually care if the values are null.

tests/FakeItEasy.Specs/CallDescriptionsInAssertionSpecs.cs
```diff
@@ -184,7 +184,7 @@ public void AssertedCallDescriptionWithCustomValueFormatter(IFoo fake, Exception
                 .x(() => fake = A.Fake<IFoo>());
 
             "And the fake has a method with a parameter that has a custom argument value formatter"
-                .See<IFoo>(foo => foo.Bar(default(HasCustomValueFormatter)));
+                .See((IFoo foo, HasCustomValueFormatter hcvf) => foo.Bar(hcvf));
```
thomaslevesque: Again, we don't actually care if a null is passed to `Bar`. And it makes things less clear, because `See<>` no longer explicitly indicates the interface type...

blairconrad: Arguably the original wasn't even that clear. It would've been better if the typeparam was a `HasCustomValueFormatter`.

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
blairconrad: Whoops. I put the comment on the wrong line. Although I'm pretty sure it would just return a `""` here.

tests/FakeItEasy.Specs/ArgumentMatchingSpecs.cs
blairconrad: Mostly because I didn't want to pull in the language change yet. If you think the time is right, I can take that tack instead.

blairconrad:
> To be honest, I see little value in enabling nullable ref types in specs or tests. In most (if not all) of the cases you fixed, we don't actually care if the values are null, since it's just test types and methods.

It's true. But I don't mind shutting up the warnings.

> On the other hand, it would be very useful in the library itself, as it could avoid bugs.

Indeed. I was just warming up.

Withdrawn.

thomaslevesque:
> It's true. But I don't mind shutting up the warnings.

Or just not enabling ref nullability in the specs and tests :wink:

blairconrad:
> Or just not enabling ref nullability in the specs and tests 

Oh, yes.  I assumed that that's how we'd go, given your preference.

Issue #1618, opened 2019-09-05
GitHub API calls fail during deployment


blairconrad:
Seen when [trying to deploy 5.2.0](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy#L385).

I suspect this was caused by #1609.

thomaslevesque:
It says "bad credentials" for GitHub. I used the token that was declared in the ["Github Releases" environment in AppVeyor](https://ci.appveyor.com/account/FakeItEasy/environment/17406/settings), maybe it was expired.

thomaslevesque:
I can confirm the GitHub token was not valid (tried from a local script), and that the script can read AppVeyor environment variables (#1619).

I generated a new GitHub token and set it in AppVeyor environment variables.

blairconrad:
Oh, nice. Thanks!

blairconrad:
Fixed, as shown by release of 5.3.0-beta.1

blairconrad:
This change has been released as part of [FakeItEasy 5.3.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/5.3.0).

PR #1619, opened 2019-09-05
[DO NOT MERGE] Test AppVeyor environment variables work as expected

thomaslevesque:
Investigating #1618


thomaslevesque:
OK, so reading environment variables from AppVeyor works

image

PR #1620, opened 2019-09-06
Pre-reduce nullable warnings

blairconrad:
Supports #1613.

I figure we can tidy up the code a little before we even turn on the annotations and internal warnings.
In my sandbox, I did activate the warnings and ran a build. Interestingly, the build fails in the tests.
But before it does, I saw 121 unique warnings about nullability. In many cases, the proper thing to do is to use a nullable reference type in place of non-nullable, but in some cases, I think we can take a different approach.

The commits in this PR do not require the new language changes and reduce the number of unique warnings from 121 to 106.
Perhaps some are silly, but probably the sillier they are, the more fun they were to write.
Don't shy away from telling to drop any or all commits!


src/FakeItEasy.Analyzer.Shared/ArgumentConstraintTypeMismatchCodeFixProvider.cs
```diff
@@ -111,38 +111,34 @@ private static async Task<Document> MakeNotNullConstraintAsync(CodeFixContext co
             var constraintNode = GetConstraintNode(diagnostic, root);
 
             // The A<T> type
-            var aType = (constraintNode as MemberAccessExpressionSyntax)?.Expression as GenericNameSyntax;
+            var aType = (GenericNameSyntax)((MemberAccessExpressionSyntax)constraintNode).Expression;
```
blairconrad: Maybe this is a bad change. Is there a case where a code fix will be provided source that doesn't match the expected form?
That is, will we ever _not_ get a `GenericNameSyntax`?

src/FakeItEasy/Creation/FakeAndDummyManager.cs
```diff
@@ -5,19 +5,15 @@ namespace FakeItEasy.Creation
...
+    internal class FakeAndDummyManager
     {
         private static readonly ConcurrentDictionary<Type, Func<ProxyOptions, IFakeOptions>> FakeOptionsFactoryCache = new ConcurrentDictionary<Type, Func<ProxyOptions, IFakeOptions>>();
 
         private readonly IFakeObjectCreator fakeCreator;
         private readonly ImplicitOptionsBuilderCatalogue implicitOptionsBuilderCatalogue;
         private readonly IDummyValueResolver dummyValueResolver;
 
-        public DefaultFakeAndDummyManager(IDummyValueResolver dummyValueResolver, IFakeObjectCreator fakeCreator, ImplicitOptionsBuilderCatalogue implicitOptionsBuilderCatalogue)
+        public FakeAndDummyManager(IDummyValueResolver dummyValueResolver, IFakeObjectCreator fakeCreator, ImplicitOptionsBuilderCatalogue implicitOptionsBuilderCatalogue)
```
blairconrad: We weren't taking advantage of the interface, and I didn't feel like adding a method to _it_ and the default implementation!

blairconrad:
Huh. I had an extra change that affected `CreationResult`. I seem to have lost it.
I'll try to recreate tomorrow, unless you've just me down by then, @thomaslevesque.

blairconrad:
Okay. Now down to 97.

src/FakeItEasy.Analyzer.Shared/ArgumentConstraintTypeMismatchCodeFixProvider.cs
```diff
@@ -111,38 +111,34 @@ private static async Task<Document> MakeNotNullConstraintAsync(CodeFixContext co
...
+            var constraintType = aType.TypeArgumentList.Arguments.FirstOrDefault();
 
-                // The A<?> type
-                var nullableAType = aType.ReplaceNode(constraintType, nullableConstraintType);
+            // The T? type
+            var nullableConstraintType = SyntaxFactory.NullableType(constraintType);
 
-                // A<T?>.That
-                var thatNode =
+            // The A<?> type
```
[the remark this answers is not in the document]
blairconrad: Happy to change.

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
```diff
@@ -317,7 +317,7 @@ public static T NullCheckedMatches<T>(this IArgumentConstraintManager<T> manager
             Guard.AgainstNull(manager, nameof(manager));
 
             return manager.Matches(
-                x => ((object)x) != null && predicate(x),
+                x => (!ReferenceEquals(x, null) && predicate(x)),
```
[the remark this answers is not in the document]
blairconrad: Oh, I like your signature change. I think we should release the first version with nullable annotations as a new major version anyhow.

I have the same sort of memory as you regarding the `!=`, but I just tried it and see no ill effects. In practice, I doubt that `T` is a value type!

Do you prefer a change to `x != null`?

src/FakeItEasy/Creation/FakeAndDummyManager.cs
blairconrad: The new `CreateFake` overload right downstairs. As you say, eliding the interface isn't strictly necessary here. I can move to a separate PR if you like.

src/FakeItEasy/Creation/FakeAndDummyManager.cs
```diff
@@ -25,6 +25,15 @@ public object CreateDummy(Type typeOfDummy, LoopDetectingResolutionContext resol
             return this.dummyValueResolver.TryResolveDummyValue(typeOfDummy, resolutionContext).Result;
         }
 
+        public object CreateFake(
+            Type typeOfFake,
+            LoopDetectingResolutionContext resolutionContext)
+        {
+            var proxyOptions = this.BuildProxyOptions(typeOfFake, null);
```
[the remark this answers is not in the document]
blairconrad: `BuildProxyOptions` doesn't mind getting a null. It's the signal for "don't even try to build up some proxy options (unless there are some implicit ones)". I can make a non-null special value to pass in. Will add a commit for you to assess.

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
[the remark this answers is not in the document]
thomaslevesque: No, `ReferenceEquals` is fine. It's just that seeing this method made me wonder why we didn't constrain `T` to be a ref type.
I think there's an unnecessary pair of parentheses, though ;)

src/FakeItEasy/Creation/FakeAndDummyManager.cs
thomaslevesque: Ah, yes. No, you can leave it there, it's fine

src/FakeItEasy/Creation/FakeAndDummyManager.cs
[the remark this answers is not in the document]
thomaslevesque: If the end goal is to reduce nullability warnings, yes, I think passing a non-null value would be better. Maybe we could have a static default options builder delegate somewhere that just does nothing.

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
```diff
@@ -317,7 +317,7 @@ public static T NullCheckedMatches<T>(this IArgumentConstraintManager<T> manager
             Guard.AgainstNull(manager, nameof(manager));
 
             return manager.Matches(
-                x => ((object)x) != null && predicate(x),
+                x => !ReferenceEquals(x, null) && predicate(x),
```
blairconrad: Parentheses gone! Thanks.

thomaslevesque: 👍

thomaslevesque:
Looks good to me. I'll merge after you rebase
Thanks!

blairconrad:
Thanks for the suggested changes! All squashed.

blairconrad:

PR #1624, opened 2019-09-12
Check for nullity using "is null" or "is object"

blairconrad:
Jared Parsons lays out the reasons in
https://twitter.com/jaredpar/status/1171478965355024384:

> ... "x is object" is either a) a null check or b) does
> not compile. In the case "x" is known to be a struct the compiler will
> emit a diagnostic.
>
> Similarly "x is null" has the same properties except it's always a
> null check vs. a non-null check


src/FakeItEasy/ServiceLocator.cs
```diff
@@ -16,10 +16,10 @@ static ServiceLocator()
         /// <typeparam name="T">Type of the service.</typeparam>
         /// <returns>An instance of the service type.</returns>
         [DebuggerStepThrough]
-        internal static T Resolve<T>()
```
blairconrad: I don't love this change (but don't really see any harm in it). The Appveyor dotnet SDK only supports matching an unconstrained typeparam against a constant at "preview" language version or higher. So making that change would also be an option, if we wanted to go there. The dotnet 3.0 preview SDK supports it at "latest" level, of course.
I suppose I could try `!(service is object)` here too.

blairconrad:
You seemed excited when I mentioned this, @thomaslevesque. Feel like making @jaredpar incrementally happier?

src/FakeItEasy/ServiceLocator.cs
thomaslevesque: You mean with the latest SDK we can do `(x is null)` even when x could be a struct? I wasn't aware of that change, and to be honest I don't think I like it...
Anyway, I think it's fine to add a `class` constraint here, since our services will always be classes.

thomaslevesque:
You'll need to squash before I can merge, though

blairconrad:

PR #1625, opened 2019-09-16
Protect out and ref parameters value producer

blairconrad:
Came out of the nullability work for #1613. I found an opportunity to remove more  nullability warnings without using the C# 8 language features. While I was looking, I noticed that the `OutAndRefParametersValueProducer` is only ever gotten from outside the class by unit tests that are covered in the specs.
I was going to remove the getter, but then we'd have a set-only property, which StyleCop would complain about.
And it seems to me that if we're going to have annoying checks and chances of failure, that code would fit better on a method than a property, so I swapped their roles.

I'll be sending another pull request to pre-reduce some nullability warnings, including the unset `OutAndRefParametersValueProducer`.


blairconrad:
Thank you!

blairconrad:

PR #1638, opened 2019-10-03
Reject multi-constraint argument constraint factories

### LINKED ISSUE #1628: Detect and reject an "argument constraint factory method" that produces two constraints
It's a weird situation to be in, but if it comes up, we should complain about it.

Specifically, this spec fails, because no exception is thrown (we just use the first constraint, not even the one that's returned from the method):

```c#
[Scenario]
public static void ConstraintFactoryThatMakesTwoConstraints(
    IHaveAnObjectParameter fake,
    Func<object> constraintFactory,
    Exception exception)
{
    "Given a fake"
        .x(() => fake = A.Fake<IHaveAnObjectParameter>());

    "And a delegate that produces two constraints"
        .x(() => constraintFactory = () =>
            {
                A<object>.That.Matches(i => i is object);
                return A<object>.That.Matches(i => i is null);
            });

    "When I try to configure a method of the fake with this delegate"
        .x(() => exception = Record.Exception(() => A.CallTo(() => fake.Bar(constraintFactory())).Returns(1)));

    "Then the call configuration throws a FakeConfigurationException"
        .x(() => exception.Should().BeAnExceptionOfType<FakeConfigurationException>());
}
```

(Of course, if we weren't using a method to create constraints, and just specifying one inline via an expression, it's only possible to create a single constraint.)

blairconrad:
Fixes #1628.

### DIFF
--- a/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
+++ b/src/FakeItEasy/Core/ArgumentConstraintTrap.cs
@@ -3,40 +3,36 @@ namespace FakeItEasy.Core
     using System;
     using System.Collections.Generic;
     using System.Diagnostics.CodeAnalysis;
+    using System.Threading;
 
     using FakeItEasy.Configuration;
 
     internal class ArgumentConstraintTrap
         : IArgumentConstraintTrapper
     {
-        [ThreadStatic]
-        private static List<IArgumentConstraint> trappedConstraints;
-
         [SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "CallTo", Justification = "It's an identifier")]
-        public static void ReportTrappedConstraint(IArgumentConstraint constraint)
-        {
-            if (trappedConstraints is null)
-            {
-                throw new InvalidOperationException("A<T>.Ignored, A<T>._, and A<T>.That can only be used in the context of a call specification with A.CallTo()");
-            }
+        private static readonly Action<IArgumentConstraint> ThrowWhenUnpreparedToTrapConstraint = constraint =>
+            throw new InvalidOperationException("A<T>.Ignored, A<T>._, and A<T>.That can only be used in the context of a call specification with A.CallTo()");
 
-            trappedConstraints.Add(constraint);
-        }
+        private static ThreadLocal<Action<IArgumentConstraint>> saveTrappedConstraintAction =
+            new ThreadLocal<Action<IArgumentConstraint>>(() => ThrowWhenUnpreparedToTrapConstraint);
+
+        public static void ReportTrappedConstraint(IArgumentConstraint constraint) =>
+            saveTrappedConstraintAction.Value.Invoke(constraint);
 
         public IArgumentConstraint TrapConstraintOrCreate(
             Action actionThatProducesConstraint,
             Func<IArgumentConstraint> defaultConstraintFactory)
         {
-            trappedConstraints = new List<IArgumentConstraint>();
-            var result = trappedConstraints;
+            var trappedConstraints = new List<IArgumentConstraint>();
 
+            saveTrappedConstraintAction.Value = trappedConstraints.Add;
             actionThatProducesConstraint.Invoke();
+            saveTrappedConstraintAction.Value = ThrowWhenUnpreparedToTrapConstraint;
 
-            trappedConstraints = null;
-
-            return result.Count == 0 ? defaultConstraintFactory.Invoke() :
-                result.Count == 1 ? result[0] :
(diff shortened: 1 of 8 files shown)


blairconrad:
Everything after f6ed4ef is refactoring, partly for fun, and partly to reduce nullability warnings while I'm in the area.
I moved a lot of the tests to the specs because their configuration was just... byzantine.

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -70,16 +70,45 @@ public interface IHaveAnObjectParameter
...
+            "Then an assertion with all the same non-params argument values should succeed"
+                .x(() => A.CallTo(() => fake.MethodWithParameterArray("foo")).MustHaveHappened());
+
+            "And an assertion with null in place of the params arguments should fail"
+                .x(() => A.CallTo(() => fake.MethodWithParameterArray("foo", (string[])null)).MustNotHaveHappened());
+
+            "And an assertion with an empty array in place of the params arguments should succeed"
+                .x(() => A.CallTo(() => fake.MethodWithParameterArray("foo", Array.Empty<string>())).MustHaveHappened());
+        }
+
```
thomaslevesque: I'm not sure I understand... Is this related to the rest of the PR in a way I'm not seeing?

blairconrad: It mirrors one of the unit tests I deleted. Or at least what I think the test was trying to achieve by returning a `null` value to the `ExpressionArgumentConstraintFactory`.

blairconrad:
Thanks for the comments, @thomaslevesque. Yup, e32bf0d was just to remove the nullability warning. Truth be told, except for the need for `.Value` when dealing with the `ThreadLocal`, I actually find the new code easier to understand, but I don't mind dropping the commit. Worst case, we can have a discussion about it later! :grin:.
Will fixup the PR later if/when I get a few spare minutes at work.

thomaslevesque:
> Truth be told, except for the need for `.Value` when dealing with the `ThreadLocal`, I actually find the new code easier to understand

I'll have another look at it, then!

src/FakeItEasy/Core/ArgumentConstraintTrap.cs
```diff
@@ -3,40 +3,36 @@ namespace FakeItEasy.Core
...
             Action actionThatProducesConstraint,
             Func<IArgumentConstraint> defaultConstraintFactory)
         {
-            trappedConstraints = new List<IArgumentConstraint>();
-            var result = trappedConstraints;
+            var trappedConstraints = new List<IArgumentConstraint>();
 
+            saveTrappedConstraintAction.Value = trappedConstraints.Add;
             actionThatProducesConstraint.Invoke();
+            saveTrappedConstraintAction.Value = ThrowWhenUnpreparedToTrapConstraint;
```
thomaslevesque: Nothing new here, but shouldn't this statement be in a finally block?

thomaslevesque:
> > Truth be told, except for the need for `.Value` when dealing with the `ThreadLocal`, I actually find the new code easier to understand
> 
> I'll have another look at it, then!

OK, I'm sold 👍

thomaslevesque:
> Will fixup the PR later if/when I get a few spare minutes at work.

Was there anything else you wanted to change? The only change I wasn't happy about was `ArgumentConstraintTrap`, but you convinced me to keep it.

blairconrad:
There was some outstanding talk in https://github.com/FakeItEasy/FakeItEasy/pull/1638#discussion_r330962741, but I take it you're okay with the test now.
Yeah, it was mostly `ArgumentConstraintTrap`. I'll update to add a `finally`.

If it suits you, I'll amend unless the change turns out to be a lot more complicated than it seems it should!

blairconrad:
Oh, and thanks for relookatting `ArgumentConstraintTrap`. I found the new words to be more intention-revealing, in addition to avoiding the nulls, but I also worried that I felt like that because I was emotionally attached to the new code.

thomaslevesque:
> There was some outstanding talk in [#1638 (comment)](https://github.com/FakeItEasy/FakeItEasy/pull/1638#discussion_r330962741), but I take it you're okay with the test now.

Ah, sorry, forgot to reply to that. Yes, it's fine.

thomaslevesque:
> If it suits you, I'll amend unless the change turns out to be a lot more complicated than it seems it should!

Amending is fine

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you!

PR #1642, opened 2019-10-03
Embed package icon

### LINKED ISSUE #1627: Embed FakeItEasy icon in package
`PackageIconUrl` is deprecated in .NET Core 3 SDK, and is replaced with `PackageIcon`, which refers to an embedded image.
https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageiconurl

We should embed the icon in the package when we switch to .NET Core 3 SDK.

Note: I'm not seeing the NU5048 warning on FIE when using the latest SDK (3.0 RC1), not sure why. I tried on another project and the warning does appear.

thomaslevesque:
Fixes #1627


blairconrad:
Lovely. Thanks!

thomaslevesque:
BTW, if you use Nuget Package Explorer to examine the package, you won't see the icon. That's because https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/820 isn't implemented yet.

Hopefully it works in the NuGet gallery, but we should probably release a beta to make sure.

PR #1643, opened 2019-10-03
Support for creating dummy tuples

### LINKED ISSUE #1637: Make Dummy ValueTuples' members Dummies, to match Tuple
Repreod in 5.2 also.

I got bitten by unexpected behavior with FakeItEasy 2.3.0 (before I updated) while I was doing some 'refactoring'. I was trying to take advantage of the new tuple language features. Unfortunately I found this not-quite-innocent-as-it-looks change actually broke my tests.

`ArmCreatableResourceFactory` below is a fake that had no behavior configured for `CreateStorageAccountAsync`, so it got the default call behavior...

WORKS: 

```c#
Tuple<StorageAccount, string> storageAccountAndId = await ArmCreatableResourceFactory.CreateStorageAccountAsync(dep, scope, resourceGroup, token, ExecCtx);
storageAccount = storageAccountAndId.Item1;
string storageAccountId = storageAccountAndId.Item2;
var x = storageAccount.AccountName;
```

DOESN'T WORK:
```c#
var (storageAccount, storageAccountId) = await ArmCreatableResourceFactory.CreateStorageAccountAsync(dep, scope, resourceGroup, token, ExecCtx);
var x = storageAccount.AccountName;
```
                
The reason it doesn't work is while the code runs, it results in different state, and when you dereference `storageAccount.AccountName`, in the second instance it is null!

The main impact of the issue is this kind of brittleness makes the code harder to refactor.

thomaslevesque:
Fixes #1637

### DIFF
--- a/docs/dummies.md
+++ b/docs/dummies.md
@@ -64,8 +64,9 @@ When FakeItEasy needs to access a Dummy of type `T`, it tries a number
 of approaches in turn, until one succeeds:
 
 1. If there's a user-supplied
-  [custom Dummy creation](custom-dummy-creation.md) mechanism for `T`,
+  [dummy factory](custom-dummy-creation.md) for `T`,
   return whatever it makes.
+1. If `T` is `String`, return an empty string.
 1. If `T` is `Task`, the returned Dummy will be an actual `Task` that
   completes immediately.
 1. If `T` is `Task<TResult>`, the returned Dummy will be an actual
@@ -76,10 +77,14 @@ of approaches in turn, until one succeeds:
   `Lazy<TValue>` whose `Value` is a Dummy of type
   `TValue`, or a default `TValue` if no Dummy can be made
   for `TValue`.
-1. If `T` is [fakeable](what-can-be-faked.md), the Dummy will be a
-  Fake `T`.
+1. If `T` is a tuple type (`Tuple<>` or `ValueTuple<>`), the Dummy will
+be a tuple whose elements are dummies, or default values when dummies
+can't be made.
+  `Activator.CreateInstance`.
 1. If `T` is a value type, the Dummy will be a `T` created via
   `Activator.CreateInstance`.
+1. If `T` is [fakeable](what-can-be-faked.md), the Dummy will be a
+  Fake `T`.
 1. If nothing above matched, then `T` is a class. Loop over all its constructors in _descending order of argument list length_.
   For each constructor, attempt to get Dummies to satisfy the argument
   list. If the Dummies can be found, use `Activator.CreateInstance` to
--- a/src/FakeItEasy/Creation/DummyValueResolver.cs
+++ b/src/FakeItEasy/Creation/DummyValueResolver.cs
@@ -28,6 +28,7 @@ namespace FakeItEasy.Creation
                     new ResolveFromDummyFactoryStrategy(dummyFactory),
                     new ResolveByCreatingTaskStrategy(),
                     new ResolveByCreatingLazyStrategy(),
+                    new ResolveByCreatingTupleStrategy(),
                     new ResolveByActivatingValueTypeStrategy(),
                     new ResolveByCreatingFakeStrategy(fakeObjectCreator),
                     new ResolveByInstantiatingClassUsingDummyValuesAsConstructorArgumentsStrategy()
@@ -188,6 +189,37 @@ namespace FakeItEasy.Creation
             {
                 return () => value;
             }
+        }
+
+        private class ResolveByCreatingTupleStrategy : ResolveStrategy
+        {
+            public override CreationResult TryCreateDummyValue(Type typeOfDummy, IDummyValueResolver resolver, LoopDetectingResolutionContext resolutionContext)
+            {
+                if (IsTuple(typeOfDummy))
+                {
+                    var argTypes = typeOfDummy.GetTypeInfo().GetGenericArguments();
+                    var args = new object[argTypes.Length];
+                    for (int i = 0; i < argTypes.Length; i++)
+                    {
+                        var argType = argTypes[i];
+                        var creationResult = resolver.TryResolveDummyValue(argType, resolutionContext);
+                        args[i] = creationResult.WasSuccessful
+                            ? creationResult.Result
+                            : argType.GetDefaultValue();
+                    }
+
+                    var dummy = Activator.CreateInstance(typeOfDummy, args);
+                    return CreationResult.SuccessfullyCreated(dummy);
+                }
+
+                return CreationResult.FailedToCreateDummy(typeOfDummy, "It is not a tuple.");
+            }
+
+            private static bool IsTuple(Type type) =>
+                type.GetTypeInfo().IsGenericType
+                && !type.GetTypeInfo().IsGenericTypeDefinition
+                && (type.FullName.StartsWith("System.Tuple`", StringComparison.Ordinal) ||
+                    type.FullName.StartsWith("System.ValueTuple`", StringComparison.Ordinal));
         }
 
         private class ResolveByInstantiatingClassUsingDummyValuesAsConstructorArgumentsStrategy : ResolveStrategy
--- a/tests/FakeItEasy.Specs/DummyCreationSpecs.cs
+++ b/tests/FakeItEasy.Specs/DummyCreationSpecs.cs
@@ -242,6 +242,7 @@ namespace FakeItEasy.Specs
     No Dummy Factory produced a result.
     It is not a Task.
     It is not a Lazy.
+    It is not a tuple.
     It is not a value type.
 
   The constructors with the following signatures were not tried:
@@ -270,6 +271,7 @@ namespace FakeItEasy.Specs
     No Dummy Factory produced a result.
     It is not a Task.
     It is not a Lazy.
+    It is not a tuple.
     It is not a value type.
     It has no public constructors.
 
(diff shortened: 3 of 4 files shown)


thomaslevesque:
The last commit is an optimization, to avoid doing reflection and using `Activator.CreateInstance` for each dummy. It does make the code harder to follow, so I can drop it if you prefer.

blairconrad:
Now that we're talking about the arguments for `Task` and `Lazy` and `ValueTuple`, I'm wondering if we _shouldn't_ special case `Tuple`. Right now we can't create a Dummy `Tuple` if we can't make Dummies for the constructor arguments, just as with any other type we'd want to Dummy or Fake.

thomaslevesque:
> Perhaps we should change the way we create the constructor arguments and take `ResolveByCreatingTaskStrategy` as our model where we use a default value if we can't make a Dummy for an argument.

You mean using `TryResolveDummyValue`? I guess I can do that. The code generation is going to get a bit hairier ;)

> And if we do, we should add `ValueTuple` to the list of special Dummy creation exceptions.

What do you mean by that?

> On that note, is there a reason to create this as a `DummyFactory` instead of a `ResolveStrategy`


Not really. A `ResolveStrategy` would probably work just as well. In fact, it might make things simpler wrt the change mentioned above.

> (I know, we do use both, and the lines are blurry.)

They are. Why are we using a dummy factory for strings, for instance?

> Regardless of what we do to make the constructor arguments, I see that we're creating a new `LoopDetectingResolutionContext` object for every constructor argument. If we're worried about performance, maybe better to just make the one (something that's taken care of for us as a `ResolveStrategy`).

Oh, I thought we had to create a new one. I'll reuse the same one, then.

> Now that we're talking about the arguments for `Task` and `Lazy` and `ValueTuple`, I'm wondering if we _shouldn't_ special case `Tuple`.

We could. At least it would Tuple and ValueTuple consistent with each other

blairconrad:
> > And if we do, we should add ValueTuple to the list of special Dummy creation exceptions.

> What do you mean by that?

Sorry. Docs. [How are the Dummies made?](https://fakeiteasy.readthedocs.io/en/stable/dummies/#how-are-the-dummies-made)

> > … `LoopDetectingResolutionContext`

> Oh, I thought we had to create a new one. I'll reuse the same one, then.

Nah. By the time one argument is resolved, the context should be empty and ready for reuse.

blairconrad:
> > (I know, we do use both, and the lines are blurry.)

> They are. Why are we using a dummy factory for strings, for instance?

Exactly. No idea. Seemed expedient? There's less incentive to move that into a resolver because we really don't need the `LoopDetectingResolutionContext` in that dummy factory, but there's no harm in moving it.

blairconrad:
> You mean using TryResolveDummyValue? I guess I can do that. The code generation is going to get a bit hairier ;)

You could just not use the code generation. Do you know how much work it saves? We're content with reflection elsewhere. Is this the start of a trend to convert all the resolver factories?

thomaslevesque:
Well, it turns out that doing it in a `ResolveStrategy` had a few benefits:
- I don't have to worry about the resolution context, it's already provided
- the dummy resolver is also provided

In fact, I didn't need to generate code for everything, just for creating a delegate for the tuple "constructor" (a quick benchmark showed it to be much faster than either `ctor.Invoke()` or `Activator.CreateInstance()`. So the overall code is much simpler.

thomaslevesque:
> Sorry. Docs. [How are the Dummies made?](https://fakeiteasy.readthedocs.io/en/stable/dummies/#how-are-the-dummies-made)

Will do.

> You could just not use the code generation. Do you know how much work it saves?

I did a micro benchmark (to take with a grain of salt, of course). Generating a delegate that calls the constructor is more than 10 times faster than calling `Activator.CreateInstance`. So I thought it was worth it.

> We're content with reflection elsewhere. Is this the start of a trend to convert all the resolver factories?

Not a trend, but if we want to improve the overall performance of FakeItEasy, there is something to be gained here, I think.

thomaslevesque:
Oh, and I forgot to mention: now we can handle both `Tuple<>` and `ValueTuple<>` with the same code!

docs/dummies.md
```diff
@@ -76,10 +76,14 @@ of approaches in turn, until one succeeds:
...
   `TValue`, or a default `TValue` if no Dummy can be made
   for `TValue`.
-1. If `T` is [fakeable](what-can-be-faked.md), the Dummy will be a
-  Fake `T`.
+1. If `T` is a tuple type (`Tuple<>` or `ValueTuple<>`), the Dummy will
+be a tuple whose elements are dummies, or default values when dummies
+can't be made.
+  `Activator.CreateInstance`.
 1. If `T` is a value type, the Dummy will be a `T` created via
   `Activator.CreateInstance`.
```
thomaslevesque: These two were out of order! We try the value type strategy before the fake strategy.

docs/dummies.md
```diff
@@ -63,8 +63,8 @@ argument values. It uses Dummies.
 When FakeItEasy needs to access a Dummy of type `T`, it tries a number
 of approaches in turn, until one succeeds:
 
-1. If there's a user-supplied
-  [custom Dummy creation](custom-dummy-creation.md) mechanism for `T`,
+1. If there's a user-supplied or built-in
+  [dummy factory](custom-dummy-creation.md) for `T`,
```
thomaslevesque: I don't know if we should mention that we have built-in dummy factories. It's more accurate like this, but maybe the user doesn't need to know. It's a separate commit, so I can easily drop it if you prefer.

thomaslevesque: Or maybe we should just add an item in the list for `String`, which is effectively a special case.

blairconrad: I wouldn't mention it. There's no user-visible consequence.

docs/dummies.md
blairconrad: Probably my fault. I reordered some of the strategies during the performance work, if I recall correctly. Thanks for correcting!

docs/dummies.md
blairconrad: Oops. I commented on another page where I didn't see the `String` comment. I would support mentioning `String` (but don't feel it's _necessary_), but don't think there's any point in mentioning build-in Dummy factories.

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -190,6 +191,58 @@ private static Func<T> CreateFunc<T>(T value)
...
         }
 
+        private class ResolveByCreatingTupleStrategy : ResolveStrategy
+        {
+            private readonly ConcurrentDictionary<Type, Func<object[], object>> tupleConstructorsCache =
+                new ConcurrentDictionary<Type, Func<object[], object>>();
+
+            public override CreationResult TryCreateDummyValue(Type typeOfDummy, IDummyValueResolver resolver, LoopDetectingResolutionContext resolutionContext)
+            {
+                if (IsTuple(typeOfDummy, "System.Tuple`") || IsTuple(typeOfDummy, "System.ValueTuple`"))
```
blairconrad: `IsTuple` feels too specific for a method that still needs to be told "System.Tuple" or "System.ValueTuple". And running this way, we're checking the generic type definition twice. I'd keep the method name and just pass in `typeOfDummy` and do the check against the two strings inside.

blairconrad:
> Generating a delegate that calls the constructor is more than 10 times faster than calling Activator.CreateInstance. So I thought it was worth it.

Even including the time required to compile to the delegate? Amazing.

> * I don't have to worry about the resolution context, it's already provided
> * the dummy resolver is also provided

Yes! Those are most of the reasons I was pushing you away from the dummy factory. Glad it worked out as well as I'd hoped.

blairconrad:
> once you squash out some of the early commits that created the Dummy Factory

Sorry. I assumed you'd want to, since they were completely cancelled out. If you don't, that's fine too.

docs/dummies.md
thomaslevesque: OK, I'll just add a few words about `String`

thomaslevesque:
> Even including the time required to compile to the delegate? Amazing.

No. Just running the already compiled delegate is faster, but of course compiling it takes a bit more time. I don't remember exactly how much. Unfortunately I didn't keep my benchmark (it was just a throw-away LINQPad query). I'll try to do it again.

thomaslevesque:
> No. Just running the already compiled delegate is faster, but of course compiling it takes a bit more time. I don't remember exactly how much. Unfortunately I didn't keep my benchmark (it was just a throw-away LINQPad query). I'll try to do it again.

OK, maybe it's not worth it after all. It takes at least 1000 dummy creations before the cost of compiling the expression is amortized. I guess I'll use `Activator.CreateInstance` then, but it makes me a little sad...

Thanks for challenging me on this!

thomaslevesque:
OK, I think I'm done.
I dropped the commits that introduced a `ValueTupleDummyStrategy`
I removed the bit about built-in dummy factories, added an item about `String`, and squashed all docs commits together
I removed all code generation and used `Activator` instead.

thomaslevesque:
> Sorry for this, but one last comment (I swear!). Your middle commit message now reads funny:
> "Use ResolveStrategy instead of dummy factory"

Oh, I thought I had renamed it. In fact, I think I edited the message in the rebase file, but it doesn't work like that, I should have used "reword"...

It's fixed now

blairconrad:
Thanks for your hard work!

>  I edited the message in the rebase file, but it doesn't work like that

I do this so so often!

thomaslevesque:
Thanks for the review!

PR #1645, opened 2019-10-04
Continue to detect illegal argument constraint creation after failed argument constraint factory invocation

### LINKED ISSUE #1644: Exception thrown from argument constraint factory prevents detection of misused A<T>._, A<T>.Ignored, or A<T>.That
If an exception is thrown from an argument constraint factory, it leaves the system in an unstable state, where `A<T>._`, `A<T>.Ignored`, or `A<T>.That` when used outside a call configuration will not be detected as an error.

blairconrad:
Fixes #1644


blairconrad:
Thank you for mentioning it! And the merge.

PR #1648, opened 2019-10-07
Stop wrapping action callback exceptions

### LINKED ISSUE #1640: Stop wrapping exceptions thrown by "action" callbacks in UserCallbackException
Since #1364, we wrap most exceptions thrown by user callbacks (`WhenArgumentsMatch`, `ReturnsLazily`, etc) in a `UserCallbackException`. This makes it easier to understand where the exception was thrown, especially for call specification callbacks (`A<T>.That.Matches(<callback>)`, `Where(<callback>`, `WhenArgumentsMatch(<callback>)`, etc).

We rolled back that behavior for `Invokes` to fix #1543 and now let the exception bubble up without wrapping it, because throwing from the `Invokes` callback is a legitimate scenario.

Now, #1635 makes it apparent that throwing from the `ReturnsLazily` callback is also a legitimate scenario, so we should also stop wrapping the exception in this case.

In fact, we should probably stop wrapping the exceptions thrown by all "action" callbacks:

- `Invokes` (already done)
- `ReturnsLazily`
- `AssignsOutAndRefParametersLazily`
- `Throws`

In the case of `Throws`, the callback is supposed to *return* an exception, not throw it; but it doesn't really matter if it throws, since we just throw the returned exception anyway, so the effect is the same in the end.

thomaslevesque:
Fixes #1640

### DIFF
--- a/src/FakeItEasy/Configuration/RuleBuilder.cs
+++ b/src/FakeItEasy/Configuration/RuleBuilder.cs
@@ -69,20 +69,7 @@ namespace FakeItEasy.Configuration
         public virtual IAfterCallConfiguredConfiguration<IVoidConfiguration> Throws(Func<IFakeObjectCall, Exception> exceptionFactory)
         {
             this.AddRuleIfNeeded();
-            this.RuleBeingBuilt.UseApplicator(call =>
-            {
-                Exception exceptionToThrow;
-                try
-                {
-                    exceptionToThrow = exceptionFactory(call);
-                }
-                catch (Exception ex) when (!(ex is FakeConfigurationException))
-                {
-                    throw new UserCallbackException(ExceptionMessages.UserCallbackThrewAnException("Exception factory"), ex);
-                }
-
-                throw exceptionToThrow;
-            });
+            this.RuleBeingBuilt.UseApplicator(call => throw exceptionFactory(call));
             return this;
         }
 
@@ -221,20 +208,7 @@ namespace FakeItEasy.Configuration
             {
                 Guard.AgainstNull(valueProducer, nameof(valueProducer));
                 this.ParentConfiguration.AddRuleIfNeeded();
-                this.ParentConfiguration.RuleBeingBuilt.UseApplicator(call =>
-                {
-                    TMember returnValue;
-                    try
-                    {
-                        returnValue = valueProducer(call);
-                    }
-                    catch (Exception ex) when (!(ex is FakeConfigurationException))
-                    {
-                        throw new UserCallbackException(ExceptionMessages.UserCallbackThrewAnException("Return value producer"), ex);
-                    }
-
-                    call.SetReturnValue(returnValue);
-                });
+                this.ParentConfiguration.RuleBeingBuilt.UseApplicator(call => call.SetReturnValue(valueProducer(call)));
                 return this;
             }
 
--- a/src/FakeItEasy/Configuration/BuildableCallRule.cs
+++ b/src/FakeItEasy/Configuration/BuildableCallRule.cs
@@ -210,14 +210,7 @@ namespace FakeItEasy.Configuration
 
             var indexes = GetIndexesOfOutAndRefParameters(fakeObjectCall);
             ICollection<object> values;
-            try
-            {
-                values = this.OutAndRefParametersValueProducer(fakeObjectCall);
-            }
-            catch (Exception ex) when (!(ex is FakeConfigurationException))
-            {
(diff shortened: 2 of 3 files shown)


src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -69,20 +69,7 @@ public IThenConfiguration<IVoidConfiguration> NumberOfTimes(int numberOfTimes)
...
-                    exceptionToThrow = exceptionFactory(call);
-                }
-                catch (Exception ex) when (!(ex is FakeConfigurationException))
-                {
-                    throw new UserCallbackException(ExceptionMessages.UserCallbackThrewAnException("Exception factory"), ex);
-                }
-
-                throw exceptionToThrow;
-            });
+            this.RuleBeingBuilt.UseApplicator(call => throw exceptionFactory(call));
```
blairconrad: I never noticed before, but throwing the return value does look weird. It's correct, but weird-looking!

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -224,15 +211,7 @@ public IAfterCallConfiguredWithOutAndRefParametersConfiguration<IReturnValueConf
...
-                    try
-                    {
-                        returnValue = valueProducer(call);
-                    }
-                    catch (Exception ex) when (!(ex is FakeConfigurationException))
-                    {
-                        throw new UserCallbackException(ExceptionMessages.UserCallbackThrewAnException("Return value producer"), ex);
-                    }
-
+                    returnValue = valueProducer(call);
```
blairconrad: nit: snug these lines up?

blairconrad:
Overall, very good. I have the most minor of comments. Something I expected the linter to complain about, but I see now that it didn't! I can withdraw, if you like.

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: What's weird about it?

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: Will do

src/FakeItEasy/Configuration/RuleBuilder.cs
blairconrad: Just your comment earlier about how we throw the thing that's returned, but if the method throws instead, we let it go and it all works out.
Maybe it's not weird. Maybe _I'm_ weird!

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -208,12 +208,7 @@ public IAfterCallConfiguredWithOutAndRefParametersConfiguration<IReturnValueConf
             {
                 Guard.AgainstNull(valueProducer, nameof(valueProducer));
                 this.ParentConfiguration.AddRuleIfNeeded();
-                this.ParentConfiguration.RuleBeingBuilt.UseApplicator(call =>
-                {
-                    TMember returnValue;
-                    returnValue = valueProducer(call);
-                    call.SetReturnValue(returnValue);
-                });
+                this.ParentConfiguration.RuleBeingBuilt.UseApplicator(call => call.SetReturnValue(valueProducer(call)));
```
thomaslevesque: Like this, or is it too much?

blairconrad: I'm content! The extra line with the uninitialized variable just bugged me. And probably would introduce a nullability warning! :grin:
I didn't care whether you initialized right away or elided the variable.

blairconrad: I was too slow to tell you I was okay with amend. Squash it and I'll merge!

thomaslevesque:
I squashed the fixup commit.

blairconrad:
You're fast!

thomaslevesque:
Too fast for AppVeyor!
GitHub needs an "auto-merge" button, to merge when all checks have complete. Azure DevOps does that, it's very handy.

blairconrad:
Exactly. I wouldn't say "no" to a "rebase with-autosquash and then merge" button ("…when all the checks are done") either!

That's a big button.

blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
Yeah, that'd be nice. But maybe too specific to our workflow. However, if we were able to make our own custom merge buttons... But maybe that's too much to ask for.

Thanks for the merge!

PR #1651, opened 2019-10-09
Prevent stack overflow when competing threads resolve same Dummy type

### LINKED ISSUE #1639: Stack overflow exception when concurrently creating two Dummies of type that takes a parameter of its own type
https://ci.appveyor.com/project/blairconrad/fakeiteasy/builds/27840892#L279

I fear this is a symptom of a poorly-fixed #1603 (what else?), but figured I'd make the specific issue referencing the build failure until we can investigate.

blairconrad:
Fixes #1639.

### DIFF
--- a/src/FakeItEasy/Creation/LoopDetectingResolutionContext.cs
+++ b/src/FakeItEasy/Creation/LoopDetectingResolutionContext.cs
@@ -9,7 +9,7 @@ namespace FakeItEasy.Creation
 
         /// <summary>
         /// Indicate that we're trying to resolve a fake or dummy <paramref name="type"/>.
-        /// Call <see cref="OnSuccessfulResolve"/> once the type has been successfully resolved,
+        /// Call <see cref="EndResolve"/> once the attempt to resolve the type is complete,
         /// so subsequent calls to this method will not detect a loop.
         /// </summary>
         /// <param name="type">The type that we're trying to resolve.</param>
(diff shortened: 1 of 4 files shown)


blairconrad:
Most of the work went into creating the quite Byzantine test!
Looking forward to your suggestions, @thomaslevesque. Meanwhile, I'll be away from computers for the rest of the day.

src/FakeItEasy/Creation/LoopDetectingResolutionContext.cs
```diff
@@ -20,10 +20,10 @@ internal class LoopDetectingResolutionContext
         public bool TryBeginToResolve(Type type) => this.typesCurrentlyBeingResolved.Add(type);
 
         /// <summary>
-        /// Indicate successful resolution of a <paramref name="type"/> object.
+        /// Indicate the end of an attempt to resolve a <paramref name="type"/> object.
         /// Subsequent calls to <see cref="TryBeginToResolve"/> will now pass.
         /// </summary>
         /// <param name="type">The type that was resolved.</param>
-        public void OnSuccessfulResolve(Type type) => this.typesCurrentlyBeingResolved.Remove(type);
+        public void EndResolve(Type type) => this.typesCurrentlyBeingResolved.Remove(type);
```
blairconrad: The changes to this file were not strictly required to fix the issue, but for a brief time, I forgot to call this after I successfully used a cached resolver strategy. Then I thought more about this and realized that, like refraining from trapping argument constraints, we should do this every time we resolve a Dummy.
Failing to remove the "in progress" marker from this container might mean that we later throw an exception indicating a recursive resolution when it's not actually happening. I think we'd need non-deterministic constructors in the classes for this to become an issue, but it's easy to guard against.

thomaslevesque: Good thinking.

tests/FakeItEasy.IntegrationTests/DummyTests.cs
```diff
@@ -58,6 +58,86 @@ public void Same_dummy_type_can_be_created_concurrently_on_different_threads()
...
+                {
+                    if (isTask1)
+                    {
+                        isTask1 = false;
+                        Argument1OfLongConstructor.Task2IsResolvingLongConstructor.Wait();
+                    }
+                }
+            }
+        }
+
```
thomaslevesque: Well, it's not as bad as I feared! I can almost understand how it works 😉

PR #1654, opened 2019-10-13
Add 'released as part of' notes to issues when releasing

### LINKED ISSUE #1652: Automatically add "this change has been released" note to issues
Every time we create a non-pre-release release, we add a note to issues (or pull requests) like this:

```
This change has been released as part of FakeItEasy 5.1.2.
```
with a link to the release notes.

It's time-consuming. We could have the release process do this for us. Humans would still need to hand-write notes to the changes that were contributed by the community, though.

blairconrad:
Fixes #1652.

### DIFF
--- a/tools/FakeItEasy.Deploy/Program.cs
+++ b/tools/FakeItEasy.Deploy/Program.cs
@@ -65,10 +65,10 @@
                 await UploadPackageToNuGetAsync(file, nugetServerUrl, nugetApiKey);
             }
 
-            var issuesInCurrentRelease = GetIssuesReferencedFromReleases(new[] { release });
+            var issueNumbersInCurrentRelease = GetIssueNumbersReferencedFromReleases(new[] { release });
             var preReleases = GetPreReleasesContributingToThisRelease(release, releases);
-            var issuesInPreReleases = GetIssuesReferencedFromReleases(preReleases);
-            var newIssueNumbers = issuesInCurrentRelease.Except(issuesInPreReleases).Select(s => Int32.Parse(s));
+            var issueNumbersInPreReleases = GetIssueNumbersReferencedFromReleases(preReleases);
+            var newIssueNumbers = issueNumbersInCurrentRelease.Except(issueNumbersInPreReleases);
 
             Console.WriteLine($"Adding 'released as part of' notes to {newIssueNumbers.Count()} issues");
             var commentText = $"This change has been released as part of [{repoName} {releaseName}](https://github.com/{repoOwner}/{repoName}/releases/tag/{releaseName}).";
--- a/tools/FakeItEasy.Deploy/Program.cs
+++ b/tools/FakeItEasy.Deploy/Program.cs
@@ -1,9 +1,11 @@
 ﻿namespace FakeItEasy.Deploy
 {
     using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.Linq;
     using System.Runtime.CompilerServices;
+    using System.Text.RegularExpressions;
     using System.Threading.Tasks;
     using FakeItEasy.Tools;
     using Octokit;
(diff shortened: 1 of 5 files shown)


blairconrad:
I went for option 3—adding a note during pre-release and trying not to repeat it later. Parsing the numbers from the previous releases was easier than my initial idea of scraping the issues themselves.

blairconrad:
If you want, you can see where I tested on https://github.com/blairconrad/FakeItEasy/releases/tag/5.5.0 and https://github.com/blairconrad/FakeItEasy/releases/tag/5.5.0-beta.1
I actually used the releases twice, so you can see notes for 5.4 and 5.5 in there. I included a PR to make sure that would get updated, and one of the issues was "previously fixed" in a beta, so it didn't get a note in the new releases.

tools/FakeItEasy.Deploy/Program.cs
```diff
@@ -63,9 +64,29 @@ public static async Task Main(string[] args)
...
 
+        private static IEnumerable<Release> GetPreReleasesContributingToThisRelease(Release release, IReadOnlyList<Release> releases)
+        {
+            if (release.Prerelease)
+            {
+                return Enumerable.Empty<Release>();
+            }
+
+            string baseName = release.Name.Split('-')[0];
+            return releases.Where(r => r.Prerelease && r.Name.StartsWith(baseName));
```
thomaslevesque: I don't think this filter is sufficient. If the release is `5.1.1`, it will match `5.1.10-beta.1`.

Of course, under normal circumstances 5.1.10-xxx wouldn't exist... But I think it wouldn't hurt to include the dash in `basename`, i.e. match only issues starting with `5.1.1-`.

tools/FakeItEasy.Deploy/Program.cs
```diff
@@ -63,9 +64,29 @@ public static async Task Main(string[] args)
                 await UploadPackageToNuGetAsync(file, nugetServerUrl, nugetApiKey);
             }
 
+            var issueNumbersInCurrentRelease = ReleaseHelpers.GetIssueNumbersReferencedFromReleases(new[] { release });
```
thomaslevesque: Maybe consider `using static ReleaseHelpers`? If you prefer to keep it explicit, it's fine by me, but I just wanted to mention the possibility.

tools/FakeItEasy.Deploy/Program.cs
blairconrad: Yeah, I thought of this last night. And was too lazy. In addition to thinking that 5.1.10-xxx wouldn't exist, even if it did, it just means we're remove issues attributed to it from the thanking mechanism for 5.1.1, which... is fine?

But I'm happy to change.

tools/FakeItEasy.Deploy/Program.cs
blairconrad: Done!

blairconrad:
Thanks for the comments, @thomaslevesque. I pushed fixups. I figured I'd wait for the rebase until I'm squashing anyhow…

tools/FakeItEasy.Deploy/Program.cs
```diff
@@ -84,8 +84,10 @@ private static IEnumerable<Release> GetPreReleasesContributingToThisRelease(Rele
                 return Enumerable.Empty<Release>();
             }
 
-            string baseName = release.Name.Split('-')[0];
-            return releases.Where(r => r.Prerelease && r.Name.StartsWith(baseName));
+            string baseName = BaseName(release);
+            return releases.Where(r => r.Prerelease && BaseName(r) == baseName);
+
+            string BaseName(Release release) => release.Name.Split('-')[0];
```
thomaslevesque: Very nice 👍

blairconrad:

PR #1665, opened 2019-10-17
Remove option to skip formatting of argument values

blairconrad:
We never set the value to `true`, and probably won't start.


blairconrad:
Thank you!

thomaslevesque:
This change has been released as part of [FakeItEasy 5.4.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/5.4.0).

PR #1666, opened 2019-10-18
Implement support for ValueTask

### LINKED ISSUE #1658: Support for ValueTask
Currently FakeItEasy has special support for `Task` and `Task<T>`:
- Dummy creation strategy: return a completed `Task`, or `Task<T>` whose result is a dummy `T`
- Special overload for configuring the return value of a method that returns `Task<T>`, where the user can just pass the task's value instead of the task itself.

We should consider supporting `ValueTask` and `ValueTask<T>` as well. But it might not be as simple as for `Task` and `Task<T>`...
- `ValueTask<T>` is only available in .NET Core (all versions) and .NET Standard 2.1
- `ValueTask` is only available in .NET Core 2.1+ and .NET Standard 2.1
- BUT there's a [NuGet package](https://www.fuget.org/packages/System.Threading.Tasks.Extensions/) supported on .NET Standard 1.0+ that exposes both types

So, if we add support for `ValueTask` only in a new .NET Standard 2.1 target, projects that target TFMs not compatible with .NET Standard 2.1 and use the package won't benefit from it. A possible approach would be to offer the `Returns` overloads for `ValueTask` in a separate package, and support dummy creation dynamically (like we did for `ValueTuple`)

Also, `ValueTask[<T>]` doesn't work exactly the same as `Task[<T>]`, there are a few subtleties. For instance, you can't await the same `ValueTask[<T>]` more than once, which will have to be taken into account for dummy creation (we'll have to return a new one for each call).

thomaslevesque:
Fixes #1658

I'm afraid it got a bit out of hand... There's much more in there than I initially intended. I'll add review comments to make things clearer.


src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -0,0 +1,62 @@
...
+    <None Include="$(OutputPath)/netstandard1.6/$(AssemblyName).pdb" Pack="true" PackagePath="lib/netstandard1.6" Visible="false" />
+    <None Include="$(OutputPath)/netstandard2.0/$(AssemblyName).pdb" Pack="true" PackagePath="lib/netstandard2.0" Visible="false" />
+  </ItemGroup>
+
+  <!-- Exclude files from netstandard2.1 folder in package -->
+  <Target Name="RemoveNetStd21Files" AfterTargets="_WalkEachTargetPerFramework">
+    <ItemGroup>
+      <_BuildOutputInPackage Remove="$(MSBuildThisFileDirectory)$(IntermediateOutputPath)netstandard2.1/$(AssemblyName).*" Pack="false" />
+    </ItemGroup>
+  </Target>
```
thomaslevesque: That's the ugly bit... I want the package to declare a netstandard2.1 TFM (so that it didn't pick the netstandard2.0 TFM on netcoreapp3.0), but I want that TFM to be empty. So I have to include it in `TargetFrameworks`, but I need to exclude its output.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -0,0 +1,62 @@
...
+
+  <!-- Include a placeholder file in netstandard2.1 folder in package -->
+
+  <Target Name="AddNetStd21Placeholder" BeforeTargets="_WalkEachTargetPerFramework">
+    <WriteLinesToFile File="$(IntermediateOutputPath)_._" Overwrite="true" Lines="" />
+    <ItemGroup>
+      <None Include="$(IntermediateOutputPath)_._" Pack="true" PackagePath="lib/netstandard2.1" Visible="false" />
+    </ItemGroup>
+  </Target>
```
thomaslevesque: This generates a placeholder file named `_._` so that the netstandard2.1 folder isn't actually empty (that would cause it to be excluded from the zip nupkg). Could have been made simpler, but I didn't want to add the placeholder file in the sources themselves.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -0,0 +1,62 @@
...
+
+  <!-- Warning to users targeting netstandard2.1 that this package is no longer required -->
+  <ItemGroup>
+    <None Include="build/FakeItEasy.Extensions.ValueTask.targets" Pack="true" PackagePath="build/netstandard2.1" />
+  </ItemGroup>
```
thomaslevesque: Same approach we used in FakeItEasy.Analyzer. It's intended for users who upgrade their project to netcoreapp3.0, to warn them that they no longer need the package.

I'm not sure that's actually necessary. Without this, they will just keep the package, which will have no effect.

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
```diff
@@ -77,7 +77,11 @@ public static class ArgumentConstraintManagerExtensions
         /// <returns>A dummy argument value.</returns>
         public static string Contains(this IArgumentConstraintManager<string> manager, string value)
         {
+#if FEATURE_STRING_CONTAINS_COMPARISONTYPE
+            return manager.NullCheckedMatches(x => x.Contains(value, StringComparison.CurrentCulture), x => x.Write("string that contains ").WriteArgumentValue(value));
+#else
             return manager.NullCheckedMatches(x => x.Contains(value), x => x.Write("string that contains ").WriteArgumentValue(value));
+#endif
```
thomaslevesque: This fixes a [CA1307](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1307) warning that appeared when I added the netstandard2.1, because new overloads of `String.Contains` with a `StringComparison` parameter have been added

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -32,10 +32,20 @@
   <!-- Release -->
 
   <PropertyGroup Condition="'$(Configuration)' == 'Release'">
-    <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\FakeItEasy.xml</DocumentationFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
thomaslevesque: Not actually related to the rest PR, but since I was touching that file anyway... I can remove it if you prefer

src/FakeItEasy/Properties/AssemblyInfo.cs
```diff
@@ -5,6 +5,7 @@
 [assembly: CLSCompliant(true)]
 
 [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
+[assembly: InternalsVisibleTo("FakeItEasy.Extensions.ValueTask, PublicKey=00240000048000009400000006020000002400005253413100040000010001002f50d82092713be482c885861c984334606da0f54c78738a3dd0862fb2dfc6080f0780132cc65d88f0f0c70af74e8a53430962395bfc1a36fab08b7a2549d387e805c13cc84acd884447ec8c4dcfb6216df720f0998380f9c906b5de8141798d64661f036d47274e6ecb76c9cde5f4cf2b521040601e44b3914fbeb9f39127f9")]
```
thomaslevesque: On one hand, I'm not very happy about that, but on the other hand, I can't think of any real drawback. FakeItEasy.Extensions.ValueTask is a first-party package, so maybe its fine if it has a privileged status compared to third-party ones? Anyway, it makes things a bit simpler and avoids some code duplication.

src/FakeItEasy/ReturnValueConfigurationExtensions.cs
```diff
@@ -117,7 +117,7 @@ public static void ReturnsNextFromSequence<T>(this IReturnValueConfiguration<Tas
             configuration.ReturnsNextFromQueue(new ConcurrentQueue<Task<T>>(values.Select(TaskHelper.FromResult)));
         }
 
-        private static void ReturnsNextFromQueue<T>(this IReturnValueConfiguration<T> configuration, ConcurrentQueue<T> queue)
+        internal static void ReturnsNextFromQueue<T>(this IReturnValueConfiguration<T> configuration, ConcurrentQueue<T> queue)
```
thomaslevesque: To make it accessible to FakeItEasy.Extensions.ValueTask. If you think it's not worth it, I can just duplicate the code and remove the `InternalsVisibleTo`

src/FakeItEasy/ValueTaskReturnValueConfigurationExtensions.StronglyTyped.tt
```diff
@@ -0,0 +1,71 @@
...
+
+                return new ValueTask<TReturnType>(valueProducer(<#= argumentList #>));
+            });
+        }
+
+<#
+    }
+#>
+    }
+}
```
thomaslevesque: Copy-pasta from ReturnValueConfigurationExtensions.StronglyTyped.tt

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
```diff
@@ -471,6 +472,95 @@ public void ValueTupleCreation((Foo foo, string s) dummy)
...
+
+            "Then it returns a task"
+                .x(() => dummy.Should().BeOfType<Task<string>>());
+
+            "And the task is completed successfully"
+                .x(() => dummy.Status.Should().Be(TaskStatus.RanToCompletion));
+
+            "And the task's result is a dummy of the value type"
+                .x(() => dummy.Result.Should().BeOfType<string>());
+        }
```
thomaslevesque: `Task` and `Lazy` dummies were covered in other specs (`TaskSpecs` and `LazySpecs`), but in an indirect way. I thought it best to cover them in this file, along with other dummy creation specs.

tests/FakeItEasy.Specs/ValueTaskReturnValueSpecs.cs
```diff
@@ -0,0 +1,166 @@
+namespace FakeItEasy.Specs
+{
+    using System;
+    using System.Threading.Tasks;
+    using FluentAssertions;
+    using Xbehave;
+
+    public class ValueTaskReturnValueSpecs
```
thomaslevesque: The `Task` equivalent was covered (at great length) in unit tests, but we tend to prefer specs now. At some point we might want to migrate the `Task` tests to specs.

tests/FakeItEasy.Specs/ValueTaskReturnValueSpecs.cs
```diff
@@ -0,0 +1,166 @@
...
+        }
+
+        [Scenario]
+        public void ValueTaskLazilyComputed(IReturnValueTask fake, Func<int> valueProducer, ValueTask<int> task)
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<IReturnValueTask>());
+
+            "And a value producer configured to return a specific value"
+                .x(() => valueProducer = A.Fake<Func<int>>(o => o.ConfigureFake(f => A.CallTo(() => f()).Returns(42))));
```
thomaslevesque: This is to be able to assert that the value producer was called. If you think it's overkill, I'll simplify it.

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
blairconrad: Feels slightly redundant, but it unlikely to cause problems. What do you think about absorbing the other specs into this?

thomaslevesque: Well, there isn't much more is the existing specs. It's just that instead of using `A.Dummy<T>()`, they call unconfigured methods returning a T. They also test the case where T is not dummyable. If you want, I can add specs to cover the non-dummyable case, and remove LazySpecs and TaskSpecs.

blairconrad: I sort of like that idea, if it's not too much trouble.

thomaslevesque: Will do

thomaslevesque:
> * I struggled with where to put this comment, so you're getting it here. I think the new package's metadata can be improved. Here's a diff against FakeItEasy:
>   image
>   
>   1. there's no title
>   2. I think authors and owners should be "Thomas Levesque, FakeItEasy contributors"
>   3. The description is not the best!

Somehow I completely overlooked the package metadata... I'll update them.

> * Before seeing the PR I was wondering what our versioning scheme was going to be, and whether we would lock the new package to various FIE versions, or release with a dependency on the lowest viable FIE. The latter allows more flexibility, but the lockstep approach has treated us well for the Analyzer, so I think you made the right choice.

To be honest, I didn't make any choice! I just didn't think about that at all. MinVer is referenced from src/Directory.Build.props, so it applies to everything inside src.
 
> * # DOCS OR IT DIDN'T HAPPEN!

I wonder if I ever will remember to update the docs without you reminding me. This time I almost didn't forget; I thought about it at some point, but then it slipped my mind....

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
thomaslevesque: Actually... I did a quick check, and it doesn't work on a project that targets multiple TFMs including netstandard2.1. The warning appears because of netstandard2.1, even though the package is necessary for the other TFMs.

I can add a condition on the target that emits the warning, like this:
```
Condition="'$(TargetFrameworks)' == '' or '$(TargetFrameworks)' == '$(TargetFramework)'"
```
It seems to work on a basic scenario, but it the project has multiple TFMs supporting netstandard2.1 (currently only netcoreapp3.0 supports it).

So, either I make that fix, which isn't foolproof, or I just remove the warning completely. What do you prefer?

thomaslevesque:
While looking at the docs, I realized I forgot something else: the default behavior for cancellation. We should return a cancelled `ValueTask` if the arguments contain a cancelled `CancellationToken`.

thomaslevesque:
Oh, and `ThrowsAsync` too!

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
```diff
@@ -52,6 +52,30 @@ public void ApproveApiNetStd21()
...
+            ApproveExtensionsValueTaskApi("netstandard1.6");
+        }
+
+        [Fact]
+        [UseReporter(typeof(DiffReporter))]
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public void ApproveExtensionsValueTaskApiNetStd20()
+        {
+            ApproveExtensionsValueTaskApi("netstandard2.0");
+        }
```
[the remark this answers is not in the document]
thomaslevesque: I did. But ApprovalTests relies on the name of the test, so I'm not sure it would work with a theory...

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -32,10 +32,20 @@
...
     <DebugType>pdbonly</DebugType>
   </PropertyGroup>
+
+  <!-- .NET Standard 2.1 -->
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.1'">
+    <DefineConstants>$(DefineConstants);FEATURE_REFLECTION_GETASSEMBLIES;FEATURE_EXCEPTION_DISPATCH_INFO;FEATURE_ARRAY_EMPTY;FEATURE_PARAMETERINFO_CUSTOMATTRIBUTES_PROPERTY</DefineConstants>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
thomaslevesque: I thought about this too, but I had a different approach in mind: centralize the list of features per framework somewhere (e.g. Directory.Build.props) so that we don't have to repeat them in every project.
But negative constants would work too, with the advantage you described.
And both approaches could be combined.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -0,0 +1,57 @@
...
+    <Compile Include="../FakeItEasy/ValueTaskReturnValueConfigurationExtensions.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\FakeItEasy\FakeItEasy.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.1'">
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.3" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
thomaslevesque: I think you're right. I used `!=`, because the alternative was to list all other TFMs. I can change it if you prefer.

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -167,5 +167,9 @@
       <DependentUpon>ReturnValueConfigurationExtensions.StronglyTyped.tt</DependentUpon>
     </Compile>
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' != 'netstandard2.1'">
+    <Compile Remove="ValueTaskReturnValueConfigurationExtensions.cs" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
thomaslevesque: Well, I think we'll detect it... But maybe we should explicitly list other TFMs

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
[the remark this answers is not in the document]
thomaslevesque: Worth a try, though!

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
blairconrad: I don't entirely understand. Partly I think due to some minor typos in your comment, and mostly because I would've been barely on the edge of understanding to begin with. I'm getting the gist that sometimes emitting a warning will not work. I say if we're at all unsure about what circumstances will properly trigger the warning, we're better off without, and we can just provide a good description/docs of the package.

src/FakeItEasy/FakeItEasy.csproj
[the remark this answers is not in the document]
blairconrad: In this case, with the `Remove`, maybe it's safer. We're unlikely to add a new TFM in the future that doesn't include this extension.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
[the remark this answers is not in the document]
blairconrad: On reflection, I'm not so worried about this package, since we're probably not going to add another TFM anyhow. I think it's fine.

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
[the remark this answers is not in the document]
blairconrad: > But ApprovalTests relies on the name of the test,

I did not know this. Or had forgotten!

blairconrad:
> I wonder if I ever will remember to update the docs without you reminding me

As I was writing my comment, I said to my wife, "@thomaslevesque will laugh at himself and comment about how he never remembers the docs". Thanks for making me look smart!

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
thomaslevesque: Sorry, there was a typo and an incomplete sentence, I fixed it (I think).

Maybe an example will help. Suppose I have a test project that targets net472 and netcoreapp3.0. If I reference FakeItEasy.Extensions.ValueTask in that project, I will see the warning, because netcoreapp3.0 doesn't need the package. But net472 _does_ need it, so we shouldn't emit the warning.

Now, with the fix I suggest:
- it will work as expected (i.e. no warning) if my test project still targets net472 and netcoreapp3.0
- BUT if I change my test project to target netcoreapp3.0 and netcoreapp3.1, it won't show the warning, even though it should (because neither netcoreapp3.0 nor netcoreapp3.1 need the package)

So, as I said, the fix isn't foolproof.

> I say if we're at all unsure about what circumstances will properly trigger the warning, we're better off without, and we can just provide a good description/docs of the package.

Sounds good to me.

src/FakeItEasy/FakeItEasy.csproj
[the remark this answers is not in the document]
thomaslevesque: 👍

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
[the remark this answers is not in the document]
thomaslevesque: OK, but I think I'm going to change it anyway, to be consistent with FakeItEasy.csproj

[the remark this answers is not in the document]
blairconrad: I'm not surprised. And prefer the consistency. I just didn't want to ask for more work.

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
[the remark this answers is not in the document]
thomaslevesque: I'll give it a try anyway. If it doesn't work, at least the `ApproveApi` method can probably be refactored to accept the project name, instead of duplicating it.

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
```diff
@@ -17,42 +17,73 @@ public class ApiApproval
...
+
+        private static void ApproveApi(string projectName, string frameworkVersion)
         {
             string codeBase = Assembly.GetExecutingAssembly().CodeBase;
             UriBuilder uri = new UriBuilder(new Uri(codeBase));
             string assemblyPath = Uri.UnescapeDataString(uri.Path);
             var containingDirectory = Path.GetDirectoryName(assemblyPath);
             var configurationName = new DirectoryInfo(containingDirectory).Parent.Name;
-            var assemblyFile = $@"..\..\..\..\..\src\FakeItEasy\bin\{configurationName}\{frameworkVersion}\FakeItEasy.dll";
+            var assemblyFile = $@"..\..\..\..\..\src\{projectName}\bin\{configurationName}\{frameworkVersion}\{projectName}.dll";
```
thomaslevesque: Note that I removed the `NoInlining` attribute from this method, because it wasn't necessary. It only matters on the test method itself, which ApprovalTests uses to infer the name of the approved/received files.

(and this method would never have been anyway, it's too long)

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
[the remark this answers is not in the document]
thomaslevesque: Got it working! ApprovalTests does *not* make it easy to control the approved/received file names... I spent at least half an hour reverse engineering it to understand how to achieve this.

I'll send a separate PR for this. This one is already big enough, and doesn't need the additional confusion.

thomaslevesque:
@blairconrad thanks for your feedback. I think I addressed everything. I left all the fixup commits to make it easier to follow the changes.

docs/faking-async-methods.md
```diff
@@ -27,3 +27,19 @@ Of course, you can still configure calls to `async` methods as you would normall
 ```csharp
 A.CallTo(() => foo.Bar()).Returns(Task.FromResult("bar"));
 ```
+
+There are also convenience overloads of `Returns` and `ReturnsLazily` that let you specify a value rather
+than a task, and configure the method to return a completed task whose result is the specified value:
+
+```csharp
+A.CallTo(() => foo.Bar()).Returns("bar");
+```
```
thomaslevesque: The convenience extension methods for `Task<T>` were not mentioned, so I added them

docs/faking-async-methods.md
```diff
@@ -27,3 +27,19 @@ Of course, you can still configure calls to `async` methods as you would normall
...
+
+These overloads of `Returns` and `ReturnsLazily` also exist for `ValueTask<T>`; however, unless you're
+using the .NET Standard 2.1 version of FakeItEasy, they're in a separate package:
+[`FakeItEasy.Extensions.ValueTask`](https://www.nuget.org/packages/FakeItEasy.Extensions.ValueTask).
+
+## Throwing exceptions
+
+To configure an async method to throw an exception, see
+[Throwing exceptions from an async method](throwing-exceptions.md#throwing-exceptions-from-an-async-method).
```
thomaslevesque: Not sure if *Throwing exceptions from an async method* belongs on this page, or on "Throwing exceptions"... So I just added a link.

docs/faking-async-methods.md
blairconrad: Nice.

docs/faking-async-methods.md
```diff
@@ -27,3 +27,19 @@ Of course, you can still configure calls to `async` methods as you would normall
...
+There are also convenience overloads of `Returns` and `ReturnsLazily` that let you specify a value rather
+than a task, and configure the method to return a completed task whose result is the specified value:
+
+```csharp
+A.CallTo(() => foo.Bar()).Returns("bar");
+```
+
+These overloads of `Returns` and `ReturnsLazily` also exist for `ValueTask<T>`; however, unless you're
+using the .NET Standard 2.1 version of FakeItEasy, they're in a separate package:
+[`FakeItEasy.Extensions.ValueTask`](https://www.nuget.org/packages/FakeItEasy.Extensions.ValueTask).
```
blairconrad: This is a good note. However, I worry that we'll forget to update it when we add a .NET Standard 3.2 package or whatever. Would saying that you need to use the new package whenever using a version of FakeItEasy that targets a framework that lacks an inbuilt `ValueTask` make sense?

tests/FakeItEasy.Specs/ThrowsAsyncSpecs.cs
```diff
@@ -48,7 +48,7 @@ public static void NoResultPassExceptionInstance(IFoo fake, Task task)
                 .x(() => task.Status.Should().Be(TaskStatus.Faulted));
 
             "And the task's exception is the configured exception"
-                .x(() => task.Exception?.InnerException.Should().BeAnExceptionOfType<MyException>());
```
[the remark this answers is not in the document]
thomaslevesque: Ah, sorry. I wanted to comment on it, but I forgot.

This code was wrong, because the `?.` short-circuits the whole expression, so `.Should().BeAnException()` would not be executed if `task.Exception` was null.

docs/faking-async-methods.md
thomaslevesque: Good point. I'll do that.

thomaslevesque: In fact, it's not so simple... .NET Core 1.0+ has `ValueTask<T>`, and .NET Core 2.1+ has `ValueTask`. But if the user's project targets .NET Core 2.1, it will pick the .NET Standard 2.0 version of FakeItEasy, which doesn't have the methods (since they're only in the .NET Standard 2.1 version). I'll try to think of an explanation that makes sense.

thomaslevesque: Done. It's not ideal, but I can't think of a better way to phrase it that is both accurate and easy to understand.

blairconrad: Lovely. Thanks.

thomaslevesque:
I'll try to squash today, but with so many fixup commits, I'm getting conflicts at every step, and in the end the resulting isn't identical. I'm afraid it's going to be a chore...

blairconrad:
Oh no! I better take a copy first, to keep you honest. 😉 
But seriously, you do have a lot of fixups targeting the same initial commit. have you considered trying to non `--autosquash` some of them into bigger fixups and then rebasing? Sometimes that can cut down on the number of times you have to resolve the same conflict.
I don't even mind if you end up with fewer "primary commits" than you have now, if it makes it easier on you. Whatever you like!

thomaslevesque:
I think I'll rebase manually, cherry-picking the main commits and manually applying the fixups

thomaslevesque:
OK, it's done.

The issues while squashing were mostly due to 2 things:
- one fixup didn't refer to the right original commit, and was applied to early
- FakeItEasy.csproj had inconsistent line endings (it's already the case on master; not sure when or how it happened)

I added a last commit to make line endings consistent in FakeItEasy.csproj

blairconrad:
Thanks, @thomaslevesque. Fantastic!

thomaslevesque:
Thanks for the merge, and the feedback!

PR #1667, opened 2019-10-18
Upgrade analyzers to .NET Standard 2.0 and test them on .NET Core

thomaslevesque:
Cherry-picked the commits from my current work on https://github.com/FakeItEasy/FakeItEasy/issues/1662, I hope it didn't depend on anything I didn't include...

PR #1668, opened 2019-10-18
Target framework sets

### LINKED ISSUE #1662: Support building only part of the target frameworks
With the multiplication of target frameworks for the library and tests, the build times have been getting longer and longer. Although all these TFMs are there for a good reason, it can significantly slow down the feedback loop when working on the project.

As discussed in chat with @blairconrad, we should consider a way to build only some TFMs.

A possible way to do this would be to define "target framework sets" that define the target frameworks for each project. The default set would be the "full" set, including all currently supported target frameworks. We could then introduce other sets with only some of the target frameworks for faster development cycle.

thomaslevesque:
Fixes #1662

To be merged after #1667


thomaslevesque:
This PR is on top of https://github.com/FakeItEasy/FakeItEasy/pull/1667 so that it builds successfully. The first 2 commits won't be part of it after https://github.com/FakeItEasy/FakeItEasy/pull/1667 is merged.

Directory.Build.props
```diff
@@ -17,4 +17,75 @@
...
+    <IntegrationTestsTargetFrameworks>netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</IntegrationTestsTargetFrameworks>
+    <UnitTestsTargetFrameworks>netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</UnitTestsTargetFrameworks>
+    <AnalyzerTestsTargetFrameworks>netcoreapp2.1;netcoreapp3.0</AnalyzerTestsTargetFrameworks>
+    <TestHelpersTargetFrameworks>netstandard1.6;netstandard2.0</TestHelpersTargetFrameworks>
+    <ApprovalTestsTargetFrameworks>netcoreapp3.0</ApprovalTestsTargetFrameworks>
+  </PropertyGroup>
+
+  <!-- .NET Framework only -->
+  <PropertyGroup Condition=" '$(TargetFrameworkSet)' == 'netfx' ">
+    <FakeItEasyTargetFrameworks>net45</FakeItEasyTargetFrameworks>
```
blairconrad: and net40 as well?

Directory.Build.props
```diff
@@ -17,4 +17,75 @@
...
+    <ExampleTargetFrameworks>net45</ExampleTargetFrameworks>
+    <SpecsTargetFrameworks>net461;netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</SpecsTargetFrameworks>
+    <IntegrationTestsTargetFrameworks>net461;netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</IntegrationTestsTargetFrameworks>
+    <UnitTestsTargetFrameworks>net461;netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</UnitTestsTargetFrameworks>
+    <AnalyzerTestsTargetFrameworks>net461;netcoreapp2.1;netcoreapp3.0</AnalyzerTestsTargetFrameworks>
+    <TestHelpersTargetFrameworks>net461;netstandard1.6;netstandard2.0</TestHelpersTargetFrameworks>
+    <ApprovalTestsTargetFrameworks>net461</ApprovalTestsTargetFrameworks>
+  </PropertyGroup>
+
+  <!-- .NET Core 3 / .NET Standard 2.0 -->
```
blairconrad: .NET Standard 2.1?

Directory.Build.props
```diff
@@ -17,4 +17,75 @@
...
+  <PropertyGroup Condition=" '$(TargetFrameworkSet)' == 'netcore3.0' ">
+    <FakeItEasyTargetFrameworks>netstandard2.1</FakeItEasyTargetFrameworks>
+    <ValueTaskExtensionsTargetFrameworks>netstandard2.1</ValueTaskExtensionsTargetFrameworks>
+    <ExampleTargetFrameworks>netcoreapp3.0</ExampleTargetFrameworks>
+    <SpecsTargetFrameworks>netcoreapp3.0</SpecsTargetFrameworks>
+    <IntegrationTestsTargetFrameworks>netcoreapp3.0</IntegrationTestsTargetFrameworks>
+    <UnitTestsTargetFrameworks>netcoreapp3.0</UnitTestsTargetFrameworks>
+    <AnalyzerTestsTargetFrameworks>netcoreapp3.0</AnalyzerTestsTargetFrameworks>
+    <TestHelpersTargetFrameworks>netstandard2.0</TestHelpersTargetFrameworks>
```
blairconrad: I'm not sure about this. See my summary.

tests/FakeItEasy.Tests.TestHelpers/FakeItEasy.Tests.TestHelpers.csproj
```diff
@@ -0,0 +1,20 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net461;netstandard1.6;netstandard2.0</TargetFrameworks>
+    <SignAssembly>true</SignAssembly>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard1.6'">
+    <DefineConstants>$(DefineConstants);FEATURE_NETCORE_REFLECTION;USE_RUNTIMELOADER;FEATURE_EXCEPTION_DISPATCH_INFO;FEATURE_ARRAY_EMPTY;FEATURE_PARAMETERINFO_CUSTOMATTRIBUTES_PROPERTY</DefineConstants>
```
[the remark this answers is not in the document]
thomaslevesque: Oh. I must have copied this from FakeItEasy.csproj... We don't use `FEATURE_ARRAY_EMPTY` in the tests anyway. I'll fix this and leave only the necessary ones.

tests/FakeItEasy.Tests.TestHelpers/ObjectAssertionsExtensions.cs
```diff
@@ -17,7 +17,10 @@ public static class ObjectAssertionsExtensions
         /// <param name="assertion">A FluentAssertions assertion that has been initiated on a subject.</param>
         public static void BeAFake(this ReferenceTypeAssertions<object, ObjectAssertions> assertion)
         {
-            Guard.AgainstNull(assertion, nameof(assertion));
+            if (assertion == null)
```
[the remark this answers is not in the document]
thomaslevesque: Yes. At this point. If you prefer, I can add a `InternalsVisibleTo` attribute to FIE

Directory.Build.props
thomaslevesque: Since the goal was to make the build faster, I thought it would make sense to include as few TFMs as possible. Is there a scenario where we would want to build all netfx targets, and nothing else?
But the target sets I defined here are mostly examples anyway. The idea is that we can add new ones as needed.

Directory.Build.props
thomaslevesque: Yes! Forgot to update after I rebased

thomaslevesque:
> In particular, the netcore3.0 set doesn't work for me

Oops. Seems broken for me too. Must have broke when I rebased. I'll look into it.

Directory.Build.props
blairconrad: Ah. I was taking "netcoreall" for comparison, without paying attention to the "all" on the end of that.
I don't mind excluding net40 from the set, but maybe in this case, rename to net45, since we're closer to a "netcore2.1" situation than a "netcoreall" situation?

tests/FakeItEasy.Tests.TestHelpers/ObjectAssertionsExtensions.cs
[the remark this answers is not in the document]
blairconrad: I don't mind either way. What you have is fine. Thanks for asking.

thomaslevesque:
In general, what do you think about this approach?
I like your idea to add a parameter to the build script.
We could also enable a user-defined props file that would be included in Directory.Build.props if it exists, and ignored by Git. e.g.

```xml
    <Import Project="$(MSBuildThisFileDirectory)/FakeItEasy.user.props" Condition="Exists('$(MSBuildThisFileDirectory)/FakeItEasy.user.props')" />
```

This way, there would be no need to temporarily edit Directory.Build.props

Directory.Build.props
thomaslevesque: Good point. I don't think a `netfxall` is very useful. I intended `netcoreall` to be the default for non-Windows build (could be set automatically based on the OS)

thomaslevesque:
As a possible improvement, I could move the "target framework set definitions" to separate files, which we import based on the value of `TargetFrameworkSet` property:

    <Import Project="targets/$(TargetFrameworkSet).props" />

This would be mostly to make the Directory.Build.props file look cleaner

Also, I'm not sure about the name of the concept. "Target framework set" is what I came up with, but let me know if you have a better idea. "Build profile", maybe?

blairconrad:
I think the approach is good. I'm not sure I'll use the feature that often, but who knows? I use `-s` more than I expected.


I liked adding a parameter to the build script as well. The problem is that it's not discoverable as is, and as you and M. Ralph noted, adding parameters that aren't handled by Bullseye can be problematic. I toyed with the idea of adding targets to set the targetframeworkset (e.g. `build netcoreall default`, but the downside is that as soon as you include the target, you lose the defaultiness of the default target. Maybe that's  an acceptable downside? If just running tests it isn't apparent: `build netcoreall specs`. And honestly, most of the time that I'll be trying to speed up the build, I will for sure be cutting out the approval and packing targets!


> We could also enable a user-defined props file that would be included in Directory.Build.props if it exists, and ignored by Git.

Yes, please. Better than temporarily editing anything. Until we edit _that_ file and forget to unedit! :laughing: 

> could move the "target framework set definitions" to separate files

I'm ambivalent. Cleaner main file, but more files to look at. Do whichever you think is best.

> I'm not sure about the name of the concept. "Target framework set" is what I came up with, but let me know if you have a better idea. "Build profile", maybe?

I found "target framework set" to be comprehensible, so I was content. But since you put "build profile" in front of my face, I find I like it more…

thomaslevesque:
> I'm not sure I'll use the feature that often, but who knows?

I think I would. In fact, I think I will almost always set the target framework set to `netcore3.0`, to have a fast feedback loop while coding, and only switch back to `full` before committing.

> I liked adding a parameter to the build script as well. The problem is that it's not discoverable as is, and as you and M. Ralph noted, adding parameters that aren't handled by Bullseye can be problematic.

Look [here](https://github.com/thomaslevesque/Hamlet/blob/master/tools/build/Build.cs) for a workaround. Not ideal, but it gets the job done.

>  I toyed with the idea of adding targets to set the targetframeworkset (e.g. `build netcoreall default`, but the downside is that as soon as you include the target, you lose the defaultiness of the default target. Maybe that's an acceptable downside? If just running tests it isn't apparent: `build netcoreall specs`.

It's a good option too. But I think I like something like `build.cmd --targets=netcoreall` better than `build.cmd netcoreall default`

> Yes, please. Better than temporarily editing anything. Until we edit _that_ file and forget to unedit! 😆

Will do 👍 
 
> I'm ambivalent. Cleaner main file, but more files to look at. Do whichever you think is best.

I'll try it and we'll see how it goes.

> I found "target framework set" to be comprehensible, so I was content. But since you put "build profile" in front of my face, I find I like it more…

Build profile it is, then!

thomaslevesque:
I pushed new commits for the changes we discussed, let me know what you think.

thomaslevesque:
(I didn't implement the changes in the build script yet)

profiles/netcoreall.props
```diff
@@ -0,0 +1,14 @@
+<Project>
+  <!-- All .NET Core / .NET Standard TFMs -->
+  <PropertyGroup>
+    <FakeItEasyTargetFrameworks>netstandard1.6;netstandard2.0;netstandard2.1</FakeItEasyTargetFrameworks>
+    <ValueTaskExtensionsTargetFrameworks>netstandard1.6;netstandard2.0;netstandard2.1</ValueTaskExtensionsTargetFrameworks>
+    <ExampleTargetFrameworks>netcoreapp2.1</ExampleTargetFrameworks>
```
blairconrad: I expected 2.1 and 3.0.
If the thinking was that they're just examples, so we probably don't need to build both, maybe default to the newer framework? I can live with it as is, though.

(Actually, this brings up a similar question for the full profile, and which frameworks we should target for the examples and the approval tests - if we're just going to target one in each build profile, should it be the latest/most portable? Or maybe it's all fine.)

thomaslevesque: > If the thinking was that they're just examples, so we probably don't need to build both

Yes, that's what I had in min

> maybe default to the newer framework?

👍 

> (Actually, this brings up a similar question for the full profile, and which frameworks we should target for the examples and the approval tests - if we're just going to target one in each build profile, should it be the latest/most portable? Or maybe it's all fine.)

The latest for each profile is a good choice, I think.

thomaslevesque:
I changed the TFM for the example projects, squashed fixup commits, and updated the docs

profiles/netcoreall.props
```diff
@@ -0,0 +1,14 @@
...
+  <!-- All .NET Core / .NET Standard TFMs -->
+  <PropertyGroup>
+    <FakeItEasyTargetFrameworks>netstandard1.6;netstandard2.0;netstandard2.1</FakeItEasyTargetFrameworks>
+    <ValueTaskExtensionsTargetFrameworks>netstandard1.6;netstandard2.0;netstandard2.1</ValueTaskExtensionsTargetFrameworks>
+    <ExampleTargetFrameworks>netcoreapp3.0</ExampleTargetFrameworks>
+    <SpecsTargetFrameworks>netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</SpecsTargetFrameworks>
+    <IntegrationTestsTargetFrameworks>netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</IntegrationTestsTargetFrameworks>
+    <UnitTestsTargetFrameworks>netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</UnitTestsTargetFrameworks>
+    <AnalyzerTestsTargetFrameworks>netcoreapp2.1;netcoreapp3.0</AnalyzerTestsTargetFrameworks>
+    <TestHelpersTargetFrameworks>netstandard1.6;netstandard2.0;netstandard2.1</TestHelpersTargetFrameworks>
```
blairconrad: Why a netstandard2.1 here but not in full?

how_to_build.md
```diff
@@ -44,6 +46,42 @@ After the build has completed, the build artifacts will be located in `artifacts
...
 * View the full list of options:
 
     `build.cmd -?`
+
+### Building only a subset of the supported target frameworks
+
+FakeItEasy targets multiple versions of .NET (.NET Framework 4.0 and 4.5, .NET
+Standard 1.6, 2.0 and 2.1), and the tests also run on multiple frameworks (.NET
+Framework 4.6.1 and several versions of .NET Core). A consequence is that a full
+build can take a significant time. When working on the code, you might want a
```
blairconrad: "take a significant time" hits my ear funny. I'd've written "take a significant amount of time". Or perhaps "take significant time".

profiles/netcoreall.props
thomaslevesque: A mistake. TestHelpers doesn't need a netstandard2.1 target

how_to_build.md
thomaslevesque: Ah, thanks. Not being an native English speaker, it can sometimes be hard to tell what will sound funny to a native speaker...

thomaslevesque: And it also shouldn't be netstandard2.1 for the netcore3.0 profile

profiles/netcoreall.props
thomaslevesque: And it also shouldn't be netstandard2.1 for the netcore3.0 profile

thomaslevesque:
Fixed and squashed

profiles/netcore3.0.props
```diff
@@ -0,0 +1,14 @@
...
+  <!-- .NET Core 3 / .NET Standard 2.1 -->
+  <PropertyGroup>
+    <FakeItEasyTargetFrameworks>netstandard2.1</FakeItEasyTargetFrameworks>
+    <ValueTaskExtensionsTargetFrameworks>netstandard2.1</ValueTaskExtensionsTargetFrameworks>
+    <ExampleTargetFrameworks>netcoreapp3.0</ExampleTargetFrameworks>
+    <SpecsTargetFrameworks>netcoreapp3.0</SpecsTargetFrameworks>
+    <IntegrationTestsTargetFrameworks>netcoreapp3.0</IntegrationTestsTargetFrameworks>
+    <UnitTestsTargetFrameworks>netcoreapp3.0</UnitTestsTargetFrameworks>
+    <AnalyzerTestsTargetFrameworks>netcoreapp3.0</AnalyzerTestsTargetFrameworks>
+    <TestHelpersTargetFrameworks>netstandard2.0</TestHelpersTargetFrameworks>
```
blairconrad: When I use the `FakeItEasy.user.props` file to set the build profile to netcore3.0 (which was fun), and run `build integ`, I again get

```
D:\Sandbox\FakeItEasy\tests\FakeItEasy.Tests.TestHelpers\FakeItEasy.Tests.TestHelpers.csproj :
error NU1201: Project FakeItEasy is not compatible with netstandard2.0 (.NETStandard,Version=v2.0).
Project FakeItEasy supports: netstandard2.1 (.NETStandard,Version=v2.1) [D:\Sandbox\FakeItEasy\FakeItEasy.sln]
```

tests/FakeItEasy.Tests.TestHelpers/FakeItEasy.Tests.TestHelpers.csproj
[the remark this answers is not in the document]
blairconrad: I think we both forgot about this comment as well.

profiles/netcore3.0.props
thomaslevesque: Argh... OK, I'll *really* review and test those TFMs before I push a new fix.

tests/FakeItEasy.Tests.TestHelpers/FakeItEasy.Tests.TestHelpers.csproj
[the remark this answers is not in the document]
thomaslevesque: Oops! Sorry

thomaslevesque:
> It's the review that never ends. Sorry.

My fault... It should be fixed now
Thanks for the careful review!

blairconrad:
Hey. Packing fails in a clean sandbox with a non-full build. e.g. when I build netcore3.0. Perhaps we should take a page out of the main package's book and use something like

```
<None Include="$(OutputPath)/**/$(AssemblyName).pdb" Pack="true" PackagePath="lib" Visible="false" />
```

thomaslevesque:
> Packing fails in a clean sandbox with a non-full build. e.g. when I build netcore3.0.

Oh :(
I had tried all profiles, but only with the `build` target.

> Perhaps we should take a page out of the main package's book and use something like

Yes, I think I'll do that. Although I intended to drop that at some point and use `AllowedOutputExtensionsInPackageBuildOutputFolder` instead, as mentioned [here](https://github.com/dotnet/sourcelink#alternative-pdb-distribution).

thomaslevesque:
> Perhaps we should take a page out of the main package's book and use something like

In fact, it doesn't work, because it also takes netstandard2.1, which should be excluded...

blairconrad:
```xml
  <None Include="$(OutputPath)/net45/$(AssemblyName).pdb" Pack="true" PackagePath="lib/net45" Visible="false" Condition="Exists('$(OutputPath)/net45/$(AssemblyName).pdb')" />
    <None Include="$(OutputPath)/netstandard1.6/$(AssemblyName).pdb" Pack="true" PackagePath="lib/netstandard1.6" Visible="false" Condition="Exists('$(OutputPath)/netstandard1.6/$(AssemblyName).pdb')" />
    <None Include="$(OutputPath)/netstandard2.0/$(AssemblyName).pdb" Pack="true" PackagePath="lib/netstandard2.0" Visible="false" Condition="Exists('$(OutputPath)/netstandard2.0/$(AssemblyName).pdb')" />
```
?

thomaslevesque:
It's an option. But if the PDBs are there from a previous build, it would include them even when they're not in the selected TFMs. I have another option in mind, let me try...

thomaslevesque:
> I have another option in mind, let me try...

Well, my thing didn't work. I found another way, which isn't great, but the end result is slightly cleaner than before

> But if the PDBs are there from a previous build, it would include them even when they're not in the selected TFMs

Actually, it doesn't really matter, because we will only ever publish packages produced by the full build anyway.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -1,8 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netstandard1.6;netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>$(ValueTaskExtensionsTargetFrameworks)</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
```
thomaslevesque: This causes the Pack task to *not* include .dll and .xml files from the build output. I include them manually, which gives me control over what goes into the package.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -1,8 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netstandard1.6;netstandard2.0;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>$(ValueTaskExtensionsTargetFrameworks)</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <BeforePack>AddNetStd21Placeholder;$(BeforePackOn)</BeforePack>
```
thomaslevesque: Cleaner than depending on `_WalkEachTargetPerFramework`, which is a private target

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -40,23 +42,16 @@
...
-  <!-- Include PDB in package -->
+  <!-- Selectively include build output for package -->
   <ItemGroup>
-    <None Include="$(OutputPath)/net45/$(AssemblyName).pdb" Pack="true" PackagePath="lib/net45" Visible="false" />
-    <None Include="$(OutputPath)/netstandard1.6/$(AssemblyName).pdb" Pack="true" PackagePath="lib/netstandard1.6" Visible="false" />
-    <None Include="$(OutputPath)/netstandard2.0/$(AssemblyName).pdb" Pack="true" PackagePath="lib/netstandard2.0" Visible="false" />
+    <None Include="$(OutputPath)/**/$(AssemblyName).dll" Pack="true" PackagePath="lib" Visible="false" />
+    <None Include="$(OutputPath)/**/$(AssemblyName).pdb" Pack="true" PackagePath="lib" Visible="false" />
+    <None Include="$(OutputPath)/**/$(AssemblyName).xml" Pack="true" PackagePath="lib" Visible="false" />
+    <None Remove="$(OutputPath)/netstandard2.1/*" />
```
thomaslevesque: I include everything, then exclude what I don't want

blairconrad:
Nice job, @thomaslevesque! Thanks for your work.

thomaslevesque:
Thanks for the careful review!

PR #1669, opened 2019-10-20
Refactor approval tests

thomaslevesque:
As discussed in #1666


thomaslevesque:
By default ApprovalTests uses the name of the test to name the approved API file, but now all TFMs share the same name, so I had to customize the file names. I thought it was a good opportunity to reorganize those files. Let me know if you don't like it.

tests/FakeItEasy.Tests.Approval/ApiApproval.cs
```diff
@@ -1,79 +1,40 @@
...
+        [Theory]
+        [InlineData("net40")]
+        [InlineData("net45")]
+        [InlineData("netstandard1.6")]
+        [InlineData("netstandard2.0")]
+        [InlineData("netstandard2.1")]
         [UseReporter(typeof(DiffReporter))]
         [MethodImpl(MethodImplOptions.NoInlining)]
-        public void ApproveApi45()
+        public void ApproveFakeItEasyApi(string frameworkVersion)
```
blairconrad: When I started reading, I was expecting to see a test parameterized by project name and frameworkVersion. This is good, though!

thomaslevesque: I pushed a new commit implementing that change

thomaslevesque:
Thanks for the merge!

Next in line, #1662, which I just rebased. I'm afraid I have a few more coming after that...

thomaslevesque:
This change has been released as part of [FakeItEasy 5.4.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/5.4.0).

PR #1672, opened 2019-10-23
Remove ILMerge

### LINKED ISSUE #1258: Stop merging Castle.Core
Recently we found [a bug ](https://github.com/FakeItEasy/FakeItEasy/issues/1257) that  users could've worked around by using a newer version of Castle.Core than we bundle in the .NET Framework flavour of our package.

@thomaslevesque  commented (https://github.com/FakeItEasy/FakeItEasy/issues/1257#issuecomment-342503486):

> this issue makes me wonder if we should really ILMerge Castle.Core... We probably discussed it before, but there's a compelling argument against it: in such a case when there's a bug in Castle.Core, the end-user can't just work around the issue by upgrading to the next version of Castle.Core where the issue is fixed. They have to wait for us to release a version that uses the newer Castle.Core.

I responded:

>  I see your point. I think the original reason for merging was to work around situations where users' production code required a different version of Castle.Core than we liked. What's the lesser of the two evils?

@thomaslevesque reresponded:

> I'm not sure... if they want a more recent version, fine, but if they need an older one, they're in trouble. Maybe we should offer a separate package where Castle.Core is not ILMerged. In fact, if we do that, it would be more consistent to stop using ILMerge in the main package, and do it only in the new one (which would only target net40/net45, since ILMerge doesn't work with .NET Standard). Anyway, it's just a random idea, and a bit off-topic...

And now we're here!

thomaslevesque:
Fixes #1258

😃


thomaslevesque:
Such a small change, so many benefits :)

blairconrad:
I don't understand the connection between removing ILMerge and dc2ad275af73989d145173cf86af9d8f797402c0.

thomaslevesque:
> I don't understand the connection between removing ILMerge and [dc2ad27](https://github.com/FakeItEasy/FakeItEasy/commit/dc2ad275af73989d145173cf86af9d8f797402c0).

Look at the comment in the removed TypeExtensions class. The `GetTypeInformation()` method only existed because of ILMerge

blairconrad:
Ah, thanks @thomaslevesque. Really should've opened that diff.

thomaslevesque:
Thanks!

PR #1673, opened 2019-10-23
Use SourceLink

### LINKED ISSUE #1417: Consider switching to SourceLink instead of PdbGit

thomaslevesque:
Fixes #1417

### DIFF
--- a/src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.projitems
+++ b/src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.projitems
@@ -13,12 +13,6 @@
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)FakeItEasy.Analyzer.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-    <DebugType>full</DebugType>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
-    <DebugType>pdbonly</DebugType>
-  </PropertyGroup>
   <ItemGroup>
     <Compile Include="$(MSBuildThisFileDirectory)..\*.cs" />
   </ItemGroup>
--- a/src/FakeItEasy/FakeItEasy.csproj
+++ b/src/FakeItEasy/FakeItEasy.csproj
@@ -5,6 +5,7 @@
     <AssemblyName>FakeItEasy</AssemblyName>
     <FileAlignment>512</FileAlignment>
     <SignAssembly>true</SignAssembly>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
   </PropertyGroup>
 
   <PropertyGroup Label="Package properties">
@@ -61,12 +62,6 @@
   <PropertyGroup Condition="'$(TargetFramework)' == 'net40'">
     <DefineConstants>$(DefineConstants);FEATURE_REFLECTION_GETASSEMBLIES;FEATURE_REFLECTIONONLYLOAD;FEATURE_EVENT_ARGS_MUST_EXTEND_EVENTARGS</DefineConstants>
   </PropertyGroup>
-
-  <!-- Include PDB in package -->
-
-  <ItemGroup>
-    <None Include="$(OutputPath)/**/$(AssemblyName).pdb" Pack="true" PackagePath="lib" Visible="false" />
-  </ItemGroup>
 
   <ItemGroup>
     <!-- T4 template support -->
--- a/src/FakeItEasy/FakeItEasy.csproj
+++ b/src/FakeItEasy/FakeItEasy.csproj
@@ -22,17 +22,8 @@
     <PackageReference Include="Castle.Core" Version="4.3.1" />
   </ItemGroup>
 
-  <!-- Debug -->
-
-  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-    <DebugType>full</DebugType>
-  </PropertyGroup>
-
-  <!-- Release -->
-
   <PropertyGroup Condition="'$(Configuration)' == 'Release'">
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <DebugType>pdbonly</DebugType>
   </PropertyGroup>
 
   <!-- .NET Standard 2.1 -->
--- a/tools/FakeItEasy.Build/Program.cs
+++ b/tools/FakeItEasy.Build/Program.cs
@@ -17,20 +17,6 @@ namespace FakeItEasy.Build
             "src/FakeItEasy.Analyzer.VisualBasic/FakeItEasy.Analyzer.VisualBasic.csproj"
         };
 
-        private static readonly Pdb[] Pdbs =
-        {
-            "src/FakeItEasy/bin/Release/net40/FakeItEasy.pdb",
-            "src/FakeItEasy/bin/Release/net45/FakeItEasy.pdb",
-            "src/FakeItEasy/bin/Release/netstandard1.6/FakeItEasy.pdb",
-            "src/FakeItEasy/bin/Release/netstandard2.0/FakeItEasy.pdb",
-            "src/FakeItEasy/bin/Release/netstandard2.1/FakeItEasy.pdb",
-            "src/FakeItEasy.Extensions.ValueTask/bin/Release/net45/FakeItEasy.Extensions.ValueTask.pdb",
-            "src/FakeItEasy.Extensions.ValueTask/bin/Release/netstandard1.6/FakeItEasy.Extensions.ValueTask.pdb",
-            "src/FakeItEasy.Extensions.ValueTask/bin/Release/netstandard2.0/FakeItEasy.Extensions.ValueTask.pdb",
-            "src/FakeItEasy.Analyzer.CSharp/bin/Release/FakeItEasy.Analyzer.Csharp.pdb",
-            "src/FakeItEasy.Analyzer.VisualBasic/bin/Release/FakeItEasy.Analyzer.VisualBasic.pdb"
-        };
-
         private static readonly IReadOnlyDictionary<string, string[]> TestSuites = new Dictionary<string, string[]>
         {
             ["unit"] = new[]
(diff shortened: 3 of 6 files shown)


src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -71,12 +64,6 @@
     <DefineConstants>$(DefineConstants);FEATURE_REFLECTION_GETASSEMBLIES;FEATURE_REFLECTIONONLYLOAD;FEATURE_EVENT_ARGS_MUST_EXTEND_EVENTARGS</DefineConstants>
   </PropertyGroup>
 
-  <!-- Include PDB in package -->
-
-  <ItemGroup>
-    <None Include="$(OutputPath)/**/$(AssemblyName).pdb" Pack="true" PackagePath="lib" Visible="false" />
-  </ItemGroup>
-
```
thomaslevesque: No longer necessary thanks to `AllowedOutputExtensionsInPackageBuildOutputFolder`. We can't do the same in `FakeItEasy.Extensions.ValueTask`, though, because it would include the PDB for `netstandard2.1`, and I couldn't find a way to exclude it.

If you prefer, I can move that to a separate PR. I did it in this one, because the `AllowedOutputExtensionsInPackageBuildOutputFolder` approach is mentioned in the SourceLink instructions.

src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.projitems
```diff
@@ -12,12 +12,7 @@
     <PackageTags>FakeItEasy;analyzer</PackageTags>
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)FakeItEasy.Analyzer.ruleset</CodeAnalysisRuleSet>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-    <DebugType>full</DebugType>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
-    <DebugType>pdbonly</DebugType>
```
blairconrad: Does this mean that users on < .NET Framework 4.7.1 will have degraded stack traces? https://github.com/dotnet/sdk/issues/1238#issuecomment-303203347 suggests so, if I read it right.

blairconrad: Also, could this be hoisted to `src/Directory.build.props`?

src/FakeItEasy/FakeItEasy.csproj
blairconrad: I am content as is. Thanks for the explanation.

tools/FakeItEasy.Build/Program.cs
```diff
@@ -73,28 +59,13 @@ public static void Main(string[] args)
 
             Target(
                 "pack",
-                DependsOn("build", "pdbgit"),
```
blairconrad: pretty

src/FakeItEasy.Analyzer.Shared/FakeItEasy.Analyzer.Shared.projitems
thomaslevesque: > Does this mean that users on < .NET Framework 4.7.1 will have degraded stack traces? [dotnet/sdk#1238 (comment)](https://github.com/dotnet/sdk/issues/1238#issuecomment-303203347) suggests so, if I read it right.

I wasn't aware of this. Do we care?
We could keep emitting legacy PDBs for net40 and net45, if you prefer.

> Also, could this be hoisted to `src/Directory.build.props`?

Sure

blairconrad: Hmm. Either it's not so bad, or I did it wrong. I just removed a `Record.Exception` and ran a spec:

```
  X FakeItEasy.Specs.StrictFakeSpecs.ArgumentMismatchVoid() [03] When I call the method with non-matching arguments [1ms]
  Error Message:
   FakeItEasy.ExpectationException : Call to unconfigured method of strict fake: FakeItEasy.Specs.StrictFakeSpecs+IMyInterface.DoIt(id: a762f030-d39e-4316-92b1-a51eff3dc51f).
  Stack Trace:
     at FakeItEasy.Core.StrictFakeRule.Apply(IInterceptedFakeObjectCall fakeObjectCall) in D:\Sandbox\FakeItEasy\src\FakeItEasy\Core\StrictFakeRule.cs:line 11
   at FakeItEasy.Core.FakeManager.ApplyBestRule(IInterceptedFakeObjectCall fakeObjectCall) in D:\Sandbox\FakeItEasy\src\FakeItEasy\Core\FakeManager.cs:line 254
   at FakeItEasy.Core.FakeManager.FakeItEasy.Core.IFakeCallProcessor.Process(InterceptedFakeObjectCall fakeObjectCall) in D:\Sandbox\FakeItEasy\src\FakeItEasy\Core\FakeManager.cs:line 167
   at FakeItEasy.Creation.CastleDynamicProxy.CastleDynamicProxyGenerator.ProxyInterceptor.Intercept(IInvocation invocation) in D:\Sandbox\FakeItEasy\src\FakeItEasy\Creation\CastleDynamicProxy\CastleDynamicProxyGenerator.cs:line 193
   at Castle.Proxies.ObjectProxy_3.DoIt(Guid id)
   at FakeItEasy.Specs.StrictFakeSpecs.<>c__DisplayClass1_0.<ArgumentMismatchVoid>b__2() in D:\Sandbox\FakeItEasy\tests\FakeItEasy.Specs\StrictFakeSpecs.cs:line 30
   at Xbehave.StringExtensions.<>c__DisplayClass0_0.<x>b__0(IStepContext c)
   at Xbehave.Execution.StepInvoker.<RunAsync>b__6_1()
   at Xbehave.Execution.Invoker.<>c__DisplayClass0_1.<<Invoke>b__1>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
  ! FakeItEasy.Specs.StrictFakeSpecs.ArgumentMismatchVoid() [04] Then the fake throws an expectation exception [1ms]
  ! FakeItEasy.Specs.StrictFakeSpecs.ArgumentMismatchVoid() [05] And the exception message describes the call [1ms]
```

thomaslevesque: Ah, but you didn't *really* run it on a framework older than 4.7.2... .NET Framework updates are in-place. So if you have e.g. 4.7.2, you no longer have the previous version, and you run everything on 4.7.2. So we can't really test it unless we find an old machine...

thomaslevesque: I'm a bit torn... On one hand, if we say we support net40/net45, we should *really* support them, including providing a good debug experience. On the other hand, the PDBs for net4x are huge (1.8 MB each, 750 KB compressed; the whole package is 2.32 MB, vs. only 800 KB with portable PDBs).

Of course, we don't *have* to embed the PDBs in the package. We could upload them to a symbol server. This is now the recommended approach. As far as I can tell , all the tools support this (NuGet, SourceLink, AppVeyor). And it would avoid the headache of manually including the PDBs.

thomaslevesque: What I like about having the PDB in the package is that they don't need to be downloaded during debug, since they're already present. But it does make the package significantly larger.

blairconrad: > Ah, but you didn't really run it on a framework older than 4.7.2.

I _knew_ I was right to think I was wrong!

Man. Now I want to see how bad the experience is. The stack trace down to the _user_'s code shouldn't be affected, right? It might just go wonky inside FakeItEasy. Then one could test on a modern machine and see the error?

thomaslevesque: > The stack trace down to the _user_'s code shouldn't be affected, right? It might just go wonky inside FakeItEasy.

I *think* so... I can spin up an Azure VM and try, if you're worried

blairconrad: Ah, I was looking for such an option earlier and had bad search results. That's why I mentioned AppVeyor. It seems a reasonable step, since you were also concerned about _really_ supporting the older frameworks.

blairconrad: I don't want to put you to work, though. But I bet you'll get it done faster than I would...

thomaslevesque: > Ah, I was looking for such an option earlier and had bad search results. That's why I mentioned AppVeyor. It seems a reasonable step, since you were also concerned about _really_ supporting the older frameworks.

Are you talking about the Azure VM? Or packaging the PDBs separately?

blairconrad: Oh, the Azure VM.

Packaging the PDBs separately would be fine, I guess, but has the downside you mention and also slightly recomplicates the csprojs, no?

OH! I missed your note before that SourceLink works with legacy pdbs.
If we don't want to chase down a VM and view a stacktrace with the portable PDB at this point, we could revert to legacy pdbs for the .NET framework TFMs and then optionally convert to portable PDBs and/or package the PDBs separately.

thomaslevesque: > and also slightly recomplicates the csprojs, no?

On the contrary, it would probably make them simpler. The supported way is to package the PDBs separately; we have to jump through hoops to put them in the same package.

> OH! I missed your note before that SourceLink works with legacy pdbs.
> If we don't want to chase down a VM and view a stacktrace with the portable PDB at this point, we could revert to legacy pdbs for the .NET framework TFMs and then optionally convert to portable PDBs and/or package the PDBs separately.

I have the VM running, shouldn't be long to test. If the result isn't good, we can consider keeping the legacy PDBs.

thomaslevesque: After discussion in the chat, we're switching to portable PDB for all TFMs

thomaslevesque:
Thanks!

PR #1674, opened 2019-10-23
Create lazy dummies lazily

### LINKED ISSUE #1656: Resolve Dummy Lazy values lazily
I thought of this while reviewing docs for dummies and thinking about how many dummy types/objects might get created.

Wouldn't it be neat if this passed, and in fact a dummy type for WantsToBeDummyed was not actually generated, if Lazy.Value was never called? [A definite possibility when faking objects with 'lazy dependencies].

```
namespace DoesntMatter
{
    public class WantsToBeDummyed
    {
        public static bool ynstanshyated;

        public WantsToBeDummyed()
        {
            ynstanshyated = true;
        }

        public virtual long CanOverryde()
        {
            return 3L;
        }
    }             

    class Program
    {
        static void Main(string[] args)
        {
            Assert.False(WantsToBeDummyed.ynstanshyated);
            var dummy = A.Dummy<Lazy<WantsToBeDummyed>>();
            Assert.False(WantsToBeDummyed.ynstanshyated); // what, its true?!

            var ynnerDummy = dummy.Value;
            Assert.True(WantsToBeDummyed.ynstanshyated);
        }
    }
}
```

thomaslevesque:
Fixes #1656

### DIFF
--- a/src/FakeItEasy/Creation/DummyValueResolver.cs
+++ b/src/FakeItEasy/Creation/DummyValueResolver.cs
@@ -180,6 +180,11 @@ namespace FakeItEasy.Creation
 
         private class ResolveByCreatingLazyStrategy : ResolveStrategy
         {
+            private static readonly MethodInfo CreateLazyDummyGenericDefinition =
+                typeof(ResolveByCreatingLazyStrategy).GetMethod(
+                    nameof(CreateLazyDummy),
+                    BindingFlags.Static | BindingFlags.NonPublic);
+
             public override CreationResult TryCreateDummyValue(
                 Type typeOfDummy,
                 IDummyValueResolver resolver,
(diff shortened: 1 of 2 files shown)


tests/FakeItEasy.Specs/DummyCreationSpecs.cs
```diff
@@ -510,6 +510,22 @@ public void LazyCreationNonDummyable(Lazy<NonDummyable> dummy)
                 .x(() => dummy.Value.Should().BeNull());
         }
 
+        [Scenario]
+        public void LazyCreationIsLazy(Lazy<LazilyCreated> dummy)
+        {
+            "Given a type that tracks instance creation"
+                .x(() => LazilyCreated.IsInstanceCreated = false);
```
blairconrad: I fear that there's a small chance that the `GenericDummyCreationSpecs` and `NonGenericDummyCreationSpecs` will run concurrently and `IsInstanceCreated` will be reset, spoiling the last assert.

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
```diff
@@ -510,6 +510,22 @@ public void LazyCreationNonDummyable(Lazy<NonDummyable> dummy)
...
         }
 
+        [Scenario]
+        public void LazyCreationIsLazy(Lazy<LazilyCreated> dummy)
+        {
+            "Given a type that tracks instance creation"
+                .x(() => LazilyCreated.IsInstanceCreated = false);
+
+            "When a dummy lazy of this type is requested"
+                .x(() => dummy = A.Dummy<Lazy<LazilyCreated>>());
```
blairconrad: I see you took a page out of my book and spelled `this.CreateDummy` funny.

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
thomaslevesque: Ah, good catch. I thought there'd be no problem because this was only used in a single test, but I forgot it was actually two tests...
A simple fix would be to assign both `GenericDummyCreationSpecs` and `NonGenericDummyCreationSpecs` to the same xunit collection. Would that work for you, or do you have something more elaborate in mind?

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
thomaslevesque: Oops!

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
thomaslevesque: I added a `[Collection]` attribute to the base class (which applies it to the derived classes as well, since it's inherited), that should do it.

blairconrad: Good one. I was considering atomic counting. Yours is superior.

thomaslevesque:
> Do squash it up, good sir.

Done!

blairconrad:
Thank you!

thomaslevesque:
Thanks!

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -188,32 +193,25 @@ private class ResolveByCreatingLazyStrategy : ResolveStrategy
                 if (typeOfDummy.GetTypeInfo().IsGenericType && typeOfDummy.GetGenericTypeDefinition() == typeof(Lazy<>))
                 {
                     var typeOfLazyResult = typeOfDummy.GetGenericArguments()[0];
-                    var creationResult = resolver.TryResolveDummyValue(typeOfLazyResult, resolutionContext);
-                    object lazyResult = creationResult.WasSuccessful
-                        ? creationResult.Result
-                        : typeOfLazyResult.GetDefaultValue();
-
                     var funcType = typeof(Func<>).MakeGenericType(typeOfLazyResult);
```
TimLovellSmith: In the diff view, it looks like funcType is instantiated but no longer used - is that right?

TimLovellSmith:
Neato!

src/FakeItEasy/Creation/DummyValueResolver.cs
blairconrad: I think you're correct, @TimLovellSmith.

thomaslevesque: Indeed. Thanks @TimLovellSmith!

PR #1682, opened 2019-10-26
Build on Linux

### LINKED ISSUE #1605: Support building the project on Linux or macOS
Currently the project can't be built on Linux pr macOS because of the .NET Framework targets (`net40` and `net45` for the library, `net461` for the tests).

I think it might be possible to build on Linux by referencing the [.NET Framework reference assemblies](https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies) package (the `net461` tests will still run only on Windows, though). The package is not stable yet, though, so we might want to wait until it is before referencing it.

Another possible approach is to use MSBuild conditions on the `TargetFrameworks` elements to not include .NET Framework targets when building on Linux or macOS. It's something we can do right now, since it doesn't rely on a not-yet-stable package. However, this approach won't generate a full, publishable package, since the netfx targets will be missing. It would just be a convenience for contributors using Linux or macOS.

thomaslevesque:
Fixes #1605


appveyor.yml
```diff
@@ -1,5 +1,7 @@
 version: '{build}'
-image: Visual Studio 2019
+image:
+- Visual Studio 2019
+- Ubuntu
```
thomaslevesque: Currently not running. We need to configure it to run build.sh instead of build.cmd

profiles/netcore2+.props
```diff
@@ -0,0 +1,14 @@
...
+    <ValueTaskExtensionsTargetFrameworks>netstandard2.0;netstandard2.1</ValueTaskExtensionsTargetFrameworks>
+    <ExampleTargetFrameworks>netcoreapp3.0</ExampleTargetFrameworks>
+    <SpecsTargetFrameworks>netcoreapp2.1;netcoreapp3.0</SpecsTargetFrameworks>
+    <IntegrationTestsTargetFrameworks>netcoreapp2.1;netcoreapp3.0</IntegrationTestsTargetFrameworks>
+    <UnitTestsTargetFrameworks>netcoreapp2.1;netcoreapp3.0</UnitTestsTargetFrameworks>
+    <AnalyzerTestsTargetFrameworks>netcoreapp2.1;netcoreapp3.0</AnalyzerTestsTargetFrameworks>
+    <TestHelpersTargetFrameworks>netstandard2.0;netstandard2.1</TestHelpersTargetFrameworks>
+    <ApprovalTestsTargetFrameworks>netcoreapp3.0</ApprovalTestsTargetFrameworks>
+  </PropertyGroup>
+</Project>
```
thomaslevesque: Doesn't seem possible to install .NET Core 1.0 runtime on a recent Ubuntu, so I added a new profile to build just .NET Core 2+

tests/FakeItEasy.Tests.TestHelpers/ObjectAssertionsExtensions.cs
```diff
@@ -73,6 +73,16 @@ public static void BeAFake(this ReferenceTypeAssertions<object, ObjectAssertions
...
         }
 
+        public static ExceptionAssertions<TExpectedException> WithMessageModuloLineEndings<TExpectedException>(
+            this ExceptionAssertions<TExpectedException> assertion,
+            string expectedMessage,
+            string because = null,
+            params object[] becauseArgs)
+            where TExpectedException : Exception
+        {
+            return assertion.WithMessage(expectedMessage.NormalizeLineEndings(), because, becauseArgs);
```
thomaslevesque: This isn't perfect, because it doesn't normalize the actual message, only the expected one. Seems to work, though... As long as we use Environment.NewLine, I think it should be OK

tests/FakeItEasy.Tests.TestHelpers/StringExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            if (string.IsNullOrEmpty(s))
+            {
+                return s;
+            }
+
+#if FEATURE_STRING_CONTAINS_COMPARISONTYPE
+            if (!s.Contains('\n', StringComparison.Ordinal))
+#else
+            if (s.IndexOf('\n') < 0)
```
thomaslevesque: I don't use this on .NET Standard 2.1, because it causes a warning. But now that I think of it, maybe we don't need a .NET Standard 2.1 target for the test helpers...

appveyor.yml
thomaslevesque: Now it runs; I changed the image to Ubuntu1804, rather than latest, to avoid surprises

tests/FakeItEasy.Tests.TestHelpers/StringExtensions.cs
thomaslevesque: Ah, I just remembered why we needed it... It references FakeItEasy, which is only built on netstandard2.1 in the netcore3.0 profile.

appveyor.yml
```diff
@@ -1,22 +1,33 @@
...
+  deploy_script:
+    - cmd: tools\deploy.cmd
+- matrix:
+    only:
+      - image: Ubuntu1804
+  build_script:
+    - sh: ./build.sh
 
 on_failure:
   - ps: $root = (Resolve-Path artifacts); [IO.Directory]::GetFiles($root.Path, '*.*', 'AllDirectories') | % { Push-AppveyorArtifact $_ -FileName $_.Substring($root.Path.Length + 1) -DeploymentName failed-build }
```
blairconrad: I'm thinking that at best, this will not fire on ubuntu. At worst, it'll cause a more confusing error. What think you?

profiles/netcore2+.props
blairconrad: I can't even figure out how to install the runtime on my wsl!

appveyor.yml
thomaslevesque: > I'm thinking that at best, this will not fire on ubuntu

I'm not so sure... The Ubuntu images come with Powershell Core installed. But maybe we need to switch to `pwsh` instead of `ps`.

thomaslevesque: Actually, it works with `ps`. See [this failed build](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/builds/28409128/job/03xbdnjcylmfutv5)

tests/FakeItEasy.IntegrationTests/Assertions/ExceptionMessagesTests.cs
```diff
@@ -33,7 +33,7 @@ public void Exception_message_should_be_correctly_formatted()
     1: FakeItEasy.IntegrationTests.IFoo.Bar() 2 times
     ...
     3: FakeItEasy.IntegrationTests.IFoo.Bar(argument: ""test"")
-    4: FakeItEasy.IntegrationTests.IFoo.Bar(argument: 4/5/1977 12:00:00 AM, argument2: ""birthday"")
+    4: FakeItEasy.IntegrationTests.IFoo.Bar(argument: 4/5/*77 12:00:00 AM, argument2: ""birthday"")
```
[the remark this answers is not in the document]
thomaslevesque: > keep everything like you have it, because it's fine

That's one of my favorite options :)

> just not test with a Date. I don't feel like it's an integral part of the test

Probably the best thing to do.

>     use culture `""` instead of `"en-US"`

I don't think it makes a difference... The format used for short date is based on the system locale, which uses two digits for the year

>     add an `ArgumentValueFormatter<Date>` to ensure a consistent format. Maybe `ToString("s", System.Globalization.CultureInfo.InvariantCulture)`

I don't think it would work, for the same reason as above. We'd  have to specify the full format explicitly.

>     do that, but in the core library. It's a change, but maybe even a feature! (We could slip that in a separate PR for visibility.)

I'm not convinced it would be a good thing... I think people like having dates in their own culture in error messages.

tests/FakeItEasy.Tests.TestHelpers/StringAssertionsExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    public static class StringAssertionsExtensions
+    {
+        public static AndConstraint<StringAssertions> BeModuloLineEndings(
+            this StringAssertions assertion,
+            string expected,
+            string because = null,
+            params object[] becauseArgs)
+        {
+            return assertion.Subject.NormalizeLineEndings().Should().Be(expected.NormalizeLineEndings(), because, becauseArgs);
```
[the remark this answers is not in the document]
thomaslevesque: Good idea 👍

tests/FakeItEasy.Tests.TestHelpers/StringExtensions.cs
thomaslevesque: Yes, because in netstandard2.1 there's an overload that takes a `StringComparison`, which should be used instead

tests/FakeItEasy.Tests.TestHelpers/StringExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+#endif
+            {
+                return s;
+            }
+
+            return Regex.Replace(s, "\r?\n", Environment.NewLine);
```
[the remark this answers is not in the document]
thomaslevesque: I think I'd rather stay with the safest choice. As you said, it wouldn't cover everything, only multiline string literals in tests.

tests/FakeItEasy.Tests/TestHelpers/StringEnumerableExtensions.cs
```diff
@@ -8,14 +9,14 @@ namespace FakeItEasy.Tests.TestHelpers
     public static class StringEnumerableExtensions
     {
         /// <summary>
-        /// Joins a number of individual lines with Windows-style newline characters
+        /// Joins a number of individual lines with default newline characters
```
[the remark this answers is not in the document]
thomaslevesque: 👍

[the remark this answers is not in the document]
thomaslevesque: Or I could drop this method entirely!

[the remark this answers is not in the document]
thomaslevesque: I removed  `AsTextBlock` in a separate commit, which I can drop if you prefer to keep the method

how_to_build.md
```diff
@@ -5,26 +5,50 @@ It also replicates the build on the Continuous Integration build server and is t
...
-"latest" or "at least" forms of guidance. If it seems you have an incompatible version of the software, prefer to upgrade rather than downgrade.
+The build requires that a few pieces of software be installed on the host computer. We're somewhat aggressive about adoptiong new language features and the like, so rather than specifying exactly which versions are required, we'll tend toward "latest" or "at least" forms of guidance. If it seems you have an incompatible version of the software, prefer to upgrade rather than downgrade.
+
+### On Windows
 
 Ensure that the following are installed:
 
 1. a recent version of Visual Studio 2019 (currently this means 16.3 or later) or the Build Tools for Visual Studio 2019
 
-1. a recent version of the .NET Core 1.0 Runtime (currently this means 1.0.12 or later)
```
thomaslevesque: We don't actually need a "recent version". I have 1.0.5 and it works fine

tests/FakeItEasy.Tests.TestHelpers/StringAssertionsExtensions.cs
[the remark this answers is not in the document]
thomaslevesque: In fact, that doesn't work, because exception messages sometimes include messages from Castle.Core, which uses `\r\n`...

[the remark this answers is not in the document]
thomaslevesque: Scratch that. The `\r\n` comes from our code, not from Castle.Core

tests/FakeItEasy.IntegrationTests/Assertions/ExceptionMessagesTests.cs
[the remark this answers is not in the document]
blairconrad: > I think people like having dates in their own culture in error messages.

Heathens.

blairconrad:
Lovely. Squash it up, sir!

thomaslevesque:
> Squash it up, sir!

Done!

blairconrad:
:penguin:

PR #1684, opened 2019-10-28
Add target to approve new public API

blairconrad:
… I get tired of copying files manually.

### DIFF
--- a/tools/FakeItEasy.Build/Program.cs
+++ b/tools/FakeItEasy.Build/Program.cs
@@ -63,6 +63,16 @@ namespace FakeItEasy.Build
                 forEach: ProjectsToPack,
                 action: project => Run("dotnet", $"pack {project.Path} --configuration Release --no-build --output {Path.GetFullPath("artifacts/output")}"));
 
+            Target(
+                "force-approve",
+                () =>
+                    {
+                        foreach (var received in Directory.EnumerateFiles("tests/FakeItEasy.Tests.Approval/ApprovedApi", "*.received.txt", SearchOption.AllDirectories))
+                        {
+                            File.Copy(received, received.Replace(".received.txt", ".approved.txt"), overwrite: true);
+                        }
+                    });
+
             RunTargetsAndExit(args, messageOnly: ex => ex is NonZeroExitCodeException);
         }
 


blairconrad:
Technically I think this target should be called "approve" and the current approve something like "check-api", but what can you do?
Also "check-api" is not the best name.

thomaslevesque:
> Technically I think this target should be called "approve" and the current approve something like "check-api", but what can you do?
> Also "check-api" is not the best name.

I don't mind changing the name of the existing target to use `approve` for the new one. `check-api` isn't great, but I don't have a better idea...

tools/FakeItEasy.Build/Program.cs
```diff
@@ -1,5 +1,6 @@
 namespace FakeItEasy.Build
 {
+    using System;
```
thomaslevesque: What do you need this for? As far as I can tell, the new code only uses classes from `System.IO`

tools/FakeItEasy.Build/Program.cs
```diff
@@ -63,6 +64,16 @@ public static void Main(string[] args)
...
 
+            Target(
+                "accept-api",
+                () =>
+                    {
+                        foreach (var received in Directory.EnumerateFiles("tests/FakeItEasy.Tests.Approval", "*.received.txt", SearchOption.AllDirectories))
+                        {
+                            File.Copy(received, received.Replace(".received.txt", ".approved.txt"), overwrite: true);
+                        }
+                    });
```
thomaslevesque: How do we use this? It assumes the .received files have already been generated... So, run `approve`, which will fail, then run `accept-api`?

tools/FakeItEasy.Build/Program.cs
```diff
@@ -63,6 +64,16 @@ public static void Main(string[] args)
                 forEach: ProjectsToPack,
                 action: project => Run("dotnet", $"pack {project.Path} --configuration Release --no-build --output {Path.GetFullPath("artifacts/output")}"));
 
+            Target(
+                "accept-api",
+                () =>
+                    {
+                        foreach (var received in Directory.EnumerateFiles("tests/FakeItEasy.Tests.Approval", "*.received.txt", SearchOption.AllDirectories))
```
thomaslevesque: Maybe specify `/ApprovedApi` in the path, to further reduce the search scope ?

tools/FakeItEasy.Build/Program.cs
thomaslevesque: Maybe we should add some kind of flag (e.g. an environment variable) to change the behavior of the approval tests. If the flag is set, don't compare received against approved, but just copy received to approved?

blairconrad: > So, run approve, which will fail, then run accept-api?

I usually build everything, find the failure on `approve` and then manually copy. I was just trying to ease the last step. Do you do something different?

> Maybe we should add some kind of flag (e.g. an environment variable) to change the behavior of the approval tests. If the flag is set, don't compare received against approved, but just copy received to approved?

Could do, I suppose. I guarantee you 100% I will forget.

tools/FakeItEasy.Build/Program.cs
blairconrad: In initial development, I was just spewing messages to the console. I forgot to remove the `using` when I changed that.

blairconrad:
> I don't mind changing the name of the existing target to use approve for the new one. check-api isn't great, but I don't have a better idea...

validate-api feels slightly better…

or we could leave as is. Or drop the whole thing if you like another approach, such as the environment variables.

tools/FakeItEasy.Build/Program.cs
thomaslevesque: > I usually build everything, find the failure on `approve` and then manually copy. I was just trying to ease the last step. Do you do something different?

For me, each test that fail cause my diff tool (Devart Code Compare) to open the received and approved files, then I just need hit the "Copy all changes to right" button. Doesn't it do that for you?

> Could do, I suppose. I guarantee you 100% I will forget.

Yes, I'd probably forget too. Your way is fine, I think

thomaslevesque:
> validate-api feels slightly better…
> 
> or we could leave as is. Or drop the whole thing if you like another approach, such as the environment variables.

Or keep the name of the current target, and name the new one `force-approve`?

tools/FakeItEasy.Build/Program.cs
blairconrad: For me, it pops up copies of... something. WinMerge? I dunno. Then I panic and close them all and copy over! I could use the differ to copy.

blairconrad:
I think I made your suggested changes. And I squashed. If you like it enough, we could merge. If you have any doubts, we could not!!

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you!

PR #1686, opened 2019-11-04
Add and remove null parameter checks

blairconrad:
Some public methods that will not allow null argument values lack null checks.
Perhaps some have arguably too many.

### DIFF
--- a/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
+++ b/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
@@ -77,6 +77,9 @@ namespace FakeItEasy
         /// <returns>A dummy argument value.</returns>
         public static string Contains(this IArgumentConstraintManager<string> manager, string value)
         {
+            Guard.AgainstNull(manager, nameof(manager));
+            Guard.AgainstNull(value, nameof(value));
+
 #if FEATURE_STRING_CONTAINS_COMPARISONTYPE
             return manager.NullCheckedMatches(x => x.Contains(value, StringComparison.CurrentCulture), x => x.Write("string that contains ").WriteArgumentValue(value));
 #else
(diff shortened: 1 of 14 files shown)


src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -222,22 +229,17 @@ public IReturnValueConfiguration<TMember> WhenArgumentsMatch(Func<ArgumentCollec
                 return this;
             }
 
-            public UnorderedCallAssertion MustHaveHappened(int numberOfTimes, Times timesOption)
-            {
-                Guard.AgainstNull(timesOption, nameof(timesOption));
+            public UnorderedCallAssertion MustHaveHappened(int numberOfTimes, Times timesOption) =>
+                this.ParentConfiguration.MustHaveHappened(numberOfTimes, timesOption);
```
blairconrad: I thought it reasonable, when we're just deferring to another public method with the same parameters, to let the second method do the null check. But if you worry that we might then lose the check if ever we refactor, I am happy to replace. And to add checks where we don't have them.

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
```diff
@@ -301,6 +331,7 @@ public static T Matches<T>(this IArgumentConstraintManager<T> manager, Func<T, b
         [SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "Appropriate for Linq expressions.")]
         public static T Matches<T>(this IArgumentConstraintManager<T> scope, Expression<Func<T, bool>> predicate)
         {
+            Guard.AgainstNull(scope, nameof(scope));
```
thomaslevesque: Maybe rename it to `manager`? The naming is quite inconsistent in those methods.

src/FakeItEasy/Configuration/RuleBuilder.cs
thomaslevesque: 👍

blairconrad:
Thanks, @thomaslevesque. I have another commit ready to rename all the
scopes. I'd hoped to include in a separate PR but can add here if you
prefer.

thomaslevesque:
>  I have another commit ready to rename all the scopes. I'd hoped to include in a separate PR but can add here if you prefer.

Another PR is fine!

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thanks, @thomaslevesque!

PR #1688, opened 2019-11-06
Add targets to create and manage FakeItEasy.user.props

blairconrad:
Augments #1662

### DIFF
--- a/tools/FakeItEasy.Build/Program.cs
+++ b/tools/FakeItEasy.Build/Program.cs
@@ -3,6 +3,9 @@ namespace FakeItEasy.Build
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
+    using System.Text;
+    using System.Text.RegularExpressions;
+    using System.Xml.Linq;
     using SimpleExec;
     using static Bullseye.Targets;
     using static SimpleExec.Command;
@@ -72,6 +75,53 @@ namespace FakeItEasy.Build
                             File.Copy(received, received.Replace(".received.txt", ".approved.txt"), overwrite: true);
                         }
                     });
+
+            Target(
+                "initialize-user-properties",
+                () =>
+                    {
+                        if (!File.Exists("FakeItEasy.user.props"))
+                        {
+                            var defaultUserProps = @"
+<Project>
+  <PropertyGroup>
+    <BuildProfile></BuildProfile>
+  </PropertyGroup>
+</Project>".Trim();
+                            File.WriteAllText("FakeItEasy.user.props", defaultUserProps, Encoding.UTF8);
+                        }
+                    });
+
+            foreach (var profile in Directory.EnumerateFiles("profiles", "*.props").Select(Path.GetFileNameWithoutExtension))
+            {
+                Target(
+                    "use-profile-" + profile,
+                    DependsOn("initialize-user-properties"),
+                    () =>
+                        {
+                            var xmlDoc = XDocument.Load("FakeItEasy.user.props");
+
+                            var buildProfileElement = xmlDoc.Root.Elements("PropertyGroup").Elements("BuildProfile").FirstOrDefault();
+                            if (buildProfileElement is null)
+                            {
+                                var propertyGroupElement = xmlDoc.Root.Element("PropertyGroup");
+                                if (propertyGroupElement is null)
+                                {
+                                    propertyGroupElement = new XElement("PropertyGroup");
+                                    xmlDoc.Root.Add(propertyGroupElement);
+                                }
+
+                                buildProfileElement = new XElement("BuildProfile");
+                                propertyGroupElement.Add(buildProfileElement);
+                            }
+
+                            if (buildProfileElement.Value != profile)
+                            {
+                                buildProfileElement.Value = profile;
+                                xmlDoc.Save("FakeItEasy.user.props");
+                            }
+                        });
+            }
 
             RunTargetsAndExit(args, messageOnly: ex => ex is NonZeroExitCodeException);
         }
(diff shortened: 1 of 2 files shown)


tools/FakeItEasy.Build/Program.cs
```diff
@@ -73,6 +75,38 @@ public static void Main(string[] args)
...
+
+            foreach (var profile in Directory.EnumerateFiles("profiles", "*.props").Select(Path.GetFileNameWithoutExtension))
+            {
+                Target(
+                    "use-profile-" + profile,
+                    DependsOn("initialize-user-properties"),
+                    () =>
+                        {
+                            var originalText = File.ReadAllText("FakeItEasy.user.props", Encoding.UTF8);
+                            var newText = Regex.Replace(originalText, "(?<=<BuildProfile>)[^>]*(?=</BuildProfile>)", profile);
```
thomaslevesque: I... don't even understand how that regex works. How about using Linq to XML? I think the code would be almost as short, and much easier to understand.

blairconrad: Done. And at the risk of tooting your horn, I think it's an improvement!

tools/FakeItEasy.Build/Program.cs
```diff
@@ -98,11 +99,12 @@ public static void Main(string[] args)
                     DependsOn("initialize-user-properties"),
                     () =>
                         {
-                            var originalText = File.ReadAllText("FakeItEasy.user.props", Encoding.UTF8);
-                            var newText = Regex.Replace(originalText, "(?<=<BuildProfile>)[^>]*(?=</BuildProfile>)", profile);
-                            if (newText != originalText)
+                            var xmlDoc = XDocument.Load("FakeItEasy.user.props", LoadOptions.PreserveWhitespace);
```
thomaslevesque: Good thinking, preserving the whitespace!

tools/FakeItEasy.Build/Program.cs
```diff
@@ -98,11 +99,12 @@ public static void Main(string[] args)
                     DependsOn("initialize-user-properties"),
                     () =>
                         {
-                            var originalText = File.ReadAllText("FakeItEasy.user.props", Encoding.UTF8);
-                            var newText = Regex.Replace(originalText, "(?<=<BuildProfile>)[^>]*(?=</BuildProfile>)", profile);
-                            if (newText != originalText)
+                            var xmlDoc = XDocument.Load("FakeItEasy.user.props", LoadOptions.PreserveWhitespace);
+                            var buildProfileElement = xmlDoc.Descendants("BuildProfile").FirstOrDefault();
```
thomaslevesque: I tend to avoid using `Descendants`, because it can match elements unrelated to the one we're looking for (e.g. a `BuildProfile` element in an `ItemGroup`). I usually try to specify the path more precisely, e.g. `xmlDoc.Root.Elements("PropertyGroup").Elements("BuildProfile").FirstOrDefault()`.

Anyway, the current code doesn't handler the case where the `PropertyGroup` or `BuildProfile` element is missing... You probably need something like this:

```csharp
var buildProfileElement = xmlDoc.Root.Elements("PropertyGroup").Elements("BuildProfile").FirstOrDefault();
if (buildProfileElement is null)
{
    var propertyGroupElement = xmlDoc.Root.Element("PropertyGroup");
    if (propertyGroupElement is null)
    {
        propertyGroupElement = new XElement("PropertyGroup");
        xmlDoc.Root.Add(propertyGroupElement);
    }
    buildProfileElement = new XElement("BuildProfile");
    propertyGroupElement.Add(buildProfileElement);
}
...
```

Well, it ended up being much longer than the original version... But to be fair, it also handles more cases!

blairconrad: I figured there'd be a low chance that the file would exist and be that malformed. However, I wasn't considering that users would fail to be notified that the transformation wasn't made. Glad you thought of it.

Pushed a fixup. I deviated from your suggestions slightly

1. I started checking for missing elements at PropertyGroup, not BuildProfile
2. I assumed only a singled BuildProfile element in the PropertyGroup
3. (more a deviation from my original), I stopped preserving whitespace on load, since it easily allowed me to indent the output on save. Without that change, newly-added elements were on the same line as their parents

blairconrad: … and added a new fixup, since you seem to not have digested the last one yet. I extracted `GetOrAddElement`.

tools/FakeItEasy.Build/Program.cs
```diff
@@ -73,9 +76,57 @@ public static void Main(string[] args)
...
+            {
+                Target(
+                    "use-profile-" + profile,
+                    DependsOn("initialize-user-properties"),
+                    () =>
+                        {
+                            var xmlDoc = XDocument.Load("FakeItEasy.user.props");
+
+                            var propertyGroupElement = GetOrAddElement(xmlDoc.Root, "PropertyGroup");
+                            var buildProfileElement = GetOrAddElement(propertyGroupElement, "BuildProfile");
```
thomaslevesque: This assumes the `BuildProfile` element is in the first `PropertyGroup`... it might not be the case. That's why I first looked for `BuildProfile` in my suggestion.

blairconrad: Your error- and future-proofing is infuriating!
Infuriatingly correct and appropriate.

I've pushed a thing. It's exactly your thing, so it probably satisfies.
(Oh, as before, I'm no longer preserving whitespace, as it seemed the lesser of the two formatting evils.)

Sorry to have taken so much of your time with my whim.

thomaslevesque: > (Oh, as before, I'm no longer preserving whitespace, as it seemed the lesser of the two formatting evils.)

👍 

> Sorry to have taken so much of your time with my whim.

You haven't!

thomaslevesque:
Thanks, @blairconrad!

thomaslevesque:
Oh, I can't merge yet... you need to squash first

blairconrad:
Thanks. Squashed. I enjoyed your comments!

PR #1692, opened 2019-11-07
Suppress logo when testing and packing

blairconrad:
We do when building, so why not for the other targets?


thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you!

PR #1695, opened 2019-11-11
Run approval tests under netcoreapp3.0 in the full build profile

### LINKED ISSUE #1694: API approval gives different results for Windows and Linux default build profiles
… depending on the assemblies' public APIs.
Discovered when adding a public `MaybeNull` attribute to `A.Dummy<T>`. This obviously changed the public API, but on Windows, the attribute did not need to be added to `netstandard2.1.approved.txt`. I was surprised, but went with it. Then the Linux build [failed](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/builds/28743699/job/8m8valctd5025lwl). There are a few confounding problems with ApprovalTests, but the reason was that the `MaybeNull` did appear in `netstandard2.1.received.txt`.

It turns out that the difference is that the `full` build profile runs the approval tests as net461, but the `coreapp2+` profile, which is used for the Linux builds, runs the approval tests as netcoreapp3.0.

The safest approach would be to run the approval tests under all TFMs and safe the results separately, but that seems like a lot of work.

I'd consider running the approval tests under netcoreapp3.0 for the full build profile.  Every other build profile that targets netstandard2.1 (possibly with others) does so, and I think it makes sense to 
1. be consistent with the Linux build, and
2. use the newest framework

blairconrad:
Fixes #1694


blairconrad:
Thanks, sir!

PR #1697, opened 2019-11-13
Record ReturnValue on CompletedFakeObjectCall

### LINKED ISSUE #1696: ReturnValue not set
It seems like the ReturnValue of ICompletedFakeObjectCall is not being set. This test fails:

```c#
[Fact]
public void RetrnValueIsSet()
{
    var fake = A.Fake<IInterface>();

    var returnValue = fake.Function();

    Fake.GetCalls(fake).Single().ReturnValue.ShouldBe(returnValue);
}

public interface IInterface
{
    object Function();
}
```

Am I misunderstanding something - or is this a bug? 

Using FakeItEasy version 5.4.0 on netcore 3.0.

thomaslevesque:
Fixes #1696 

~~WIP because I want to try another approach~~ Scratch that, my idea won't work...

### DIFF
--- a/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
+++ b/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
@@ -5,17 +5,16 @@ namespace FakeItEasy.Core
 
     internal class CompletedFakeObjectCall : ICompletedFakeObjectCall
     {
-        public CompletedFakeObjectCall(IInterceptedFakeObjectCall interceptedCall, object[] arguments, object returnValue)
+        public CompletedFakeObjectCall(IInterceptedFakeObjectCall interceptedCall, object[] arguments)
         {
             this.FakedObject = interceptedCall.FakedObject;
             this.Method = interceptedCall.Method;
             this.Arguments = new ArgumentCollection(arguments, this.Method);
             this.ArgumentsAfterCall = interceptedCall.Arguments;
-            this.ReturnValue = returnValue;
             this.SequenceNumber = SequenceNumberManager.GetNextSequenceNumber();
         }
 
-        public object ReturnValue { get; }
+        public object ReturnValue { get; set; }
 
         public MethodInfo Method { get; }
 
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -159,8 +159,8 @@ namespace FakeItEasy.Core
                 listenerNode.Value.OnBeforeCallIntercepted(fakeObjectCall);
             }
 
-            var readonlyCall = fakeObjectCall.AsReadOnly();
-            this.RecordCall(readonlyCall);
+            var callToRecord = fakeObjectCall.ToCompletedCall();
+            this.RecordCall(callToRecord);
 
             try
             {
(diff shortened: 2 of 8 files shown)


src/FakeItEasy/Core/InterceptedFakeObjectCall.cs
```diff
@@ -21,10 +21,15 @@ internal abstract class InterceptedFakeObjectCall : IInterceptedFakeObjectCall
         public abstract MethodInfo Method { get; }
 
         /// <summary>
-        /// Freezes the call so that it can no longer be modified.
+        /// Gets or sets the return value of the call.
+        /// </summary>
+        public abstract object ReturnValue { get; internal set; }
```
thomaslevesque: I had to choose between adding a `GetReturnValue()` method, or change to a property. The accessor method pair felt to "Javaesque" for my taste...

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -159,18 +159,19 @@ void IFakeCallProcessor.Process(InterceptedFakeObjectCall fakeObjectCall)
                 listenerNode.Value.OnBeforeCallIntercepted(fakeObjectCall);
             }
 
-            var readonlyCall = fakeObjectCall.AsReadOnly();
-            this.RecordCall(readonlyCall);
+            var callToRecord = fakeObjectCall.ToCompletedCall();
```
thomaslevesque: Because it's no longer true that `CompletedFakeObjectCall` is read-only

src/FakeItEasy/Core/InterceptedFakeObjectCall.cs
blairconrad: The class is internal, so we could just leave the set public…

thomaslevesque: Good point!

src/FakeItEasy/Core/CompletedFakeObjectCall.cs
```diff
@@ -15,7 +15,7 @@ public CompletedFakeObjectCall(IInterceptedFakeObjectCall interceptedCall, objec
             this.SequenceNumber = SequenceNumberManager.GetNextSequenceNumber();
```
blairconrad: I can't make  a comment any higher, but do we expect the `returnValue` provided in the constructor to survive? Could it be dispensed with?

src/FakeItEasy/Creation/CastleDynamicProxy/CastleInvocationCallAdapter.cs
```diff
@@ -46,11 +46,17 @@ public CastleInvocationCallAdapter(IInvocation invocation)
...
         public override object FakedObject => this.invocation.Proxy;
 
+        public override object ReturnValue
+        {
+            get => this.invocation.ReturnValue;
+            internal set => this.invocation.ReturnValue = value;
+        }
+
         /// <summary>
         /// Freezes the call so that it can no longer be modified.
```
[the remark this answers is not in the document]
thomaslevesque: Oops

src/FakeItEasy/Core/CompletedFakeObjectCall.cs
thomaslevesque: Yes, I'll remove it. Thanks!

thomaslevesque:
> I wonder if it isn't worthwhile testing a faked delegate as well?

It is!

thomaslevesque:
> And I just thought of another thing `ICompletedFakeObjectCall.ReturnValue` has a slightly misleading comment, as I noted in the issue. Perhaps a good idea to fix up while you're here, if you'd be so kind, to indicate that it's the value that was returned, not the value _set to be returned_.

Will do

thomaslevesque:
> Squash it up!

Done

blairconrad:
Thanks, @thomaslevesque!

thomaslevesque:
Thanks for the review!

PR #1700, opened 2019-11-13
Have prepare-release manage support releases

### LINKED ISSUE #1698: prepare_release tool should handle support branches
Currently `prepare_release` will always copy and mutate the vNext milestone, release checklist issue, and release. We should be able to have a second (hopefully only second) set of these available for support work. So prepare_release should be augmented to create these as well as to use them for releasing.

Proposed new UI:

```
prepare_release next 6.0.0 vNext    # releases 6.0.0 from the current vNext set
prepare_release fork 5.vNext vNext  # creates a new set of release artifacts, using vNext as template
prepare_release next 5.4.1 5.vNext  # releases 5.4.1 from the current 5.vNext set
```

blairconrad:
Fixes #1698

### DIFF
--- a/tools/FakeItEasy.PrepareRelease/Program.cs
+++ b/tools/FakeItEasy.PrepareRelease/Program.cs
@@ -28,14 +28,13 @@ namespace FakeItEasy.PrepareRelease
             var existingReleaseName = args[2];
 
             var gitHubClient = GetAuthenticatedGitHubClient();
+            var existingMilestone = await gitHubClient.GetExistingMilestone(existingReleaseName);
+            var issuesInExistingMilestone = await gitHubClient.GetIssuesInMilestone(existingMilestone);
+            var existingReleaseIssue = GetExistingReleaseIssue(issuesInExistingMilestone, existingReleaseName);
 
             if (action == "next")
             {
                 var nextReleaseName = existingReleaseName;
-                var existingMilestone = await gitHubClient.GetExistingMilestone(existingReleaseName);
-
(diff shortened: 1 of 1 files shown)


blairconrad:
Given the annoyance of testing the tool, I didn't get super fancy with the implementation, and made two not-that-well-separated commits, mostly to make the diffing easier. Even assuming you want no more changes, I'm happy to squash.

Oh, I used the tool to create milestone 5.vNext. So that's something, right?

thomaslevesque:
Oh, wait! What about the docs?
prepare_release isn't mentioned in the "public" docs in the repo itself, but it's mentioned
- [in the wiki](https://github.com/FakeItEasy/FakeItEasy/wiki/Managing-releases) (doesn't seem to need an update, though)
- in the release issues, which need to be updated

Anyway, that's not something that can be done in a PR, so I'll merge anyway. I'll let you make the changes if you don't mind, I don't trust myself to correctly understand how to use the updated script...

blairconrad:
Thanks for the review and merge! Yeah, I thought about the docs too late, but came to the same realization that you did. I will for sure update.

blairconrad:
I did end up editing the wiki anyhow, and the two release checklist issues. Thanks!
It occurs to me that the tool _could_ automatically replace `vNext`s with `5.vNext`s in the release issue checklist when forking.

thomaslevesque:
> I did end up editing the wiki anyhow

Nice! Good idea to explicitly state the approach for support branches, so that we don't need to think too much about it next time.

> It occurs to me that the tool _could_ automatically replace `vNext`s with `5.vNext`s in the release issue checklist when forking.

Yes, it would be better, I think

PR #1703, opened 2019-11-14
Merge 5.4.1 into master

blairconrad:


blairconrad:
image

BUSTED! I'll branch from master and try to merge that.

thomaslevesque:
Ah, forgot about that... It's annoying. Means we can never merge a support branch directly into master, since it will never be up to date.

blairconrad:
We could relax the restriction about only merging branches that are up to date. I like it, but we always knew it might cause problems. Then again, I was mostly thinking of contributors, not us!
But I didn't mind doing the extra work.

thomaslevesque:
> We could relax the restriction about only merging branches that are up to date. I like it, but we always knew it might cause problems. Then again, I was mostly thinking of contributors, not us!
> But I didn't mind doing the extra work.

Let's keep it the restriction for now. If it becomes painful, we can always revisit

PR #1705, opened 2019-11-14
Add nullability annotations to public API

blairconrad:
Contributes to #1613. There'd be at least one followup to finish up the internal-only annotations and warnings.


blairconrad:
Hi, @thomaslevesque. As discussed, I'm attempting to merge my public API changes into the long-lived nullability branch. 
I think you've seen at least up to "Allow Dummies to be null" - a2c6bb9, and the others contain changes we've discussed. You might even have seen.

I'd consider squashing down to 3 commits, keeping the first two as is, since I really was just flailing randomly and expect that changes from various stated activities are all intermingled.

If the GitHub UI won't let us merge as a fast-forward merge, we can always merge by hand and push…

blairconrad:
I miscounted. I liked 2 commits. At least they both compile.

thomaslevesque:
Thanks, @blairconrad !

blairconrad:
Thank you!

Issue #1691, opened 2019-11-07
Heed nullability warnings in test code


blairconrad:
A companion to #1613. We talked about this once before and rejected it, as it seemed like a lot of overhead for little gain, but as I was typing up #1690, I realized that half the benefit of #1613 supposes that our clients will be heeding nullability warnings in _their_ test code, so perhaps we should revisit. I don't think there's any rush, though. We could enable after we're done implementing the warnings for our production code.

thomaslevesque:
👍 for enabling nullability checks in our tests

> We could enable after we're done implementing the warnings for our production code.

Yes, but I think we should do it before we release 6.0.0. It might help us detect issues in the public API nullability annotations.

afakebot:
This change has been released as part of [FakeItEasy 6.0.0-beta.1](https://github.com/FakeItEasy/FakeItEasy/releases/tag/6.0.0-beta.1).

PR #1707, opened 2019-11-16
Quell nullability warnings in src

blairconrad:
Contributes to #1613

### DIFF
--- a/src/FakeItEasy.Analyzer.CSharp/FakeItEasy.Analyzer.CSharp.csproj
+++ b/src/FakeItEasy.Analyzer.CSharp/FakeItEasy.Analyzer.CSharp.csproj
@@ -6,7 +6,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <FileAlignment>512</FileAlignment>
     <SignAssembly>true</SignAssembly>
-    <DefineConstants>$(DefineConstants);CSHARP</DefineConstants>
+    <DefineConstants>$(DefineConstants);CSHARP;LACKS_NULLABLE_ATTRIBUTES</DefineConstants>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
 
--- a/src/FakeItEasy/Configuration/FakeConfigurationManager.cs
+++ b/src/FakeItEasy/Configuration/FakeConfigurationManager.cs
@@ -31,7 +31,7 @@ namespace FakeItEasy.Configuration
             this.AssertThatMemberCanBeIntercepted(parsedCallExpression);
 
             var rule = this.ruleFactory.Invoke(parsedCallExpression);
-            var fake = Fake.GetFakeManager(parsedCallExpression.CallTarget);
+            var fake = Fake.GetFakeManager(parsedCallExpression.CallTarget!);
 
             return this.configurationFactory.CreateConfiguration(fake, rule);
         }
@@ -46,7 +46,7 @@ namespace FakeItEasy.Configuration
             this.AssertThatMemberCanBeIntercepted(parsedCallExpression);
 
             var rule = this.ruleFactory.Invoke(parsedCallExpression);
-            var fake = Fake.GetFakeManager(parsedCallExpression.CallTarget);
+            var fake = Fake.GetFakeManager(parsedCallExpression.CallTarget!);
 
             return this.configurationFactory.CreateConfiguration<T>(fake, rule);
         }
@@ -67,7 +67,7 @@ namespace FakeItEasy.Configuration
             GuardAgainstNonFake(parsedCallExpression.CallTarget);
             this.AssertThatMemberCanBeIntercepted(parsedCallExpression);
 
-            var fake = Fake.GetFakeManager(parsedCallExpression.CallTarget);
+            var fake = Fake.GetFakeManager(parsedCallExpression.CallTarget!);
             var parsedSetterCallExpression = PropertyExpressionHelper.BuildSetterFromGetter<TValue>(parsedCallExpression);
 
             return new PropertySetterConfiguration<TValue>(
@@ -76,7 +76,7 @@ namespace FakeItEasy.Configuration
                     this.CreateVoidArgumentValidationConfiguration(fake, newParsedSetterCallExpression));
         }
 
-        private static void GuardAgainstNonFake(object target)
+        private static void GuardAgainstNonFake(object? target)
         {
             if (target is object)
             {
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -65,7 +65,7 @@ namespace FakeItEasy.Core
         /// <param name="proxy">The faked proxy object.</param>
         /// <param name="fakeObjectName">The name of the fake object.</param>
         /// <returns>An instance of <see cref="FakeManager"/>.</returns>
-        internal delegate FakeManager Factory(Type fakeObjectType, object proxy, string fakeObjectName);
+        internal delegate FakeManager Factory(Type fakeObjectType, object proxy, string? fakeObjectName);
 
         /// <summary>
         /// Gets the faked proxy object.
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -101,7 +101,7 @@ namespace FakeItEasy.Core
         internal string FakeObjectDisplayName =>
             string.IsNullOrEmpty(this.FakeObjectName)
                 ? "Faked " + this.FakeObjectType
-                : this.FakeObjectName;
+                : this.FakeObjectName!;
 
         /// <summary>
         /// Adds a call rule to the fake object.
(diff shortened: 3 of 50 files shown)


blairconrad:
This could close issue 1613, or we could fold #1691 into that issue and fix it when I clean up the tests. I've started that work. Only 334 warnings to go!

src/FakeItEasy/Creation/ProxyGeneratorResult.cs
```diff
@@ -66,23 +66,22 @@ public ProxyGeneratorResult(object generatedProxy)
...
 
-            this.ProxyWasSuccessfullyGenerated = true;
             this.GeneratedProxy = generatedProxy;
         }
 
         /// <summary>
         /// Gets a value indicating whether the proxy was successfully created.
         /// </summary>
-        public bool ProxyWasSuccessfullyGenerated { get; }
+        public bool ProxyWasSuccessfullyGenerated => this.GeneratedProxy is object;
```
thomaslevesque: I thought about something the other day: we *could* use `is notnull` instead of `is object`! We'd just need a `using notnull = System.Object;` in every file... Of course, it could lead to weird things, such as `notnull.Equals(...)` or `new notnull()`, but I like the idea anyway ;)

src/FakeItEasy/Creation/ProxyOptions.cs
```diff
@@ -25,7 +26,8 @@ internal class ProxyOptions : IProxyOptions
 
         public IEnumerable<Expression<Func<Attribute>>> Attributes => this.attributes;
 
-        public string Name { get; set; }
+        [DisallowNull]
```
thomaslevesque: Is there any benefit to adding this attribute here? We never explicitly set it to null, but we could; it's internal anyway. Also we don't check if the assigned value is null, so the attribute basically a lie... Unless it's necessary to quell a warning, I'd be inclined to remove it.

src/FakeItEasy/Expressions/CallExpressionParser.cs
```diff
@@ -11,19 +11,12 @@ internal class CallExpressionParser : ICallExpressionParser
     {
         public ParsedCallExpression Parse(LambdaExpression callExpression)
         {
-            var methodExpression = callExpression.Body as MethodCallExpression;
-            if (methodExpression is object)
+            return callExpression.Body switch
```
thomaslevesque: I saw what you did here! 👍

src/FakeItEasy/Expressions/CallExpressionParser.cs
```diff
@@ -39,7 +32,7 @@ public ParsedCallExpression Parse(LambdaExpression callExpression, object fake)
         private static ParsedCallExpression ParseInvocationExpression(InvocationExpression expression)
         {
             var target = expression.Expression.Evaluate();
-            var method = target.GetType().GetMethod("Invoke");
+            var method = target!.GetType().GetMethod("Invoke");
```
thomaslevesque: Can we be sure `target` will never be null? It seems a pretty big assumption, when invoking an expression we know nothing about... Not sure what to do if it's null, though... I guess we'll just throw `NullReferenceException`. Which makes sense, since it would mean the user did something like this:
```csharp
Action action = null;
A.CallTo(() => action())...
```

src/FakeItEasy/Expressions/ParsedCallExpression.cs
```diff
@@ -16,7 +16,7 @@ internal class ParsedCallExpression
         {
             this.CalledMethod = calledMethod;
             this.ArgumentsExpressions = argumentsExpressions;
-            this.callTarget = new Lazy<object>(() => callTargetExpression?.Evaluate());
+            this.callTarget = new Lazy<object>(() => callTargetExpression?.Evaluate() !);
```
thomaslevesque: Two things:
- Why the `?` (I know, it was already there)?. `callTargetExpression` is of type `Expression` (not nullable), so it should never be null. And if it is, it's an error.
- How do we know `Evaluate` won't return null?

src/FakeItEasy/Expressions/CallExpressionParser.cs
thomaslevesque: Anyway, do we really need to evaluate the expression here? We only use `target` to get its type, but we could get the type from `expression.Expression.Type`. Just tried it, it seems to work.

src/FakeItEasy/ServiceLocator.cs
```diff
@@ -18,20 +18,20 @@ static ServiceLocator()
         [DebuggerStepThrough]
         internal static T Resolve<T>() where T : class
         {
-            var service = Service<T>.Instance;
+            var service = Service<T>.Instance!;
```
thomaslevesque: Why is this necessary? We check for null on the next line

src/FakeItEasy.Analyzer.CSharp/FakeItEasy.Analyzer.CSharp.csproj
```diff
@@ -6,7 +6,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <FileAlignment>512</FileAlignment>
     <SignAssembly>true</SignAssembly>
-    <DefineConstants>$(DefineConstants);CSHARP</DefineConstants>
+    <DefineConstants>$(DefineConstants);CSHARP;LACKS_NULLABLE_ATTRIBUTES</DefineConstants>
```
thomaslevesque: Is this necessary? We don't need the `#if`s, since there's a single TFM... Or is it future-proofing if we target netstandard2.1 someday?

thomaslevesque: Ah, I understand. The included file has `#if`s anyway, so we need the symbol

thomaslevesque:
@blairconrad The latest changes look good thanks! There's just one unresolved comment left. Maybe it's fine, but I think there might be cases where `callTargetExpression` will evaluate to null.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -101,7 +101,7 @@ public virtual IEnumerable<IFakeObjectCallRule> Rules
         internal string FakeObjectDisplayName =>
             string.IsNullOrEmpty(this.FakeObjectName)
                 ? "Faked " + this.FakeObjectType
-                : this.FakeObjectName;
+                : this.FakeObjectName!;
```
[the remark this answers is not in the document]
thomaslevesque: No, it's fine, I think

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -147,7 +147,7 @@ bool IsAccessibleImpl(Type t)
             // Return the result if the return type is non-void
             if (!isVoid)
             {
-                bodyExpressions.Add(result);
+                bodyExpressions.Add(result!);
```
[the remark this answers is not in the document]
thomaslevesque: I'm not sure it makes it more readable... Especially the `Insert`s at the end, which are a bit confusing. Anyway, I commented because I was surprised, but I wasn't asking for a change

src/FakeItEasy/Creation/ProxyGeneratorResult.cs
thomaslevesque: That was the idea

src/FakeItEasy/Expressions/ParsedCallExpression.cs
blairconrad: You're right, `callTargetExpression` could evaluate to `null`. In particular when clients try to configure a static method. We catch the nullness/staticness later, but for a while, we can have a null call target.
I've fixed up. The nullability propagated bit. In 4 places I resorted to the `!` operator. In 3 of them, it's fairly obvious that it's safe. In `PropertyExpressionHelper`, I added comments to clear it up a bit…

Longer term, we might want to consider refactoring the determination of whether a method can be intercepted, but I'm leery of mucking about too much in the middle of the nullability transformation...

thomaslevesque: > I'm leery of mucking about too much in the middle of the nullability transformation...

Totally agree!

src/FakeItEasy/Configuration/FakeConfigurationManager.cs
```diff
@@ -76,7 +76,7 @@ public IPropertySetterAnyValueConfiguration<TValue> CallToSet<TValue>(Expression
                     this.CreateVoidArgumentValidationConfiguration(fake, newParsedSetterCallExpression));
         }
 
-        private static void GuardAgainstNonFake(object target)
+        private static void GuardAgainstNonFake(object? target)
```
thomaslevesque: Too bad there isn't an attribute to say "this method will throw if the argument is null". This way the compiler will now that the argument can't be null after the call. R# has an attribute like that ([example](https://github.com/thomaslevesque/Linq.Extras/blob/528233fef751cdfefbec3bb10096b96c41cd4d9e/src/Linq.Extras/Internal/ArgumentExtensions.cs#L10)).

blairconrad:
Thanks, @thomaslevesque. Excellent comments!

thomaslevesque:
Thanks @blairconrad!

PR #1708, opened 2019-11-19
(Re)-allow constraining arguments to be collections that contain null

blairconrad:
I broke this in #1686. Fortunately, we've not released, so we can just fold the fix in. On that basis, labelling as enhancement, like the original was.


blairconrad:
Thanks, @thomaslevesque.

I'd feel better about the catch if it wasn't necessary due to my silly error. I'm like one of those superheroes that has to neutralize an arch-nemesis that they created.

Oh, I found it while working on nullability woes in the tests. AND we'll need to update the API to add some `?`s. May I rebase nullability on top of master now so we get cleaner commits later?

thomaslevesque:
> May I rebase nullability on top of master now so we get cleaner commits later?

Sure!

PR #1718, opened 2019-11-26
Add CallsWrappedMethod

### LINKED ISSUE #1717: Add support for calling the wrapped object's method
Currently, when using a wrapping fake like this:

```csharp
var foo = A.Fake<IFoo>(o => o.Wrapping(new Foo()));
```

There's no way to explicitly call the method of the wrapped object. It's the default behavior, but as soon as a call is configured, this behavior is lost and there's no way to restore it. For instance, if we do this:

```csharp
A.CallTo(() => foo.Bar()).Invokes(() => DoSomething());
```

The `Bar` method will return a default value, instead of the value from the wrapped object.

A possible workaround is to do something like this:

```csharp
var realFoo = new Foo();
var foo = A.Fake<IFoo>(o => o.Wrapping(realFoo));
A.CallTo(() => foo.Bar()).Invokes(() => DoSomething()).ReturnsLazily(() => realFoo.Bar());
```

It's reasonably simple in this specific case, but if the method had arguments it would be much uglier.

A better alternative would be to introduce a new `CallsWrappedMethod` method:

```csharp
var foo = A.Fake<IFoo>(o => o.Wrapping(new Foo()));
A.CallTo(() => foo.Bar()).Invokes(() => DoSomething()).CallsWrappedMethod();
```

`CallsWrappedMethod` would throw if the fake is not a wrapping fake, of course.

thomaslevesque:
Fixes #1717 

Should probably be merged after the nullability branch is merged, as there might be additional nullability related changes to make.


blairconrad:
> Should probably be merged after the nullability branch is merged, as there might be additional nullability related changes to make.

Thanks, @thomaslevesque. And to that end, I'm not going to look at the code or even the _docs changes which I'm sure you made_ for a little bit, while I focus on the nullability changes. Yup. Not even going to see which files you changed.

thomaslevesque:
> the _docs changes which I'm sure you made_

Oops. I actually thought about it, searched the current docs for `CallsBaseMethod`, and found none, so I thought it would be ok to not document `CallsWrappedMethod`... But now I realized Visual Studio lied to me (or, more likely, that I don't know how to search...)

blairconrad:
> so I thought it would be ok to not document `CallsWrappedMethod`

I feel attacked

thomaslevesque:
No need to be sorry!
I'll update the PR to add nullability annotations after we merge the nullability branch

thomaslevesque:
> Except maybe this isn't the way to start checking in `.vscode`?
> I assume this was accidental.

Oops! That's what I get from working when I should be sleeping.
Fixed

PR #1720, opened 2019-11-27
Add nullability annotations to public API

blairconrad:
Supports #1613.

### DIFF
--- a/src/FakeItEasy/Fake.cs
+++ b/src/FakeItEasy/Fake.cs
@@ -75,7 +75,7 @@ namespace FakeItEasy
         /// <param name="fakeManager">The fake manager, or <c>null</c> if <paramref name="potentialFake"/> is not actually a faked object.</param>
         /// <returns><c>true</c> if <paramref name="potentialFake"/> is a faked object, else <c>false</c>.</returns>
         [DebuggerStepThrough]
-        public static bool TryGetFakeManager(object potentialFake, out FakeManager fakeManager)
+        public static bool TryGetFakeManager(object potentialFake, [NotNullWhen(true)] out FakeManager? fakeManager)
         {
             Guard.AgainstNull(potentialFake, nameof(potentialFake));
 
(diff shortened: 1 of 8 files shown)


blairconrad:
Should include all the API annotations we need. I've quelled nullability warnings up through the integration tests with no further changes to the public API and only one alteration to an internal class in src. I have those changes ready to go after this is merge. Then I'll work on Specs and tools.

Oh, if/when you merge, don't forget the 3rd merge option!

src/FakeItEasy/Fake.cs
```diff
@@ -75,7 +75,7 @@ public static void ClearRecordedCalls(object fakedObject)
         /// <param name="fakeManager">The fake manager, or <c>null</c> if <paramref name="potentialFake"/> is not actually a faked object.</param>
         /// <returns><c>true</c> if <paramref name="potentialFake"/> is a faked object, else <c>false</c>.</returns>
         [DebuggerStepThrough]
-        public static bool TryGetFakeManager(object potentialFake, out FakeManager fakeManager)
+        public static bool TryGetFakeManager(object potentialFake, [NotNullWhen(true)]out FakeManager? fakeManager)
```
thomaslevesque: I notice you didn't add a space after the attribute. In at least one other place, you did. I think it would be better with a space. If you don't want to change it now, we can do it in a subsequent PR.

blairconrad: Oh. I wasn't thinking about it one way or another. Happy to add. Amend suits?

thomaslevesque: > Amend okay?

Perfect

thomaslevesque:
Thanks, @blairconrad

blairconrad:
Amended and pushed!

… I guess I was late to the party there. I'd been watching the build, and got distracted by work. Thanks for the comment and merge!

PR #1722, opened 2019-11-28
Heed nullability warnings in tools code

blairconrad:


tools/Directory.Build.props
```diff
@@ -2,6 +2,11 @@
 
     <Import Project="../Directory.Build.props" />
 
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>netcoreapp3.0</TargetFramework>
```
blairconrad: not strictly necessary, but I figured we'd get better annotations from the libraries, like knowing what `IsNullOrEmpty` means for its arguments.

tools/ReleaseHelpers.cs
```diff
@@ -12,12 +13,10 @@ public static ICollection<int> GetIssueNumbersReferencedFromReleases(IEnumerable
...
             {
-                foreach (Match match in Regex.Matches(release.Body, @"\(\s*#(?<issueNumber>[0-9]+)(,\s*#(?<issueNumber>[0-9]+))*\s*\)"))
+                foreach (var capture in Regex.Matches(release.Body, @"\(\s*#(?<issueNumber>[0-9]+)(,\s*#(?<issueNumber>[0-9]+))*\s*\)")
+                                             .SelectMany(match => match.Groups["issueNumber"].Captures))
                 {
-                    foreach (Capture capture in match.Groups["issueNumber"].Captures)
-                    {
-                        issuesReferencedFromRelease.Add(int.Parse(capture.Value, NumberStyles.Integer, NumberFormatInfo.InvariantInfo));
-                    }
+                    issuesReferencedFromRelease.Add(int.Parse(capture.Value, NumberStyles.Integer, NumberFormatInfo.InvariantInfo));
```
blairconrad: There were warnings all in here, and I noticed that VS wanted to Linq up part of the query, so I let it, and... the warnings went away!

tools/Directory.Build.props
thomaslevesque: :+1:

blairconrad:
Thanks, @thomaslevesque. I figured we both needed a palate cleanser before what should be my last nullability PR: Specs. I'll start working on those tonight.

PR #1723, opened 2019-11-29
Annotate WithNonVoidReturnType to include nullable reference return types

blairconrad:
Contributes to #1613.


blairconrad:
One last annotation that I found we needed while working on cleaning up the specs.

thomaslevesque:
Thanks @blairconrad!

PR #1725, opened 2019-12-01
[WIP] Internalize MayBeNull attribute

blairconrad:
Contributes to #1613


blairconrad:
Only the last commit actually applies. Will rebase once #1724 merges.

blairconrad:
Note that approved API descriptions all changed, with the exception of netstandard2.1.

tests/FakeItEasy.Tests.Approval/ApprovedApi/FakeItEasy/net40.approved.txt
```diff
@@ -198,9 +198,7 @@ namespace FakeItEasy
     [System.Runtime.CompilerServices.NullableAttribute(0)]
     public class static FakeObjectCallExtensions
     {
-        [return: System.Diagnostics.CodeAnalysis.MaybeNullAttribute()]
```
thomaslevesque: It's annoying that we're losing the annotations, because it means we won't see changes to the nullability attributes... But the alternative is to expose a copy of the attributes that could conflict with an existing one.
There's a new version of PublicApiGenerator (released just hours ago!) that adds support for nullable ref types, maybe it does the right thing? Worth trying, I think.

thomaslevesque:
Shouldn't this PR be merged into nullability rather than master?

blairconrad:
Bother. Thanks

PR #1727, opened 2019-12-01
Add nullability annotations and heed warnings

### LINKED ISSUE #1613: Expose nullability annotations (and enable nullability warnings internally in production code)

### LINKED ISSUE #1691: Heed nullability warnings in test code

blairconrad:
Fixes #1613 and fixes #1691.


blairconrad:
Note that I'm merging from blairconrad/nullability. FakeItEasy/nullability still exists for reference (and has all the same content).

blairconrad:
Thank you, @thomaslevesque. It was a lot of work for you too. Thanks for the excellent reviews. Sorry for the bumbling delays.

PR #1730, opened 2019-12-02
Upgrade StyleCop.Analyzers to 1.2.0-beta.113

blairconrad:
Because it's nice to upgrade.
Also because the new rules understand constructs like 
`var eventInfo = GetEvent(fakeObjectCall.Method)!`, which I'd longed for while working on the nullability branch.

### DIFF
--- a/tests/FakeItEasy.Specs/DummyCreationSpecs.cs
+++ b/tests/FakeItEasy.Specs/DummyCreationSpecs.cs
@@ -870,7 +870,7 @@ namespace FakeItEasy.Specs
     {
         protected override T CreateDummy<T>()
         {
-            return (T)Sdk.Create.Dummy(typeof(T)) !;
+            return (T)Sdk.Create.Dummy(typeof(T))!;
         }
 
         protected override IList<T> CreateCollectionOfDummy<T>(int count)
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -11,7 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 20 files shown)


tests/FakeItEasy.Specs/DummyCreationSpecs.cs
```diff
@@ -455,7 +455,7 @@ public void TupleCreation(Tuple<Foo, string> dummy)
         }
 
         [Scenario]
-        public void ValueTupleCreation((Foo foo, string s) dummy)
+        public void ValueTupleCreation((Foo Foo, string S) dummy)
```
thomaslevesque: Did StyleCop give a warning for this? It's annoying... I tend to prefer camelCase name for tuple members. If you agree, we can probably configure the rule.

blairconrad: It did warn. I have no strong opinion. Will reconfigure for camelCase.

blairconrad:
Updated. Adding `stylecop.json` is not straightforward!

Also I had missed a warning, so I got that too.

Working on FxCop for another PR…

thomaslevesque:
> Adding `stylecop.json` is not straightforward!

Ah, sorry, I didn't intend to give you much work... Well, at least it's done now, if we ever need to configure something else!

blairconrad:
Squashed. Thank you!

And the `settings.json` wasn't _that_ much work. But the documentation could be clearer. "The easiest way to…" is, as far as I can tell, "The only documented way to…", but I looked around longer than I should've to find the alternatives, because I didn't want to fire up VS2019 and force another spurious warning!

As you say, now we have the template.

thomaslevesque:
Thanks @blairconrad!

PR #1732, opened 2019-12-03
Upgrade Microsoft.CodeAnalysis.FxCopAnalyzers to 2.9.8

blairconrad:


blairconrad:
I created separate commits this time because FxCop recommendations are often contentious!

src/FakeItEasy/Configuration/BuildableCallRule.cs
```diff
@@ -77,7 +77,7 @@ public void UseApplicator(Action<IInterceptedFakeObjectCall> newApplicator)
         {
             if (this.wasApplicatorSet)
             {
-                throw new InvalidOperationException("The behavior for this call has already been defined");
+                throw new InvalidOperationException(ExceptionMessages.CallBehaviorAlreadyDefined);
```
blairconrad: General comment on this commit: technically FxCop just complained about string literals in the constructors. I hunted down other constructed arguments and moved them to `ExceptionMessages` as well.

src/FakeItEasy.Analyzer.Shared/ArgumentConstraintAnalyzerBase.cs
```diff
@@ -12,7 +12,7 @@ namespace FakeItEasy.Analyzer
     using Microsoft.CodeAnalysis.VisualBasic.Syntax;
 #endif
 
-    public abstract class ArgumentConstraintAnalyzerBase : DiagnosticAnalyzer
+    internal abstract class ArgumentConstraintAnalyzerBase : DiagnosticAnalyzer
```
blairconrad: Other analyzers were internal, so I figured why not?

src/FakeItEasy.ruleset
```diff
@@ -5,6 +5,7 @@
     <Rule Id="CA1020" Action="None" />
     <Rule Id="CA1200" Action="None" />
     <Rule Id="CA1305" Action="None" />
+    <Rule Id="CA2237" Action="None" /> <!-- Add [Serializable] to [...] as this type implements ISerializable -->
```
blairconrad: Opinionated suppression: I don't see much point in marking our exceptions as serializable. Open to change.

thomaslevesque: I'm in complete agreement

src/FakeItEasy.Analyzer.Shared/ArgumentConstraintAnalyzerBase.cs
blairconrad: Hmm. I only have 2017 at work, and it seems that _none_ of the analyzers work in 5.5.0. When I roll back to 4.9.2, `NonVirtualSetupAnalyzer`, which is internal works, as does `ArgumentConstraintOutsideCallSpecAnalyzer`, which is public.

thomaslevesque: That's annoying... So, it doesn't matter that they're internal, but we must have broken something at some point. I'll try with VS2019.

thomaslevesque: Works in VS2019.
We upgraded the analyzers to .NET Standard 2.0 in 5.4.0 (#1667); that's probably the cause. I didn't expect that they would stop working in VS2017... Well, people stuck on VS2017 can keep using an older version, I guess. But that means we probably should decouple the analyzer version from the library version...

src/FakeItEasy/ExceptionMessages.cs
```diff
@@ -32,6 +33,27 @@ internal static class ExceptionMessages
...
         public static string NotAWrappingFake => "The configured fake is not a wrapping fake.";
 
+        public static string CallBehaviorAlreadyDefined =>
+            "The behavior for this call has already been defined";
+
+        public static string OutAndRefBehaviorAlreadyDefined =>
+            "How to assign out and ref parameters has already been defined for this call";
+
+        public static string NumberOfTimesNotGreaterThanZero =>
+            "The number of times to occur is not greater than zero.";
```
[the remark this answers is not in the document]
blairconrad: I wouldn't say "should", but "could" be, and I'll change if you really want, but I just transferred the existing message, and many of the others are phrased as complaints, like "The configured fake is not a wrapping fake." on line 34, which could be "The configured fake must be a wrapping fake."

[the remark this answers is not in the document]
thomaslevesque: You're right, of course. If we want to revisit the wording, we'll do it in another PR.

blairconrad:

thomaslevesque:
Thanks!

PR #1734, opened 2019-12-05
Use alternate when argument value formatter throws or returns null

### LINKED ISSUE #1690: Use builtin formatters when IArgumentValueFormatter.GetArgumentValueAsString throws an exception or returns null
From a conversation in #1683. At that point I suggested a throwing a `UserCallbackException`, but the documentation for that class says it's "An exception thrown when a user-provided callback throws an exception", so if we want to throw, we should either consider a different exception or update the docs for that class.
However, Maybe throwing isn't the best course of action. We're already in an error condition, so derailing that just to tell the user about a different error could confuse things. Perhaps we should just construct a non-null message in its place. Clients who turn nullability warnings on in their test projects will be alerted that `null` isn't the best return value anyhow.

blairconrad:
Fixes #1690.

### DIFF
--- a/src/FakeItEasy/Core/ArgumentValueFormatter.cs
+++ b/src/FakeItEasy/Core/ArgumentValueFormatter.cs
@@ -12,13 +12,10 @@ namespace FakeItEasy.Core
 
     internal class ArgumentValueFormatter
     {
-        private readonly ConcurrentDictionary<Type, IArgumentValueFormatter> cachedFormatters;
         private readonly IEnumerable<IArgumentValueFormatter> typeFormatters;
 
         public ArgumentValueFormatter(IEnumerable<IArgumentValueFormatter> typeFormatters)
         {
-            this.cachedFormatters = new ConcurrentDictionary<Type, IArgumentValueFormatter>();
-
             this.typeFormatters = typeFormatters.Concat(
                 new IArgumentValueFormatter[]
                     {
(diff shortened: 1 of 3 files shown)


blairconrad:
I tried a _slightly_ different tack. Rather than falling back to a builtin formatter, I just fell back up the list of available formatters until we find one that applies and doesn't return null.

I think we could take this behaviour, but I did have some alternate thoughts. I'm just thinking out loud here

1. is it worthwhile caching the formatters? Ideally, we rarely use them because users' tests mostly pass. The caching introduces some complexity and also runs the risk that we cache a formatter that works on _one particular value_ and less well on others.
2. we _could_ fall back in a similar way for the formatters that throw. Arguably returning `null` or throwing are both error conditions. We could make a best-case effort to format the value regardless.

blairconrad:
Just for comparison, here's how a non-caching-falling-back-in-the-face-of-exceptions formatter would look:
https://github.com/blairconrad/FakeItEasy/tree/alternate-fallback

I'm not married to it. It was just a fun exercise.

thomaslevesque:
> * is it worthwhile caching the formatters? Ideally, we rarely use them because users' tests mostly pass. The caching introduces some complexity and also runs the risk that we cache a formatter that works on _one particular value_ and less well on others.

Agreed. I don't think the cost of reflection is going to be noticeable in real world usage, and the behavior is actually better without caching.

> * we _could_ fall back in a similar way for the formatters that throw. Arguably returning `null` or throwing are both error conditions. We could make a best-case effort to format the value regardless.

Yes, I think formatters that throw should be handled the same as formatters that return null.

Otherwise, I have no comment on the current implementation, but you're going to replace it anyway, so I'll wait for your next commit ;)

src/FakeItEasy/Core/ArgumentValueFormatter.cs
```diff
@@ -37,16 +34,26 @@ public virtual string GetArgumentValueAsString(object? argumentValue)
...
-                throw new UserCallbackException(ExceptionMessages.UserCallbackThrewAnException($"Custom argument value formatter '{formatter}'"), ex);
+                try
+                {
+                    var formattedValue = formatter.GetArgumentValueAsString(argumentValue);
+                    if (formattedValue is object)
+                    {
+                        return formattedValue;
+                    }
+                }
+                catch when (formatter.GetType().GetTypeInfo().Assembly != typeof(ArgumentValueFormatter).GetTypeInfo().Assembly)
```
thomaslevesque: I think there are several places where we do this kind of check... Maybe we should have a `IsFakeItEasyType` method somewhere. (out of scope for this PR, of course)

thomaslevesque:
Thanks, @blairconrad!

src/FakeItEasy/Core/ArgumentValueFormatter.cs
blairconrad: I cannot find another place!
We _do_ save away the the assembly in `TypeCatalogue` though…

blairconrad:
Thanks for the feedback and merge, @thomaslevesque!

PR #1738, opened 2019-12-06
Quiet warning about catching general exception in ArgumentValueFormatter

blairconrad:
Completes #1734.

I don't know how I missed the warning!

### DIFF
--- a/src/FakeItEasy/Core/ArgumentValueFormatter.cs
+++ b/src/FakeItEasy/Core/ArgumentValueFormatter.cs
@@ -2,8 +2,8 @@ namespace FakeItEasy.Core
 {
     using System;
     using System.Collections;
-    using System.Collections.Concurrent;
     using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


src/FakeItEasy/Core/ArgumentValueFormatter.cs
```diff
@@ -25,6 +25,7 @@ public ArgumentValueFormatter(IEnumerable<IArgumentValueFormatter> typeFormatter
                     });
         }
 
+        [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Any type of exception may be encountered.")]
```
thomaslevesque: I don't think this warning makes sense when there's a `when` clause anyway... This should probably be changed in FxCop analyzers.

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you.

Issue #1735, opened 2019-12-05
Move analyzers to a separate repository


thomaslevesque:
When we first released the FakeItEasy analyzers, we decided to version them together with the library. For every release of the library, the analyzers were also released, even if they had not changed.

This approach has several issues:
- It conveys the wrong message, as we make many releases of the analyzers that differ only by their version number
- The analyzers don't really follow SemVer. We only bump the major version if there's a breaking change in the library.
- Since the analyzers are in the same solution as the library, it increases build times

We should consider moving the analyzers to a separate repository, and version them independently of the library.

thomaslevesque:
To do:
- [ ] Move shared build and release tools to a new repository
  - [ ] Create a new repos for shared tools
  - [ ] Copy shared tools to new repo
  - [ ] Include shared tools in FakeItEasy using submodules and remove original tools
- [ ] Create new repo for analyzers ("FakeItEasy.Analyzers"?)
- [ ] Copy source and tests of analyzers to new repo
  - [ ] Keep git history? Possible, but is it worth it?
- [ ] Configure AppVeyor to build and release analyzers
- [ ] Remove source and tests of analyzers from main library repo
- [ ] Decide what to do about the docs: put analyzer docs in the new repo, or keep them in the main repo? There's only one page regarding the analyzers, and it might be difficult to publish it to the same ReadTheDocs site if it's in a different repo. On the other hand, the "stable" docs are only updated when we release a new version of the library, so... Maybe use [subprojects](https://docs.readthedocs.io/en/stable/subprojects.html)?

thomaslevesque:
Regarding the tools, it's going to be a bit of a mess. The scripts have hard-coded relative paths to the artifacts, and hard-coded values (e.g. repo name in PrepareRelease).
I think the right approach is to parameterize the C# scripts, and just have specific cmd scripts in the consuming repo.

blairconrad:
> I think the right approach is to parameterize the C# scripts, and just have specific cmd scripts in the consuming repo.

Yeah. I'd assume as much as well.

thomaslevesque:
> * [ ]  Keep git history? Possible, but is it worth it?

@blairconrad unless you think it's important, I'd rather not preserve the history. It's possible, but it's going to be a pain.

blairconrad:
I'm not sure it's super important, but mind if I try something?

thomaslevesque:
> I'm not sure it's super important, but mind if I try something?

[...]

PR #1741, opened 2019-12-09
Purge all vestiges of the FakeItEasy Analyzer

blairconrad:
Contributes to #1735.


mkdocs.yml
```diff
@@ -37,7 +37,7 @@ pages:
 - Scanning for Extension Points: scanning-for-extension-points.md
 - Bootstrapper: bootstrapper.md
 - Advanced usage: advanced-usage.md
-- Analyzer: analyzer.md
+- Analyzer: https://fakeiteasyanalyzers.readthedocs.io/
```
blairconrad: 1. I figured there's no point in directing people to our obsolete analyzers.md page
2. Based on the subproject directory structure, I saw no way to make a decent relative link to the analyzer docs, so I used a raw URL like anybody else would

mkdocs.yml
```diff
@@ -4,7 +4,7 @@ theme: readthedocs
 repo_url: https://github.com/FakeItEasy/FakeItEasy
 extra_css: [css/extra-highlight.css]
 
-pages:
+nav:
```
blairconrad: I got a warning when I ran locally, so I updated. I could pretend that this is the change that lets me link to a URL instead of a relative .md file down on line 40, but honestly, I did not try.

mkdocs.yml
thomaslevesque: I didn't know there was a top-level URL for subprojects. I only knew of this one: https://fakeiteasy.readthedocs.io/projects/fakeiteasyanalyzers/en/latest/.

Maybe you could link `/projects/fakeiteasyanalyzers/` instead? this way it would still appear "under" the FakeItEasy root doc site. And maybe we could change the alias to just "analyzers".

profiles/full.props
```diff
@@ -7,7 +7,6 @@
     <SpecsTargetFrameworks>net461;netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</SpecsTargetFrameworks>
     <IntegrationTestsTargetFrameworks>net461;netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</IntegrationTestsTargetFrameworks>
     <UnitTestsTargetFrameworks>net461;netcoreapp1.0;netcoreapp2.1;netcoreapp3.0</UnitTestsTargetFrameworks>
-    <AnalyzerTestsTargetFrameworks>net461;netcoreapp2.1;netcoreapp3.0</AnalyzerTestsTargetFrameworks>
```
blairconrad: Forgot these the first time 'round.

mkdocs.yml
blairconrad: subprojects are projects in their own right, so there's a top-level.
I didn't even pick the original subproject URL, and when I tried to change, it got all snippy with me. Ended up deleting and readding after you asked for the change.

PR #1744, opened 2019-12-11
Remove unnecessary assembly attributes and add missing ones

thomaslevesque:
#### First commit: remove unnecessary assembly attributes

- `AssemblyTrademark`: we set it to an empty string anyway
- `NeutralResourcesLanguage`: the assembly doesn't have any embedded resource
- `AssemblyCulture`: this is normally used only for satellite assemblies. If I understand the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assemblycultureattribute?view=netframework-4.8) correctly, there's no reason at all to include it

Now if only we could remove `ComVisible`, we could get rid of CommonAssemblyInfo.cs. But there's no MSBuild attribute to control it, and if we don't include it, the assembly is COM-visible by default. Of course it's possible to do this, but it's a bit ugly:

```xml
    <AssemblyAttribute Include="System.Runtime.InteropServices.ComVisible">
      <_Parameter1>false</_Parameter1>
    </AssemblyAttribute>
```

#### Second commit: add `ComVisible(false)` to FakeItEasy.Extensions.ValueTask

It didn't include CommonAssemblyInfo.cs, so it didn't have `ComVisible(false)`, and was COM-visible by default

---

Now that there's only a single attribute in CommonAssemblyInfo.cs, maybe it's not worth the trouble of having a shared file. We could just do the ugly trick mentioned above in Directory.Build.props.


thomaslevesque:
Oh, wait, I messed up my commits. I'll fix it.

thomaslevesque:
> Oh, wait, I messed up my commits. I'll fix it.

Done.



> We could just do the ugly trick mentioned above in Directory.Build.props.

Actually, it won't work because of https://github.com/microsoft/msbuild/issues/2281 :(

blairconrad:
> Remove unnecessary assembly attributes and add missing ones

Why do we want to add missing unnecessary assembly attributes?

Sorry. I'll stop.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -23,6 +23,7 @@
     <Compile Include="../FakeItEasy/ValueTaskReturnValueConfigurationExtensions.StronglyTyped.cs" />
     <Compile Include="../FakeItEasy/ValueTaskAsyncReturnValueConfigurationExtensions.cs" />
     <Compile Include="../FakeItEasy/ValueTaskAsyncReturnValueConfigurationExtensions.StronglyTyped.cs" />
+    <Compile Include="../*.cs" />
```
blairconrad: I'm perhaps being silly, but could this be added to the `Directory.Build.props`, since it's copied in both `.csproj`s now?

thomaslevesque: Sure, I can do that.

src/Directory.Build.props
```diff
@@ -21,8 +21,12 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
   </ItemGroup>
 
+    <ItemGroup>
```
blairconrad: nit: indentation.
Otherwise, I'm thrilled. Don't send a fixup. Just squash locally or amend or whatever.

blairconrad: Sorry! I made more work for you!

thomaslevesque: Oops. Will fix.

thomaslevesque: Done

blairconrad:
Thanks, M. Levesque!

PR #1747, opened 2020-01-10
Add Contains overload with a StringComparison

### LINKED ISSUE #1681: Add ArgumentConstraintManagerExtensions.Contains overload that takes a StringComparison
Currently we force the comparison to use the current culture.

thomaslevesque:
Fixes #1681

### DIFF
--- a/docs/argument-constraints.md
+++ b/docs/argument-constraints.md
@@ -65,7 +65,8 @@ used. There are a few built-in matchers:
 |IsEqualTo(other)|object equality using `object.Equals`|
 |IsSameAs(other)|object identity - like `object.ReferenceEquals`|
 |IsInstanceOf(type)|an argument that can be assigned to a variable of type `type`|
-|Contains(string)|substring match|
+|Contains(string)|substring match with ordinal string comparison|
+|Contains(string, comparisonType)|substring match with the specified comparison type|
 |StartsWith(string)|substring at beginning of string|
 |EndsWith(string)|substring at end of string|
 |IsNullOrEmpty()|`null` or `""`|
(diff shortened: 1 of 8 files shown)


thomaslevesque:
> We're adding a new feature _and_ fixing the "bug" you mentioned in the issue at the same time.

True. I thought about separating, and eventually I didn't because. I'll change it if I can do it quickly (if not, it will have to wait, because I'm leaving for the weekend)

thomaslevesque:
Oops, looks like I didn't read your comment properly... I made a separate commit in the same branch, instead of a separate PR. If you really want a separate PR, I'll do it when I get back home.

> (and even potentially port the fix to the 5.x stream, if we feel like it)

Ah, I didn't think about this. I tought it only affected 6.x, but now I realize it was part of the `ValueTask` work, so it's on 5.x. We could fix it on 5.x, but to be honest, I'm not sure it's worth the trouble; it's unlikely to affect anyone, and 6.0.0 is coming soon.

OK, I'm out. See you later

blairconrad:
I don't want to make more work (appearances to the contrary), but I keep landing on the conclusion that we should split the fix and the enhancement.  I agree that we can hold off (and possibly never implement) the fix on the 5.x stream.

Oh. It might be worthwhile including the new overload in the docs (and indicate which StringComparison the 1-parameter version uses while we're at it…)

thomaslevesque:
Rebased

thomaslevesque:
> Oh. It might be worthwhile including the new overload in the docs (and indicate which StringComparison the 1-parameter version uses while we're at it…)

Done

thomaslevesque:
Thanks!

PR #1752, opened 2020-01-22
Embed PDBs in assemblies

### LINKED ISSUE #1685: Update source stepping documentation
The [Source Stepping](https://fakeiteasy.readthedocs.io/en/stable/source-stepping/) documentation page is out of date. It should be updated with instructions regarding SourceLink.

[This article](https://www.hanselman.com/blog/ExploringNETCoresSourceLinkSteppingIntoTheSourceCodeOfNuGetPackagesYouDontOwn.aspx) by Scott Hanselman has some configuration instructions.

### LINKED ISSUE #1751: Embed PDBs in assemblies
Currently, source-stepping with SourceLink doesn't work in some scenarios, due to differences in the SDK behavior. For instance, the .NET Core 3 SDK copies the assemblies from referenced packages to the output folder, but not the PDB (see https://github.com/dotnet/sdk/issues/1458). So the debugger doesn't find the PDB, so it can't find the source.

A simple workaround would be to use the embedded PDB format, which is basically the same as the portable format we're currently using, except that it's embedded in the DLL.

**Benefits**
- The PDB will always be available, it will no longer depends on whether the SDK copied it or not, leading to a better debugging experience for users.
- Packaging will become simpler, since we don't have to worry about packaging the PDB

**Downsides**
- The assemblies will be larger.

I don't think the increased assembly size is a major issue ; the PDB is already downloaded with the package anyway. And size itself doesn't really matter, since FakeItEasy is mostly a dev dependency (not shipped with apps). Anyway, the PDB files aren't that big now (about 80KB, when the assembly is around 250KB)

thomaslevesque:
Fixes #1751
Fixes #1685 

Two birds with one stone! I hope you don't mind...


src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -40,7 +40,6 @@
   <!-- Selectively include build output for package -->
   <ItemGroup>
     <None Include="$(OutputPath)/**/$(AssemblyName).dll" Pack="true" PackagePath="lib" Visible="false" />
-    <None Include="$(OutputPath)/**/$(AssemblyName).pdb" Pack="true" PackagePath="lib" Visible="false" />
```
blairconrad: I am surprised there's no equivalent to this in FakeItEasy.csproj. Maybe the AllowedOutputExtensionsInPackageBuildOutputFolder takes its place?

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -5,7 +5,6 @@
     <AssemblyName>FakeItEasy</AssemblyName>
     <FileAlignment>512</FileAlignment>
     <SignAssembly>true</SignAssembly>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
```
blairconrad: I am surprised there's no equivalent to this in FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj. Maybe the explicit mention of the .pdb takes its place?

FakeItEasy.sln
```diff
@@ -69,6 +69,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{072853A5-3
 		docs\assigning-out-and-ref-parameters.md = docs\assigning-out-and-ref-parameters.md
 		docs\bootstrapper.md = docs\bootstrapper.md
 		docs\calling-base-methods.md = docs\calling-base-methods.md
+		docs\calling-wrapped-methods.md = docs\calling-wrapped-methods.md
```
blairconrad: Ha. Oops. Nice catch.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
thomaslevesque: > Maybe the AllowedOutputExtensionsInPackageBuildOutputFolder takes its place?

Yes. We have to manually include the output in Extensions.ValueTask, because we don't want the netstandard2.1 output.

src/FakeItEasy/FakeItEasy.csproj
thomaslevesque: Yes

FakeItEasy.sln
thomaslevesque: I couldn't find source-stepping.md in the solution, so I had to add it, and looked for other missing ones while I was at it.

src/FakeItEasy/FakeItEasy.csproj
thomaslevesque: yes

thomaslevesque:
@blairconrad do you think the documentation should include instructions for VS Code / Rider?

thomaslevesque:
> do you think the documentation should include instructions for VS Code / Rider?

For future reference:

VS Code:
- Just set `justMyCode` to false in the launch configuration (launch.json)

Rider:
- In Settings > Build, Execution, Deployment > Debugger, check "Enable external source debug"

blairconrad:
Oh, since you have the instructions, I think it would be a nice addition. But not required.

If you've unsure about the longevity of the instructions, they could be hedged with a "maintainers do not use but…" or similar note.

thomaslevesque:
> Oh, since you have the instructions, I think it would be a nice addition. But not required.

Done

> If you've unsure about the longevity of the instructions, they could be hedged with a "maintainers do not use but…" or similar note.

Good idea, I did that

docs/source-stepping.md
```diff
@@ -5,11 +5,32 @@ but occasionally, it may help to see what FakeItEasy is doing under the hood.
 FakeItEasy uses [SourceLink](https://github.com/dotnet/sourcelink) to make it
 a breeze to step into the FakeItEasy source code.
 
-To enable SourceLink to work in Visual Studio:
```
blairconrad: "To enable SourceLink to work," sounds funny to me. Actually on reread, so does "To enable SourceLink to work in Visual Studio:", but slightly less so. I think you could drop "to work".

Whatever change you decided to make (if any!) feel free to amend.

thomaslevesque: Thanks. Amended!

PR #1754, opened 2020-01-25
Remove docs about SideWaffle snippets

### LINKED ISSUE #1753: Remove the documentation page about SideWaffle, or update SideWaffle snippets
The FakeItEasy snippets provided by SideWaffle, which are [mentioned in the documentation](https://fakeiteasy.readthedocs.io/en/stable/snippets-for-fakeiteasy-in-sidewaffle/), are outdated (last update was in 2016). In particular, call assertions that specify the number of calls use the retired `Repeated` API (so our doc is lying, since it shows the new API).

We should either:
- Update the snippets in SideWaffle. Easily done, but I'm not sure it's worth the trouble, because
  - there has been no new releases for almost 4 years, so I don't think it's still maintained
  - it doesn't support VS2017 and 2019
- Update the docs to mention that some snippets are outdated
- Remove the page entirely

thomaslevesque:
Fixes #1753


mkdocs.yml
```diff
@@ -38,7 +38,6 @@ nav:
 - Bootstrapper: bootstrapper.md
 - Advanced usage: advanced-usage.md
 - Analyzer: /projects/analyzers
-- Snippets for FakeItEasy in SideWaffle: snippets-for-fakeiteasy-in-sidewaffle.md
```
blairconrad: And remove it from the sln?

thomaslevesque: I *knew* I had to be forgetting something...

thomaslevesque: Amended

blairconrad:
Thanks, @thomaslevesque!

PR #1760, opened 2020-04-08
Improve error message for explicit implementation of interface method

### LINKED ISSUE #1759: Improve error message when configuring explicitly implemented interface method
Given this interface and class:
```csharp
public interface IFoo
{
    int Bar();
}

public class Foo : IFoo
{
    // Explicitly implemented method
    int IFoo.Bar() => 123;
}
```

When trying to configure the `IFoo.Bar()` method on a fake `Foo`:
```csharp
var foo = A.Fake<Foo>();
A.CallTo(() => ((IFoo)foo).Bar()).Returns(42);
```
The following exception is thrown:

> FakeConfigurationException: 
> 
>   The current proxy generator can not intercept the method UserQuery+IFoo.Bar() for the following reason:
>     - Can not create proxy for method Int32 UserQuery.IFoo.Bar() because it or its declaring type is not accessible. Make it public, or internal and mark your assembly with [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")] attribute, because assembly LINQPadQuery is not strong-named.

---

The problem is easy enough to fix: just specify `options.Implements<IFoo>()` in the fake creation options.

However, the error message is misleading and confusing. It seems to imply that the interface or method is not public, which isn't true, and doesn't point to the actual cause of the problem: the fact that `IFoo.Bar()` is implemented explicitly and can't be overriden unless `Foo` explicitly implements `IFoo` as well.

The message in this scenario should to be improved to something along these lines:

> Foo implements IFoo.Bar explicitly; in order to fake IFoo.Bar, the faked Foo must implement IFoo explicitly using options.Implements\<IFoo\>() in the fake creation options.

thomaslevesque:
Fixes #1759

### DIFF
--- a/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
+++ b/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
@@ -1,4 +1,4 @@
-﻿namespace FakeItEasy.Creation.CastleDynamicProxy
+namespace FakeItEasy.Creation.CastleDynamicProxy
 {
     using System;
     using System.Diagnostics.CodeAnalysis;
@@ -37,7 +37,11 @@
 #else
                 var explicitImplementation = method.Name.Contains('.');
 #endif
-                if (!explicitImplementation)
+                if (explicitImplementation)
+                {
+                    return "The base type implements this interface method explicitly. In order to be able to intercept this method, the fake must specify that it implements this interface in the fake creation options.";
+                }
+                else
                 {
                     return "Non-virtual members can not be intercepted. Only interface members and virtual, overriding, and abstract members can be intercepted.";
                 }
--- a/src/FakeItEasy/Configuration/OutAndRefParameters.StronglyTyped.cs
+++ b/src/FakeItEasy/Configuration/OutAndRefParameters.StronglyTyped.cs
@@ -183,7 +183,7 @@ namespace FakeItEasy.Configuration
                 ValueProducerSignatureHelper.AssertThatValueProducerSignatureSatisfiesCallSignature(
                     call.Method, valueProducer.GetMethodInfo(), NameOfOutRefLazilyFeature);
 
-                return valueProducer(call.GetArgument<T1>(0));
+                return valueProducer(call.GetArgument<T1>(0)!);
             });
         }
 
@@ -209,7 +209,7 @@ namespace FakeItEasy.Configuration
                 ValueProducerSignatureHelper.AssertThatValueProducerSignatureSatisfiesCallSignature(
                     call.Method, valueProducer.GetMethodInfo(), NameOfOutRefLazilyFeature);
 
-                return valueProducer(call.GetArgument<T1>(0), call.GetArgument<T2>(1));
+                return valueProducer(call.GetArgument<T1>(0)!, call.GetArgument<T2>(1)!);
             });
         }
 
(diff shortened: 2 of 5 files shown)


src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
```diff
@@ -37,7 +37,11 @@ public bool MethodCanBeInterceptedOnInstance(MethodInfo method, object? callTarg
 #else
                 var explicitImplementation = method.Name.Contains('.');
 #endif
-                if (!explicitImplementation)
+                if (explicitImplementation)
+                {
+                    return "The base type implements this interface method explicitly. In order to be able to intercept this method, the fake must specify that it implements this interface in the fake creation options.";
```
thomaslevesque, line 42: The message doesn't mention the the actual types and methods involved, for two reasons:
- It would be difficult to get all the required pieces of information here (we already get them at some other point, but passing them to this point would require many changes, and finding them again would be expensive)
- The other messages returned by this method are all generic, so it would be inconsistent to make this one specific. The full error message contains the necessary information anyway.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
```diff
@@ -37,7 +37,11 @@ public bool MethodCanBeInterceptedOnInstance(MethodInfo method, object? callTarg
 #else
                 var explicitImplementation = method.Name.Contains('.');
 #endif
-                if (!explicitImplementation)
```
thomaslevesque, line 40: I'm not sure why we allowed explicit interface implementations here before. As far as I can tell, there's no way to fake the method if we reach that point: the method is sealed, it doesn't matter if it's an explicit implementation or not (except for returning a different message).

If the fake implements the interface, we enter the first `if` and return null on line 30, so we don't reach this point.

thomaslevesque:
I have hundreds of warnings in the T4-generated files... They were gone in previous builds, and now they're back. I also have them locally.
The only thing that changed, as far as I can tell, is the .NET Core SDK.

thomaslevesque:
Confirmed: if I force the use of the 3.0 SDK instead of 3.1, I don't have any warnings.
You had [reported](https://developercommunity.visualstudio.com/content/problem/810959/in-t4-generated-code-with-nullability-on-many-the.html) the problem in VS Dev Community, but it looks like they won't do anything about it... Did you also report it on Github? I can't remember

thomaslevesque:
Oh, actually it's not the same warning!
Before, we had this:
> warning CS8632: The annotation for nullable reference types should only be used in code within a ‘#nullable’ annotations context
Now we have this:
> warning CS8604: Possible null reference argument for parameter 'arg' in 'object?[] Func<T1, object?[]>.Invoke(T1 arg)'.

(if I remove the `#nullable enable`, I get the old warning again)

thomaslevesque:
The new warning seems legit. We have this:

```csharp
return valueProducer(call.GetArgument<T1>(0)!);
```

`valueProducer` is a `Func<T1, object?[]>`, and `GetArgument` has a `[return: MaybeNull]` attribute. So we could be passing a null T1 argument for a non-null T1 parameter. But I don't know how to say "`T1` in `Func<T1, object?[]>` can be null". I can't use `T1?`, because there's no constraint on `T1`.
I can just add `!` after the call to `GetArgument`, but it's frustrating...

thomaslevesque:
OK, so according to https://github.com/dotnet/roslyn/issues/37829, it's by design that `<Nullable>enabled</Nullable>` is ignored in generated code.
What has changed is that the compiler now reports more nullability warnings.

blairconrad:
Perfect, @thomaslevesque. Thanks!

PR #1761, opened 2020-04-09
Document configuration of explicit interface implementation

thomaslevesque:
Relates to #1759

I'm not completely happy with the wording, feel free to suggest improvements!

### DIFF
--- a/docs/specifying-a-call-to-configure.md
+++ b/docs/specifying-a-call-to-configure.md
@@ -51,6 +51,17 @@ A.CallTo(() => deepThought.Invoke("What is the answer to life, the universe, and
 A.CallTo(() => deepThought("What is the answer to life, the universe, and everything?")).Returns(42);
 ```
 
+## Specifying a call to an explicitly implemented interface member
+
+An explicitly implemented member is not directly visible on the concrete class. Instead, it has to be called
+(or overridden) via the interface. In addition, since fakes don't automatically intercept explicitly implemented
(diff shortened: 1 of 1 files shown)


docs/specifying-a-call-to-configure.md
```diff
@@ -51,6 +51,17 @@ A.CallTo(() => deepThought.Invoke("What is the answer to life, the universe, and
 A.CallTo(() => deepThought("What is the answer to life, the universe, and everything?")).Returns(42);
 ```
 
+## Specifying a call to an explicitly implemented interface member
+
+An explicitly implemented member is not directly visible on the concrete, it has to be called via the interface.
+In addition, since it's not possible to just override an explicitly implemented member in a derived class, you
+need to explicitly specify that the fake implements the interface:
```
blairconrad, line 58:
```suggestion
An explicitly implemented member is not directly visible on the concrete class. Instead, it has to be called
(or overridden) via the interface. In addition, since fakes don't automatically intercept explicitly implemented
interfaces, you need to explicitly specify that the fake implements the interface:
```

thomaslevesque: Perfect, thanks for the suggestion!

thomaslevesque: I applied your suggestion but now I need to squash...

blairconrad:
Thanks, @thomaslevesque!

PR #1763, opened 2020-04-21
Document that Dummies are made from public constructors

blairconrad:


blairconrad:
Thanks for the merge. I was linking to the docs from https://stackoverflow.com/a/61331733/1199 and noticed that we didn't have the restriction!

PR #1769, opened 2020-05-07
Strict fake options

### LINKED ISSUE #1765: Add creation option that would allow calling object class methods on a strict Fake
A strict Fake correctly prevents unconfigured calls, however this applies not only to methods defined in the faked class or interface, but also to methods inherited from the object class - Equals, GetHashCode and ToString. While this behaviour can be justified and might be desirable in certain scenarios, it is quite counterintuitive and impractical in the general case, because several actions that don't seem to violate strictness are not possible. This includes:
- checking whether a Fake is a member of a collection (requires calling Equals)
- adding a Fake to a hashed collection e.g. HashSet (requires calling GetHashCode)
- displaying a Fake name e.g. in error messages, even for explicitly named Fakes (requires calling ToString).

It is possible to work around this problem by configuring the Fake to call appropriate base methods or even wrap such configuration in a relatively convenient extension, but it is likely to cause issues e.g. if the configuration of a Fake is cleared by ClearConfiguration.

I would like to know whether Fakes are actually intended to be this strict. If so, would it be possible to provide a different option that would allow using base object methods without additional configuration?

thomaslevesque:
Fixes #1765

### DIFF
--- a/docs/strict-fakes.md
+++ b/docs/strict-fakes.md
@@ -23,3 +23,25 @@ A.CallTo(() => foo.Bar()).Returns("bar");
 
 Strict fakes are useful when it is important to ensure that no calls
 are made to your fake other than the ones you are expecting.
+
+## Object members
+
+It can sometimes be inconvenient that *all* methods throw an exception
+if not configured. You might want to allow calls to methods inherited
+from `System.Object` (`Equals`, `GetHashCode` and `ToString`), because
+they're used all the time, often implicitly, and in most cases there's
+no real value in configuring them manually.
+
+To achieve this, pass a `StrictFakeOptions` value to the `Strict`
+method when you create the fake:
+
+```csharp
+// Allow calls to all object methods
+var foo = A.Fake<IFoo>(x => x.Strict(StrictFakeOptions.AllowObjectMethods));
+
+// Allow calls to ToString
+var foo = A.Fake<IFoo>(x => x.Strict(StrictFakeOptions.AllowToString));
+
+// Allow calls to Equals and GetHashCode
+var foo = A.Fake<IFoo>(x => x.Strict(StrictFakeOptions.AllowEquals | StrictFakeOptions.AllowGetHashCode));
+```
--- a/src/FakeItEasy/Core/StrictFakeRule.cs
+++ b/src/FakeItEasy/Core/StrictFakeRule.cs
@@ -1,14 +1,43 @@
 namespace FakeItEasy.Core
 {
+    using static ObjectMembers;
+
     internal class StrictFakeRule : IFakeObjectCallRule
     {
+        private readonly StrictFakeOptions options;
+
+        public StrictFakeRule(StrictFakeOptions options)
+        {
+            this.options = options;
+        }
+
         public int? NumberOfTimesToCall => null;
 
-        public bool IsApplicableTo(IFakeObjectCall fakeObjectCall) => true;
+        public bool IsApplicableTo(IFakeObjectCall fakeObjectCall)
+        {
+            if (fakeObjectCall.Method.IsSameMethodAs(EqualsMethod))
+            {
+                return !this.HasOption(StrictFakeOptions.AllowEquals);
+            }
+
+            if (fakeObjectCall.Method.IsSameMethodAs(GetHashCodeMethod))
+            {
+                return !this.HasOption(StrictFakeOptions.AllowGetHashCode);
+            }
+
+            if (fakeObjectCall.Method.IsSameMethodAs(ToStringMethod))
+            {
+                return !this.HasOption(StrictFakeOptions.AllowToString);
+            }
+
+            return true;
+        }
 
         public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
         {
             throw new ExpectationException(ExceptionMessages.CallToUnconfiguredMethodOfStrictFake(fakeObjectCall));
         }
+
+        private bool HasOption(StrictFakeOptions flag) => (flag & this.options) == flag;
     }
 }
--- a/src/FakeItEasy/FakeOptionsExtensions.cs
+++ b/src/FakeItEasy/FakeOptionsExtensions.cs
@@ -17,12 +17,27 @@ namespace FakeItEasy
         /// <returns>An options object.</returns>
         public static IFakeOptions<T> Strict<T>(this IFakeOptions<T> options) where T : class
         {
+            return options.Strict(StrictFakeOptions.None);
+        }
+
+        /// <summary>
+        /// Makes the fake strict. This means that any call to the fake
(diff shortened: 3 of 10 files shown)


src/FakeItEasy/Core/StrictFakeRule.cs
```diff
@@ -1,14 +1,43 @@
...
+
+            return true;
+        }
 
         public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
         {
             throw new ExpectationException(ExceptionMessages.CallToUnconfiguredMethodOfStrictFake(fakeObjectCall));
         }
+
+        private bool HasOption(StrictFakeOptions flag) => (flag & this.options) == flag;
```
thomaslevesque, line 41: Because `Enum.HasFlag` is slow (it works on `Enum` rather than a specific enum, which causes the argument value to be boxed)

src/FakeItEasy/FakeOptionsExtensions.cs
```diff
@@ -16,13 +16,27 @@ public static class FakeOptionsExtensions
         /// <param name="options">Options used to create the fake object.</param>
         /// <returns>An options object.</returns>
         public static IFakeOptions<T> Strict<T>(this IFakeOptions<T> options) where T : class
+        {
+            return options.Strict(StrictFakeOptions.None);
```
thomaslevesque, line 20: Not checking for null here because we do it in the other overload

src/FakeItEasy/StrictFakeOptions.cs
```diff
@@ -0,0 +1,37 @@
+namespace FakeItEasy
+{
+    using System;
+
+    /// <summary>
+    /// Options for strict fakes.
+    /// </summary>
+    [Flags]
+    public enum StrictFakeOptions
```
thomaslevesque, line 9: Maybe we could find a better name for this. It's unlikely that we'll ever add new options for strict fakes, and if we do, an enum might not be enough. We *could* rename it to `AllowedMembers` or `StrictFakeAllowedMembers`...
I initially named it `StrictFakeBehavior`, but FxCop reminded me that a `[Flags]` enum should have a plural name, and `StrictFakeBehaviors` didn't sound right.

src/FakeItEasy/Core/StrictFakeRule.cs
blairconrad: I did not know this! Thanks.

src/FakeItEasy/Core/StrictFakeRule.cs
```diff
@@ -1,14 +1,43 @@
...
+                return !this.HasOption(StrictFakeOptions.AllowGetHashCode);
+            }
+
+            if (fakeObjectCall.Method.IsSameMethodAs(ToStringMethod))
+            {
+                return !this.HasOption(StrictFakeOptions.AllowToString);
+            }
+
+            return true;
+        }
```
blairconrad, line 34: I'm surprised this wasn't a super fancy pattern-matching switch!

src/FakeItEasy/FakeOptionsExtensions.cs
```diff
@@ -16,13 +16,27 @@ public static class FakeOptionsExtensions
...
         /// <returns>An options object.</returns>
         public static IFakeOptions<T> Strict<T>(this IFakeOptions<T> options) where T : class
+        {
+            return options.Strict(StrictFakeOptions.None);
+        }
+
+        /// <summary>
+        /// Makes the fake strict. This means that any call to the fake
+        /// that has not been explicitly configured will throw an exception,
+        /// except calls to the System.Object methods specified in <c>options</c>.
```
blairconrad, line 26:
```suggestion
        /// except calls to the System.Object methods specified in <c>strictOptions</c>.
```

Alternatively, a `paramref``? I'm not fussy.

src/FakeItEasy/StrictFakeOptions.cs
blairconrad: …Behaviors certainly isn't common usage, but I don't think it's the worst thing in the world.
Truth be told, I don't really understand your concern here. You seem to be worried both about the name and the inflexibility of the enum. For the latter, we could always implement an options-building scheme like the "outer" options, but it seems heavy.

I'm not particularly worried about what you have done, but am prepared to be convinced to worry.

src/FakeItEasy/StrictFakeOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <summary>
+        /// Default behavior. No unconfigured method can be called, including those
+        /// inherited from <see cref="Object" />.
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Calls to <see cref="Object.Equals(object)"/> are allowed.
+        /// </summary>
```
blairconrad, line 19: Should we mention that they'd behave just as if the fake weren't strict?

docs/strict-fakes.md
```diff
@@ -23,3 +23,25 @@ A.CallTo(() => foo.Bar()).Returns("bar");
...
+if not configured. You might want to allow calls to methods inherited
+from `System.Object` (`Equals`, `GetHashCode` and `ToString`), because
+they're used all the time, often implicitly, and in most cases there's
+no real value in configuring them manually.
+
+To achieve this, pass a `StrictFakeOptions` value to the `Strict`
+method when you create the fake:
+
+```csharp
+// Allow calls all object methods
```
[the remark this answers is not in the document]
thomaslevesque: Doh!

src/FakeItEasy/Core/StrictFakeRule.cs
thomaslevesque: Yeah, I know... It didn't work out in this case

src/FakeItEasy/FakeOptionsExtensions.cs
thomaslevesque: Oh! I didn't know about paramref. I'll look into it

src/FakeItEasy/StrictFakeOptions.cs
thomaslevesque: I'm not really worried about the inflexibility of the enum, it's just that if we ever need to add more options for which an enum isn't a good fit, the name will already be taken.

> For the latter, we could always implement an options-building scheme like the "outer" options, but it seems heavy.

Yeah... probably overkill, especially for something we don't expect to ever need

If you're not worried, I'm not worried!

src/FakeItEasy/StrictFakeOptions.cs
thomaslevesque: Sure!

thomaslevesque:
Thanks for the merge!

PR #1770, opened 2020-05-08
Throw exception when bootstrapper returns null

### LINKED ISSUE #1689: Throw informative exception when IBootstrapper.GetAssemblyFileNamesToScanForExtensions returns null
From a conversation in #1683. At that point I suggested a `UserCallbackException`, but the documentation for that class says it's "An exception thrown when a user-provided callback throws an exception", so we should either consider a different exception or update the docs for that class.

thomaslevesque:
Fixes #1689


thomaslevesque:
Thanks for the merge

PR #1772, opened 2020-05-30
Support StringComparison in EndsWith and StartsWith constraints

### LINKED ISSUE #1771: Add ArgumentConstraintManagerExtensions.StartsWith and EndsWith overloads that take a StringComparison
As proposed by @thomaslevesque  in https://github.com/FakeItEasy/FakeItEasy/issues/1681#issuecomment-635425902

blairconrad:
Fixes #1771.

### DIFF
--- a/tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringEndsWithTests.cs
+++ b/tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringEndsWithTests.cs
@@ -45,7 +45,9 @@ namespace FakeItEasy.Tests.ArgumentConstraintManagerExtensions
 
         protected override void CreateConstraint(INegatableArgumentConstraintManager<string> scope)
         {
+#pragma warning disable CA1307 // Specify StringComparison
             scope.EndsWith("table");
+#pragma warning restore CA1307 // Specify StringComparison
         }
     }
 }
--- a/tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringEndsWithTests.cs
+++ b/tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringEndsWithTests.cs
@@ -17,7 +17,9 @@ namespace FakeItEasy.Tests.ArgumentConstraintManagerExtensions
                 "bear",
                 "chicken",
                 "lorem ipsum",
-                (object?)null);
+                (object?)null,
+                "tabletop",
+                "stabled");
         }
 
         public static IEnumerable<object?[]> ValidValues()
--- a/tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringStartsWithWithComparisonTypeTests.cs
+++ b/tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringStartsWithWithComparisonTypeTests.cs
@@ -1,13 +1,14 @@
 namespace FakeItEasy.Tests.ArgumentConstraintManagerExtensions
 {
+    using System;
     using System.Collections.Generic;
     using FakeItEasy.Tests.TestHelpers;
     using Xunit;
 
-    public class StringStartsWithTests
+    public class StringStartsWithWithComparisonTypeTests
         : ArgumentConstraintTestBase<string>
     {
-        protected override string ExpectedDescription => "string that starts with \"abc\"";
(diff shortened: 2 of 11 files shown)


tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringEndsWithTests.cs
```diff
@@ -45,7 +45,9 @@ public override void IsValid_should_return_true_for_valid_values(object validVal
 
         protected override void CreateConstraint(INegatableArgumentConstraintManager<string> scope)
         {
+#pragma warning disable CA1307 // Specify StringComparison
             scope.EndsWith("table");
+#pragma warning restore CA1307 // Specify StringComparison
```
thomaslevesque, line 50: I didn't realize this warning also applied to types other than `String`. Doesn't that mean that existing user code calling this method will start showing this warning too?

tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringStartsWithWithComparisonTypeTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public static IEnumerable<object?[]> InvalidValues()
+        {
+            return TestCases.FromObject(
+                "foo",
+                "bar",
+                "biz",
+                "baz",
+                "lorem ipsum",
+                (object?)null);
```
thomaslevesque, line 21: Nitpick: could include values that contain "abc" but do not start with it. Feel free to apply or ignore the suggestion!

tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringEndsWithTests.cs
blairconrad: Oh, that's a very good point. I didn't think about the warning flowing through. I'll look into it.

tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringStartsWithWithComparisonTypeTests.cs
blairconrad: It's a fair point.

tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/StringEndsWithTests.cs
blairconrad: It seems the analyzer is looking at the shape of the methods and what sibling methods exist with or without the comparison type. 
I don't know how we could avoid triggering the warning for clients, short of either changing the signature (reorder the parameters?), or replacing the original constraint with a differently-named one ("EndsWithExactly" or similar). I don't love either option, and I don't really want to subject a bunch of presumably happy users to a new warning, especially when I think the method that causes the warning is the better choice in general.
I'm at a loss.

blairconrad: >  or replacing the original constraint with a differently-named one ("EndsWithExactly" or similar)

Of renaming the new ones, I suppose.  "EndsWithUsingComparison" or similar. Unsurprisingly, I don't love this either.

Given that this was an owner-requested feature, I'm losing enthusiasm! But will still think on it.

thomaslevesque: I don't think it's a very major issue. Only users who use FxCop will see the warning, and those users are typically used to this kind of warning and know how to fix them. It's annoying, but easy to fix.
Alternatively, we could consider this a breaking change (since it can produce new warnings in code that didn't have them), and release this change in a new major version. But it's probably overkill... I'm more inclined to just ignore the problem.

blairconrad:
Okay, @thomaslevesque . I'm convinced. Let's merge it. It's not like we don't have the problem with `.Contains(string)` anyhow. Thanks!

thomaslevesque:
You need to squash, @blairconrad :wink:

blairconrad:
So I do! :blush:

Thanks for the review and merge.

thomaslevesque:
Thanks @blairconrad!

Issue #1762, opened 2020-04-10
Release 6.1.0


thomaslevesque:
- [x] Address all coding and style violations detected during a *Release* build (send a regular pull request which must be merged before continuing)
- [x] ensure completeness of draft [release notes](https://github.com/FakeItEasy/FakeItEasy/releases), including non-owner contributors, if any (move release notes forward from any pre-releases to the current release)
- [x] run `tools\prepare_release.cmd next x.y.z vNext` to:
    - change "vNext" names to appropriate x.y.z version for the next GitHub Release, milestone, and release checklist (this issue)
    - create a new draft GitHub Release, milestone, and release checklist
- [x] tag master with the new version number and push to upstream. This will cause the build server to
    - push the nupkgs to [nuget.org](https://www.nuget.org/profiles/FakeItEasy),
    - attach all nupkgs from the tagged build to the GitHub Release,
    - add a comment to each issue and pull request in the release, saying it was published in this release
- [x] wait for NuGet package validation
- [x] de-list pre-release NuGet packages if present
- [x] publish the GitHub release
- [x] tweet, mentioning contributors, and post link as comment here for easy retweeting ;-)
- [x] post tweet in [Gitter](https://gitter.im/FakeItEasy/FakeItEasy)
- [x] add a note in each issue that has a significant contribution from a community member, thanking the contributor
- [ ] close this issue
- [ ] if this is **not** a pre-release, close the [milestone](https://github.com/FakeItEasy/FakeItEasy/milestones?state=open) for this release

blairconrad:
@thomaslevesque, I started the release process and noticed that we have [warnings](https://ci.appveyor.com/project/blairconrad/fakeiteasy/build/job/o07uxh4cwfuuf3pt#L48) in the latest build. I haven't the cycles to address now, but could work on over the weekend.

blairconrad:
```
Get FakeItEasy 6.1.0!

- new creation option allows calling object class methods on a strict fake
- better exception when GetAssemblyFileNamesToScanForExtensions returns null
- new StartsWith and EndsWith string argument constraints take StringComparison

https://github.com/FakeItEasy/FakeItEasy/releases/6.1.0
```

PR #1776, opened 2020-06-05
Fix warnings

thomaslevesque:
Relates to #1762

### DIFF
--- a/src/FakeItEasy/Creation/FakeObjectCreator.cs
+++ b/src/FakeItEasy/Creation/FakeObjectCreator.cs
@@ -66,10 +66,15 @@ namespace FakeItEasy.Creation
         }
 
         [SuppressMessage("Microsoft.Design", "CA1021:AvoidOutParameters", MessageId = "2#", Justification = "Seems appropriate here.")]
-        public bool MethodCanBeInterceptedOnInstance(MethodInfo method, object? callTarget, [NotNullWhen(false)]out string? failReason) =>
-            callTarget is object && DelegateCreationStrategy.IsResponsibleForCreating(callTarget.GetType())
-                ? this.delegateCreationStrategy.MethodCanBeInterceptedOnInstance(method, callTarget, out failReason)
-                : this.defaultCreationStrategy.MethodCanBeInterceptedOnInstance(method, callTarget, out failReason);
+        public bool MethodCanBeInterceptedOnInstance(MethodInfo method, object? callTarget, [NotNullWhen(false)] out string? failReason)
+        {
+            if (callTarget is object && DelegateCreationStrategy.IsResponsibleForCreating(callTarget.GetType()))
+            {
+                return this.delegateCreationStrategy.MethodCanBeInterceptedOnInstance(method, callTarget, out failReason);
+            }
+
+            return this.defaultCreationStrategy.MethodCanBeInterceptedOnInstance(method, callTarget, out failReason);
+        }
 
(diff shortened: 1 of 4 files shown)


src/FakeItEasy/Creation/FakeObjectCreator.cs
```diff
@@ -66,10 +66,23 @@ internal class FakeObjectCreator : IFakeObjectCreator, IMethodInterceptionValida
...
+                }
+                else
+                {
+                    return this.defaultCreationStrategy.MethodCanBeInterceptedOnInstance(method, callTarget, out failReason);
+                }
+            }
+
+            failReason = "Call target is null";
+            return false;
+        }
```
thomaslevesque, line 85: There was a CS8762 warning, because `failReason` ended up being null when `callTarget` was null, violating the `[NotNullWhen(false)]` contract.

The conditional expression *should* have worked, but produced the same warning. I opened https://github.com/dotnet/roslyn/issues/44886

src/FakeItEasy/RootModule.cs
```diff
@@ -92,7 +92,7 @@ public static void RegisterDependencies(IServiceRegistrar registrar)
             registrar.Register<IFakeObjectCallFormatter>(fakeObjectCallFormatter);
 
             StringBuilderOutputWriter StringBuilderOutputWriterFactory() =>
-                new StringBuilderOutputWriter(argumentValueFormatter);
+                new StringBuilderOutputWriter(argumentValueFormatter!);
```
thomaslevesque, line 95: Not sure why that was necessary. There's no way `argumentValueFormatter` is null at this point.

tests/FakeItEasy.Tests/Core/MethodInfoManagerTests.cs
```diff
@@ -211,7 +211,7 @@ public void DoSomething(string text)
             }
 
             [SuppressMessage("Microsoft.Usage", "CA1801:ReviewUnusedParameters", MessageId = "other", Justification = "Required for testing.")]
-            public bool Equals(Base other) => true;
+            public bool Equals([AllowNull] Base other) => true;
```
thomaslevesque, line 214: The interface method is declared with this attribute, not having it on the implementation caused a warning.

blairconrad:
Thanks, @thomaslevesque. I have some concerns about the failing tests! :wink:

thomaslevesque:
It's fixed now!

src/FakeItEasy/Creation/FakeObjectCreator.cs
blairconrad: > There was a CS8762 warning, because `failReason` ended up being `null` when `callTarget` was null, violating the `[NotNullWhen(false)]` contract.

I think this comment was found to be false, right? Not complaining, just making sure I understand.

src/FakeItEasy/RootModule.cs
blairconrad: Yeah. Ah well, it's a minor annoyance.

thomaslevesque:
Thanks for the merge!

src/FakeItEasy/Creation/FakeObjectCreator.cs
thomaslevesque: > I think this comment was found to be false, right?

Yes, `callTarget` being null had nothing to do with it. It's just the compiler bug, after all.

PR #1778, opened 2020-06-05
Apply Object member rules to overridden members

### LINKED ISSUE #1777: Apply Object member rules even when the methods are overridden
Prompted by #1774, after internal discussion.

Currently, special fake rules are applied for members of `System.Object`:
- `Equals(object)`: returns true if the argument is the same instance as the fake
- `GetHashCode()`: returns the hash code of the fake manager
- `ToString()`: returns a default string representation of the form `Faked <faked object type>`

However, when the faked type overrides those methods, these rules are not applied, so a default value is returned instead (unless the methods are explicitly configured). This leads to surprising results:
- `Equals` always returns false, so the fake isn't even considered equal to itself
- `GetHashCode` always returns 0
- `ToString` always returns an empty string

Although this behavior makes sense from a technical point of view, there doesn't seem to be any scenario where it's desirable.

The special rules should still be applied even if the fake type overrides `System.Object` members.

thomaslevesque:
Fixes #1777

To be merged after 6.1.0 release

The last commit is optional, but I thought it made sense. I needed a `MethodInfo` to get access to `GetBaseDefinition`, and I realized that the methods in `MethodBaseExtensions` were *always* called on `MethodInfo`, so I changed their signature and renamed the class. I'm happy to rollback if you have an objection.

### DIFF
--- a/docs/default-fake-behavior.md
+++ b/docs/default-fake-behavior.md
@@ -49,6 +49,18 @@ This behavior can be used to
 * supply values for the system under test to use (via the getter) or to
 * verify that the system under test performed the `set` action on the Fake
 
+### Object members
+
+Virtual methods inherited from `System.Object` are faked with a special default behavior:
+
+* `Equals` uses reference equality: it returns `true` if the argument is the fake
+  itself, `false` for any other argument.
+* `GetHashCode` returns a hashcode consistent with the behavior of `Equals`.
+* `ToString` returns a string of the form "Faked &lt;type of fake&gt;"
+
+This behavior applies to all fakes, including fakes of types that override these
+methods. Like any other method, the behavior can be explictly configured.
+
 ### Cancellation tokens
 
 When a faked method that accepts a `CancellationToken` receives a canceled token
--- a/src/FakeItEasy/Core/MethodInfoManager.cs
+++ b/src/FakeItEasy/Core/MethodInfoManager.cs
@@ -41,29 +41,11 @@ namespace FakeItEasy.Core
             return MethodCache.GetOrAdd(key, k => FindMethodOnTypeThatWillBeInvokedByMethodInfo(k.Type, new FakeItEasy.Compatibility.MethodInfoWrapper(k.MethodInfo, k.Type)));
         }
 
-        private static bool HasSameBaseMethod(MethodInfo first, MethodInfo second)
-        {
-            var baseOfFirst = GetBaseDefinition(first);
-            var baseOfSecond = GetBaseDefinition(second);
-
-            return baseOfFirst.IsSameMethodAs(baseOfSecond);
-        }
-
-        private static MethodInfo GetBaseDefinition(MethodInfo method)
-        {
-            if (method.IsGenericMethod && !method.IsGenericMethodDefinition)
-            {
-                method = method.GetGenericMethodDefinition();
-            }
-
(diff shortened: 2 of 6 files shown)


src/FakeItEasy/Core/MethodInfoManager.cs
```diff
@@ -41,29 +41,11 @@ public virtual bool WillInvokeSameMethodOnTarget(Type target, MethodInfo first,
             return MethodCache.GetOrAdd(key, k => FindMethodOnTypeThatWillBeInvokedByMethodInfo(k.Type, new FakeItEasy.Compatibility.MethodInfoWrapper(k.MethodInfo, k.Type)));
         }
 
-        private static bool HasSameBaseMethod(MethodInfo first, MethodInfo second)
```
thomaslevesque, line 44: Moved to an extension method in MethodInfoExtensions

tests/FakeItEasy.Specs/ObjectMembersSpecs.cs
```diff
@@ -0,0 +1,146 @@
...
+                .x(() => equals = fake.Equals(otherFake));
+
+            "Then it returns false"
+                .x(() => equals.Should().BeFalse());
+        }
+
+        [Scenario]
+        public static void OverriddenGetHashCode(Foo fake, FakeManager manager, int fakeHashCode)
+        {
+            "Given a fake of a type that overrides Equals"
```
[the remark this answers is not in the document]
thomaslevesque: Oops, thanks!

thomaslevesque:
> 1. I think we should document the [default behaviour](https://fakeiteasy.readthedocs.io/en/stable/default-fake-behavior/) for object methods. Really we should've before, but now seems like a good time to remedy the lack

Good suggestion, as usual!

> 2. What happens with `new` methods? Will they benefit from the default behaviour you provide, or since they are `new`, will they behave like any other faked virtual (assuming virtual) method?

They should behave like any other faked method. I'll add tests to make sure of this.

thomaslevesque:
All done! Thanks for the review

docs/default-fake-behavior.md
```diff
@@ -49,6 +49,18 @@ This behavior can be used to
 * supply values for the system under test to use (via the getter) or to
 * verify that the system under test performed the `set` action on the Fake
 
+### Object members
+
+Virtual methods inherited from `System.Object` are faked with a special default behavior:
+
+* `Equals` uses reference equality: it returns `true` if the argument is the fake
+  itself, `false` for any other argument.
+* `GetHashCode` returns the hashcode of its `FakeManager`.
```
blairconrad, line 58: Technically true, but I wonder if it's more than we need to tell people. What if we want to change the implementation?
Perhaps it's just enough to indicate that it's consistent with our implementation of `Equals`? Who even knows what a `FakeManager is`!

blairconrad: If you want to keep as is, I'm okay with it. If you want to change, feel free to amend rather than fixup.

thomaslevesque: Yeah, I'm not too happy with this either...

> Perhaps it's just enough to indicate that it's consistent with our implementation of `Equals`?

I think I like it! I'll make the change

thomaslevesque: Done

Issue #1781, opened 2020-06-24
DoesNothing(); not available


ejwoerlee:
When I try to use the DoesNothing on a fake, this functionality is not available. 
Using FakeItEasy v6.1.0/Autofac.Extra.FakeitEasy v6.0.0

A.CallTo(() => aFake.SomeMethodThatShouldDoNothing()) **.DoesNothing();**

** DoesNothing() ** doesn't show up in intellisense

A.CallTo(() => aFake.SomeMethodThatShouldDoNothing()) **.Throws<Exception>();** works fine

thomaslevesque:
Hi @ejwoerlee,

What's the return type of `SomeMethodThatShouldDoNothing`?

`DoesNothing` only applies to `void` methods. It doesn't really make sense for non-void methods, because you need to configure a return value. So you can do something like this instead:

```csharp
A.CallTo(() => aFake.SomeMethodThatShouldDoNothing()).Returns(default)
```

ejwoerlee:
This is the function I want to return nothing in a fake:

`public async Task ConnectAsync(string host, int port)
 {
     await TcpClient.ConnectAsync(host, port); 
 }`
and
`public NetworkStream GetStream()
        {
            if (tcpClient == null)
            {
                throw new ArgumentNullException(nameof(tcpClient));
            }

            return TcpClient.GetStream();
        }`

blairconrad:
Thanks, @ejwoerlee. As @thomaslevesque  noted, `DoesNothing` is only available when configuring `void` methods. Neither of these methods are `void`. In this case, you need to supply a return value via `.Returns` (or an action like throws).

Just to make sure we're not getting ahead of ourselves though, why do you feel a need to specify `DoesNothing`? Have you previously configured these methods to do something and you don't want them to anymore? Or is the Fake strict?

If neither applies, the default behaviour for a faked method will be to return a default value and not perform any other side effect.

**However**, I am compelled to note that in your definitions above, these methods are not marked either `virtual` or `abstract`, so FakeItEasy [will not be able to intercept them at all](https://fakeiteasy.readthedocs.io/en/stable/what-can-be-faked/#what-members-can-be-overridden). The class's production behaviour will always apply.

thomaslevesque:
I think the API might be missing something. `Task` is the async equivalent of `void`, so it's a bit strange that we have `DoesNothing` for methods that return `void`, but not for methods that return `Task`. Maybe we should add one...

[...]

PR #1782, opened 2020-06-24
Explain that DoesNothing only applies to void members

blairconrad:
Inspired by #1781. I also made some other notes to make the usage a little more clear. (I hope).


blairconrad:
Thank you, @thomaslevesque!

PR #1785, opened 2020-07-16
Add DoesNothing method for task-returning methods

### LINKED ISSUE #1784: Consider adding DoesNothing for methods that return Task
Inspired by #1781

Currently the `DoesNothing` method is only available on void methods. But for async methods, `Task` is the conceptual equivalent of `void`, so it would make sense to expose a `DoesNothing` or `DoesNothingAsync` method to configure methods that return a `Task`.

Current workaround to configure a `Task`-returning method to do nothing:

```csharp
A.CallTo(() => fake.FooAsync()).Returns(Task.CompletedTask);
```

thomaslevesque:
Fixes #1784

### DIFF
--- a/docs/doing-nothing.md
+++ b/docs/doing-nothing.md
@@ -16,6 +16,6 @@ If the [Fake is strict](strict-fakes.md), an unconfigured call will throw an exc
 Or, `DoesNothing` can be used to change the behavior that an already-configured call is supposed to have. For example, if a call is [set to throw an exception](throwing-exceptions.md), that can be overridden. For more on this kind of thing, see how to [override the behavior for a call](changing-behavior-between-calls.md#overriding-the-behavior-for-a-call).
 
 **Note that `DoesNothing` is only applicable when configuring members that have a
-`void` return.** To override behaviour on already-configured (or strict) members
-that return values, you must instead
-[configure a preferred return value](specifying-return-values.md).
+`void` return (or `Task`, which is the async equivalent of `void`).** To override
+behavior on already-configured (or strict) members that return values, you must
+instead [configure a preferred return value](specifying-return-values.md).
--- a/src/FakeItEasy/ReturnValueConfigurationExtensions.cs
+++ b/src/FakeItEasy/ReturnValueConfigurationExtensions.cs
@@ -117,6 +117,18 @@ namespace FakeItEasy
             configuration.ReturnsNextFromQueue(new ConcurrentQueue<Task<T>>(values.Select(TaskHelper.FromResult)));
         }
 
+        /// <summary>
+        /// Configures the specified call to do nothing when called.
+        /// </summary>
+        /// <param name="configuration">The call configuration to extend.</param>
+        /// <returns>A configuration object.</returns>
+        public static IAfterCallConfiguredConfiguration<IReturnValueConfiguration<Task>> DoesNothing(this IReturnValueConfiguration<Task> configuration)
+        {
+            Guard.AgainstNull(configuration, nameof(configuration));
+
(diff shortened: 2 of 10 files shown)


blairconrad:
I think everything looks good, @thomaslevesque. Thanks!
I checked the docs to see if we needed to update anything, and I'm not sure we do. [Doing Nothing](https://fakeiteasy.readthedocs.io/en/stable/doing-nothing/) doesn't reference `void` methods like I thought it would (actually, that's what brought attention to this issue, isn't it), so there's not strictly a need to update it. We _could_, I suppose update to indicate that it's only available on `void`- and `Task`-returning methods, but I'll not insist on it. Leaving here, and if you want me to merge in my morning, I will do so.

thomaslevesque:
> [Doing Nothing](https://fakeiteasy.readthedocs.io/en/stable/doing-nothing/) doesn't reference `void` methods like I thought it would

Yes [it does](https://github.com/FakeItEasy/FakeItEasy/pull/1782), but it's in the [latest](https://fakeiteasy.readthedocs.io/en/latest/doing-nothing/) version, not stable.
I'll update the docs

docs/doing-nothing.md
```diff
@@ -16,6 +16,6 @@ If the [Fake is strict](strict-fakes.md), an unconfigured call will throw an exc
 Or, `DoesNothing` can be used to change the behavior that an already-configured call is supposed to have. For example, if a call is [set to throw an exception](throwing-exceptions.md), that can be overridden. For more on this kind of thing, see how to [override the behavior for a call](changing-behavior-between-calls.md#overriding-the-behavior-for-a-call).
 
 **Note that `DoesNothing` is only applicable when configuring members that have a
-`void` return.** To override behaviour on already-configured (or strict) members
-that return values, you must instead
-[configure a preferred return value](specifying-return-values.md).
\ No newline at end of file
+`void` return (or `Task`, which is the async equivalent of `void`).** To override
+behavior on already-configured (or strict) members that return values, you must
+instead [configure a preferred return value](specifying-return-values.md).
```
thomaslevesque, line 21: Note: I changed "behaviour" to "behavior", since we had settled on using the US spelling.

src/FakeItEasy/ReturnValueConfigurationExtensions.cs
```diff
@@ -122,7 +122,7 @@ public static void ReturnsNextFromSequence<T>(this IReturnValueConfiguration<Tas
         /// </summary>
         /// <param name="configuration">The call configuration to extend.</param>
         /// <returns>A configuration object.</returns>
-        public static IAfterCallConfiguredWithOutAndRefParametersConfiguration<IReturnValueConfiguration<Task>> DoesNothing(this IReturnValueConfiguration<Task> configuration)
+        public static IAfterCallConfiguredConfiguration<IReturnValueConfiguration<Task>> DoesNothing(this IReturnValueConfiguration<Task> configuration)
```
thomaslevesque, line 125: The return type was wrong (not the same as in `IDoNothingConfiguration`)
I made a fixup commit so that you can see the difference

blairconrad:
> > [Doing Nothing](https://fakeiteasy.readthedocs.io/en/stable/doing-nothing/) doesn't reference `void` methods like I thought it would
> 
> Yes [it does](https://github.com/FakeItEasy/FakeItEasy/pull/1782), but it's in the [latest](https://fakeiteasy.readthedocs.io/en/latest/doing-nothing/) version, not stable.
> I'll update the docs

Whoops! Glad you're on the ball. Nice catch on "behaviour" as well.

thomaslevesque:
Squashed

blairconrad:
Thanks, Champ!

PR #1797, opened 2020-11-11
Fix wrapping fake events

### LINKED ISSUE #1795: Registering a handler on a fake Wrapping an event does not work
Hello,

I want to be able to tell whether a method has been called on an object, but still use the actual implementation of this object.

To do this, I'm creating a fake for the interface, wrapping an instance of its implementation.

It works for methods, but not for events: trying to add an event handler on the faked event results in the concrete event never receiving the call to register the event handler.

Demo (the test fails):
```csharp
[TestClass]
public class UnitTest1
{
	[TestMethod]
	public void TestMethod1()
	{
		var success = false;
		var instance = new MyClass();

		var fake = A.Fake<IInterface>(x => x.Wrapping(instance));

		fake.Event += (s, e) => success = true;

		fake.RaiseEvent();

		if (!success)
			Assert.Fail();
	}

	public interface IInterface
	{
		void RaiseEvent();
		event EventHandler Event;
	}

	public class MyClass : IInterface
	{
		public event EventHandler Event;
		public void RaiseEvent() => Event?.Invoke(this, EventArgs.Empty);
	}
}
```

I did not find any documentation about this behavior. Is this expected?
Is there any way around that?

thomaslevesque:
Fixes #1795

### DIFF
--- a/docs/raising-events.md
+++ b/docs/raising-events.md
@@ -83,3 +83,30 @@ AddHandler fake.CustomEvent, Raise.FreeForm(Of CustomEventHandler).With(fake, sa
 
 Specifying the type of the event handler gets around the language restrictions in VB.NET.
 This method may also be used from C# if you don't want to rely on late binding.
+
+## Limitations
+
+The approach described above for raising events doesn't work in some situations:
+- Wrapping fakes
+- Fakes configured to call base methods
+
+This is because the calls (including event subscription and unsubscription) are
+forwarded to another implementation (wrapped object or base class) that
+FakeItEasy has no control over, so the fake doesn't know about the handlers and
+cannot call them.
+
+To work around this limitation, you will need to explicitly configure the calls
+to the event accessors to handle event subscription yourself, as in the example
+below:
+
+```csharp
+// Declare a delegate to store the event handlers
+EventHandler handlers = null;
+
+// Configure event subscription on the fake
+A.CallTo(robot).Where(call => call.Method.Name == "add_FellInLove").Invokes((EventHandler h) => handlers += h);
+A.CallTo(robot).Where(call => call.Method.Name == "remove_FellInLove").Invokes((EventHandler h) => handlers -= h);
+
+// Raise the event
+handlers?.Invoke(robot, EventArgs.Empty);
+```
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -23,7 +23,7 @@ namespace FakeItEasy.Core
             new DefaultReturnValueRule(),
         };
 
-        private readonly IEnumerable<IFakeObjectCallRule> postUserRules;
+        private readonly EventRule eventRule;
         private readonly LinkedList<IInterceptionListener> interceptionListeners;
         private readonly WeakReference objectReference;
 
@@ -48,7 +48,7 @@ namespace FakeItEasy.Core
             this.FakeObjectType = fakeObjectType;
             this.FakeObjectName = fakeObjectName;
 
-            this.postUserRules = SharedPostUserRules.Prepend<IFakeObjectCallRule>(new EventRule(this)).ToList();
+            this.eventRule = new EventRule(this);
             this.allUserRules = new LinkedList<CallRuleMetadata>();
 
             this.recordedCalls = new ConcurrentQueue<CompletedFakeObjectCall>();
@@ -255,7 +255,13 @@ namespace FakeItEasy.Core
                 return;
             }
 
-            foreach (var postUserRule in this.postUserRules)
+            if (this.eventRule.IsApplicableTo(fakeObjectCall))
+            {
+                this.eventRule.Apply(fakeObjectCall);
+                return;
+            }
+
+            foreach (var postUserRule in SharedPostUserRules)
             {
                 if (postUserRule.IsApplicableTo(fakeObjectCall))
                 {
--- a/src/FakeItEasy/Core/FakeManager.cs
+++ b/src/FakeItEasy/Core/FakeManager.cs
@@ -14,7 +14,7 @@ namespace FakeItEasy.Core
     /// </summary>
     public partial class FakeManager : IFakeCallProcessor
     {
-        private static readonly SharedFakeObjectCallRule[] PostUserRules =
+        private static readonly SharedFakeObjectCallRule[] SharedPostUserRules =
         {
             new ObjectMemberRule(),
             new AutoFakePropertyRule(),
@@ -23,7 +23,7 @@ namespace FakeItEasy.Core
             new DefaultReturnValueRule(),
         };
 
-        private readonly IFakeObjectCallRule[] preUserRules;
+        private readonly IEnumerable<IFakeObjectCallRule> postUserRules;
         private readonly LinkedList<IInterceptionListener> interceptionListeners;
         private readonly WeakReference objectReference;
 
@@ -48,10 +48,7 @@ namespace FakeItEasy.Core
             this.FakeObjectType = fakeObjectType;
             this.FakeObjectName = fakeObjectName;
 
-            this.preUserRules = new IFakeObjectCallRule[]
-            {
-                new EventRule(this),
-            };
+            this.postUserRules = SharedPostUserRules.Prepend<IFakeObjectCallRule>(new EventRule(this)).ToList();
             this.allUserRules = new LinkedList<CallRuleMetadata>();
 
             this.recordedCalls = new ConcurrentQueue<CompletedFakeObjectCall>();
@@ -238,15 +235,6 @@ namespace FakeItEasy.Core
         // Apply the best rule to the call. There will always be at least one applicable rule.
         private void ApplyBestRule(IInterceptedFakeObjectCall fakeObjectCall)
         {
-            foreach (var preUserRule in this.preUserRules)
-            {
-                if (preUserRule.IsApplicableTo(fakeObjectCall))
-                {
-                    preUserRule.Apply(fakeObjectCall);
-                    return;
-                }
-            }
-
             CallRuleMetadata? bestUserRule = null;
             lock (this.allUserRules)
             {
@@ -267,7 +255,7 @@ namespace FakeItEasy.Core
                 return;
             }
 
-            foreach (var postUserRule in PostUserRules)
+            foreach (var postUserRule in this.postUserRules)
             {
                 if (postUserRule.IsApplicableTo(fakeObjectCall))
                 {
--- a/src/FakeItEasy/EnumerableExtensions.cs
+++ b/src/FakeItEasy/EnumerableExtensions.cs
@@ -61,5 +61,12 @@ namespace FakeItEasy
         {
             return source as IList<object> ?? source.Cast<object>().ToList();
         }
+
+#if LACKS_PREPEND
+        public static IEnumerable<T> Prepend<T>(this IEnumerable<T> source, T item)
+        {
+            return new[] { item }.Concat(source);
+        }
+#endif
     }
 }
--- a/src/FakeItEasy/EnumerableExtensions.cs
+++ b/src/FakeItEasy/EnumerableExtensions.cs
@@ -61,12 +61,5 @@ namespace FakeItEasy
         {
             return source as IList<object> ?? source.Cast<object>().ToList();
         }
-
-#if LACKS_PREPEND
-        public static IEnumerable<T> Prepend<T>(this IEnumerable<T> source, T item)
-        {
-            return new[] { item }.Concat(source);
-        }
-#endif
     }
 }
--- a/src/FakeItEasy/ExceptionMessages.cs
+++ b/src/FakeItEasy/ExceptionMessages.cs
@@ -57,6 +57,12 @@ namespace FakeItEasy
         public static string GetAssemblyFileNamesToScanForExtensionsShouldntReturnNull =>
             "IBootstrapper.GetAssemblyFileNamesToScanForExtensions must not return null";
 
+        public static string WrappingFakeCannotRaiseEvent =>
+            "The fake cannot raise the event because it's a wrapping fake and the event subscription is managed by the wrapped object. The event can only be raised internally by the wrapped object.";
+
+        public static string CannotRaiseEventWhenCallingBaseMethod =>
+            "The fake cannot raise the event because event subscription calls the base implementation.";
+
         public static string WrongConstructorExpressionType(Type actualConstructorType, Type expectedConstructorType) =>
             $"Supplied constructor is for type {actualConstructorType}, but must be for {expectedConstructorType}.";
 
--- a/tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
+++ b/tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
@@ -289,11 +289,39 @@ namespace FakeItEasy.Specs
             "When the handler is subscribed to the fake's event"
                 .x(() => wrapper.SomeEvent += handler);
 
-            "And the event is raised"
+            "And the event is raised using a method on the wrapped object"
                 .x(() => realObject.RaiseSomeEvent());
 
             "Then the handler is called"
                 .x(() => A.CallTo(handler).MustHaveHappened());
+        }
+
+        [Scenario]
+        public static void RaisingEventOnWrappingFakeUsingRaiseSyntax(Foo realObject, IFoo wrapper, EventHandler handler, Exception exception)
+        {
+            "Given a real object"
+                .x(() => realObject = new Foo());
+
+            "And a fake wrapping this object"
+                .x(() => wrapper = A.Fake<IFoo>(o => o.Wrapping(realObject)));
+
+            "And an event handler"
+                .x(() => handler = A.Fake<EventHandler>());
+
+            "When the handler is subscribed to the fake's event"
+                .x(() => wrapper.SomeEvent += handler);
+
+            "And the event is raised using the '+= Raise' syntax"
+                .x(() => realObject.SomeEvent += Raise.WithEmpty());
+
+            "Then the handler is not called"
+                .x(() => A.CallTo(handler).MustNotHaveHappened());
+
+            "And when the event is raised using a method on the wrapped object"
+                .x(() => exception = Record.Exception(() => realObject.RaiseSomeEvent()));
+
+            "Then a NotSupportedException is thrown"
+                .x(() => exception.Should().BeAnExceptionOfType<NotSupportedException>());
         }
 
         public class Foo : IFoo
--- a/tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
+++ b/tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
@@ -16,6 +16,8 @@ namespace FakeItEasy.Specs
             void VoidMethod(string? parameter);
 
             void OutAndRefMethod(ref int @ref, out int @out);
+
+            event EventHandler? SomeEvent;
         }
 
         public interface IBar
@@ -272,6 +274,28 @@ namespace FakeItEasy.Specs
                     .WithMessage("The configured fake is not a wrapping fake."));
         }
 
+        [Scenario]
+        public static void EventSubscriptionSubscribesToWrappedEvent(Foo realObject, IFoo wrapper, EventHandler handler)
+        {
+            "Given a real object"
+                .x(() => realObject = new Foo());
+
+            "And a fake wrapping this object"
+                .x(() => wrapper = A.Fake<IFoo>(o => o.Wrapping(realObject)));
+
+            "And an event handler"
+                .x(() => handler = A.Fake<EventHandler>());
+
+            "When the handler is subscribed to the fake's event"
+                .x(() => wrapper.SomeEvent += handler);
+
+            "And the event is raised"
+                .x(() => realObject.RaiseSomeEvent());
+
+            "Then the handler is called"
+                .x(() => A.CallTo(handler).MustHaveHappened());
+        }
+
         public class Foo : IFoo
         {
             public bool NonVoidMethodCalled { get; private set; }
@@ -306,6 +330,12 @@ namespace FakeItEasy.Specs
                 @ref += 1;
                 @out = 42;
             }
+
+            public event EventHandler? SomeEvent;
+
+#pragma warning disable CA1030 // Use events where appropriate
+            public void RaiseSomeEvent() => this.SomeEvent?.Invoke(this, EventArgs.Empty);
+#pragma warning restore CA1030 // Use events where appropriate
         }
 
         public class Bar : IBar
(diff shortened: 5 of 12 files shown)


src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -23,7 +23,7 @@ public partial class FakeManager : IFakeCallProcessor
             new DefaultReturnValueRule(),
         };
 
-        private readonly IFakeObjectCallRule[] preUserRules;
```
thomaslevesque, line 26: It was the only pre-user rule, so I removed the field. We can always recreate it if we need to add a new pre-user rule.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -14,7 +14,7 @@ namespace FakeItEasy.Core
     /// </summary>
     public partial class FakeManager : IFakeCallProcessor
     {
-        private static readonly SharedFakeObjectCallRule[] PostUserRules =
+        private static readonly SharedFakeObjectCallRule[] SharedPostUserRules =
```
thomaslevesque, line 17: Up until now, all post-user rules were shared, but `EventRule` isn't, so I just made the name more explicit.

src/FakeItEasy/EnumerableExtensions.cs
```diff
@@ -61,5 +61,12 @@ public static IList<object> AsList(this IEnumerable source)
         {
             return source as IList<object> ?? source.Cast<object>().ToList();
         }
+
+#if LACKS_PREPEND
+        public static IEnumerable<T> Prepend<T>(this IEnumerable<T> source, T item)
+        {
+            return new[] { item }.Concat(source);
+        }
+#endif
```
thomaslevesque, line 70: Not strictly necessary, but the code is more readable using `Prepend`

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
```diff
@@ -272,6 +274,28 @@ public static void NotAWrappingFakeCallsWrappedMethod(IFoo fake, Exception excep
...
+            "When the handler is subscribed to the fake's event"
+                .x(() => wrapper.SomeEvent += handler);
+
+            "And the event is raised"
+                .x(() => realObject.RaiseSomeEvent());
+
+            "Then the handler is called"
+                .x(() => A.CallTo(handler).MustHaveHappened());
+        }
+
```
thomaslevesque, line 298: This should cover the scenario described by @mlaily.
We might need to check other scenarios as well, but I'm not sure which...

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
```diff
@@ -306,6 +330,12 @@ public void OutAndRefMethod(ref int @ref, out int @out)
                 @ref += 1;
                 @out = 42;
             }
+
+            public event EventHandler? SomeEvent;
+
+#pragma warning disable CA1030 // Use events where appropriate
+            public void RaiseSomeEvent() => this.SomeEvent?.Invoke(this, EventArgs.Empty);
+#pragma warning restore CA1030 // Use events where appropriate
```
thomaslevesque, line 338: IMO this CA1030 warning is useless. If I wanted this to be an event, I would have made it an event (which, in that case, wouldn't make any sense). We might want to disable this warning completely (we already had a few suppressions in the VB integration tests)

.editorconfig
```diff
@@ -22,14 +22,12 @@ insert_final_newline = true
...
 charset = utf-8-bom
 
-[project.json]
-insert_final_newline = false
-charset = utf-8-bom
-
-[*.nuget.targets]
-insert_final_newline = false
-charset = utf-8-bom
-
```
thomaslevesque, line 32: No longer necessary, probably dates back to the initial .NET Core migration.

.editorconfig
```diff
@@ -22,14 +22,12 @@ insert_final_newline = true
...
 [*.json]
 # 2-column space indentation
 indent_size = 2
+
+[*.cs]
+dotnet_style_qualification_for_event = true:silent
+dotnet_style_qualification_for_field = true:silent
+dotnet_style_qualification_for_method = true:silent
+dotnet_style_qualification_for_property = true:silent
```
thomaslevesque, line 33: VS was warning me for every `this.` prefix and suggesting to remove them, so I thought we might as well put the rule explicitly here (it's currently enforced by StyleCop). We could also add some naming rules here as well, I guess.

tests/FakeItEasy.Specs/UnconfiguredFakeSpecs.cs
```diff
@@ -200,7 +200,7 @@ public FakedClass()
 
             public virtual string StringProperty { get; set; }
 
-            public string StringPropertyValueDuringConstructorCall { get; }
+            public string? StringPropertyValueDuringConstructorCall { get; }
```
thomaslevesque, line 203: The compiler now detects that `StringProperty` might not have a non-null value when we assign it to `StringPropertyValueDuringConstructorCall` in the constructor.

thomaslevesque (commented):
The second commit is not related to the bug fix, I can move it to a separate PR if you prefer.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -48,10 +48,7 @@ internal FakeManager(Type fakeObjectType, object proxy, string? fakeObjectName)
             this.FakeObjectType = fakeObjectType;
             this.FakeObjectName = fakeObjectName;
 
-            this.preUserRules = new IFakeObjectCallRule[]
-            {
-                new EventRule(this),
-            };
+            this.postUserRules = SharedPostUserRules.Prepend<IFakeObjectCallRule>(new EventRule(this)).ToList();
```
blairconrad, line 51: I worry about the effect this will have on performance. We spent a fair amount of effort teasing apart constructed lists of rules during the big performance overhaul. I'm not going to die on this hill or anything, but maybe it's worth considering or investigating. I'm not entirely sure how expensive prepending, to-listing, and then switching to enumerating over the list is. It's less readable, but maybe we could've just renamed the preuser rules (and former postuser, why not?) and shifted the iteration lower, especially since right now our per-instance postuser rules are all grouped together?

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
blairconrad: I wouldn't mind seeing what happens when one uses `+= Raise…` to raise the event in this case. I don't expect it will work, and that could be fine, but I think it's worth trying, to see the effect. Heck, I might even keep the test with a comment indicating that we expect the behaviour.

I'd also check the interaction of event raising with `CallsBaseMethods`. And strict? I expect previously we could subscribe to events on a strict fake (and raise them). We won't be able to now. Which I think is fine?

I find I still don't have a complete handle on the interaction between these features. I'm going to need to think for a big before we merge, I think.

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
blairconrad: Sounds good. I could support it.

blairconrad (commented):
Generally looks like a good approach, @thomaslevesque. I think I raised more questions than provided opinions, though. I think I need some thinking time.

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: > I'm not entirely sure how expensive prepending, to-listing, and then switching to enumerating over the list is

Prepend is quite cheap, I think. ToList is a bit more expensive, and maybe not necessary. Enumeration is probably not very expensive.

> but maybe we could've just renamed the preuser rules (and former postuser, why not?) and shifted the iteration lower, especially since right now our per-instance postuser rules are all grouped together?

I'm not sure I follow. Could you elaborate?

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
thomaslevesque: Indeed, `+= Raise...` won't work. And I don't see how we could make it work, since we have no control on how the wrapped raises the event. Unless... we could do something like this:
- keep the default behavior that supports the `+= Raise...` syntax
- *also* forward the event subscription to the wrapped object
- when the wrapped object raises the event, the registered handlers are called directly by the wrapped object
- when `+= Raise` is used, the registered handlers are called by `EventRule`
This way, we still support both "modes". Of course, this approaches has its own problems. E.g. what if the wrapped object decides to clear the handlers? We wouldn't know about it and the handlers managed by `EventRule` would no longer be synchronized with the wrapped object. Also, the `sender` argument would be different depending on how the event was raised (maybe not a big issue, since the user can specify the sender in `Raise.With`)

> I'd also check the interaction of event raising with CallsBaseMethods.

What do you mean? `CallsBaseMethods` is not related to wrapping fakes, as far as I can tell.

> And strict? I expect previously we could subscribe to events on a strict fake (and raise them). We won't be able to now. Which I think is fine?

I think this is the correct behavior. I can add a spec to cover this.

blairconrad (commented):
Ha! Forgot to hit "Submit review!" :blush:

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: Ah, I see. In fact, since there's only one rule in `individualPostUserRules`, maybe we could just remove the array and directly call the rule; even better as far as performance is concerned, but maybe not as clean.

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
thomaslevesque: > If nothing else, if we can figure out a way to relax a strict fake so that it could start handling and raising events

Would it make sense, though? Everything would be strict except events? If we did that, why wouldn't we also add similar relaxations for properties or methods? It seems a little strange to do that just for events (unless you just want a way to restore the previous behavior, for compability?).

> in case there's a client out there who'd been making strict fakes and doing nothing but raising events.

Before this PR, `Strict` didn't affect event raising and subscription, so there was no point in making the fake strict. If anything, the existing behavior was a bug.

thomaslevesque: > True, but it's the interaction between the event rule and calling the base methods that I worry about

Ah, I see. I guess that currently, base methods aren't called for event subscription, so this would change with this PR

thomaslevesque: > I wouldn't mind seeing what happens when one uses += Raise… to raise the event in this case.

I added a spec for this case. As expected, using `+= Raise.With...` doesn't raise the event. But it does forwards the "dummy" handler to the wrapped object... So when the wrapped object actually raises the event, it throws an exception, since this handler is not actually supposed to be called. We probably need to do something to prevent that. I can see two ways to do it:
1. Split `EventRule` into two rules
  - One that handles the `+= Raise.With...` syntax (pre-user rule)
  - One that handles event subscription (post-user rule)
For a wrapping fake, `Raise.With` wouldn't actually do anything, since there would be no handlers registered on `EventRule`.
2. Handle it in `WrappedObjectRule` and throw an exception if the user tries to use `Raise.With`

I lean toward the second option, because it gives clearer feedback to the user

blairconrad: > Handle it in `WrappedObjectRule` and throw an exception if the user tries to use `Raise.With`

I also prefer this option, as the first one sounds like we'de be creating a variant of #1795.

I think we'd need a similar loophole for a Fake that `CallsBaseMethods`?

blairconrad: > > If nothing else, if we can figure out a way to relax a strict fake so that it could start handling and raising events
> 
> Would it make sense, though? Everything would be strict except events? If we did that, why wouldn't we also add similar relaxations for properties or methods? It seems a little strange to do that just for events (unless you just want a way to restore the previous behavior, for compability?).
> 
> > in case there's a client out there who'd been making strict fakes and doing nothing but raising events.
> 
> Before this PR, `Strict` didn't affect event raising and subscription, so there was no point in making the fake strict. If anything, the existing behavior was a bug.

I agree that it was a bug, but one that was very long-standing, so people may be relying on it. I was a little unclear. I wasn't suggesting we have a "Strict except events" mode. Just that we figure out how to allow clients to support event-registration and -raising on an otherwise strict Fake. It would help if we get a question. Maybe someone out there is using implicit creation options to make every Fake strict and they have some even-only interface. I know, it's not super likely.

But the main reason I mention it is that I know how to work around autoproperties on strict Fakes. We just make a setter that configures the getter to return the set value. I have _no idea_ how to turn on raising of events via `+=`. It would be nice to know if it's even possible to restore the functionality we're potentially taking away.

thomaslevesque: > I think we'd need a similar loophole for a Fake that `CallsBaseMethods`?

Do you mean `CallsWrappedMethod` ?
Yes. Unless we just let the exception be thrown when Raise.Now is called... But [the message is not great](https://github.com/FakeItEasy/FakeItEasy/blob/5b73e426467a83de6e65cc3a4337e98c57379c8a/src/FakeItEasy/ExceptionMessages.cs#L16), and anyway it's probably better to throw as early as possible.

thomaslevesque: > I wasn't suggesting we have a "Strict except events" mode. Just that we figure out how to allow clients to support event-registration and -raising on an otherwise strict Fake.

Ah, I see. To be honest, I'm not sure I like the idea. I'd rather make it a breaking change and make strict fakes really strict.

> But the main reason I mention it is that I know how to work around autoproperties on strict Fakes. We just make a setter that configures the getter to return the set value. I have no idea how to turn on raising of events via +=. It would be nice to know if it's even possible to restore the functionality we're potentially taking away.

Yes, I was thinking about that too. I don't think something like `A.CallTo(() =fake.Event += A<EventHandler>._)` is currently supported... And it probably couldn't be supported, because this would be legal in an expression anyway. We could do this instead:

```csharp
    EventHandler handler = null;
    A.CallTo(fake).Where(call => call.Method.Name == "add_Event").Invokes((EventHandler h) => handler += h);
    A.CallTo(fake).Where(call => call.Method.Name == "remove_Event").Invokes((EventHandler h) => handler -= h);
```
We might want to introduce an easier syntax, though...

thomaslevesque: Maybe `A.CallToSubscribe(fake, "Event")` instead of `A.CallTo(fake).Where(call => call.Method.Name == "add_Event")`?

blairconrad: >> I wasn't suggesting we have a "Strict except events" mode. Just that we figure out how to allow clients to support event-registration and -raising on an otherwise strict Fake.
>
>Ah, I see. To be honest, I'm not sure I like the idea. I'd rather make it a breaking change and make strict fakes really strict.

I can see I still wasn't clear. I would prefer strict Fakes to be strict as well. I just thought we should understand the technique (if there can be one) for adapting a Fake so that if one wanted to add the event-handling/raising behaviour after the fact, one could, just like one can explicitly configure other methods on a strict Fake.

> We could do this instead:
> 
> ```c#
>    EventHandler handler = null;
>    A.CallTo(fake).Where(call => call.Method.Name == "add_Event").Invokes((EventHandler h) => handler += h);
>    A.CallTo(fake).Where(call => call.Method.Name == "remove_Event").Invokes((EventHandler h) => handler -= h)
> ```

This is the sort of thing I was thinking of figuring out, yes. Then we could just raise by calling `handler`. Fantastic.
I'm not in a huge rush to find syntactic sugar, but if we did look for some, I think maybe something that does a little more work (since I think your example would still require the user to write the `Invokes` and so on…)?

>> I think we'd need a similar loophole for a Fake that `CallsBaseMethods`?
>
> Do you mean `CallsWrappedMethod`?

No, I meant that a Fake created with the `CallsBaseMethods` option would probably also need special handling, just as ones created that wrap a concrete object would.

thomaslevesque: > I can see I still wasn't clear. I would prefer strict Fakes to be strict as well. I just thought we should understand the technique (if there can be one) for adapting a Fake so that if one wanted to add the event-handling/raising behaviour after the fact, one could, just like one can explicitly configure other methods on a strict Fake.

OK, I think I understand this time.

> since I think your example would still require the user to write the Invokes

Indeed.

> No, I meant that a Fake created with the CallsBaseMethods option would probably also need special handling, just as ones created that wrap a concrete object would.

Oh. Sorry, I'm a bit slow today... Yes, that also applies to CallsBaseMethods.

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: Done

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -267,7 +255,12 @@ private void ApplyBestRule(IInterceptedFakeObjectCall fakeObjectCall)
                 return;
             }
 
-            foreach (var postUserRule in PostUserRules)
+            if (this.eventRule.IsApplicableTo(fakeObjectCall))
+            {
+                this.eventRule.Apply(fakeObjectCall);
```
blairconrad, line 260: `return`, no?

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
```diff
@@ -272,6 +274,56 @@ public static void NotAWrappingFakeCallsWrappedMethod(IFoo fake, Exception excep
...
+                .x(() => wrapper = A.Fake<IFoo>(o => o.Wrapping(realObject)));
+
+            "And an event handler"
+                .x(() => handler = A.Fake<EventHandler>());
+
+            "When the handler is subscribed to the fake's event"
+                .x(() => wrapper.SomeEvent += handler);
+
+            "And the event is raised using the '+= Raise' syntax"
+                .x(() => realObject.SomeEvent += Raise.WithEmpty());
```
blairconrad, line 315: Do we intend to try to raise the event from the `realObject`? This is an error, no? I think the more interesting behaviour is if we try to `Raise` from the Fake.

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
```diff
@@ -272,6 +274,56 @@ public static void NotAWrappingFakeCallsWrappedMethod(IFoo fake, Exception excep
...
+            "Then the handler is not called"
+                .x(() => A.CallTo(handler).MustNotHaveHappened());
+
+            "And when the event is raised using a method on the wrapped object"
+                .x(() => exception = Record.Exception(() => realObject.RaiseSomeEvent()));
+
+            "Then a NotSupportedException is thrown"
+                .x(() => exception.Should().BeAnExceptionOfType<NotSupportedException>());
+        }
+
```
blairconrad, line 326: Could we also add the sample code you had above (

```c#
EventHandler handler = null;
A.CallTo(fake).Where(call => call.Method.Name == "add_Event").Invokes((EventHandler h) => handler += h);
A.CallTo(fake).Where(call => call.Method.Name == "remove_Event").Invokes((EventHandler h) => handler -= h)
```
) or similar to either a test or an example so we can point to it in case people want to raise events from a wrapped Fake? I guess it's pretty unlikely, but still.

blairconrad (changes requested):
Looks generally good, @thomaslevesque. Thanks for the updates.
Perhaps I misunderstood our conversation (or was not clear), but shall we have Specs for the interaction of the event-Raising and a Fake that `CallsBaseMethods` as well?

thomaslevesque:
> Perhaps I misunderstood our conversation (or was not clear), but shall we have Specs for the interaction of the event-Raising and a Fake that `CallsBaseMethods` as well?

Yes, absolutely. I wasn't done with this PR, sorry if I made it look that way. I'll mark it as WIP until I'm actually done.

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: Yes! Good catch.

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
thomaslevesque: Yes. In fact, I changed that spec, but didn't yet commit the changes. The new spec ensures that using `+= Raise` on the fake throws a NotSupportedException (as opposed to throwing when the dummy handler is called)

tests/FakeItEasy.Specs/WrappingFakeSpecs.cs
thomaslevesque: Will do. We could also add it to the docs, I guess.

thomaslevesque:
@blairconrad I think I addressed all your comments. I also documented the limitation of `+= Raise` and the suggested workaround.
I removed commit e90e0b4d because it had nothing to do with the rest of the PR; I'll send it in a separate PR, probably with some other related changes.

docs/raising-events.md
```diff
@@ -83,3 +83,30 @@ AddHandler fake.CustomEvent, Raise.FreeForm(Of CustomEventHandler).With(fake, sa
...
+// Declare a delegate to store the event handlers
+EventHandler handlers = null;
+
+// Configure event subscription on the fake
+A.CallTo(robot).Where(call => call.Method.Name == "add_FellInLove").Invokes((EventHandler h) => handlers += h);
+A.CallTo(robot).Where(call => call.Method.Name == "remove_FellInLove").Invokes((EventHandler h) => handlers -= h);
+
+// Raise the event
+handlers?.Invoke(robot, EventArgs.Empty);
+```
```
blairconrad, line 112: This is beautiful.

src/FakeItEasy/ExceptionMessages.cs
```diff
@@ -57,6 +57,12 @@ internal static class ExceptionMessages
         public static string GetAssemblyFileNamesToScanForExtensionsShouldntReturnNull =>
             "IBootstrapper.GetAssemblyFileNamesToScanForExtensions must not return null";
 
+        public static string WrappingFakeCannotRaiseEvent =>
+            "The fake cannot raise the event because it's a wrapping fake and the event subscription is managed by the wrapped object. The event can only be raised internally by the wrapped object";
```
blairconrad, line 61: nit: I think you're missing a "." at the end.

blairconrad (commented):
Thought I could get through the changes, but I'm gonna stop after my complaint in the ExceptionMessages. See you in my morning.

blairconrad:
Thanks, @thomaslevesque. I made it through. Except for my one nitpick about the ending "." on the exception message, I'm content. If you're inclined to make the change, I'll merge. (Feel free to directly amend or squash or whatever.)

src/FakeItEasy/ExceptionMessages.cs
thomaslevesque: Fixed (amended the commit)

blairconrad (approved):

blairconrad:
Fantastic. Thanks, @thomaslevesque.

thomaslevesque:
Thanks for the review & merge!

PR #1801, opened 2020-11-18
Remove or update unsupported TFMs

### LINKED ISSUE #1798: Remove or update unsupported target frameworks
With the .NET 5.0 SDK, we get warnings like this in test and utility projects:

> Warning	NETSDK1138	The target framework 'netcoreapp3.0' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.

We should probably retarget to netcoreapp3.1 instead.

~~(for some reason, it doesn't complain about netcoreapp1.0...)~~

thomaslevesque:
And do some cleanup of unnecessary #ifs and compatibility code

Closes #1798


thomaslevesque:
Oops. Build failed because I forgot to push the changed to the submodule.

thomaslevesque (commented):
Most of the changes are straightforward, so I didn't comment inline.
Here are some per-commit comments.

* f0e2e0e1 Upgrade netcoreapp3.0 targets to netcoreapp3.1
* d923e5ea Remove netstandard1.6 and netcoreapp1.0 targets
* c8190689 Remove net40 target
* 64639758 Remove FEATURE_NETCORE_REFLECTION

    This symbol was (almost) only used by the netstandard1.6 and netcoreapp1.0 targets. It was mostly used to conditionally import `System.Reflection`, and for a few minor differences in the reflection API.
    The commit title isn't strictly true: I didn't completely remove the symbol. It's still used in the integration tests, to skip a test that isn't relevant in .NET Core. I replaced it with THROWS_ON_DUPLICATE_ASSEMBLY_LOAD in a subsequent commit.

* bc432ccd Remove USE_RUNTIMELOADER

    Was only used in netstandard1.6

* 7f51f55c Remove FEATURE_EXCEPTION_DISPATCH_INFO

    All supported frameworks have this feature now (we didn't have the symbol for net45 before, but we should have).

* ec24a142 Remove FEATURE_PARAMETERINFO_CUSTOMATTRIBUTES_PROPERTY

    All supported framework have this feature now.

* 148c66c7 Remove FEATURE_REFLECTION_GETASSEMBLIES

    All supported framework have this feature now.

* fa18f708 Remove FEATURE_EVENT_ARGS_MUST_EXTEND_EVENTARGS

    None of the supported framework have this constraint now.

* f6c1d0b7 Change FEATURE_ARRAY_EMPTY to LACKS_ARRAY_EMPTY

    Only net45 lacks `Array.Empty`, so it's probably better this way.

* f1153ef6 Remove GetMethodInfo, GetRuntimeInterfaceMap,

    These extension methods are no longer necessary (and maybe never were? not sure...)

* 1348757b Remove GetTypeInfo extension method

    This extension method just returned its argument, so that we wouldn't have to use #if to call `GetTypeInfo()` or not. But now we (almost) never need to call `GetTypeInfo()`. There are just a few cases where we need it for things that are exposed on `TypeInfo` and not on `Type`.

    (in fact, I'm not sure the code compiled after the previous commit... I should probably squash this one with its parent; I'll leave it like this for now as it makes reviewing easier)

* e4288c3f Remove MethodInfoWrapper

    It was used as a workaround because `ReflectedType` was missing in pre-release versions of .NET Core 1.0. It was added back in RTM, but we forgot to remove this.

* 3d652f62 Update how_to_build, remove netcore2+ profile

    netcore2+ and netcoreall were identical after removing netstandard1.6 and netcore1.0.
    Also some changes I forgot to make in the first commit

* 07ed8209 Replace FEATURE_NETCORE_REFLECTION with THROWS_ON_DUPLICATE_ASSEMBLY_LOAD

    More explicit

blairconrad:
> (in fact, I'm not sure the code compiled after the previous commit... I should probably squash this one with its parent; I'll leave it like this for now as it makes reviewing easier)

Yeah, I had wondered about that when I read the previous commit. I was pretty sure we needed the extension methods.

thomaslevesque:
> Yeah, I had wondered about that when I read the previous commit. I was pretty sure we needed the extension methods.

Just tried it: both f1153ef6 and 1348757b build successfully.
The reason I wasn't sure was because of some `using System.Reflection` I had to add in other parts of the code.
It's a bit confusing because of the `GetTypeInfo()` extension method that returns a `Type` rather than a `TypeInfo`...

blairconrad (approved):

blairconrad:
Nice, @thomaslevesque. Apply whatever you squashing you feel like, and I'll merge it. Thanks for this. I'm not sure whether it was tedious or fun.

thomaslevesque:
Squashed

> I'm not sure whether it was tedious or fun.

Mostly fun, actually. Removing code always feels good!
image

And in fact, I think there are other things we could remove. Like FEATURE_THREAD_CURRENTCULTURE. All supported frameworks now support setting either Thread.CurrentCulture or CultureInfo.CurrentCulture.

blairconrad:
> Mostly fun, actually. Removing code always feels good!

That was my guess, but then came (I'm sure) the checking of the docs, etc.

> And in fact, I think there are other things we could remove. Like FEATURE_THREAD_CURRENTCULTURE. All supported frameworks now support setting either Thread.CurrentCulture or CultureInfo.CurrentCulture.

Oooh! Fancy.

blairconrad (approved):

thomaslevesque:
Thanks for the merge!

PR #1802, opened 2020-11-21
Remove compilation constant FEATURE_THREAD_CURRENTCULTURE

blairconrad:
Continues #1801 a bit, as suggested by @thomaslevesque  in https://github.com/FakeItEasy/FakeItEasy/pull/1801#issuecomment-731448162

### DIFF
--- a/tests/FakeItEasy.Tests.TestHelpers/UsingCultureAttribute.cs
+++ b/tests/FakeItEasy.Tests.TestHelpers/UsingCultureAttribute.cs
@@ -23,28 +23,16 @@ namespace FakeItEasy.Tests.TestHelpers
 
         public override void After(MethodInfo methodUnderTest)
         {
-#if FEATURE_THREAD_CURRENTCULTURE
-            Thread.CurrentThread.CurrentCulture = this.originalCulture;
-            Thread.CurrentThread.CurrentUICulture = this.originalUiCulture;
-#else
             CultureInfo.CurrentCulture = this.originalCulture;
             CultureInfo.CurrentUICulture = this.originalUiCulture;
-#endif
         }
 
         public override void Before(MethodInfo methodUnderTest)
         {
-#if FEATURE_THREAD_CURRENTCULTURE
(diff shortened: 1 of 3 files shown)


tests/FakeItEasy.Tests.TestHelpers/UsingCultureAttribute.cs
```diff
@@ -23,28 +23,16 @@ public UsingCultureAttribute(string cultureName)
 
         public override void After(MethodInfo methodUnderTest)
         {
-#if FEATURE_THREAD_CURRENTCULTURE
             Thread.CurrentThread.CurrentCulture = this.originalCulture;
             Thread.CurrentThread.CurrentUICulture = this.originalUiCulture;
-#else
-            CultureInfo.CurrentCulture = this.originalCulture;
-            CultureInfo.CurrentUICulture = this.originalUiCulture;
-#endif
```
thomaslevesque, line 32: Any specific reason to keep `Thread.CurrentThread.CurrentCulture` instead of `CultureInfo.CurrentCulture`? Both work, but `CultureInfo.CurrentCulture` has always felt more natural to me.
Doesn't really matter though; if you prefer to keep it this way, I'll merge as is.

blairconrad:
I... Did not think. Just removed branch with constant. Will rework.

thomaslevesque (approved):

blairconrad:
Flipped. I just amended, rather than fixupping. That would've added confusion, I think.

tests/FakeItEasy.Tests.TestHelpers/UsingCultureAttribute.cs
blairconrad: Swapped!

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you. I like your version better too.

blairconrad:
Thanks for
1. suggesting
2. improving my implementation
3. merging,

@thomaslevesque!

blairconrad:

PR #1803, opened 2020-11-21
Add .NET 5.0 target

thomaslevesque:


how_to_build.md
```diff
@@ -94,6 +94,7 @@ available:
 * `net45`: builds only the .NET Framework 4.5 target framework
 * `netcore2.1`: builds only .NET Core 2.1 / .NET Standard 2.0 target frameworks
 * `netcore3.1`: builds only .NET Core 3.1 / .NET Standard 2.1 target frameworks
+* `net5.0`: builds only .NET 5.0 target frameworks
 * `netcoreall`: the default profile on non-Windows OSes, builds all supported
   .NET Core / .NET Standard target frameworks
```
blairconrad, line 99: Is 5.0 technically .NET Core? I don't mind 5.0 building from the `netcoreall` profile, but wonder if it wouldn't be better to mention 5.0 explicitly here and elsewhere where we talk about the supported TFMs.

Also, up on lines 84/85/86, we don't say that we build on .NET 5.0 (nor do we say we test on it, unless you think .NET Core _does_ include .NET 5.0…)

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
```diff
@@ -6,6 +6,10 @@
     <SignAssembly>true</SignAssembly>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net5.0'">
+    <DefineConstants>$(DefineConstants);FEATURE_STRING_CONTAINS_COMPARISONTYPE</DefineConstants>
```
blairconrad, line 10: A perfectly good change. Have we reached the point where we should swap to indicate that `String.Contains` lacks comparison type? We can wait until the next supported TFM change.

blairconrad (changes requested):

how_to_build.md
thomaslevesque: > Is 5.0 technically .NET Core?

Good question. "Technically", I'd say yes. The removal of "Core" from the name is just branding; the technology remains fundamentally the same as .NET Core. But it's true that the name `netcoreall` doesn't make it obvious that it includes .NET 5.0.

So... `netcoreand5.0` ?
Or maybe `fullexceptnetfx` ? i.e. instead of mentioning all included TFMs, we just mention the one that is excluded

Another possible approach: change `full` so that it includes net45 on Windows but not on Linux/MacOS. This way, the default profile is always `full`. And, as an added benefit, we don't need to come up with an awkward name for `netcoreall`.

> Also, up on lines 84/85/86, we don't say that we build on .NET 5.0 (nor do we say we test on it, unless you think .NET Core does include .NET 5.0…)

That's an oversight, I'll fix it.

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
thomaslevesque: > Have we reached the point where we should swap to indicate that `String.Contains` lacks comparison type?

In this project, yes (2 TFMs have the feature, 1 doesn't)
But in FakeItEasy.csproj, we have 2 TFMs with the feature and 2 without.

I guess we can swap now.

blairconrad: For me, it's more about expected future work. I like the idea of new TFMs having cleaner `.csproj`s. Make the old TFMs take more work! But it's not a huge deal. Whether or when we change doesn't rile me up too much.

how_to_build.md
blairconrad: > Another possible approach: change full so that it includes net45 on Windows but not on Linux/MacOS

I like it.

tests/FakeItEasy.IntegrationTests/FakeItEasy.IntegrationTests.csproj
thomaslevesque: > I like the idea of new TFMs having cleaner `.csproj`s.

Me too. Let's do it.

how_to_build.md
thomaslevesque: > I like it.

Great. I'll see if it's possible (I don't see why not, but who knows...)

thomaslevesque:
Oh, I almost forgot... I also have 103 nullability warnings to fix before this can be merged 😠

blairconrad:
> Oh, I almost forgot... I also have 103 nullability warnings to fix before this can be merged 😠

I _thought_ you mentioned that. I was surprised to only see a few.

Also, you touted the extra warnings as a benefit earlier, I thought. No frowning for you!

thomaslevesque:
> Also, you touted the extra warnings as a benefit earlier, I thought.

I do think they're a benefit. Doesn't mean I'm happy at having to fix them :wink:

thomaslevesque:
I just pushed a large commit to fix most nullability warnings. I'll post a review later today to explain the non-obvious ones.
There's one warning that I didn't fix, because I think it requires discussion:

```csharp
// In FakeManager
public virtual object Object => this.objectReference.Target;
```
`WeakReference.Target` can be null if the object was garbage collected, so the `Object` property should be nullable. But if we do so, about 20 new warnings appear, because in the rest of the code we never consider that it could be null. I'm not sure what we should do when it's null... Do nothing? Throw an exception? We probably need to decide on a case-by-case basis. I'll look into it in more details later, but right now I have to go out.

thomaslevesque:
> WeakReference.Target can be null if the object was garbage collected,

It's actually pretty easy to observe, actually:

```csharp
    var fake = A.Fake<IFoo>();
    var manager = Fake.GetFakeManager(fake);
    fake = null;
    GC.Collect();
    GC.WaitForPendingFinalizers();
    Console.WriteLine(manager.Object is null); // true
```

src/FakeItEasy/Configuration/ArgumentCollection.cs
```diff
@@ -125,7 +125,7 @@ public T Get<T>(string argumentName)
             return this.arguments;
         }
 
-        private string[] GetArgumentNames() => this.Method.GetParameters().Select(x => x.Name).ToArray();
+        private string[] GetArgumentNames() => this.Method.GetParameters().Select(x => x.Name!).ToArray();
```
thomaslevesque, line 128: `Name` can only be null for return parameters

src/FakeItEasy/Configuration/RuleBuilder.cs
```diff
@@ -294,7 +294,7 @@ public bool Matches(IFakeObjectCall call)
                        ReferenceEquals(this.builder.manager.Object, call.FakedObject);
             }
 
-            public override string ToString() =>
+            public override string? ToString() =>
```
thomaslevesque, line 297: To match the base definition in Object

src/FakeItEasy/Core/ArgumentConstraintTrap.cs
```diff
@@ -18,7 +18,7 @@ internal class ArgumentConstraintTrap
             new ThreadLocal<Action<IArgumentConstraint>>(() => ThrowWhenUnpreparedToTrapConstraint);
 
         public static void ReportTrappedConstraint(IArgumentConstraint constraint) =>
-            saveTrappedConstraintAction.Value.Invoke(constraint);
+            saveTrappedConstraintAction.Value!.Invoke(constraint);
```
thomaslevesque, line 21: Can't actually be null since a value factory is passed to the ThreadLocal constructor, and we never assign null to `Value`.

src/FakeItEasy/Core/ArgumentValueFormatter.cs
```diff
@@ -98,7 +98,7 @@ protected override string GetStringValue(object argumentValue)
 
                 return Fake.TryGetFakeManager(argumentValue, out var manager)
                     ? manager.FakeObjectDisplayName
-                    : argumentValue.ToString();
+                    : argumentValue.ToString() ?? string.Empty;
```
thomaslevesque, line 101: because ToString() can return null

src/FakeItEasy/Core/BootstrapperLocator.cs
```diff
@@ -34,7 +34,7 @@ public static IBootstrapper FindBootstrapper()
 
             var bootstrapperType = candidateTypes.FirstOrDefault() ?? typeof(DefaultBootstrapper);
 
-            return (IBootstrapper)Activator.CreateInstance(bootstrapperType);
+            return (IBootstrapper)Activator.CreateInstance(bootstrapperType)!;
```
thomaslevesque, line 37: Activator.CreateInstance can only return null if the type argument is `Nullable<T>`.

src/FakeItEasy/Core/CallRuleMetadata.cs
```diff
@@ -49,7 +49,7 @@ public bool HasNotBeenCalledSpecifiedNumberOfTimes()
         /// </summary>
         public void RecordCall() => ++this.CalledNumberOfTimes;
 
-        public override string ToString()
+        public override string? ToString()
```
thomaslevesque, line 52: To match the base definition in Object

src/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
```diff
@@ -21,7 +21,7 @@ public FakeManager GetFakeManager(object proxy)
         {
             Guard.AgainstNull(proxy, nameof(proxy));
 
-            FakeManager result = this.TryGetFakeManager(proxy);
+            FakeManager? result = this.TryGetFakeManager(proxy);
```
thomaslevesque, line 24: Can be null if the object isn't a fake

src/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
```diff
@@ -36,12 +36,11 @@ public FakeManager GetFakeManager(object proxy)
         /// </summary>
         /// <param name="proxy">The proxy to get the manager from.</param>
         /// <returns>The fake manager, or <c>null</c> if <paramref name="proxy"/> is not actually a faked object.</returns>
-        public FakeManager TryGetFakeManager(object proxy)
+        public FakeManager? TryGetFakeManager(object proxy)
```
thomaslevesque, line 39: Can be null if the object isn't a fake

src/FakeItEasy/Core/DelegateRaiser.cs
```diff
@@ -17,7 +17,7 @@ internal class DelegateRaiser<TEventHandler> : IEventRaiserArgumentProvider wher
         /// <param name="argumentProviderMap">A map from event handlers to supplied arguments to use when raising.</param>
         public DelegateRaiser(object?[] arguments, EventHandlerArgumentProviderMap argumentProviderMap)
         {
-            var delegateMethod = typeof(TEventHandler).GetMethod("Invoke");
+            var delegateMethod = typeof(TEventHandler).GetMethod("Invoke")!;
```
thomaslevesque, line 20: The Invoke method always exists on a delegate

src/FakeItEasy/Core/DynamicDummyFactory.cs
```diff
@@ -11,7 +11,7 @@ namespace FakeItEasy.Core
     internal class DynamicDummyFactory
     {
         private readonly IEnumerable<IDummyFactory> allDummyFactories;
-        private readonly ConcurrentDictionary<Type, IDummyFactory> cachedDummyFactories;
+        private readonly ConcurrentDictionary<Type, IDummyFactory?> cachedDummyFactories;
```
thomaslevesque, line 14: Can be null if no dummy factory matches the type

src/FakeItEasy/Core/DynamicRaiser.cs
```diff
@@ -28,7 +28,7 @@ public override bool TryConvert(ConvertBinder binder, out object result)
                 throw new InvalidCastException(ExceptionMessages.UnableToCast(this.GetType(), targetType));
             }
 
-            var method = targetType.GetMethod("Invoke");
+            var method = targetType.GetMethod("Invoke")!;
```
thomaslevesque, line 31: The Invoke method always exists on a delegate

src/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
```diff
@@ -14,7 +15,7 @@ public void AddArgumentProvider(Delegate eventHandler, IEventRaiserArgumentProvi
             this.map[eventHandler] = argumentProvider;
         }
 
-        public bool TryTakeArgumentProviderFor(Delegate eventHandler, out IEventRaiserArgumentProvider argumentProvider)
+        public bool TryTakeArgumentProviderFor(Delegate eventHandler, [NotNullWhen(true)] out IEventRaiserArgumentProvider? argumentProvider)
```
thomaslevesque, line 18: Can be null if no argumentProvider exists for this handler

src/FakeItEasy/Core/EventHandlerArgumentProviderMap.cs
```diff
@@ -28,7 +29,7 @@ public bool TryTakeArgumentProviderFor(Delegate eventHandler, out IEventRaiserAr
         /// </summary>
         private class EventRaiserDelegateComparer : IEqualityComparer<Delegate>
         {
-            public bool Equals(Delegate leftDelegate, Delegate rightDelegate)
+            public bool Equals(Delegate? leftDelegate, Delegate? rightDelegate)
```
thomaslevesque, line 32: To match the base definition in `IEqualityComparer<T>`

src/FakeItEasy/Core/FakeCallEqualityComparer.cs
```diff
@@ -6,7 +6,7 @@ namespace FakeItEasy.Core
     internal class FakeCallEqualityComparer
         : IEqualityComparer<IFakeObjectCall>
     {
-        public bool Equals(IFakeObjectCall x, IFakeObjectCall y)
+        public bool Equals(IFakeObjectCall? x, IFakeObjectCall? y)
```
thomaslevesque, line 9: To match the base definition in `IEqualityComparer<T>`

src/FakeItEasy/Core/FakeManager.CancellationRule.cs
```diff
@@ -37,14 +37,14 @@ public override void Apply(IInterceptedFakeObjectCall fakeObjectCall)
                     if (taskResultType == null)
                     {
                         var canceledTask = TaskHelper.Canceled();
-                        var ctor = returnType.GetConstructor(new[] { typeof(Task) });
+                        var ctor = returnType.GetConstructor(new[] { typeof(Task) })!;
```
thomaslevesque, line 40: We know for a fact that ValueTask has a constructor that accepts a Task

src/FakeItEasy/Core/FakeManager.CancellationRule.cs
```diff
@@ -37,14 +37,14 @@ public override void Apply(IInterceptedFakeObjectCall fakeObjectCall)
...
-                        var ctor = returnType.GetConstructor(new[] { typeof(Task) });
+                        var ctor = returnType.GetConstructor(new[] { typeof(Task) })!;
                         var valueTask = ctor.Invoke(new object[] { canceledTask });
                         fakeObjectCall.SetReturnValue(valueTask);
                     }
                     else
                     {
                         var canceledTask = TaskHelper.Canceled(taskResultType);
-                        var ctor = returnType.GetConstructor(new[] { canceledTask.GetType() });
+                        var ctor = returnType.GetConstructor(new[] { canceledTask.GetType() })!;
```
thomaslevesque, line 47: We know for a fact that ValueTask has a constructor that accepts a Task

src/FakeItEasy/Core/FakeManager.CancellationRule.cs
```diff
@@ -83,7 +83,8 @@ private static bool IsValueTask(Type type, out Type? valueType)
 
                 if (type.IsGenericType &&
                     !type.IsGenericTypeDefinition &&
-                    type.FullName.StartsWith("System.Threading.Tasks.ValueTask`", StringComparison.Ordinal))
+                    type.FullName is string fullName &&
+                    fullName.StartsWith("System.Threading.Tasks.ValueTask`", StringComparison.Ordinal))
```
thomaslevesque, line 87: FullName can be null in [some situations](https://docs.microsoft.com/en-us/dotnet/api/system.type.fullname?view=net-5.0)

src/FakeItEasy/Core/FakeManager.EventRule.cs
```diff
@@ -58,10 +58,9 @@ private void HandleEventCall(EventCall eventCall)
             {
                 if (eventCall.IsEventRegistration())
                 {
-                    IEventRaiserArgumentProvider argumentProvider;
                     if (EventHandlerArgumentProviderMap.TryTakeArgumentProviderFor(
                         eventCall.EventHandler,
-                        out argumentProvider))
+                        out var argumentProvider))
```
thomaslevesque, line 63: argumentProvider should be nullable but wasn't

src/FakeItEasy/Core/FakeManager.EventRule.cs
```diff
@@ -88,9 +87,7 @@ private void AddEventListener(EventCall call)
 
             private void AddHandler(object key, Delegate handler)
             {
-                Delegate result;
-
-                if (this.RegisteredEventHandlers.TryGetValue(key, out result))
+                if (this.RegisteredEventHandlers.TryGetValue(key, out var result))
```
thomaslevesque, line 90: result should be nullable but wasn't

src/FakeItEasy/Core/FakeManager.EventRule.cs
```diff
@@ -104,9 +101,7 @@ private void AddHandler(object key, Delegate handler)
 
             private void RemoveHandler(object key, Delegate handler)
             {
-                Delegate registration;
-
-                if (this.RegisteredEventHandlers.TryGetValue(key, out registration))
+                if (this.RegisteredEventHandlers.TryGetValue(key, out var registration))
```
thomaslevesque, line 104: registration should be nullable but wasn't

src/FakeItEasy/Core/FakeManager.EventRule.cs
```diff
@@ -123,9 +118,7 @@ private void RemoveHandler(object key, Delegate handler)
 
             private void RaiseEvent(EventCall call, IEventRaiserArgumentProvider argumentProvider)
             {
-                Delegate raiseMethod;
-
-                if (this.RegisteredEventHandlers.TryGetValue(call.Event, out raiseMethod))
+                if (this.RegisteredEventHandlers.TryGetValue(call.Event, out var raiseMethod))
```
thomaslevesque, line 121: raiseMethod should be nullable but wasn't

src/FakeItEasy/Core/FakeManager.EventRule.cs
```diff
@@ -145,8 +138,8 @@ private void RaiseEvent(EventCall call, IEventRaiserArgumentProvider argumentPro
 
             private class EventCall
             {
-                private static readonly Func<EventInfo, MethodInfo> GetAddMethod = e => e.GetAddMethod(true);
-                private static readonly Func<EventInfo, MethodInfo> GetRemoveMethod = e => e.GetRemoveMethod(true);
+                private static readonly Func<EventInfo, MethodInfo> GetAddMethod = e => e.GetAddMethod(true)!;
+                private static readonly Func<EventInfo, MethodInfo> GetRemoveMethod = e => e.GetRemoveMethod(true)!;
```
thomaslevesque, line 142: No idea when GetAddMethod or GetRemoveMethod can return null... I don't think it's possible to declare an event without add or remove in C#, but maybe directly in IL? Anyway, I don't know what we should do in such a case...

src/FakeItEasy/Core/FakeManager.EventRule.cs
```diff
@@ -191,7 +184,7 @@ private EventCall(EventInfo @event, MethodInfo callingMethod, Delegate eventHand
                         return null;
                     }
 
-                    var eventInfos = eventAdderOrRemover.DeclaringType.GetEvents(
+                    var eventInfos = eventAdderOrRemover.DeclaringType!.GetEvents(
```
thomaslevesque, line 187: DeclaringType can only be null for non-nested types, module-level methods, or dynamic methods.

src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
```diff
@@ -11,8 +11,8 @@ public partial class FakeManager
         private class PropertyBehaviorRule
             : IFakeObjectCallRule
         {
-            private readonly MethodInfo propertyGetter;
-            private readonly MethodInfo propertySetter;
+            private readonly MethodInfo? propertyGetter;
+            private readonly MethodInfo? propertySetter;
```
thomaslevesque, line 15: The code seems to handle the case when either of these fields is null

src/FakeItEasy/Core/FakeManager.PropertyBehaviorRule.cs
```diff
@@ -55,7 +55,7 @@ public void Apply(IInterceptedFakeObjectCall fakeObjectCall)
             private static PropertyInfo GetProperty(MethodInfo propertyGetterOrSetter)
             {
                 return
-                    (from property in propertyGetterOrSetter.DeclaringType.GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
+                    (from property in propertyGetterOrSetter.DeclaringType!.GetProperties(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic)
```
thomaslevesque, line 58: DeclaringType can only be null for non-nested types, module-level methods, or dynamic methods.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -70,6 +70,7 @@ internal FakeManager(Type fakeObjectType, object proxy, string? fakeObjectName)
         /// <summary>
         /// Gets the faked proxy object.
         /// </summary>
+        /// <remarks>Can be null if the proxy object has been collected by the garbage collector.</remarks>
```
thomaslevesque, line 73: That's the one I didn't fix yet, but I added the comment anyway...

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -142,7 +143,10 @@ public virtual void RemoveRule(IFakeObjectCallRule rule)
             lock (this.allUserRules)
             {
                 var ruleToRemove = this.allUserRules.FirstOrDefault(x => x.Rule.Equals(rule));
-                this.allUserRules.Remove(ruleToRemove);
+                if (ruleToRemove is object)
+                {
+                    this.allUserRules.Remove(ruleToRemove);
+                }
```
thomaslevesque, line 149: Remove doesn't expect a null (although I suspect it wouldn't have any ill effect)

src/FakeItEasy/Core/IFakeManagerAccessor.cs
```diff
@@ -21,7 +21,7 @@ internal interface IFakeManagerAccessor
         /// </summary>
         /// <param name="proxy">The proxy to get the manager from.</param>
         /// <returns>The fake manager, or <c>null</c> if <paramref name="proxy"/> is not actually a faked object.</returns>
-        FakeManager TryGetFakeManager(object proxy);
+        FakeManager? TryGetFakeManager(object proxy);
```
thomaslevesque, line 24: Can be null if proxy isn't a fake

src/FakeItEasy/Core/ImplicitOptionsBuilderCatalogue.cs
```diff
@@ -11,7 +11,7 @@ namespace FakeItEasy.Core
     internal class ImplicitOptionsBuilderCatalogue
     {
         private readonly IEnumerable<IFakeOptionsBuilder> allFakeOptionsBuilders;
-        private readonly ConcurrentDictionary<Type, IFakeOptionsBuilder> cachedFakeOptionsBuilders;
+        private readonly ConcurrentDictionary<Type, IFakeOptionsBuilder?> cachedFakeOptionsBuilders;
```
thomaslevesque, line 14: Can be null if no IFakeOptionsBuilder matches the type

src/FakeItEasy/Core/MethodInfoManager.cs
```diff
@@ -53,7 +53,7 @@ select MakeGeneric(typeMethod, method)).FirstOrDefault()
 
         private static MethodInfo? GetMethodOnInterfaceTypeImplementedByMethod(Type type, MethodInfo method)
         {
-            var reflectedType = method.ReflectedType;
+            var reflectedType = method.ReflectedType!;
```
thomaslevesque, line 56: RefectedType can only be null for non-nested types, module-level methods, or dynamic methods.

src/FakeItEasy/Core/MethodInfoManager.cs
```diff
@@ -88,7 +88,7 @@ where method.HasSameBaseMethodAs(methodTargetPair.TargetMethod)
         {
             var baseDefinition = method.GetBaseDefinition();
 
-            if (!baseDefinition.DeclaringType.IsInterface || !TypeImplementsInterface(type, baseDefinition.DeclaringType))
+            if (!baseDefinition.DeclaringType!.IsInterface || !TypeImplementsInterface(type, baseDefinition.DeclaringType))
```
thomaslevesque, line 91: DeclaringType can only be null for non-nested types, module-level methods, or dynamic methods.

src/FakeItEasy/Core/MethodInfoManager.cs
```diff
@@ -139,11 +139,9 @@ public override int GetHashCode()
             }
 
             [SuppressMessage("Microsoft.Usage", "CA2231:OverloadOperatorEqualsOnOverridingValueTypeEquals", Justification = "The type is used privately only.")]
-            public override bool Equals(object obj)
+            public override bool Equals(object? obj)
```
thomaslevesque, line 142: To match the base definition in Object

src/FakeItEasy/Core/Raise.of.TEventArgs.cs
```diff
@@ -73,13 +73,13 @@ internal Raise(TEventArgs e, EventHandlerArgumentProviderMap argumentProviderMap
         }
 
 #pragma warning disable CA1822 // Mark members as static
-        private void Now(object sender, TEventArgs e)
+        private void Now(object? sender, TEventArgs e)
```
thomaslevesque, line 76: To match the definition of `EventHandler<TEventArgs>`

src/FakeItEasy/Core/Raise.of.TEventArgs.cs
```diff
@@ -73,13 +73,13 @@ internal Raise(TEventArgs e, EventHandlerArgumentProviderMap argumentProviderMap
...
 #pragma warning disable CA1822 // Mark members as static
-        private void Now(object sender, TEventArgs e)
+        private void Now(object? sender, TEventArgs e)
 #pragma warning restore CA1822 // Mark members as static
         {
             throw new NotSupportedException(ExceptionMessages.NowCalledDirectly);
         }
 
-        private void Now(object sender, EventArgs e)
+        private void Now(object? sender, EventArgs e)
```
thomaslevesque, line 82: To match the definition of `EventHandler`

src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -47,7 +47,7 @@ public void Load(IEnumerable<string> extraAssemblyFiles)
                     // Just keep the types that were successfully loaded, and warn about the rest.
                     foreach (var type in ex.Types.Where(t => t is object))
                     {
-                        this.availableTypes.Add(type);
+                        this.availableTypes.Add(type!);
```
thomaslevesque, line 50: We now it's not null since we filtered out the nulls

src/FakeItEasy/Core/TypeCatalogue.cs
```diff
@@ -161,10 +161,10 @@ private static void WarnFailedToGetSomeTypes(Assembly assembly, ReflectionTypeLo
                 int notLoadedCount = ex.Types.Count(t => t is null);
                 writer.Write($"{notLoadedCount} type(s) were not loaded for the following reasons:");
                 writer.WriteLine();
-                foreach (var loaderException in ex.LoaderExceptions)
+                foreach (var loaderException in ex.LoaderExceptions.Where(ex => ex is object))
                 {
                     writer.Write(" - ");
-                    writer.Write(loaderException.GetType());
+                    writer.Write(loaderException!.GetType());
```
thomaslevesque, line 167: We now it's not null since we filtered out the nulls
(but we had a potential bug there before I added the filter)

src/FakeItEasy/Core/TypeCatalogueInstanceProvider.cs
```diff
@@ -32,7 +32,7 @@ public IEnumerable<T> InstantiateAllOfType<T>()
             {
                 try
                 {
-                    result.Add((T)Activator.CreateInstance(type));
+                    result.Add((T)Activator.CreateInstance(type)!);
```
thomaslevesque, line 35: Activator.CreateInstance can only return null if the type argument is `Nullable<T>`.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
```diff
@@ -169,7 +169,7 @@ public override int GetHashCode()
                 return HashCode;
             }
 
-            public override bool Equals(object obj)
+            public override bool Equals(object? obj)
```
thomaslevesque, line 172: To match the base definition in Object

src/FakeItEasy/Creation/CreationResult.cs
```diff
@@ -215,7 +215,7 @@ private ResolvedConstructorComparer()
 
             public static ResolvedConstructorComparer Default { get; } = new ResolvedConstructorComparer();
 
-            public bool Equals(ResolvedConstructor x, ResolvedConstructor y)
+            public bool Equals(ResolvedConstructor? x, ResolvedConstructor? y)
```
thomaslevesque, line 218: To match the base definition in IEqualityComparer<T>

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -23,7 +23,7 @@ internal static class DelegateProxyGenerator
         {
             Guard.AgainstNull(typeOfProxy, nameof(typeOfProxy));
 
-            var invokeMethod = typeOfProxy.GetMethod("Invoke");
+            var invokeMethod = typeOfProxy.GetMethod("Invoke")!;
```
thomaslevesque, line 26: The Invoke method always exists on a delegate

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -129,7 +129,7 @@ bool IsAccessibleImpl(Type t)
             {
                 // If the return type is non void, cast the result of eventRaiser.Raise()
                 // to the real return type and assign to the result variable
-                call = Expression.Assign(result, Expression.Convert(call, delegateMethod.ReturnType));
+                call = Expression.Assign(result!, Expression.Convert(call, delegateMethod.ReturnType));
```
thomaslevesque, line 132: We know it's not null because isVoid is false

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -151,7 +151,7 @@ bool IsAccessibleImpl(Type t)
                 bodyExpressions.Add(result!);
             }
 
-            var variables = isVoid ? new[] { arguments } : new[] { arguments, result };
+            var variables = isVoid ? new[] { arguments } : new[] { arguments, result! };
```
thomaslevesque, line 154: We know it's not null because isVoid is false

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -171,7 +171,7 @@ private static NewArrayExpression WrapParametersInObjectArray(ParameterExpressio
 
         private class DelegateCallInterceptedEventRaiser
         {
-            public static readonly MethodInfo RaiseMethod = typeof(DelegateCallInterceptedEventRaiser).GetMethod(nameof(Raise));
+            public static readonly MethodInfo RaiseMethod = typeof(DelegateCallInterceptedEventRaiser).GetMethod(nameof(Raise))!;
```
thomaslevesque, line 174: We know the method exists

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -53,7 +53,7 @@ public CreationResult TryResolveDummyValue(Type typeOfDummy, LoopDetectingResolu
 
             try
             {
-                return this.strategyCache.TryGetValue(typeOfDummy, out ResolveStrategy cachedStrategy)
+                return this.strategyCache.TryGetValue(typeOfDummy, out ResolveStrategy? cachedStrategy)
```
thomaslevesque, line 56: Null if not in cache

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -155,15 +155,16 @@ private class ResolveByCreatingTaskStrategy : ResolveStrategy
 
                 if (typeOfDummy.IsGenericType &&
                     !typeOfDummy.IsGenericTypeDefinition &&
-                    typeOfDummy.FullName.StartsWith("System.Threading.Tasks.ValueTask`", StringComparison.Ordinal))
+                    typeOfDummy.FullName is string fullName &&
+                    fullName.StartsWith("System.Threading.Tasks.ValueTask`", StringComparison.Ordinal))
```
thomaslevesque, line 159: FullName can be null in some cases

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -155,15 +155,16 @@ private class ResolveByCreatingTaskStrategy : ResolveStrategy
...
                 {
                     var typeOfTaskResult = typeOfDummy.GetGenericArguments()[0];
                     var creationResult = resolver.TryResolveDummyValue(typeOfTaskResult, resolutionContext);
                     object? taskResult = creationResult.WasSuccessful
                         ? creationResult.Result
                         : typeOfTaskResult.GetDefaultValue();
 
-                    var ctor = typeOfDummy.GetConstructor(new[] { typeOfTaskResult });
+                    var ctor = typeOfDummy.GetConstructor(new[] { typeOfTaskResult })!;
```
thomaslevesque, line 167: We know this constructor exists

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -183,7 +184,7 @@ private class ResolveByCreatingLazyStrategy : ResolveStrategy
             private static readonly MethodInfo CreateLazyDummyGenericDefinition =
                 typeof(ResolveByCreatingLazyStrategy).GetMethod(
                     nameof(CreateLazyDummy),
-                    BindingFlags.Static | BindingFlags.NonPublic);
+                    BindingFlags.Static | BindingFlags.NonPublic)!;
```
thomaslevesque, line 187: We know this method exists

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -327,7 +329,11 @@ private static bool TryCreateDummyValueUsingConstructor(ConstructorInfo construc
                 catch (TargetInvocationException e)
                 {
                     result = default;
-                    resolvedConstructor.ReasonForFailure = e.InnerException.Message;
+                    if (e.InnerException?.Message is string message)
+                    {
+                        resolvedConstructor.ReasonForFailure = message;
```
thomaslevesque, line 334: ReasonForFailure is `string?`, but has a `[DisallowNull]` attribute

src/FakeItEasy/Creation/FakeAndDummyManager.cs
```diff
@@ -63,7 +63,7 @@ public bool TryCreateDummy(Type typeOfDummy, LoopDetectingResolutionContext reso
 
         private static Func<ProxyOptions, IFakeOptions> GetFakeOptionsFactory(Type typeOfFake)
         {
-            var method = typeof(FakeAndDummyManager).GetMethod(nameof(CreateFakeOptions), BindingFlags.NonPublic | BindingFlags.Static).MakeGenericMethod(typeOfFake);
+            var method = typeof(FakeAndDummyManager).GetMethod(nameof(CreateFakeOptions), BindingFlags.NonPublic | BindingFlags.Static)!.MakeGenericMethod(typeOfFake);
```
thomaslevesque, line 66: We know this method exists

src/FakeItEasy/ExpressionExtensions.cs
```diff
@@ -44,7 +44,7 @@ internal static class ExpressionExtensions
         // - method calls
         // - array creation (including params arrays)
         // - quoted expressions
-        private static bool TryFastEvaluate(Expression expression, out object? result)
+        private static bool TryFastEvaluate(Expression? expression, out object? result)
```
thomaslevesque, line 47: The code handles the case where it's null, so it's clearly nullable

src/FakeItEasy/ExpressionExtensions.cs
```diff
@@ -124,7 +124,7 @@ private static bool TryFastEvaluate(Expression expression, out object? result)
 
                 case ExpressionType.NewArrayInit:
                     var newArrayExpression = (NewArrayExpression)expression;
-                    var arrayItems = Array.CreateInstance(newArrayExpression.Type.GetElementType(), newArrayExpression.Expressions.Count);
+                    var arrayItems = Array.CreateInstance(newArrayExpression.Type.GetElementType()!, newArrayExpression.Expressions.Count);
```
thomaslevesque, line 127: GetElementType() is not null because we know the type is an array type

src/FakeItEasy/Expressions/CallConstraintDescriber.cs
```diff
@@ -66,7 +66,7 @@ private static void WriteArgumentsListString(IOutputWriter writer, MethodInfo me
                     }
 
                     var parameter = parameters[index];
-                    writer.Write(parameter.Name).Write(": ");
+                    writer.Write(parameter.Name!).Write(": ");
```
thomaslevesque, line 69: It's not a return parameter, so Name is not null

src/FakeItEasy/Expressions/CallExpressionParser.cs
```diff
@@ -61,7 +61,7 @@ private static ParsedCallExpression ParsePropertyCallExpression(MemberExpression
             }
 
             return new ParsedCallExpression(
-                calledMethod: property.GetGetMethod(true),
+                calledMethod: property.GetGetMethod(true)!,
```
thomaslevesque, line 64: The property comes from a MemberExpression, and since an expression tree can't have an assignment, it has to be a read access, so there has to be a getter (unless, of course, the expression was created manually... should we handle this case?)

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -146,7 +146,7 @@ private static Type GetGenericTypeDefinition(Type type)
 
         private static void CheckConstraintIsCompatibleWithParameterType(ITypedArgumentConstraint constraint, Type parameterType)
         {
-            parameterType = parameterType.IsByRef ? parameterType.GetElementType() : parameterType;
+            parameterType = parameterType.IsByRef ? parameterType.GetElementType()! : parameterType;
```
thomaslevesque, line 149: We know it's a byref type, so there has to be an element type

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -171,7 +171,7 @@ private IArgumentConstraint GetArgumentConstraintFromExpression(Expression expre
             {
                 throw new UserCallbackException(
                     ExceptionMessages.UserCallbackThrewAnException($"Argument constraint expression <{expression}>"),
-                    ex.InnerException);
+                    ex.InnerException!);
```
thomaslevesque, line 174: TargetInvocationException always has an InnerException

src/FakeItEasy/Expressions/ExpressionArgumentConstraintFactory.cs
```diff
@@ -187,7 +187,7 @@ private IArgumentConstraint GetArgumentConstraintFromExpression(Expression expre
         private IArgumentConstraint CreateParamArrayConstraint(NewArrayExpression expression, Type parameterType)
         {
             var result = new List<IArgumentConstraint>();
-            var itemType = parameterType.GetElementType();
+            var itemType = parameterType.GetElementType()!;
```
thomaslevesque, line 190: we know parameterType is an array type, so there has to be an element type

src/FakeItEasy/Expressions/ParsedCallExpression.cs
```diff
@@ -11,7 +11,7 @@ internal class ParsedCallExpression
 
         public ParsedCallExpression(
             MethodInfo calledMethod,
-            Expression callTargetExpression,
+            Expression? callTargetExpression,
```
thomaslevesque, line 14: Could be null if it's a static method

src/FakeItEasy/Guard.cs
```diff
@@ -15,7 +16,7 @@ internal static class Guard
         /// <param name="argumentName">Name of the argument.</param>
         /// <exception cref="ArgumentNullException">The specified argument was null.</exception>
         [DebuggerStepThrough]
-        public static void AgainstNull([ValidatedNotNull]object argument, string argumentName)
+        public static void AgainstNull([ValidatedNotNull, NotNull] object? argument, string argumentName)
```
thomaslevesque, line 19: It's a bit counter-intuitive IMO, but `[NotNull]` is a post-condition: it doesn't prevent passing null for `argument`, but means that if the method returns, then `argument` is not null.
I'm not sure what `[ValidatedNotNull]` is; looks like it comes from Code Contracts, which I'm not sure is still supported...

src/FakeItEasy/ObjectMembers.cs
```diff
@@ -1,12 +1,12 @@
...
     using System.Reflection;
 
     internal static class ObjectMembers
     {
-        public static readonly MethodInfo EqualsMethod = typeof(object).GetMethod(nameof(object.Equals), new[] { typeof(object) });
-        public static readonly MethodInfo ToStringMethod = typeof(object).GetMethod(nameof(object.ToString), Type.EmptyTypes);
-        public static readonly MethodInfo GetHashCodeMethod = typeof(object).GetMethod(nameof(object.GetHashCode), Type.EmptyTypes);
+        public static readonly MethodInfo EqualsMethod = typeof(object).GetMethod(nameof(object.Equals), new[] { typeof(object) })!;
+        public static readonly MethodInfo ToStringMethod = typeof(object).GetMethod(nameof(object.ToString), Type.EmptyTypes)!;
+        public static readonly MethodInfo GetHashCodeMethod = typeof(object).GetMethod(nameof(object.GetHashCode), Type.EmptyTypes)!;
```
thomaslevesque, line 10: These methods are known to exist

src/FakeItEasy/OutputWriterExtensions.cs
```diff
@@ -50,7 +50,7 @@ public static IOutputWriter Write(this IOutputWriter writer, object value)
             Guard.AgainstNull(writer, nameof(writer));
             Guard.AgainstNull(value, nameof(value));
 
-            writer.Write(value.ToString());
+            writer.Write(value.ToString() ?? string.Empty);
```
thomaslevesque, line 53: ToString() could theoretically return null

src/FakeItEasy/ReturnValueConfigurationExtensions.cs
```diff
@@ -131,7 +131,7 @@ internal static void ReturnsNextFromQueue<T>(this IReturnValueConfiguration<T> c
             configuration.ReturnsLazily(x =>
             {
                 queue.TryDequeue(out T returnValue);
-                return returnValue;
+                return returnValue!;
```
thomaslevesque, line 134: This can't actually be null. TryDequeue will always return true, since we execute this only `queue.Count` times.
(and the ConcurrentQueue isn't exposed publicly, so it can't be modified somewhere else)

src/FakeItEasy/TaskHelper.cs
```diff
@@ -13,7 +13,7 @@ internal static class TaskHelper
         private static readonly MethodInfo CreateGenericCanceledTaskGenericDefinition =
             typeof(TaskHelper).GetMethod(
                 nameof(CreateGenericCanceledTask),
-                BindingFlags.Static | BindingFlags.NonPublic);
+                BindingFlags.Static | BindingFlags.NonPublic)!;
```
thomaslevesque, line 16: We know the method exists

src/FakeItEasy/TaskHelper.cs
```diff
@@ -53,7 +53,7 @@ public static Task Canceled(Type resultType)
                 type =>
                 {
                     var method = CreateGenericCanceledTaskGenericDefinition.MakeGenericMethod(type);
-                    return (Task)method.Invoke(null, ArrayHelper.Empty<object>());
+                    return (Task)method.Invoke(null, ArrayHelper.Empty<object>())!;
```
thomaslevesque, line 56: We know the method won't return null

tests/FakeItEasy.Tests.TestHelpers/TestCases.cs
```diff
@@ -15,13 +15,13 @@ public static class TestCases
...
-        public static IEnumerable<object[]> FromProperties<T>(params T[] cases)
+        public static IEnumerable<object?[]> FromProperties<T>(params T[] cases)
         {
             var properties = typeof(T).GetProperties(BindingFlags.Instance | BindingFlags.Public);
 
-            Func<T, object[]> objectToPropertiesProjection = x =>
+            Func<T, object?[]> objectToPropertiesProjection = x =>
             {
-                var result = new object[properties.Length];
+                var result = new object?[properties.Length];
```
thomaslevesque, line 24: Properties of T could return null

tests/FakeItEasy.Tests.TestHelpers/UsingCultureAttribute.cs
```diff
@@ -23,8 +23,17 @@ public UsingCultureAttribute(string cultureName)
...
+            {
+                CultureInfo.CurrentCulture = originalCulture;
+                this.originalCulture = null;
+            }
+
+            if (this.originalUiCulture is CultureInfo originalUiCulture)
+            {
+                CultureInfo.CurrentUICulture = originalUiCulture;
+                this.originalUiCulture = null;
+            }
```
thomaslevesque, line 36: We knew After would never be called before Before, but the compiler didn't...

thomaslevesque (commented):
I added comments (probably too many) to explain the changes

thomaslevesque:
> But if we do so, about 20 new warnings appear, because in the rest of the code we never consider that it could be null. I'm not sure what we should do when it's null... Do nothing? Throw an exception? We probably need to decide on a case-by-case basis. I'll look into it in more details later, but right now I have to go out.

OK, it's not as bad as I thought. Of the 23 warnings, only 4 are in the library itself, the rest are in unit tests. And in every case, there's no chance that the fake has been collected, so we can just use the null-forgiving operator. I'll push a commit with the remaining fixes, and the approved API changes.

thomaslevesque:
BTW, API approval tests are broken on AppVeyor. When received.txt doesn't match approved.txt, we get a null-reference exception (see build output [here](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/builds/36448858/job/0ldkoc7i9f4bqt0e#L189)). This happens on AppVeyor (both Windows and Linux), and I can reproduce it locally in WSL, but not in Windows.

thomaslevesque:
Oh. Looks like it's a problem you reported 4 years ago...
https://github.com/approvals/ApprovalTests.Net/issues/159

src/FakeItEasy/Configuration/StartConfiguration.cs
```diff
@@ -26,7 +26,7 @@ public IReturnValueArgumentValidationConfiguration<TMember> CallsTo<TMember>(Exp
         {
             Guard.AgainstNull(callSpecification, nameof(callSpecification));
 
-            var parsedCallExpression = this.expressionParser.Parse(callSpecification, this.manager.Object);
+            var parsedCallExpression = this.expressionParser.Parse(callSpecification, this.manager.Object!);
```
thomaslevesque, line 29: This method is called from `Fake<T>`, which has a strong reference to the fake object (proxy). So there's no chance that the fake could have been collected, so FakeManager.Object won't return null.

src/FakeItEasy/Core/FakeManager.EventRule.cs
```diff
@@ -120,7 +120,7 @@ private void RaiseEvent(EventCall call, IEventRaiserArgumentProvider argumentPro
             {
                 if (this.RegisteredEventHandlers.TryGetValue(call.Event, out var raiseMethod))
                 {
-                    var arguments = argumentProvider.GetEventArguments(this.fakeManager.Object);
+                    var arguments = argumentProvider.GetEventArguments(this.fakeManager.Object!);
```
thomaslevesque, line 123: This code runs when we use the `+= Raise...` syntax, so the calling code must have a strong reference to the fake object (proxy). So there's no chance that the fake could have been collected, so FakeManager.Object won't return null.

tests/FakeItEasy.Tests/Core/FakeManagerTests.cs
```diff
@@ -39,7 +39,7 @@ public void Event_listeners_that_are_removed_should_not_be_invoked_when_event_is
         public void Method_call_should_return_default_value_when_there_is_no_matching_interception_and_return_type_is_value_type()
         {
             var fake = this.CreateFakeManager<IFoo>();
-            var result = ((IFoo)fake.Object).Baz();
+            var result = ((IFoo)fake.Object!).Baz();
```
thomaslevesque, line 42: Here, it looks like the fake *could* have been collected, so `fake.Object` could return null. However, in `CreateFakeManager`, we call `MakeSureThatWeakReferenceDoesNotGetCollected`, which adds a strong reference to the fake in a list. So `fake.Object` won't return null.

src/FakeItEasy/Core/FakeManager.EventRule.cs
blairconrad: I think this is fine.

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -268,7 +270,7 @@ private class ResolveByInstantiatingClassUsingDummyValuesAsConstructorArgumentsS
                 // of constructors enumerable.
                 var consideredConstructors = new List<ResolvedConstructor>();
 
-                if (this.cachedConstructors.TryGetValue(typeOfDummy, out ConstructorInfo cachedConstructor))
+                if (this.cachedConstructors.TryGetValue(typeOfDummy, out ConstructorInfo? cachedConstructor))
```
blairconrad, line 273: no need to change but I'd've guessed `out var`

src/FakeItEasy/Expressions/CallExpressionParser.cs
blairconrad: Let's wait for the bug report!

src/FakeItEasy/Guard.cs
blairconrad: > It's a bit counter-intuitive IMO, but [NotNull] is a post-condition: it doesn't prevent passing null for argument, but means that if the method returns, then argument is not null.

:+1:

src/FakeItEasy/OutputWriterExtensions.cs
blairconrad: seems rude of it

tests/FakeItEasy.Tests.TestHelpers/UsingCultureAttribute.cs
blairconrad: Ugh. I mean, it's right, but ugh.

tests/FakeItEasy.Tests/RaiseTests.cs
```diff
@@ -125,8 +125,7 @@ public void Should_not_leak_handlers_when_raising()
             this.foo.SomethingHappened += raisingHandler;
 
             // Assert
-            IEventRaiserArgumentProvider ignoredArgumentProvider;
-            eventHandlerArgumentProvider.TryTakeArgumentProviderFor(raisingHandler, out ignoredArgumentProvider)
+            eventHandlerArgumentProvider.TryTakeArgumentProviderFor(raisingHandler, out _)
```
blairconrad, line 128: :+1:

blairconrad (approved):
Nice, job, @thomaslevesque. I'm pretty sure this one was tedious!
I found the many comments to be helpful. Thank you.

src/FakeItEasy/Creation/DummyValueResolver.cs
thomaslevesque: I went for minimal changes, I guess

thomaslevesque:
> I'm pretty sure this one was tedious!

Yes, this one definitely was!

thomaslevesque:
Thanks for the merge, @blairconrad

blairconrad:

blairconrad:
@thomaslevesque wondered what prompted us to undertake this initiative. Here's him suggesting it in a gitter chat

>I wonder if we should add a net5.0 target to the library.
> Technically, the netstandard2.1 assembly will work fine on net5.0, but netstandard no longer really makes sense in net5.0+; it's still supported, but isn't necessary anymore.
> Targeting net5.0 directly will remove the dependency on netstandard
> And it gives more complete nullability warnings, because the BCL has been annotated for nullability (if I add a net5.0 target, I get 103 new warnings)

PR #1805, opened 2020-11-24
Move to 'not null' pattern

### LINKED ISSUE #1796: Consider using 'not null' pattern
At some point, we adopted the pattern `x is object` for all null checks. But now C# 9 introduced a much more explicit syntax: `x is not null`. We should consider adopting this syntax instead.

@blairconrad any objection?

Potapy4:
These are fixes #1796

### DIFF
--- a/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
+++ b/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
@@ -49,7 +49,7 @@ namespace FakeItEasy
         {
             Guard.AgainstNull(manager, nameof(manager));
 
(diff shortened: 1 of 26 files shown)


thomaslevesque (approved):
Hi @Potapy4,

Thanks for the PR!
I can see there are a few instances of `is object` left, but that's OK:
- most of them are in Linq expressions, where pattern matching isn't supported
- there's one instance of `is object <variable>`, which is fine as it is

Looks good to me. @blairconrad, I'm not merging yet in case you want to take a look first.

blairconrad:
Thanks, @Potapy4. Looks very good to me. We appreciate the contribution.

PR #1806, opened 2020-11-26
Deprecate ClearConfiguration

### LINKED ISSUE #1767: Deprecate ClearConfiguration
As discussed in  #1764, it's somewhat confusing, and untangling it is probably not worth the effort. The preferred method of getting a Fake with no configuration is to just make a new one.

blairconrad:
Fixes #1767.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -4,16 +4,11 @@ FakeItEasy exposes a few APIs that aren't commonly needed, but can be useful in 
 
 ## Clearing the configuration of a fake
 
-The `Fake.ClearConfiguration` method removes all configured rules from a fake, leaving it in the default configuration.
-
-```csharp
-var foo = A.Fake<IFoo>();
-Assert.Equal(0, foo.Bar());
-A.CallTo(() => foo.Bar()).Returns(42);
-Assert.Equal(42, foo.Bar());
-Fake.ClearConfiguration(foo);
-Assert.Equal(0, foo.Bar());
-```
+To "unconfigure" a fake so it has default behavior, discard the fake and create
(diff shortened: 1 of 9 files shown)


docs/advanced-usage.md
```diff
@@ -4,16 +4,8 @@ FakeItEasy exposes a few APIs that aren't commonly needed, but can be useful in
...
-```csharp
-var foo = A.Fake<IFoo>();
-Assert.Equal(0, foo.Bar());
-A.CallTo(() => foo.Bar()).Returns(42);
-Assert.Equal(42, foo.Bar());
-Fake.ClearConfiguration(foo);
-Assert.Equal(0, foo.Bar());
-```
+The best way to remove a fake's configuration is to discard the fake and create
+a new one to replace it.
```
thomaslevesque, line 8: Saying that it's the "best way" suggests there's an other way. Maybe you could mention that `ClearConfiguration` exists, but is deprecated? Or just don't say that it's the best way, and don't mention `ClearConfiguration`?

thomaslevesque (changes requested):
Thanks @blairconrad, looks good to me. I just made one suggestion about the doc changes.

docs/advanced-usage.md
blairconrad: You're right. I've given it another go (I just amended the commit; no changes to any other files). I'm keen to hear your thoughts.

thomaslevesque: Perfect!

thomaslevesque (approved):

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you, @thomaslevesque. An astute comment and alacritous merging.

PR #1807, opened 2020-11-27
Update supporting package dependencies

blairconrad:
Updates non-production dependencies.
That is, System.Threading.Tasks.Extensions and Castle.Core were intentionally untouched.


blairconrad:
I cleaned up new warnings and errors introduced by the dependencies. I'd be inclined to squash everything into one commit, but left them separate for easy reviewing/manipulation for now.

Directory.Build.props
```diff
@@ -11,8 +11,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.113" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="2.9.8" PrivateAssets="all" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.261" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.1" PrivateAssets="all" />
```
thomaslevesque, line 15: Actually, I think we no longer need this package.
These analyzers are included in the .NET 5.0 SDK, with most of the rules disabled by default, but we can enable them with a property like this:
```xml
<AnalysisMode>AllEnabledByDefault</AnalysisMode>
```

tests/FakeItEasy.Tests/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidatorTests.cs
```diff
@@ -42,7 +42,9 @@ public CastleDynamicProxyInterceptionValidatorTests()
                     () => object.Equals("foo", "bar"),
                     "Static methods can not be intercepted."),
                 new NonInterceptableTestCase(
-                    () => "foo".Count(),
+                #pragma warning disable CA1806 // Do not ignore method results
+                    () => "foo".Distinct(),
+                #pragma warning restore CA1806 // Do not ignore method results
```
thomaslevesque, line 47: I don't understand this change... Why replace Count with Distinct?

thomaslevesque (commented):
Thanks @blairconrad. Looks good to me, I just have one suggestion and one question.

tests/FakeItEasy.Tests/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidatorTests.cs
blairconrad: Because `Count()` gives a warning about us not using `Length`. And a similar warning about us not using `Length`, but slightly different AND CA1806. I preferred one pragma to 3!

blairconrad: I should've said.

thomaslevesque: Ah, I see. Probably doesn't matter here, but `Distinct()` has no effect unless the result is enumerated, so it looked a bit strange

thomaslevesque (dismissed):

tests/FakeItEasy.Tests/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidatorTests.cs
blairconrad: Yeah, we just need to identify a method. I can comment up, if you like.

blairconrad:
Added commits to drop FxCop and update the build project's tools (oops). Interestingly, this highlighted a couple more minor warnings. I'll note them.

tests/FakeItEasy.Tests/Core/MethodInfoManagerTests.cs
```diff
@@ -213,7 +213,7 @@ public void DoSomething(string text)
             [SuppressMessage("Microsoft.Usage", "CA1801:ReviewUnusedParameters", MessageId = "other", Justification = "Required for testing.")]
             public bool Equals([AllowNull] Base other) => true;
 
-            public override bool Equals(object? other) => true;
+            public override bool Equals(object? obj) => true;
```
blairconrad, line 216: because the base method calls it that!

tests/FakeItEasy.Tests/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidatorTests.cs
```diff
@@ -42,9 +42,9 @@ public CastleDynamicProxyInterceptionValidatorTests()
                     () => object.Equals("foo", "bar"),
                     "Static methods can not be intercepted."),
                 new NonInterceptableTestCase(
-                #pragma warning disable CA1806 // Do not ignore method results
+                    #pragma warning disable CA1806 // Do not ignore method results
                     () => "foo".Distinct(),
-                #pragma warning restore CA1806 // Do not ignore method results
+                    #pragma warning restore CA1806 // Do not ignore method results
```
blairconrad, line 47: last minute change - I thought the indentation looked better

tests/FakeItEasy.Tests/Expressions/CallExpressionParserTests.cs
```diff
@@ -28,7 +28,7 @@ public CallExpressionParserTests()
         public void Should_return_parsed_expression_with_instance_method_set()
         {
             // Arrange
-            var call = Call(() => string.Empty.Equals(null, StringComparison.CurrentCulture));
+            var call = Call(() => string.Empty.Equals(null, StringComparison.Ordinal));
```
blairconrad, line 31: compiler preferred. It doesn't really matter, since we're just using it to specify the signature…

tests/FakeItEasy.Tests/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidatorTests.cs
thomaslevesque: No need. Thanks for the explanation

thomaslevesque (approved):
Looks good to me. The title still says [WIP], are you expecting to make more changes?

blairconrad:
Thanks for review. I'd intended to squash to 1, since I wasn't happy with
the way the commits told a story. Probably no value in having multiple. But
if you prefer separate, I can just remove WIP..

blairconrad:
Renamed in case you want the commits as-are. Otherwise, I'll squash when you say.

thomaslevesque:
Squashed would be better I think

blairconrad:
Squashed, but I'm seeing warnings on the build server that I don't see here. Hold off, please.

blairconrad:
Nope, I'm getting them. Or at least some. I'm a little confused.

thomaslevesque:
Wow, that's a lot of warnings. Did you figure it out?
Maybe it's caused by using the SDK analyzers instead of the package? When I made the switch in a project at work, I saw a few extra warnings as well.
And there are many AD0001 warnings, which indicate bugs in the analyzers...

tests/FakeItEasy.IntegrationTests.VB/RaisingEventsTests.vb
```diff
@@ -11,7 +11,7 @@ Public Interface IHaveEvents
     Event NonGenericEventHander As EventHandler
     Event GenericEventHander As EventHandler(Of MyEventArgs)
 
-    <SuppressMessage("Microsoft.Design", "CA1009:DeclareEventHandlersCorrectly",
+    <SuppressMessage("Microsoft.Design", "CA1003:DeclareEventHandlersCorrectly",
```
blairconrad, line 14: The code seems to have changed?

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
```diff
@@ -104,7 +104,7 @@ public class Foo { }
             var emitResult = compilation.Emit(this.AssemblyDependencyPath);
             if (!emitResult.Success)
             {
-                throw new Exception("Failed to create assembly - " + emitResult.Diagnostics[0]);
+                throw new InvalidOperationException("Failed to create assembly - " + emitResult.Diagnostics[0]);
```
blairconrad, line 107: Many instances of `Exception` being too generic. I struggled, but picked some specific ones.

tests/FakeItEasy.Specs/ConfigurationSpecs.cs
```diff
@@ -873,6 +873,7 @@ public Task BarAsync()
             }
         }
 
+        [SuppressMessage("Microsoft.Design", "CA1033:InterfaceMethodsShouldBeCallableByChildTypes", Justification = "Required for testing.")]
```
blairconrad, line 876: They wanted it to be `sealed` which would rather defeat the purpose.

tests/FakeItEasy.Specs/EventRaisingSpecs.cs
```diff
@@ -9,10 +9,13 @@ namespace FakeItEasy.Specs
     using Xbehave;
     using Xunit;
 
+    [SuppressMessage("Microsoft.Naming", "CA1711:IdentifiersShouldNotHaveIncorrectSuffix", Justification = "Best name to describe the type.")]
```
blairconrad, line 12: [CA1711](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1711). When I switched everything to be just "Handler" instead of "EventHandler", `CustomEventHandler` came up with a [CA1710](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1710). I figured I could
1. annotate the 3 classes, 
2. mix the naming styles, or
3. change all and annotate CustomEventHandler

I picked this, but am happy to change.

blairconrad (commented):
> Wow, that's a lot of warnings. Did you figure it out?


I had to make changes or clean the output to reproduce. I thought we were set up to rebuild each time, so am surprised.

> And there are many AD0001 warnings, which indicate bugs in the analyzers...

I quelled all but those and pushed. 
I found https://github.com/dotnet/roslyn-analyzers/issues/3613 talking about the AD0001 errors, but it claims to be fixed in May. I'm currently lost, but will look more.

tests/FakeItEasy.IntegrationTests.VB/RaisingEventsTests.vb
thomaslevesque: Looks like there are two different warnings:
[CA1003](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1003): Use generic event handler instances
[CA1009](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1009): Declare event handlers correctly

I assume the analyzers changed to emit the new one in this case.

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
thomaslevesque: Me, I would have just disabled the warning for the tests...

tests/FakeItEasy.Specs/EventRaisingSpecs.cs
thomaslevesque: Whatever works! Again, I think I would have just disabled the warning in test code.

thomaslevesque:
Shouldn't we disable AD0001 globally? It indicates a bug in the analyzer, so there's nothing we can do about it (except perhaps report it)

blairconrad:
> Shouldn't we disable AD0001 globally? It indicates a bug in the analyzer, so there's nothing we can do about it (except perhaps report it)

It's something I'm considering. There's also the possibility we're getting an indirect reference to an older assembly or something. I'm inclined to dig a little more before just ignoreing.

blairconrad:
Doing desultory investigation between bouts of paid work. Confirmed that the problem only appears when building the .net 5.0 assemblies.

tests/FakeItEasy.IntegrationTests.VB/RaisingEventsTests.vb
blairconrad: You're right. I should update the um... title? Description?

tests/FakeItEasy.Specs/EventRaisingSpecs.cs
blairconrad: Oh, I can change to disable globally within tests if you prefer.

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
blairconrad: Oh, I can change to disable globally within tests if you prefer.

tests/FakeItEasy.IntegrationTests.VB/RaisingEventsTests.vb
thomaslevesque: Probably, yes. I assume this is some kind of identifier, but I don't know where it comes from.
*or* you could just remove the `[SuppressMessage]` attribute and use the IDE to generate the `#pragma warning disable` for you.

tests/FakeItEasy.IntegrationTests/ExternalAssemblyGenerator.cs
thomaslevesque: Sounds good to me.

blairconrad:
Hey, @thomaslevesque , I didn't get much more investigating of the AD error done today, but I just now pushed 4060628bbf59e34a551b3ae7a52526fae66429da which I think addresses your other comments. One difference from your suggestion is that instead of `#pragma`ing `RaisingEventsTests.vb`, I noticed that we were already ignoring CA1003 in Specs so I hoisted up to all tests. I hope that's acceptable. If you've any concerns, I'll revert and pragma instead. 
Sorry for the roundabout PR. I'm enjoying your review, though.

thomaslevesque:
> I noticed that we were already ignoring CA1003 in Specs so I hoisted up to all tests. I hope that's acceptable.

Perfect 👍

src/CommonAssemblyInfo.cs
```diff
@@ -1,3 +1,5 @@
+using System;
 using System.Runtime.InteropServices;
 
+[assembly: CLSCompliant(true)]
```
thomaslevesque, line 4: Is this actually true? I mean, there are a number of [rules](https://docs.microsoft.com/en-us/dotnet/standard/language-independence-and-language-independent-components#cls-compliance-rules) to follow in order to be CLS compliant. Since we never worried about it before, I'm not sure we actually followed them...

thomaslevesque: Oh, we already had this on the FakeItEasy assembly. Never mind.

thomaslevesque:
Thanks @blairconrad, looks good to me. I'll merge once you squash the fixups and removed the [WIP] (assuming you're done)

blairconrad:
> I'll merge once you squash the fixups and removed the [WIP] (assuming you're done)

We still have the AD0001. Beginning to feel lazy and may just suppress as you say. I'm not sure what we're doing wrong here.

thomaslevesque:
> We still have the AD0001. Beginning to feel lazy and may just suppress as you say. I'm not sure what we're doing wrong here.

I don't think we're doing anything wrong. IMO it's a bug in the analyzer, so we can't do anything about it.

blairconrad:
Thanks, @thomaslevesque. Pushed the suppression. If you're happy I'll squash everything down to 1 commit
Um, assuming our dependencies haven't released new versions by now, it's been so long. 😉

blairconrad:
I've been reading, and it seems that EditorConfig files are the wave of the present for configuring what to do with which analysis rules, replacing .rulesets. I'm intrigued, but would rather explore in a different PR.

thomaslevesque:
> I've been reading, and it seems that EditorConfig files are the wave of the present for configuring what to do with which analysis rules, replacing .rulesets.

Indeed. However, some diagnostics can't be configured in .editorconfig at this point... Basically, any diagnostic that doesn't specify a location in code (e.g. [CA1014: Mark assemblies with CLSCompliantAttribute](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1014)) must still be configured in a ruleset (or just suppressed in the project with `<NoWarn>`)

> I'm intrigued, but would rather explore in a different PR.

👍

thomaslevesque:
> Quell AD0001 in VB projects

Oh! I hadn't even realized that these warnings were only in VB projects...

Looks good to me, squash it!

blairconrad:
Squashed!

thomaslevesque (approved):
Thanks @blairconrad!

blairconrad:
Thanks for the adjustments and merge!

(I'm starting to convert to .editorconfigs in the background.)

blairconrad:

PR #1808, opened 2020-12-04
Use editorconfig to suppress warnings

blairconrad:
No functional change or anything. Just a different way to suppress warnings. I made many commits so it'd be easy to drop some if we wanted. Or the whole thing!
I wouldn't propose merging with the same list of commits. Would squash. Either to 1 commit or to 1 up to and including "4a6e886b Promote opt-in warnings to top level from src" and one for the rest, maybe? I dunno. 1 is probbaly better. 
Anyhow, here's some explanation:

The bulk of the work. Mostly just transformed the rulesets into .editorconfigs. Possibly moved a setting here and there when they seemed oddly- scoped.
* eed5edc2 Replace rulesets with .editorconfig

Removing some inline suppressions that were covered by the .editorconfigs (and probably the rulsets)
* 9df0ae0d Remove superfluous CA1003 (Use generic event handler instances)
* ae76f0f9 Remove superfluous CA1822 (Member does not access instance data and can be marked as static)
* 495336dd Remove superfluous SA1402 (File may only contain a single type)
* 1144ef7f Remove superfluous SA1401 (Fields must be private)
* 09219386 Remove superfluous CA1501 (Do not declare visible instance fields)

Not necessary, I suppose, but they didn't trigger any warnings. Don't know why we'd want to break the rules for samples or tests...
* 4a6e886b Promote opt-in warnings to top level from src

The controversial section, I think. In some cases, I thought we'd be better off not needing a suppression. Happy to undo any.
* c7a61b8b Fix SA1107 and SA1128 instead of suppressing
* 51ef79d4 Fix CA1716 (Identifiers should not match keywords) instead of suppressing
* 202135fd Fix CA1724 (Type names should not match namespaces) instead of suppressing
* 593f8e70 Fix SA1121 (Use built-in type alias) instead of suppressing
* 69fac344 Fix CA1200 (Avoid using cref tags with a prefix) instead of suppressing
* 98af1a6b Remove empty .editorconfig
    A separate commit because I find it annoying to edit the .sln (I don't like opening VS anymore) and thought if we drop one of the other commits, we could just drop this too.

We had a broad suppression that only needed to be applied narrowly
* d7a88321 Move CA1801 (Parameter is never used) suppression to pragma


src/FakeItEasy/IHideObjectMembers.cs
```diff
@@ -6,9 +6,9 @@ namespace FakeItEasy
 
     /// <summary>
     /// Hides standard Object members to make fluent interfaces
-    /// easier to read. Found in the source of Autofac: <see cref="!:https://code.google.com/p/autofac/"/>
+    /// easier to read. Found in the source of Autofac: <see href="https://code.google.com/p/autofac/"/>
```
blairconrad, line 9: the output XML was only changed in the way you'd expect from this, to include an HREF, which I understand is now allowed

src/FakeItEasy/IHideObjectMembers.cs
```diff
@@ -21,11 +21,11 @@ public interface IHideObjectMembers
         string? ToString();
 
         /// <summary>
-        /// Determines whether the specified <see cref="System.Object"/> is equal to this instance.
+        /// Determines whether the specified <see cref="object"/> is equal to this instance.
```
blairconrad, line 24: These sorts of changes made no difference to the generated docs.

thomaslevesque (changes requested):

I agree with most of the changes, except maybe this one:

> 51ef79d Fix CA1716 (Identifiers should not match keywords) instead of suppressing

I think suppressing this warning in tests is fine. It has to do with CLS compliance (I think), but the tests aren't going to be used by anyone, they don't need to be CLS compliant.

blairconrad:
> I agree with most of the changes, except maybe this one:
> 
> > [51ef79d](https://github.com/FakeItEasy/FakeItEasy/commit/51ef79d45f1e5a6b8b7acd3a755c090c657aa452) Fix CA1716 (Identifiers should not match keywords) instead of suppressing

Fair enough! I reverted the commit in ed23e72c7a73632d81240fae029b1734d6461e8b (and changed the title to WIP so we wouldn't accidentally merge). I belatedly realize that that suppresses it in Specs only, but that should be fine for now? I can pull up, if you want. If so, I could do that and squash as one, now that we know we can compare to ed23e72c7a73632d81240fae029b1734d6461e8b. Otherwise, tells me what you want and I'll do it, even if it's just squash what we have.

thomaslevesque:
> I belatedly realize that that suppresses it in Specs only, but that should be fine for now? I can pull up, if you want

We can do this later if necessary

> Otherwise, tells me what you want and I'll do it, even if it's just squash what we have.

Looks good to me, just squash!

blairconrad:
Squashed!

thomaslevesque (approved):
Thanks, @blairconrad!

blairconrad:
Thanks for the merge!

blairconrad:

PR #1817, opened 2021-02-13
Allow specifying number of calls directly after Invokes

### LINKED ISSUE #1812: Improve interface of Invokes to allow specifing number of calls
Currently it is not possible to directly specify a number of calls when configuring Invokes, e.g. given the following interface:

```c#
public interface ITest
{
	int Test(int testNumber);
}
```
a configuration like:
```c#
[Test]
public void Test()
{
	var counter = 0;

	var test = A.Fake<ITest>();

	A.CallTo(() => test.Test(1))
		.Invokes((int testNumber) => counter++)
		.Once();
}
```
does not compile.

However, all it takes is a cast and it works flawlessly:

```c#
[Test]
public void Test()
{
	var counter1 = 0;
	var counter2 = 0;

	var test = A.Fake<ITest>();

	((IAfterCallConfiguredConfiguration<IReturnValueConfiguration<int>>)A.CallTo(() => test.Test(1))
		.Invokes((int testNumber) => counter1++))
		.Once()
		.Then
		.Invokes(() => counter2++);

	test.Test(1);
	test.Test(1);
	test.Test(1);

	Assert.That(counter1, Is.EqualTo(1));
	Assert.That(counter2, Is.EqualTo(2));
}
```

I suppose this is unintended and the number of calls specification should be available directly, without a cast.

blairconrad:
Fixes #1812.

I should play a little more with the API to make sure there aren't any surprises. Also considering

* looking hard again at `PropertySetterAfterCallbackConfiguredAdapter`. I didn't enjoy adding it. It's just private, so probably no big deal, but still
* adding a test (integration? Maybe a spec?) to ensure that `Invokes(…).MustHaveHappened(…)` is illegal. I think this could be accomplished using a [CSharpCompilation class](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.csharpcompilation?view=roslyn-dotnet), but would be keen to hear other options, including "don't bother"

Oh, I'd consider squashing too. The separate commits didn't end up adding anything, as I I think there's no interaction between them. But it's not a big deal either way.


thomaslevesque (approved):
Thanks @blairconrad, looks good to me! I only made one (very) minor comment.

> looking hard again at PropertySetterAfterCallbackConfiguredAdapter. I didn't enjoy adding it. It's just private, so probably no big deal, but still

Yeah, it's a bit ugly, but that's not a new problem, and I'm not sure we can do anything about it...

> adding a test (integration? Maybe a spec?) to ensure that Invokes(…).MustHaveHappened(…) is illegal. I think this could be accomplished using a CSharpCompilation class, but would be keen to hear other options, including "don't bother"

Sounds like a good idea, if it's reasonably easy. If not, I'd say "don't bother".

src/FakeItEasy/Configuration/ExceptionThrower.StronglyTyped.cs
```diff
@@ -1087,5 +1087,197 @@ public IAfterCallConfiguredConfiguration<IPropertySetterConfiguration> Throws<T1
         }
 
         }
+
+        private partial class PropertySetterAfterCallbackConfiguredAdapter
+        {
```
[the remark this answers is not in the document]
blairconrad: Indeed. I was surprised as well, but once I saw it was already missing in the companion class, I moved on…
```c#
        private partial class PropertySetterAdapter
        {
        /// <summary>
        /// Throws the specified exception when the currently configured
```

I can consider for a cleanup commit.

blairconrad (commented):
Looking at the "code that doesn't compile" question, it appears to be not that hard. 

As for `PropertySetterAfterCallbackConfiguredAdapter`, yeah, it might be here to stay...

src/FakeItEasy/Configuration/IVoidAfterCallbackConfiguredConfiguration.cs
```diff
@@ -0,0 +1,16 @@
+namespace FakeItEasy.Configuration
+{
+    /// <summary>
+    /// Provides configuration methods for methods that does not have a return value.
+    /// </summary>
```
thomaslevesque, line 5: Oh, I forgot to mention this in my initial review: shouldn't this interface have a (slightly) different description than `IVoidConfiguration`?

blairconrad: Yes. Thanks! And I'll fix `IPropertySetterAfterCallbackConfiguredConfiguration` as well.

src/FakeItEasy/Configuration/ExceptionThrower.StronglyTyped.cs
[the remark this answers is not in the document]
blairconrad: Cleanupped!

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -30,7 +30,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'net45' OR '$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.3" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
```
blairconrad, line 33: Arguably out of scope, and I can drop (or split out to another PR), but we have upgraded some other dependencies as part of 7.0, and Microsoft.CodeAnalysis.CSharp 3.8.0 depends on it. I thought cleaner just to align, but whatever you like.

src/FakeItEasy/Configuration/IVoidAfterCallbackConfiguredConfiguration.cs
blairconrad: Fixed!

tests/FakeItEasy.Specs/IllegalFluentApiSpecs.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        private static IEnumerable<Microsoft.CodeAnalysis.MetadataReference> references = AppDomain.CurrentDomain.GetAssemblies()
+            .Where(a => !a.IsDynamic)
+            .Where(CompilationRequiresAssembly)
+            .Select(a => MetadataReference.CreateFromFile(a.Location))
+            .ToList();
+
+        [Scenario]
+        [UsingCulture("en-US")]
+        [Example(
```
blairconrad, line 51: It's hard to pick examples. I have included the illegal `Invokes`…`MustHaveHappened`. And some other semi-random stuff. Making an exhaustive list would be exhausting! But if you want more or less of something, I'm happy to change.

tests/FakeItEasy.Specs/IllegalFluentApiSpecs.cs
```diff
@@ -0,0 +1,189 @@
...
+
+            "Then the compilation fails"
+                .x(() => diagnostics.Should().ContainSingle().Which.GetMessage().Should()
+                    .Contain(expectedError));
+        }
+
+        private static bool CompilationRequiresAssembly(Assembly a)
+        {
+            var assemblyName = a.GetName().Name;
+            return
```
blairconrad, line 154: I figure this'll have to change. We normally avoid TFMish preprocessing directives, preferring to focus on features, but
1. naming is hard, but we could entertain something like REQUIRES_MSCORLIB_REFERENCE
2. in this case it feels to me like we're really differentiating by framework

I dunno. Like I say, if you prefer something more featurey, I can do. Or if you want more generic, like `NETFRAMEWORK`, `NETCOREAPP`, ... I could pursue. I picked the specific TFM-ish things so we only add what's absolutely required. This'll require us to examine required references if we support another target.

tests/FakeItEasy.Specs/ThenSpecs.cs
```diff
@@ -17,6 +17,43 @@ public interface IFoo
             int Baz();
         }
 
+        [Scenario]
+        public static void VoidMethodWithInvokesAndDoesNothing(
```
blairconrad, line 21: I'd originally dropped all cases of `Invokes(…).DoesNothing()` in the tests, but realized we could then accidentally remove the possibility from the API, which we could do, if you like, but I'd rather do consciously and remove the last tests then.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
blairconrad: Sorry, had committer's regret right after I shut the computer off for the night. Previous version updates have been for internal tools only. If we want to bump this dependency, it's more appropriate in a separate PR, which we could merge independently. Since it looks like I've beaten the reviewers in, I'm going to drop the commit.

tests/FakeItEasy.Specs/IllegalFluentApiSpecs.cs
blairconrad: Oh, one option that I forgot to mention. It makes the code prettier, but feels gross to me: no `#if`s. The extra references probably aren't going to hurt anything, just make the tests do a little more work.

tests/FakeItEasy.Specs/IllegalFluentApiSpecs.cs
thomaslevesque: Wow. You included a *lot* of stuff here. To be honest, I didn't review every example, I'm just going to trust you on this ;)

tests/FakeItEasy.Specs/IllegalFluentApiSpecs.cs
thomaslevesque: > Oh, one option that I forgot to mention. It makes the code prettier, but feels gross to me: no #ifs.

I like the idea. The code will be much easier to understand, I think.

tests/FakeItEasy.Specs/ThenSpecs.cs
thomaslevesque: Sounds good to me

thomaslevesque (changes requested):
Thanks @blairconrad. Just a few more changes, and it should be good to merge

blairconrad (commented):
Thanks, @thomaslevesque. I think I've resolved everything!

thomaslevesque (dismissed):
Thanks @blairconrad, looks good to me!
Squash what needs to be squashed, and I'll merge it

blairconrad:
Squashed, @thomaslevesque. Thanks for the review and merge!

thomaslevesque (approved):

Issue #1819, opened 2021-03-05
Introduce `An` for argument constraints


thomaslevesque:
Because `A<Apple>` doesn't sound right, it should be `An<Apple>`.

This change brings no new functionality, it just allows the API to better imitate English grammar.

- [x] Implementation #1820 

Note: if we do this, FakeItEasy.Analyzers should be updated to take this into account, as covered by FakeItEasy/FakeItEasy.Analyzers#25.

Documentation to update:
- [x] [Argument constraints](https://fakeiteasy.readthedocs.io/en/stable/argument-constraints/)

blairconrad:
Sounds like a reasonable idea, and I love the checklist!
OMSaG! You already have PR.

blairconrad:
@thomaslevesque, I closed this issue so we can release a build. The Analyzers tests require compilable code.
I suppose an alternative would be to add the AppVeyor nuget feed and depend on a nightly build. Regardless, I think closing this is the right move, since the feature is complete.

afakebot:
This change has been released as part of [FakeItEasy 7.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/7.0.0).

blairconrad:
This change has also been released as part of [FakeItEasy 7.0.0](https://github.com/FakeItEasy/FakeItEasy/releases/tag/7.0.0).

PR #1820, opened 2021-03-05
An<T> argument constraints

thomaslevesque:
Contributes to #1819

(not closing the issue, because we also need to update FakeItEasy.Analyzers)

### DIFF
--- a/docs/argument-constraints.md
+++ b/docs/argument-constraints.md
@@ -53,7 +53,7 @@ An underscore (`_`) can be used as a shorthand for `Ignored` as well:
 A.CallTo(() => foo.Bar("hello", A<int>._)).MustHaveHappened();
 ```
 
-# More convenience matchers
+## More convenience matchers
 
 If more complicated constraints are needed, the `That` method can be
 used. There are a few built-in matchers:
@@ -110,6 +110,20 @@ events in the context of an `A.CallTo`.
 
 So, tempting as it might be to save one of the constraints away in a
 handy variable, don't do it.
+
+## Using correct grammar
+
+FakeItEasy's API attempts to imitate the English language, so that call
+configurations and assertions read naturally. In that spirit, it's
+possible to use `An<T>` instead of `A<T>` for types whose name starts
+with a vowel sound. For instance:
+
+```csharp
+An<Apple>.That.Matches(a => a.Color == "Red")
+```
+
+`A<T>` and `An<T>` are exact synonyms and can be used exactly the same
+way.
 
 # `out` parameters
 
--- a/src/FakeItEasy/A.of.T.cs
+++ b/src/FakeItEasy/A.of.T.cs
@@ -8,7 +8,6 @@ namespace FakeItEasy
     /// Provides an API entry point for constraining arguments of fake object calls.
     /// </summary>
     /// <typeparam name="T">The type of argument to validate.</typeparam>
-    [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = nameof(A), Justification = "It is spelled correctly.")]
     public static class A<T>
     {
         /// <summary>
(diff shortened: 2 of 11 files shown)


docs/argument-constraints.md
```diff
@@ -53,7 +53,7 @@ An underscore (`_`) can be used as a shorthand for `Ignored` as well:
 A.CallTo(() => foo.Bar("hello", A<int>._)).MustHaveHappened();
 ```
 
-# More convenience matchers
+## More convenience matchers
```
thomaslevesque, line 56: Unrelated to the rest of the PR, but I think this heading had the wrong level

docs/argument-constraints.md
```diff
@@ -111,6 +111,20 @@ events in the context of an `A.CallTo`.
 So, tempting as it might be to save one of the constraints away in a
 handy variable, don't do it.
 
+## Using correct grammar
+
+FakeItEasy's API attempts to imitate the English language, so that call
+configurations and assertions read naturally. In that spirit, it's
+possible to use `An<T>` instead of `A<T>` for types whose name starts
+with a vowel. For instance:
```
blairconrad, line 119: Arguably "starts with a vowel _sound_."  `An<Apple>` is a great use of this. `An<OuijaBoard>` less so.

docs/argument-constraints.md
```diff
@@ -111,6 +111,20 @@ events in the context of an `A.CallTo`.
...
+configurations and assertions read naturally. In that spirit, it's
+possible to use `An<T>` instead of `A<T>` for types whose name starts
+with a vowel. For instance:
+
+```csharp
+An<Apple>.That.Matches(a => a.Color == "Red")
+```
+
+`An<T>` is an exact synonym of `A<T>` and can be used exactly the same
+way.
```
blairconrad, line 126: Maybe `A<T>` is a perfect synonym of `An<T>`. You're going to give `An<T>` a complex…
:grin:

src/FakeItEasy/An.of.T.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    using System;
+    using System.ComponentModel;
+    using System.Diagnostics.CodeAnalysis;
+
+    /// <summary>
+    /// Provides an API entry point for constraining arguments of fake object calls.
+    /// </summary>
+    /// <typeparam name="T">The type of argument to validate.</typeparam>
+    /// <remarks>This is a synonym of <see cref="A{T}"/>, to enable proper grammar when the type of T starts with a vowel.</remarks>
```
blairconrad, line 11: ditt my comment about "vowel sound"

src/FakeItEasy/An.of.T.cs
```diff
@@ -0,0 +1,44 @@
...
+    [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = nameof(A), Justification = "It is spelled correctly.")]
```
blairconrad, line 12: `nameof(An)`?

blairconrad (commented):
Generally very nice, @thomaslevesque. 

In addition to my line-level comments, is it worthwhile highlighting the existence of `An` in a remarks on the `A` xml docs?
If you'd rather not, I'm okay with it.

src/FakeItEasy/An.of.T.cs
blairconrad: Or maybe the suppression can be removed?

docs/argument-constraints.md
thomaslevesque: Indeed. Should I rephrase?

> `A<T>` and `An<T>` are exact synonyms, and can be used exactly the same way.

src/FakeItEasy/An.of.T.cs
thomaslevesque: Oops. Copy paste mistake.
I *think* the suppression could be removed; at least Visual Studio thinks so.

docs/argument-constraints.md
thomaslevesque: I rephrased. I can roll it back if you don't like it.

thomaslevesque:
> is it worthwhile highlighting the existence of An in a remarks on the A xml docs?

Almost missed that comment. Yes, that's a good idea, I think.

src/FakeItEasy/A.of.T.cs
```diff
@@ -8,7 +8,7 @@ namespace FakeItEasy
     /// Provides an API entry point for constraining arguments of fake object calls.
     /// </summary>
     /// <typeparam name="T">The type of argument to validate.</typeparam>
-    [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly", MessageId = nameof(A), Justification = "It is spelled correctly.")]
+    /// <remarks>If desired, <see cref="An{T}"/> may be used to specify constraints on types whose names begin with a vowel sound.</remarks>
```
thomaslevesque, line 11: I took a page from you in MakeItEasy (but didn't use `<seealso>`, because I don't think it's supposed to contain text)

docs/argument-constraints.md
blairconrad: I like the new way. égalité !

src/FakeItEasy/A.of.T.cs
blairconrad: Maybe yours is better. The docs for these docs are lacking.

blairconrad (approved):

blairconrad (approved):

PR #1826, opened 2021-03-30
Use tools-shared that copies release checklist labels

blairconrad:


blairconrad:
Thanks for the merge!

PR #1832, opened 2021-05-03
Throw FakeCreationException when using Sdk.Create to fake a value type with fake creation options

### LINKED ISSUE #1830: Create.Fake throws ArgumentException instead of FakeCreationException when passed a value type and fake creation options
Simple repro:

```csharp
public struct MyStruct { }

...

// Throws FakeCreationException, as expected
Create.Fake(typeof(MyStruct));

// Throws ArgumentException
// GenericArguments[0], 'UserQuery+MyStruct', on 'FakeItEasy.Creation.IFakeOptions CreateFakeOptions[T](FakeItEasy.Creation.ProxyOptions)' violates the constraint of type 'T'.
Create.Fake(typeof(MyStruct), options => { });
```

blairconrad:
Fixes #1830.

### DIFF
--- a/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
+++ b/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
@@ -115,12 +115,6 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
 
         public static bool CanGenerateProxy(Type typeOfProxy, [NotNullWhen(false)] out string? failReason)
         {
-            if (typeOfProxy.IsValueType)
-            {
-                failReason = DynamicProxyMessages.ProxyIsValueType(typeOfProxy);
-                return false;
-            }
-
             if (typeOfProxy.IsSealed)
             {
                 failReason = DynamicProxyMessages.ProxyIsSealedType(typeOfProxy);
--- a/src/FakeItEasy/ExceptionMessages.cs
+++ b/src/FakeItEasy/ExceptionMessages.cs
@@ -118,5 +118,8 @@ namespace FakeItEasy
 
         public static string NotAnInterface(Type interfaceType) =>
             $"The specified type {interfaceType} is not an interface";
+
+        public static string FailedToFakeValueType(Type fakeType) =>
+            $"Failed to create Fake of type {fakeType} because it's a value type.";
     }
 }
--- a/src/FakeItEasy/Creation/CastleDynamicProxy/DynamicProxyMessages.cs
+++ b/src/FakeItEasy/Creation/CastleDynamicProxy/DynamicProxyMessages.cs
@@ -13,9 +13,6 @@
         public static string ProxyIsSealedType(Type type) =>
             $"The type of proxy {type} is sealed.";
 
-        public static string ProxyIsValueType(Type type) =>
-            $"The type of proxy must be an interface or a class but it was {type}.";
-
         public static string ProxyTypeWithNoDefaultConstructor(Type type) =>
             $"No usable default constructor was found on the type {type}.";
     }
--- a/src/FakeItEasy/Creation/FakeAndDummyManager.cs
+++ b/src/FakeItEasy/Creation/FakeAndDummyManager.cs
@@ -1,6 +1,7 @@
 namespace FakeItEasy.Creation
 {
     using System;
+    using FakeItEasy.Core;
 
     internal class FakeAndDummyManager
     {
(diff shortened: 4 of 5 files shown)


src/FakeItEasy/ExceptionMessages.cs
```diff
@@ -118,5 +118,8 @@ public static string CallToUnconfiguredMethodOfStrictFake(IFakeObjectCall call)
 
         public static string NotAnInterface(Type interfaceType) =>
             $"The specified type {interfaceType} is not an interface";
+
+        public static string FailedToFakeNonReferenceType(Type fakeType) =>
+            $"Failed to create Fake of type {fakeType} because it is neither an interface nor a class.";
```
thomaslevesque, line 123: nor a delegate
Or maybe just "because it's a value type"?

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
```diff
@@ -115,12 +115,6 @@ internal static class CastleDynamicProxyGenerator
 
         public static bool CanGenerateProxy(Type typeOfProxy, [NotNullWhen(false)] out string? failReason)
         {
-            if (typeOfProxy.IsValueType)
-            {
-                failReason = DynamicProxyMessages.ProxyIsValueType(typeOfProxy);
-                return false;
-            }
-
```
thomaslevesque, line 123: Are we certain we can never reach this point if `typeOfProxy` is a value type?

thomaslevesque:
Thanks @blairconrad. Two minor comments, but otherwise it looks good to me.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
blairconrad: I was last night! Let's recheck.

`CastleDynamicProxyGenerator.CanGenerateProxy` is called from

* `CastleDynamicProxyGenerator.GenerateClassProxy` is called from
    * `FakeObjectCreator.DefaultCreationStrategy.GenerateProxy` is called from
        * `FakeObjectCreator.DefaultCreationStrategy.TryCreateFakeWithDummyArgumentsForConstructor` is called from
            * `FakeObjectCreator.DefaultCreationStrategy.CreateFake`, which see one line below!
        *  `FakeObjectCreator.DefaultCreationStrategy.CreateFake` which see one line below!
* `FakeObjectCreator.DefaultCreationStrategy.CreateFake` is called from
    * `FakeObjectCreate.CreateFakeWithoutLoopDetection` is called from
        * `DummyValueResolver.ResolveByCreatingFakeStrategy.TryCreateDummyValue` _but only after `ResolveByActivatingValueTypeStrategy` did not create a Fake_. 🤔 I see now that we guard against the type of the Dummy being `void` there. I'm not even sure why we have that, nor how it might come up. So there might be a hole.
        * `FakeObjectCreator.CreateFake` is called from
            * `FakeAndDummyManager.CreateFake`, which has the new guard against trying to fake value types

As you can see, I am no longer so sure. The `ResolveByActivatingValueTypeStrategy` check (where we see that the Dummy we're trying to make isn't `void`) is something I missed last night. It's possible that this will only come up when we're making a default return value for a void-returning method, but I'm not 100% sure. I will investigate more. An easy workaround would be to add a `VoidDummyFactory` not unlike the `StringDummyFactory` and then we wouldn't need the special case.

(Also, I specifically made this change a separate commit so we could drop it if it proved problematic. I would be okay with that.)

src/FakeItEasy/ExceptionMessages.cs
blairconrad: Delegates are classes, no?
Regardless, I'm happier with "because it's a value type".

blairconrad (commented):
Thanks for the comments, @thomaslevesque. Reworded the error message as you suggested. Still  looking at the potential for dummying a `void` type.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyGenerator.cs
thomaslevesque: > I see now that we guard against the type of the Dummy being void there. I'm not even sure why we have that, nor how it might come up.

Like this:
```csharp
Create.Dummy(typeof(void));
```
But `A.Dummy<void>()` wouldn't work.

> An easy workaround would be to add a VoidDummyFactory

What would it do exactly? We can never create an instance of `void`... we should just fail to generate a dummy (which already happens).

As far as I can tell, we're good to go.

thomaslevesque (approved):
Looks good to me!

blairconrad (commented):
Squashed.

thomaslevesque:
Thanks, @blairconrad!

blairconrad:
Thanks for improvements and merge, @thomaslevesque!

PR #1834, opened 2021-05-06
Remove extraneous text in dummies doc page

thomaslevesque:
I used the Github UI to make this change, and it looks like it created the branch on this repo instead of my fork... sorry about that.

### DIFF
--- a/docs/dummies.md
+++ b/docs/dummies.md
@@ -79,7 +79,6 @@ of approaches in turn, until one succeeds:
 1. If `T` is a tuple type (`Tuple<>` or `ValueTuple<>`), the Dummy will
 be a tuple whose elements are dummies, or default values when dummies
 can't be made.
-  `Activator.CreateInstance`.
 1. If `T` is a value type, the Dummy will be a `T` created via
   `Activator.CreateInstance`.
 1. If `T` is [fakeable](what-can-be-faked.md), the Dummy will be a


thomaslevesque:
We could probably improve this page by not mentioning `Activator.CreateInstance`. Not every developer is familiar with this method, so explaining something by saying it uses `Activator.CreateInstance` might not be meaningful to everyone.

For instance:

> If T is a value type, the Dummy will be a T created via Activator.CreateInstance

could be rewritten like this:

> If T is a value type, the Dummy will be the default value for that type

And 

> If the Dummies can be found, use Activator.CreateInstance to create the Dummy, supplying the Dummies as the argument list.

could be:

> If the Dummies can be found, create an instance by calling the constructor with these dummies as arguments

blairconrad:
> I used the Github UI to make this change, and it looks like it created the branch on this repo instead of my fork... sorry about that.

Ah, yes. That's a thing it does if you have commit rights. Whatcha gonna do?

blairconrad (approved):
This is a good change. I'd happily take it as is. If you want to pile on the other suggested changes, I like that too!

thomaslevesque:
Let's leave this PR as is. I'll create an issue for the other changes

blairconrad:
Thanks for the change. And the merge!

PR #1842, opened 2021-05-28
Event API improvements

### LINKED ISSUE #1841: Improve the API for events
There are currently a few [limitations](https://fakeiteasy.readthedocs.io/en/stable/raising-events/#limitations) regarding events. When the fake forwards calls to the base class or to a wrapped object, it can't manage event subscriptions and raise events. In addition, when the fake is strict, automatic event subscription management doesn't work; this is by design, but there's no easy way to restore the default event behavior.

We should provide new APIs to:
1. make it simpler to manually configure event subscription or unsubscription
2. allow the user to explicitly enable the default event behavior
3. allow the user to create strict fakes that still manage events automatically

## 1. Make it simpler to manually configure event subscription or unsubscription

Add a new overload of `A.CallTo` that accepts an `EventAction`
```csharp
// Configure subscription to a specific event
A.CallTo(fake, EventAction.Add("EventName"))
// Configure unsubscription from a specific event
A.CallTo(fake, EventAction.Remove("EventName"))
// Configure subscription to any event
A.CallTo(fake, EventAction.Add())
// Configure unsubscription from any event
A.CallTo(fake, EventAction.Remove())
```
These calls would be equivalent to using `A.CallTo(fake).WithVoidReturnType().Where(...)` with the appropriate predicate.

The workaround mentioned in the documentation could then be simplified to this:

```csharp
// Declare a delegate to store the event handlers
EventHandler handlers = null;

// Configure event subscription on the fake
A.CallTo(robot, EventAction.Add("FellInLove")).Invokes((EventHandler h) => handlers += h);
A.CallTo(robot, EventAction.Add("FellInLove")).Invokes((EventHandler h) => handlers -= h);

// Raise the event
handlers?.Invoke(robot, EventArgs.Empty);
```

## 2. Allow the user to explicitly enable the default event behavior

The proposed new API is as follows:

```csharp
Manage.AllEvents.Of(fake);
Manage.Event("EventName").Of(fake);
```
This would restore the default behavior for events on a fake where this behavior is disabled (strict fakes, wrapping fakes and fakes that call base class methods).

Since this is a new entry point that is not easily discoverable, strict fakes should throw an exception with a specific message for events, to guide the user towards the `Manage` API.

## 3. Allow the user to create strict fakes that still manage events automatically

Add a new `ManageEvents` flag to the `StrictFakeOptions` enum:

```csharp
var fake = A.Fake<IFoo>(o => o.Strict(StrictFakeOptions.ManageEvents));
```

thomaslevesque:
Fixes #1841 

This PR contains 3 separate (but related) features. We might keep all of them, or none...

07e9d4a4: Update .editorconfig to add C# style rules. Not related to the rest of this PR, but I was getting tired of fighting Visual Studio defaults... I can move it to a separate PR if you prefer.

f4a2950b, 84d77c8d, 297ea48e, 27d2d1cf : API to simplify configuration of event accessors

6c6758cc, df35a992, 6651ce52, b8f82063 : API to enable default event behavior on a fake

a3d91e0c, 13935d3e, d560a346: add StrictFakeOptions.ManageEvents

EDIT: added two more commits:
- 3b4b1e23: add a more detailed exception message when fake accessors are called on a strict fake
- 6bed3ed4: use a single EventCallHandler per fake so that multiple calls to `Manage.Event` don't cause unexpected behavior.

I'll reorder the commits before merge, but for now I'm leaving them as they are to avoid changing the SHA1s of the other commits...

### DIFF
--- a/docs/raising-events.md
+++ b/docs/raising-events.md
@@ -100,7 +100,10 @@ Similarly, strict fakes don't handle any call unless explicitly configured,
 including event subscription or unsubscription, so FakeItEasy also can't raise
 events on strict fakes.
 
-To work around this limitation, you have two options:
+To work around this limitation, you have three options:
+
+- For a strict fake, you can [enable the default event behavior on the fake at
+creation time](strict-fakes.md#events).
 
 - You can explicitly enable the default event behavior on the fake, for a
 specific event or for all events of the fake:
--- a/docs/raising-events.md
+++ b/docs/raising-events.md
@@ -87,6 +87,7 @@ This method may also be used from C# if you don't want to rely on late binding.
 ## Limitations
 
 The approach described above for raising events doesn't work in some situations:
+
 - Wrapping fakes
 - Fakes configured to call base methods
 
@@ -94,6 +95,10 @@ This is because the calls (including event subscription and unsubscription) are
 forwarded to another implementation (wrapped object or base class) that
 FakeItEasy has no control over, so the fake doesn't know about the handlers and
 cannot call them.
+
+Similarly, strict fakes don't handle any call unless explicitly configured,
+including event subscription or unsubscription, so FakeItEasy also can't raise
+events on strict fakes.
 
 To work around this limitation, you will need to explicitly configure the calls
 to the event accessors to handle event subscription yourself, as in the example
@@ -104,8 +109,8 @@ below:
 EventHandler handlers = null;
 
 // Configure event subscription on the fake
-A.CallTo(robot).Where(call => call.Method.Name == "add_FellInLove").Invokes((EventHandler h) => handlers += h);
-A.CallTo(robot).Where(call => call.Method.Name == "remove_FellInLove").Invokes((EventHandler h) => handlers -= h);
+A.CallTo(robot, EventAction.Add("FellInLove")).Invokes((EventHandler h) => handlers += h);
+A.CallTo(robot, EventAction.Remove("FellInLove")).Invokes((EventHandler h) => handlers -= h);
 
 // Raise the event
 handlers?.Invoke(robot, EventArgs.Empty);
--- a/docs/raising-events.md
+++ b/docs/raising-events.md
@@ -100,9 +100,18 @@ Similarly, strict fakes don't handle any call unless explicitly configured,
 including event subscription or unsubscription, so FakeItEasy also can't raise
 events on strict fakes.
 
-To work around this limitation, you will need to explicitly configure the calls
-to the event accessors to handle event subscription yourself, as in the example
-below:
+To work around this limitation, you have two options:
+
+- You can explicitly enable the default event behavior on the fake, for a
+specific event or for all events of the fake:
+
+```csharp
+Manage.Event("FellInLove").Of(robot);
+Manage.AllEvents.Of(robot);
+```
+
+- If you need more control, you can explicitly configure the calls to the event
+accessors to handle event subscription yourself, as in the example below:
 
 ```csharp
 // Declare a delegate to store the event handlers
--- a/docs/specifying-a-call-to-configure.md
+++ b/docs/specifying-a-call-to-configure.md
@@ -110,6 +110,23 @@ A.CallTo(fakeShop).Where(call => call.Method.Name == "set_Address")
                   .Throws(new Exception("we can't move"));
 ```
 
+## Specifying a call to an event accessor
+
+Although calls to event accessors can be specified using the approach described
+in the previous section, FakeItEasy also provides helper methods to make this
+easier:
+
+```csharp
+// Specifies a call to the add accessor of the MyEvent event of the fake
+A.CallTo(fake, EventAction.Add("MyEvent")).Invokes((EventHandler h) => ...);
+// Specifies a call to the remove accessor of the MyEvent event of the fake
+A.CallTo(fake, EventAction.Remove("MyEvent")).Invokes((EventHandler h) => ...);
+// Specifies a call to the add accessor of any event of the fake
+A.CallTo(fake, EventAction.Add()).Invokes(...);
+// Specifies a call to the remove accessor of any event of the fake
+A.CallTo(fake, EventAction.Remove()).Invokes(...);
+```
+
 ## VB.Net
 Special syntax is provided to specify `Func`s and `Sub`s in VB, using their respective keywords:
 
--- a/src/FakeItEasy/A.cs
+++ b/src/FakeItEasy/A.cs
@@ -172,5 +172,16 @@ namespace FakeItEasy
         {
             return ConfigurationManager.CallToSet(propertySpecification);
         }
+
+        /// <summary>
+        /// Configures subscription to or unsubscription from an event of a faked object.
+        /// </summary>
+        /// <param name="fake">The fake to configure.</param>
+        /// <param name="action">An <see cref="EventAction"/> that represents the action to configure.</param>
+        /// <returns>A configuration object.</returns>
+        public static IAnyCallConfigurationWithVoidReturnType CallTo(object fake, EventAction action)
+        {
+            return CallTo(fake).MatchingEventAction(fake, action);
+        }
     }
 }
--- a/src/FakeItEasy/Core/EventCall.cs
+++ b/src/FakeItEasy/Core/EventCall.cs
@@ -38,9 +38,14 @@ namespace FakeItEasy.Core
             return true;
         }
 
-        public bool IsEventRegistration()
+        public bool IsEventSubscription()
         {
             return GetAddMethod(this.Event).Equals(this.CallingMethod);
+        }
+
+        public bool IsEventUnsubscription()
+        {
+            return GetRemoveMethod(this.Event).Equals(this.CallingMethod);
         }
 
         public bool TryTakeEventRaiserArgumentProvider([NotNullWhen(true)] out IEventRaiserArgumentProvider? argumentProvider)
--- a/src/FakeItEasy/Core/EventCallHandler.cs
+++ b/src/FakeItEasy/Core/EventCallHandler.cs
@@ -1,0 +1,111 @@
+﻿namespace FakeItEasy.Core
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+
+    internal class EventCallHandler
+    {
+        private readonly FakeManager fakeManager;
+
+        private Dictionary<object, Delegate>? registeredEventHandlersField;
+
+        public EventCallHandler(FakeManager fakeManager)
+        {
+            this.fakeManager = fakeManager;
+        }
+
+        private Dictionary<object, Delegate> RegisteredEventHandlers
+        {
+            get
+            {
+                if (this.registeredEventHandlersField is null)
+                {
+                    this.registeredEventHandlersField = new Dictionary<object, Delegate>();
+                }
+
+                return this.registeredEventHandlersField;
+            }
+        }
+
+        public void HandleEventCall(EventCall eventCall)
+        {
+            if (eventCall.IsEventSubscription())
+            {
+                if (eventCall.TryTakeEventRaiserArgumentProvider(out var argumentProvider))
+                {
+                    this.RaiseEvent(eventCall, argumentProvider);
+                }
+                else
+                {
+                    this.AddEventListener(eventCall);
+                }
+            }
+            else
+            {
+                this.RemoveEventListener(eventCall);
+            }
+        }
+
+        private void RemoveEventListener(EventCall call)
+        {
+            this.RemoveHandler(call.Event, call.EventHandler);
+        }
+
+        private void AddEventListener(EventCall call)
+        {
+            this.AddHandler(call.Event, call.EventHandler);
+        }
+
+        private void AddHandler(object key, Delegate handler)
+        {
+            if (this.RegisteredEventHandlers.TryGetValue(key, out var result))
+            {
+                result = Delegate.Combine(result, handler);
+            }
+            else
+            {
+                result = handler;
+            }
+
+            this.RegisteredEventHandlers[key] = result;
+        }
+
+        private void RemoveHandler(object key, Delegate handler)
+        {
+            if (this.RegisteredEventHandlers.TryGetValue(key, out var registration))
+            {
+                registration = Delegate.Remove(registration, handler);
+
+                if (registration is not null)
+                {
+                    this.RegisteredEventHandlers[key] = registration;
+                }
+                else
+                {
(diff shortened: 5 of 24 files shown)


docs/raising-events.md
```diff
@@ -87,6 +87,7 @@ This method may also be used from C# if you don't want to rely on late binding.
 ## Limitations
 
 The approach described above for raising events doesn't work in some situations:
+
```
thomaslevesque, line 90: The missing blank line caused the subsequent bullet points to appear on the same line as the previous paragraph.

src/FakeItEasy/A.cs
```diff
@@ -156,5 +156,17 @@ public static IPropertySetterAnyValueConfiguration<TValue> CallToSet<TValue>(Exp
...
+        /// <summary>
+        /// Configures subscription to or unsubscription from an event of a faked object.
+        /// </summary>
+        /// <param name="fake">The fake to configure.</param>
+        /// <param name="descriptor">An <see cref="EventAction"/> that represents the action to configure.</param>
+        /// <returns>A configuration object.</returns>
+        public static IAnyCallConfigurationWithVoidReturnType CallTo(object fake, EventAction descriptor)
+        {
+            Guard.AgainstNull(descriptor, nameof(descriptor));
+            return CallTo(fake).WithVoidReturnType().Where(descriptor.Matches, writer => descriptor.WriteDescription(fake, writer));
```
thomaslevesque, line 169: The methods in `A` usually defer to `ConfigurationManager`, but since this can be expressed as a call to `A.CallTo(fake)`, I thought it wasn't worth the trouble...

src/FakeItEasy/Core/EventCall.cs
```diff
@@ -38,11 +38,16 @@ public static bool TryGetEventCall(IFakeObjectCall fakeObjectCall, [NotNullWhen(
             return true;
         }
 
-        public bool IsEventRegistration()
+        public bool IsEventSubscription()
```
thomaslevesque, line 41: "Subscription" is a more common term than "Registration" for this. Also, I needed to add a method for unsubscription, and "Unregistration" sounded strange.

src/FakeItEasy/Core/EventCallHandler.cs
```diff
@@ -0,0 +1,111 @@
+﻿namespace FakeItEasy.Core
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+
+    internal class EventCallHandler
```
thomaslevesque, line 7: This contains most of the logic that was in `EventRule`, so that it can be reused in `Manage.Event...`

src/FakeItEasy/EventAction.cs
```diff
@@ -0,0 +1,91 @@
+namespace FakeItEasy
+{
+    using FakeItEasy.Core;
+
+    /// <summary>
+    /// Represents subscription to or unsubscription from an event of a fake.
+    /// </summary>
+    /// <remarks>An instance of this class can't be explicitly created. Use the static <c>Add</c> and <c>Remove</c> methods to obtain an instance.</remarks>
+    public abstract class EventAction
```
thomaslevesque, line 9: I realized belatedly that `EventAccessor` would probably be a better name for this. If you agree, I'll rename.

src/FakeItEasy/StrictFakeOptions.cs
```diff
@@ -32,6 +32,11 @@ public enum StrictFakeOptions
         /// <summary>
         /// Calls to all methods inherited from <see cref="object"/> are allowed, and behave as if the fake weren't strict.
         /// </summary>
-        AllowObjectMethods = AllowEquals | AllowGetHashCode | AllowToString
+        AllowObjectMethods = AllowEquals | AllowGetHashCode | AllowToString,
+
+        /// <summary>
+        /// Events of the fake are managed automatically (subscription, unsubscription and raising).
+        /// </summary>
+        ManageEvents = 8
```
thomaslevesque, line 40: I'm not very happy with this name. Maybe it should be `AllowEvents`, like the other options, but I wanted to convey the fact that it had a special behavior. But now that I think of it, that special behavior has nothing to do with the fake's strictness...
Let me know if you prefer a different name.

src/FakeItEasy/A.cs
blairconrad: I'm all for using our building blocks to help us.

src/FakeItEasy/EventAction.cs
blairconrad: I hadn't minded `EventAction`. I'm not sure `EventAccessor` feels better to me, but no worse, so if you prefer, I'm fine.

While reading, I was mulling over alternatives. I wonder:

```c#
A.CallTo(fake, EventSubscription.Remove())
```
? I'm not entirely sold, but present it in the interest of confusion.

blairconrad (commented):
Reviewing up to 297ea48e65bda55f6f84ecf9f1ecb9f9847f9c1c only so far. Thought I'd let you see my comments before I continued. 
So far (as usual) good stuff!

thomaslevesque:
I made the necessary changes to `Fake<T>` (see fixup commits)

docs/specifying-a-call-to-configure.md
```diff
@@ -110,6 +110,23 @@ A.CallTo(fakeShop).Where(call => call.Method.Name == "set_Address")
                   .Throws(new Exception("we can't move"));
 ```
 
+## Specifying a call to an event accessor
+
+Although calls to event accessors can be specified using the approach described
+in the previous section, FakeItEasy also provides helper methods to make this
+easier:
```
blairconrad, line 117: nit: the first two examples below _are_ the approach described above

tests/FakeItEasy.Specs/ManageEventsSpecs.cs
```diff
@@ -0,0 +1,140 @@
...
+                .x(() => fake.Bar += Raise.WithEmpty());
+
+            "And the Baz event is raised using Raise"
+                .x(() => fake.Baz += Raise.WithEmpty());
+
+            "Then the handler for Bar is called"
+                .x(() => A.CallTo(barHandler).MustHaveHappened());
+
+            "And the handler for Baz is called"
+                .x(() => A.CallTo(bazHandler).MustHaveHappened());
```
blairconrad, line 112: I believe that each raised event went to the right handler, but we could augment the test to ensure it's so.

docs/raising-events.md
```diff
@@ -95,17 +96,33 @@ forwarded to another implementation (wrapped object or base class) that
...
+To work around this limitation, you have three options:
+
+- For a strict fake, you can enable the default event behavior on the fake at
+creation time, as explained [here](strict-fakes.md#events).
+
+- You can explicitly enable the default event behavior on the fake, for a
+specific event or for all events of the fake:
+
+```
```
blairconrad, line 111:
```suggestion
```csharp
```

Otherwise we see 
image
instead of 
image


(and the highlighting is nicer in vscode :wink:)

blairconrad (commented):
Some additional minor comments. I'm up to b8f8206318879ec503a61a6d9ebb7294e32a5186 so far. Sorry for the slow progress.

docs/specifying-a-call-to-configure.md
thomaslevesque: What do you mean? The examples below use the new `EventAction` feature

thomaslevesque (commented):
I just realized I hadn't submitted my comments on your first review pass...
I'll look at the rest now

src/FakeItEasy/Manage.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// <summary>
+    /// Provides methods for managing fake events automatically.
+    /// </summary>
+    public static class Manage
+    {
+        /// <summary>
+        /// Specifies all events of the fake.
+        /// </summary>
+        /// <returns>An fluent configuration object to specify the fake.</returns>
```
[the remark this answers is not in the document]
thomaslevesque: Oops, good catch

tests/FakeItEasy.Specs/ManageEventsSpecs.cs
thomaslevesque: Ah, yes. I'll look into it. Passing specific event args should be enough to ensure this.

thomaslevesque:
I think I addressed the feedback from your last review pass, except the one about the documentation where I asked a clarification.

docs/raising-events.md
```diff
@@ -95,17 +96,33 @@ forwarded to another implementation (wrapped object or base class) that
...
-to the event accessors to handle event subscription yourself, as in the example
-below:
+Similarly, strict fakes don't handle any call unless explicitly configured,
+including event subscription or unsubscription, so FakeItEasy also can't raise
+events on strict fakes.
+
+To work around this limitation, you have three options:
+
+- For a strict fake, you can enable the default event behavior on the fake at
+creation time, as explained [here](strict-fakes.md#events).
```
blairconrad, line 106: nit: I'm not sure we SEO matters, but I think it's better to have the anchor text be something specific.
```suggestion
For a strict fake, you can [enable the default event behavior on the fake at
creation time](strict-fakes.md#events).
```

blairconrad: Aaaaand now that I'm reading this, I wonder if the "wrappers" or calls base methods" folks will want a flag!
We don't have to do now. (Or at all.)

src/FakeItEasy/ExceptionMessages.cs
```diff
@@ -63,6 +63,12 @@ internal static class ExceptionMessages
         public static string CannotRaiseEventWhenCallingBaseMethod =>
             "The fake cannot raise the event because event subscription calls the base implementation.";
 
+        public static string HandleEventsOnStrictFakes =>
+            @"Strict fakes don't handle calls to event accessors unless configured to do so. You can either:
+- configure the calls explicitly;
```
blairconrad, line 68: Does "You can either sound like we're recommending they do one of those? Should it be softened to "If you want the fake to handle the call, you can either…"? If you think not, I understand.

src/FakeItEasy/ExceptionMessages.cs
```diff
@@ -63,6 +63,12 @@ internal static class ExceptionMessages
         public static string CannotRaiseEventWhenCallingBaseMethod =>
             "The fake cannot raise the event because event subscription calls the base implementation.";
 
+        public static string HandleEventsOnStrictFakes =>
+            @"Strict fakes don't handle calls to event accessors unless configured to do so. You can either:
+- configure the calls explicitly;
+- create the fake with the StrictFakeOptions.ManageEvents flag;
+- or use Manage.Event(eventName).Of(fake) or Manage.AllEvents.Of(fake) to enable automatic event management.";
```
blairconrad, line 70: Super nit (from the future): should we use `nameof`s or the like to build the message so that we'll be forced to rewrite this if the calls move or change? I am perhaps worrying too much.

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -49,6 +49,7 @@ internal FakeManager(Type fakeObjectType, object proxy, string? fakeObjectName)
             this.FakeObjectName = fakeObjectName;
 
             this.eventRule = new EventRule(this);
+            this.EventCallHandler = new EventCallHandler(this);
```
blairconrad, line 52: It's a small thing, but we could avoid creating this until/unless we need it. Perhaps many Fakes won't have to handle events. (I know, this is an annoying request, but it's the kind of thing we shaved off during the performance improvements efforts. Or we could just not care that much about performance.)

tests/FakeItEasy.Specs/ManageEventsSpecs.cs
```diff
@@ -111,5 +111,30 @@ public void ManageAllEvents(IFoo fake, EventHandler barHandler, EventHandler baz
...
+                .x(() => fake.Bar += handler);
+
+            "And the fake is configured again to manage the Bar event"
+                .x(() => Manage.Event(nameof(IFoo.Bar)).Of(fake));
+
+            "And the Bar event is raised using Raise"
+                .x(() => fake.Bar += Raise.WithEmpty());
+
+            "Then the handler is called"
+                .x(() => A.CallTo(handler).MustHaveHappened());
```
blairconrad: Exactly once? Or is there no value in checking that we don't raise twice after managing twice?

blairconrad (approved):
Outstanding. I've made a few more churlish comments, but nothing that would prevent me from merging, if you'd rather keep things as they are.

thomaslevesque (commented):
I forgot to submit my comments, again... *sigh*

docs/specifying-a-call-to-configure.md
blairconrad: No, I am clearly having some sort of brain deficiency. I read the "set_" and "get_" above and I guess translated to "add_" and "remove_". After _carefully_ rereading, I see what you have here is perfect. 😊

src/FakeItEasy/StrictFakeOptions.cs
thomaslevesque: @blairconrad I think you missed my last comment, since the conversation was resolved. I unresolved to make it more visible.

blairconrad: I typed out a whole comment! It's lost. Sorry. 

I don't mind `ManageEvents`, since we don't need consistency in order to locate the option. However, I think `AllowEvents` is as good. There's (in my mind) no special behaviour here. Just like `AllowEquals` or the others, it's enabling the behaviour we'd have as if the fake wasn't strict.
I guess if I had to pick, I'd probably keep "Allow", but don't mind if you keep Manage.

thomaslevesque: OK, I'll rename it then.

src/FakeItEasy/Core/FakeManager.cs
thomaslevesque: While making this change, I realized that `EventCallHandler` isn't thread-safe (`EventRule` wasn't either). We just use a `Dictionary<K,V>` without any kind of locking. We should probably address this (in a later PR, since this one is big enough already).

thomaslevesque:
OK, I think I made all the requested changes. I'll rebase/squash after your review.
(given the number of fixup commits, I really hope I didn't mess up, or the rebase is going to be a nightmare...)

blairconrad (approved):
Looks good. Thanks for indulging me.

thomaslevesque:
Rebased!

blairconrad:
Thanks for the hard work, @thomaslevesque.

thomaslevesque:
Thanks for the merge

PR #1843, opened 2021-05-28
Re-enable and fix integration test in net461

thomaslevesque:
Completely unrelated to other ongoing work. I noticed that this test was skipped for all frameworks, but it was supposed to run for net461. This was due to an incorrect TFM in a MSBuild condition.

In addition, just re-enabling the test made it fail, because the exception message no longer contains the path to the assembly, it just contains its full assembly name.


blairconrad (approved):
Thanks, @thomaslevesque!

thomaslevesque:
Thanks

PR #1844, opened 2021-05-28
Add CollectionOfFake overloads that accept an action that uses fake number to configure fake

### LINKED ISSUE #1838: Add CollectionOfFake overload that accepts an action that uses fake number to configure fake
Split from #1837, as proposed by @PiotrKlecha, whose compelling example was

> A new overload for A.CollectionOfFake could also simplify assigning a unique name to each Fake:
>  ```c#
>  var fakes = A.CollectionOfFake<IInterface>(3, (o, i) => o.Named($"Fake{i}"));
> ```

PiotrKlecha:
Fixes #1838

### DIFF
--- a/src/FakeItEasy/A.cs
+++ b/src/FakeItEasy/A.cs
@@ -58,7 +58,7 @@ namespace FakeItEasy
         [SuppressMessage("Microsoft.Design", "CA1004:GenericMethodsShouldProvideTypeParameter", Justification = "Used to specify the type of fake.")]
         public static IList<T> CollectionOfFake<T>(int numberOfFakes) where T : class
         {
-            return Enumerable.Range(0, numberOfFakes).Select(_ => Fake<T>()).ToList();
+            return CollectionOfFake<T>(numberOfFakes, (o, i) => { });
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


blairconrad (commented):
Thanks, @PiotrKlecha. Liked pretty much everything. The one request I was on the verge of making was to have the docs describe what the `int` was in the `Func` in the new overloads. Then I decided to push a commit myself. I hope this doesn't offend. The reasons I went this route were twofold

1. I noticed some other "deficiencies" in the docs around the optionBuilders, which you'd conscientiously replicated, and
2. for some reason, the GitHub/Appveyor system didn't build your PR, and I figured adding a new commit would encourage them.

If you're unhappy with my commit, please drop it. And @thomaslevesque, as usual, if _you're_ unhappy, tell me!

src/FakeItEasy/A.cs
```diff
@@ -71,7 +71,20 @@ public static class A
...
+        /// Creates a collection of fakes of the specified type.
+        /// </summary>
+        /// <typeparam name="T">The type of fakes to create.</typeparam>
+        /// <param name="numberOfFakes">The number of fakes in the collection.</param>
+        /// <param name="optionsBuilder">A lambda where options for the built fake object can be specified.</param>
+        /// <returns>A collection of fake objects of the specified type.</returns>
+        public static IList<T> CollectionOfFake<T>(int numberOfFakes, Action<IFakeOptions<T>, int> optionsBuilder)
+            where T : class
+        {
+            return Enumerable.Range(1, numberOfFakes).Select(i => Fake<T>(options => optionsBuilder(options, i))).ToList();
```
thomaslevesque, line 90: Why start at index 1? That's pretty unusual...

thomaslevesque (changes requested):
Thanks @PiotrKlecha, this looks very good to me.
My only issue is the fact that the index start at 1. I think it would be less surprising to the user if it started at 0, like all indices. @blairconrad what do you think?

blairconrad:
I didn't mind, but now you mention it, having 1 passed in and then needing
to access as 0 might be confusing. Probably best to change to 0. In this
case my commit'll need to be updated as well.

PiotrKlecha:
@thomaslevesque Thank you for your review, I started the index at 1, because that was how it started in the example: https://github.com/FakeItEasy/FakeItEasy/issues/1838#issuecomment-846171643, so I thought it was the preferred option. But it's probably more intuitive to start at 0, so I'll change that.

@blairconrad Thank you for your review and update to docs, I don't mind at all.

blairconrad:
> @thomaslevesque Thank you for your review, I started the index at 1, because that was how it started in the example: [#1838 (comment)](https://github.com/FakeItEasy/FakeItEasy/issues/1838#issuecomment-846171643), so I thought it was the preferred option. But it's probably more intuitive to start at 0, so I'll change that.

I lead you astray with my bad example. Sorry about that!

PiotrKlecha:
> I lead you astray with my bad example. Sorry about that!

@blairconrad No problem, fortunately, it was easy to fix.

thomaslevesque (approved):
Looking great, thanks @PiotrKlecha!

@blairconrad I'm not merging yet, in case you want to review the latest changes.

blairconrad (approved):
Looks great. Thanks, @PiotrKlecha!

Issue #1848, opened 2021-06-24
Possible performance hotspot in FakeManager.ApplyBestRule


TimLovellSmith:
Backstory - I found my tests were not running well today. A few tests ran, and then everything stopped, while my CPU fan started to whine, so I fired up the debugger, to see what was what, and found most threads doing nothing, but this thread looked like it was doing *something*. Not once, but several times...

So from that symptom seems like it could be a bit of a performance hotspot in this scenario where I was intercepting the same call (out of various possible calls) over and over again.

One minor thing in the code today I can suggest changing: this test should probably be reversed

`if (rule.Rule.IsApplicableTo(fakeObjectCall) && rule.HasNotBeenCalledSpecifiedNumberOfTimes())`
--->
`
if (rule.HasNotBeenCalledSpecifiedNumberOfTimes() && rule.Rule.IsApplicableTo(fakeObjectCall))
`

Since doing tests on integers is fast.

Another idea of what to change would be to optimize `ObjectMemberRule.IsApplicableTo`. In the scenario where its called repeatedly with the same fakeObjectCallMethod, perhaps the result of 'ObjectMemberRule.IsApplicableTo' could be cached on the `fakeObjectCall.Method` somehow, also maybe some cheaper heuristics should be checked before calling `HasSameBaseMethodAs` -- does `fakeObjectCall.Method.Name` start with "E"? It could be "Equals". "T"? It could be "ToString". "G"? It could be "GetHashCode". None of those? None of those.

```
        public override bool IsApplicableTo(IFakeObjectCall fakeObjectCall) =>
                fakeObjectCall.Method.HasSameBaseMethodAs(EqualsMethod) ||
                fakeObjectCall.Method.HasSameBaseMethodAs(ToStringMethod) ||
                fakeObjectCall.Method.HasSameBaseMethodAs(GetHashCodeMethod);

        public static bool HasSameBaseMethodAs(this MethodInfo first, MethodInfo second)
        {
            var baseOfFirst = GetBaseDefinition(first);
            var baseOfSecond = GetBaseDefinition(second);

            return baseOfFirst.IsSameMethodAs(baseOfSecond);
        }
```

Original repro with:
```
  <package id="Castle.Core" version="4.3.1" targetFramework="net472" />
  <package id="FakeItEasy" version="6.2.1" targetFramework="net472" />
```

Stack:
```
[...]

PR #1852, opened 2021-06-28
Optimize method comparison

blairconrad:
Connects to #1848.

I hesitate to say fix, but it addresses at least one area of slowness identified there.

Has this affect on runtimes (6.1.0 is the build before we changed a lot of method comparisons to see if a call had the same base method as a given method).

 Method |             Job |      Mean |     Error |    StdDev | Ratio | RatioSD |
------- |---------------- |----------:|----------:|----------:|------:|--------:|
 Equals |           6.1.0 |  4.738 us | 0.3711 us | 0.2208 us |  1.00 |    0.00 |
 Equals |           7.1.0 | 11.110 us | 0.6914 us | 0.4115 us |  2.35 |    0.14 |
 Equals | 7.1.1-alpha.0.5 |  4.340 us | 0.3255 us | 0.2153 us |  0.91 |    0.08 |

image

and

 Method                 |             Job |     Mean |     Error |    StdDev | Ratio | RatioSD |
----------------------- |---------------- |---------:|----------:|----------:|------:|--------:|
 UnconfiguredBoolMethod |           6.1.0 | 5.421 us | 0.4837 us | 0.3199 us |  1.00 |    0.00 |
 UnconfiguredBoolMethod |           7.1.0 | 7.421 us | 0.3595 us | 0.2139 us |  1.38 |    0.10 |
 UnconfiguredBoolMethod | 7.1.1-alpha.0.5 | 4.998 us | 0.3071 us | 0.2032 us |  0.92 |    0.05 |
 
image

### DIFF
--- a/src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
+++ b/src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
@@ -1,6 +1,6 @@
 namespace FakeItEasy.Core
 {
-    using static FakeItEasy.ObjectMembers;
+    using static FakeItEasy.ObjectMethod;
 
     /// <content>Object member rule.</content>
     public partial class FakeManager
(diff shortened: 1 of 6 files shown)


src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
```diff
@@ -1,6 +1,6 @@
 namespace FakeItEasy.Core
 {
-    using static FakeItEasy.ObjectMembers;
+    using static FakeItEasy.ObjectMethod;
```
thomaslevesque, line 3: I had no idea you could do that on enums!

src/FakeItEasy/MethodInfoExtensions.cs
```diff
@@ -10,6 +11,8 @@ namespace FakeItEasy
     /// </summary>
     internal static class MethodInfoExtensions
     {
+        private static readonly ConcurrentDictionary<MethodInfo, ObjectMethod> ObjectMethodMap = new ConcurrentDictionary<MethodInfo, ObjectMethod>();
```
thomaslevesque, line 14: I think this is a good use case for [target-typed new expressions](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/target-typed-new)

thomaslevesque (approved):
Thanks, @blairconrad. The code looks good to me (just one suggestion, which you may ignore)
We don't currently have benchmarks in the code base. Maybe we should?
Anyway, it would be interesting to see the code for your benchmark. Maybe post it as a link to a gist?

blairconrad (commented):
I thought target-typed new expressions were still in our future. Thanks.

blairconrad:
Benchmarks were gathered by running 
```
.\benchmark.ps1 --filter=FakeItEasy.Benchmark.Program.SlowObjectBenchmark.UnconfiguredBoolMethod 6.1.0 7.1.0 7.1.1-alpha.0.5
.\benchmark.ps1 --filter=FakeItEasy.Benchmark.Program.SlowObjectBenchmark.Equals 6.1.0 7.1.0 7.1.1-alpha.0.5
```
(mostly because I didn't want to walk away from the computer to run all of them at once) on 
https://github.com/blairconrad/FakeItEasy.Benchmark/tree/f52025ce8cf341a7f25950c84eadcbb15adb302b

The benchmarks project is very rough, more or less the least I could do to get bars to come out. Were we to formalize the project, it may make sense to replace the plotting library with something .NETish. Maybe https://github.com/ScottPlot/ScottPlot, which I've not tried, but it looks like a good fit.

src/FakeItEasy/Core/FakeManager.ObjectMemberRule.cs
thomaslevesque: Yes, but it was a class, not an enum

blairconrad: I am not the shiniest penny in the fountain.

blairconrad:
Thanks for the review and merge, @thomaslevesque

blairconrad:
Adding to the vNext milestone, but not yet release notes. This will interrupt us when we want to release, but I'd like to see whether we do anything more for #1848 first...

Issue #1854, opened 2021-08-30
Initializing read only property of faked object does not work


Bludator:
I have been playing with fakes with `WithArgumentsForConstructor` and I found it very confusing.

First of all [the documentation](https://fakeiteasy.readthedocs.io/en/stable/creating-fakes/#explicit-creation-options):

> // Specifying arguments for constructor using expression. This is refactoring friendly!
> // The constructor seen here is never actually invoked. It is an expression and it's purpose
> // is purely to communicate the constructor arguments which will be extracted from it
> var foo = A.Fake<FooClass>(x => x.WithArgumentsForConstructor(() => new FooClass("foo", "bar")));

If one doesn't read carefully one could think that the constructor is never called and that is not true. There is even [SO question.](https://stackoverflow.com/questions/36972434/withargumentsforconstructor-extension-method-calls-constructor)
Actually the docs never say that the constructor runs (or not).

Secondly, when the constructor is called it is not possible to initialize faked (virtual/abstract) property.

Minimal (non)working example:

```C#
using FakeItEasy;
using Xunit;

public class Foo
{
	public virtual string VirtualProperty { get; }
	public virtual string VirtualPropertyWithSetter { get; set; }
	public string NotVirtualProperty { get; }
	public string NotVirtualPropertyWithSetter { get; set; }

	public Foo()
	{
		VirtualProperty = "my string"; // does nothing
		NotVirtualProperty = "my string";
		VirtualPropertyWithSetter = "my string";
		NotVirtualPropertyWithSetter = "my string";
	}
}

public class UnitTest
{
	[Fact]
	public void Test1() => Assert.Equal("my string", A.Fake<Foo>().VirtualProperty); //not passing

	[Fact]
	public void Test2() => Assert.Equal("my string", A.Fake<Foo>().VirtualPropertyWithSetter);

	[Fact]
	public void Test3() => Assert.Equal("my string", A.Fake<Foo>().NotVirtualProperty);

	[Fact]
	public void Test4() => Assert.Equal("my string", A.Fake<Foo>().NotVirtualPropertyWithSetter);
}
```

### Versions:
- FakeItEasy: 7.1.0
- .NET: 5.0

thomaslevesque:
Hi @Bludator,

Thanks for the report.

[...]

PR #1855, opened 2021-08-30
Clarify use of WithArgumentsForConstructor expression

blairconrad:
Inspired by #1854, where @Bludator had noted that the original documentation was confusing.

### DIFF
--- a/docs/creating-fakes.md
+++ b/docs/creating-fakes.md
@@ -43,9 +43,11 @@ When creating fakes you can, through a fluent interface, specify options for how
 Examples:
 
 ```csharp
-// Specifying arguments for constructor using expression. This is refactoring friendly!
-// The constructor seen here is never actually invoked. It is an expression and it's purpose
-// is purely to communicate the constructor arguments which will be extracted from it
+// Specifying arguments for constructor using an expression. This is refactoring friendly!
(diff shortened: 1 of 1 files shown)


docs/creating-fakes.md
```diff
@@ -44,8 +44,10 @@ Examples:
 
 ```csharp
 // Specifying arguments for constructor using expression. This is refactoring friendly!
-// The constructor seen here is never actually invoked. It is an expression and it's purpose
-// is purely to communicate the constructor arguments which will be extracted from it
+// Since the constructor call seen here is an expression, it is not invoked at this point
+// in the code. (In fact no instance of `FooClass` will ever be constructed.) Instead, the
```
thomaslevesque, line 48: > In fact no instance of `FooClass` will ever be constructed

I think this could be confusing too. The fake will be an instance of the proxy class, but it's still "an instance of `FooClass`", in the sense that `fake is FooClass` is true.

blairconrad: grumble grumble

working on it

blairconrad: amended

thomaslevesque (approved):
Looks good to me! Thanks @blairconrad

blairconrad:
Oh, thank you, @thomaslevesque.

PR #1864, opened 2022-01-03
Include defined Fake name in exception messages.

### LINKED ISSUE #1837: Indicate Fake name (if defined) in failing assertions
When working with a collection of Fakes it's sometimes difficult to determine which one of them violated expectations. 

I came across two cases when this becomes a problem:

1. Verifing expectations in a loop:

```cs
void TestMethod(IList<IInterface> dependencies)
{
	dependencies[1].Method1();
}

[Test]
public void TestCollection()
{
	var fakes = A.CollectionOfFake<IInterface>(3);

	TestMethod(fakes);

	foreach (var fake in fakes)
	{
		A.CallTo(() => fake.Method1())
			.MustNotHaveHappened();
	}
}
```
This causes:
```
FakeItEasy.ExpectationException : 

    Tests.IInterface.Method1()
  Expected to find it never but found it once among the calls:
    1: Tests.IInterface.Method1()
```
2. Using strict Fakes:

```cs
void TestMethod(IList<IInterface> dependencies)
{
	foreach (var dependency in dependencies)
		dependency.Method1();
}

[Test]
public void Strict()
{
	var fakes = A.CollectionOfFake<IInterface>(3, o => o.Strict());

	TestMethod(fakes);
}
```
which results in an error:
```
FakeItEasy.ExpectationException : Call to unconfigured method of strict fake: Tests.IInterface.Method1().
```

Named Fakes seem to be an intuitive and convenient way of solving this problem, but unfortunately assigned names are only displayed if a named Fake is used as an argument of a call, so at the moment they cannot help in situations described above.

It would be very helpful if Fake names were also included in error messages, e.g. with Fakes from previous examples named Dependency1 and Dependency0:
```
FakeItEasy.ExpectationException : 

  Assertion failed for the following call on Dependency1:
    Tests.IInterface.Method1()
  Expected to find it never but found it once among the calls:
    1: Tests.IInterface.Method1()
```
and
```
FakeItEasy.ExpectationException : Call to unconfigured method of strict fake Dependency0: Tests.IInterface.Method1().
```

A new overload for A.CollectionOfFake could also simplify assigning a unique name to each Fake:
```cs
var fakes = A.CollectionOfFake<IInterface>(3, (o, i) => o.Named($"Fake{i}"));
```

PiotrKlecha:
Fixes #1837.

### DIFF
--- a/src/FakeItEasy/Configuration/AnyCallCallRule.cs
+++ b/src/FakeItEasy/Configuration/AnyCallCallRule.cs
@@ -6,8 +6,14 @@ namespace FakeItEasy.Configuration
     internal class AnyCallCallRule
         : BuildableCallRule
     {
+        private readonly FakeManager manager;
         private Func<ArgumentCollection, bool> argumentsPredicate = x => true;
         private Type applicableToMembersWithReturnType = typeof(AllReturnTypes);
+
+        public AnyCallCallRule(FakeManager manager)
+        {
+            this.manager = manager;
+        }
 
         public void MakeApplicableToMembersWithReturnType(Type type) => this.applicableToMembersWithReturnType = type;
 
@@ -15,22 +21,8 @@ namespace FakeItEasy.Configuration
 
         public override void DescribeCallOn(IOutputWriter writer)
         {
-            if (this.applicableToMembersWithReturnType == typeof(AllNonVoidReturnTypes))
-            {
-                writer.Write("Any call with non-void return type to the fake object.");
-            }
-            else if (this.applicableToMembersWithReturnType == typeof(void))
-            {
-                writer.Write("Any call with void return type to the fake object.");
-            }
-            else if (this.applicableToMembersWithReturnType == typeof(AllReturnTypes))
-            {
-                writer.Write("Any call made to the fake object.");
-            }
-            else
-            {
-                writer.Write("Any call with return type ").Write(this.applicableToMembersWithReturnType).Write(" to the fake object.");
-            }
+            this.AppendCall(writer);
+            this.AppendTargetObject(writer);
         }
 
         public override void UsePredicateToValidateArguments(Func<ArgumentCollection, bool> predicate)
@@ -66,10 +58,51 @@ namespace FakeItEasy.Configuration
         }
 
         protected override BuildableCallRule CloneCallSpecificationCore() =>
-            new AnyCallCallRule
+            new AnyCallCallRule(this.manager)
             {
                 argumentsPredicate = this.argumentsPredicate
             };
+
+        private void AppendCall(IOutputWriter writer)
+        {
+            writer.Write($"Any call {GetCallDescription()}");
+
+            string GetCallDescription()
+            {
+                if (this.applicableToMembersWithReturnType == typeof(AllNonVoidReturnTypes))
+                {
+                    return "with non-void return type";
+                }
+                else if (this.applicableToMembersWithReturnType == typeof(void))
+                {
+                    return "with void return type";
+                }
+                else if (this.applicableToMembersWithReturnType == typeof(AllReturnTypes))
+                {
+                    return "made";
+                }
+                else
+                {
+                    return $"with return type {this.applicableToMembersWithReturnType}";
+                }
+            }
+        }
+
+        private void AppendTargetObject(IOutputWriter writer)
+        {
+            writer.Write($" to the fake object{GetTargetObjectName()}.");
+
+            string GetTargetObjectName()
+            {
+                var fakeName = this.manager.FakeObjectName;
+                if (string.IsNullOrEmpty(fakeName))
+                {
+                    return string.Empty;
+                }
+
+                return $" {fakeName}";
+            }
+        }
 
         private class AllNonVoidReturnTypes
         {
--- a/src/FakeItEasy/Expressions/CallConstraintDescriber.cs
+++ b/src/FakeItEasy/Expressions/CallConstraintDescriber.cs
@@ -1,5 +1,6 @@
 namespace FakeItEasy.Expressions
 {
+    using System;
     using System.Collections.Generic;
     using System.Linq;
     using System.Reflection;
@@ -15,15 +16,23 @@ namespace FakeItEasy.Expressions
         /// matcher to the supplied writer.
         /// </summary>
         /// <param name="writer">The writer on which to describe the call.</param>
+        /// <param name="callTarget">The object on which the method was called.</param>
         /// <param name="method">The method to describe.</param>
         /// <param name="argumentConstraints">The argument constraints applied to the method.</param>
-        public static void DescribeCallOn(IOutputWriter writer, MethodInfo method, IEnumerable<IArgumentConstraint> argumentConstraints)
+        public static void DescribeCallOn(IOutputWriter writer, object? callTarget, MethodInfo method, IEnumerable<IArgumentConstraint> argumentConstraints)
         {
-            writer.Write(method.DeclaringType!);
-            writer.Write(".");
+            WriteObjectType(writer, method.DeclaringType!);
             WriteMethodName(writer, method);
 
             WriteArgumentsListString(writer, method, argumentConstraints);
+
+            WriteObjectName(writer, callTarget);
+        }
+
+        private static void WriteObjectType(IOutputWriter writer, Type type)
+        {
+            writer.Write(type.ToString());
+            writer.Write(".");
         }
 
         private static void WriteMethodName(IOutputWriter writer, MethodInfo method)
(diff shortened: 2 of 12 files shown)


thomaslevesque:
Hi @PiotrKlecha,

Thanks for the PR. At first glance, it looks very good! I'll do a full review ASAP.

The build is currently failing on approval tests, probably because `FakeManager.FakeObjectName` has become virtual. Here's [how to fix it](https://github.com/FakeItEasy/FakeItEasy/blob/c2852f58e5495ca4f9fdb2d5048d454071149116/CONTRIBUTING.md#api-approval).

Note that the unit tests in FakeItEasy.Tests are mostly legacy, we rarely write new ones now; we prefer to write high-level specs instead (in FakeItEasy.Specs). So I'm not sure the unit tests you added are strictly necessary (since the change is sufficiently covered by specs), and they're the only reason for making `FakeObjectName` virtual. @blairconrad what do you think?

blairconrad:
Thanks, @PiotrKlecha. Like @thomaslevesque, I've only done a quick scan of the changes (actually, mostly the tests so far), but generally I'm pleased.

As M. Levesque says, the Specs seem to cover the new functionality. I would drop the unit test changes (and changes made only to support the unit tests, such as making `FakeObjectName` virtual).

I've a busy day, and tomorrow is my first day back at work since the 20th of December, so I may be slowed down, but I will do a thorough review as soon as I'm able.

Oh! On more thing, as you likely noted above, the branch is out of date compared to FakeItEasy/master. We'll need to rebase onto master at some point. Me, I don't care whether it's now or later, but thought you might want to mentally prepare. 😉

thomaslevesque (changes requested):
Thanks @PiotrKlecha!
In addition to the removal of unit tests, I have a few minor remarks.

thomaslevesque:
Hi @PiotrKlecha,

The `TryGetCallTarget` method seemed strange to us, since `CallTarget` doesn't usually throw. We realized that you had added this method because of failing unit tests. The problem was actually in the tests themselves, so I fixed the tests and removed the `TryGetCallTarget` method. While I was at it, I also made a small refactoring on `ExpressionCallMatcher` to avoid having 2 levels of "lazyness" for `CallTarget` (one in `ExpressionCallMatcher` and one in `ParsedCallExpression`).

I pushed the changes directly to your branch, so don't forget to pull before you make additional changes!

PiotrKlecha:
Hi @thomaslevesque, @blairconrad,

Thank you for your comments and suggestions, I'll try to do all the requested changes soon.

Regarding the `TryGetCallTarget` method, I wasn't sure if the method signatures used in failing tests could be valid in some other context. I found one method that accepts this kind of Expressions - `CompletedFakeObjectCallExtensions.Matching`, 
so I wanted the changes to be as safe as possible. Description of the call is not used there though, so perhaps it wasn't necessary.

blairconrad:
Thanks, @PiotrKlecha. Your reasoning makes sense, and it is better err on the side of caution.
As an aside, I've submitted another PR (#1865) to completely remove the unit tests that were causing you problems, since they are (to my mind) outdated. There's likely no action on your part, as @thomaslevesque will either reject it or it'll be merged into master before this PR is complete, so the upcoming rebase will incorporate it anyhow.

Sorry I haven't made any other comments on the review yet. I'll try to carve out some time in the next 24 hours.

tests/FakeItEasy.Specs/StrictFakeSpecs.cs
```diff
@@ -313,6 +313,24 @@ public static void EventsManaged(IMyInterface fake, EventHandler handler)
...
+                .x(() => fake = A.Fake<IMyInterface>(o => o.Strict().Named("Foo1")));
+
+            "And I call an unconfigured method"
+                .x(() => exception = Record.Exception(() => fake.MakeIt("argument")));
+
+            "Then it should include configured name in the thrown exception"
+                .x(() => exception
+                    .Should()
+                    .BeAnExceptionOfType<ExpectationException>()
+                    .WithMessage(@"Call to unconfigured method of strict fake: FakeItEasy.Specs.StrictFakeSpecs+IMyInterface.MakeIt(name: ""argument"") on Foo1."));
```
blairconrad, line 332: Perhaps this was discussed before, and I should've been paying more attention, but the format of this message doesn't sit quite well for me. Separating the "strict fake" and "on Foo1" makes it harder (for me) to figure out what I'm being told. I would've expected something more like
"Call to unconfigured method of strict fake Foo1: FakeItEasy.Specs.StrictFakeSpecs+IMyInterface.MakeIt(name: ""argument"")."

In my head, it makes it more clear what's named Foo1.

Am I alone in this?

blairconrad (commented):
Sorry for the late review. @thomaslevesque covered the high points already. I just had jone comment about a possible improvement to the readability of an error message. 
Thanks, @PiotrKlecha.

tests/FakeItEasy.Specs/StrictFakeSpecs.cs
thomaslevesque: @blairconrad Ideally, yes, the message you suggest would be better. In terms of implementation, I'm afraid this is going to be more complex. The fake name is embedded in the call description by `DefaultFakeObjectCallFormatter`, and `StrictFakeRule` just uses that. What you suggest would require a significant refactoring...

src/FakeItEasy/Expressions/CallConstraintDescriber.cs
```diff
@@ -15,15 +16,23 @@ internal static class CallConstraintDescriber
...
         /// <param name="writer">The writer on which to describe the call.</param>
+        /// <param name="callTarget">The object on which the method was called.</param>
         /// <param name="method">The method to describe.</param>
         /// <param name="argumentConstraints">The argument constraints applied to the method.</param>
-        public static void DescribeCallOn(IOutputWriter writer, MethodInfo method, IEnumerable<IArgumentConstraint> argumentConstraints)
+        public static void DescribeCallOn(IOutputWriter writer, object? callTarget, MethodInfo method, IEnumerable<IArgumentConstraint> argumentConstraints)
         {
-            writer.Write(method.DeclaringType!);
-            writer.Write(".");
+            WriteObjectType(writer, method.DeclaringType);
```
blairconrad, line 24: Please restore the "!". It's causing a [warning CS8604 in the build](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/builds/42130140/job/0xgev6ijlyhomatl#L67).

src/FakeItEasy/Configuration/AnyCallCallRule.cs
```diff
@@ -8,29 +8,21 @@ internal class AnyCallCallRule
     {
         private Func<ArgumentCollection, bool> argumentsPredicate = x => true;
         private Type applicableToMembersWithReturnType = typeof(AllReturnTypes);
+        private readonly FakeManager manager;
```
blairconrad, line 11: Since I'm already asking for a change in `CallConstraintDescriber` below, would you mind moving `manager` up to line 9? Having a readonly field after a non-readonly field [causes an warning in the build](https://ci.appveyor.com/project/FakeItEasy/fakeiteasy/builds/42130140/job/0xgev6ijlyhomatl#L68).

blairconrad (changes requested):
Thanks, @PiotrKlecha. I think you've addressed the major concerns we mentioned earlier, but of course I'll let @thomaslevesque chime in.

I'd thought that your latest changes would end up removing the build warnings you'd introduced earlier, but I was mistaken and two remain. I've commented on them inline. Would you mind quelling them? If it's too much bother, I can push a commit.

thomaslevesque (commented):
Thanks for the updates, @PiotrKlecha. As @blairconrad mentioned, there are a few warnings to fix, but apart from that, it looks good to me!

PiotrKlecha:
Hi @blairconrad, @thomaslevesque, I fixed the build warnings, I'll pay more attention to the build output next time. I also merged changes from the latest master, so the branch is now up to date.

thomaslevesque (approved):
Thanks @PiotrKlecha, looks good to me!

blairconrad (approved):
Thanks, @PiotrKlecha. The updated code addressed all my concerns as well.
We appreciate the work you've done, especially the back-and-forth for the revisions you've made. 

I've taken the liberty of squashing your commits and rebasing onto FakeItEasy/master, to align with the "no merges in topic branches" policy [outlined in our CONTRIBUTING.md](https://github.com/FakeItEasy/FakeItEasy/blob/96b943685cafe888297a4834367465e1f0939d8d/CONTRIBUTING.md). The squashing also had the advantage of smoothing out some "course correction"-type commits, where an approach was taken and then adjusted during the PR. 

Please don't take my action as any sort of criticism; I just didn't want to bother you once more only to have you perform this mechanical housecleaning step.

PiotrKlecha:
@blairconrad I'm totally fine with the the squashing and rebasing that you performed, it will definitely keep the project history cleaner. I'm not familiar with GitHub PR flow and I wasn't sure how rebasing the whole branch would affect this PR, since the intermediate commits appeared here. That's why I followed GitHub's suggestion about updating the branch which resulted in a merge, but I guess I should have read the contributing guidelines more carefully - I'll do it the right way next time :)

thomaslevesque:
Thanks again, @PiotrKlecha!

PR #1891, opened 2022-11-25
Replace ApprovalTests with Verify

blairconrad:


tests/FakeItEasy.Tests.Approval/ApiApproval.cs
```diff
@@ -39,23 +38,10 @@ public void ApproveApi(string projectName, string frameworkVersion)
...
-                this.SourcePath = Path.Combine(GetSourceDirectory(), "ApprovedApi", projectName);
-            }
-
-            public string SourcePath { get; }
-
-            public string Name { get; }
+            var settings = new VerifySettings();
+            settings.UseDirectory("ApprovedApi/" + projectName);
+            settings.UseFileName(frameworkVersion);
+            return Verifier.Verify(publicApi, settings);
```
thomaslevesque, line 44: Maybe use `await` instead of returning the task? Otherwise, if there's an exception thrown asynchronously, `ApproveApi` won't even be visible in the stack trace.

blairconrad: I literally have no idea! I just copied stuff! Thanks!

tests/FakeItEasy.Tests.Approval/ApprovedApi/FakeItEasy.Extensions.ValueTask/netstandard2.0.verified.txt
```diff
@@ -1,46 +1,46 @@
...
+        public static FakeItEasy.Configuration.IAfterCallConfiguredWithOutAndRefParametersConfiguration<FakeItEasy.Configuration.IReturnValueConfiguration<System.Threading.Tasks.ValueTask<TReturnType>>> ReturnsLazily<TReturnType, T1, T2, T3, T4, T5, T6, T7>(this FakeItEasy.Configuration.IReturnValueConfiguration<System.Threading.Tasks.ValueTask<TReturnType>> configuration, System.Func<T1, T2, T3, T4, T5, T6, T7, TReturnType> valueProducer) { }
+        public static FakeItEasy.Configuration.IAfterCallConfiguredWithOutAndRefParametersConfiguration<FakeItEasy.Configuration.IReturnValueConfiguration<System.Threading.Tasks.ValueTask<TReturnType>>> ReturnsLazily<TReturnType, T1, T2, T3, T4, T5, T6, T7, T8>(this FakeItEasy.Configuration.IReturnValueConfiguration<System.Threading.Tasks.ValueTask<TReturnType>> configuration, System.Func<T1, T2, T3, T4, T5, T6, T7, T8, TReturnType> valueProducer) { }
+        public static void ReturnsNextFromSequence<T>(this FakeItEasy.Configuration.IReturnValueConfiguration<System.Threading.Tasks.ValueTask<T>> configuration, params T[] values) { }
+    }
```
thomaslevesque, line 45: Looks like the line endings have changed...

tests/FakeItEasy.Tests.Approval/ModuleInitializer.cs
```diff
@@ -0,0 +1,12 @@
+namespace FakeItEasy.Tests.Approval
+{
+    using System.Runtime.CompilerServices;
+    using VerifyTests;
+    using VerifyTests.DiffPlex;
+
+    public static class ModuleInitializer
+    {
+        [ModuleInitializer]
+        public static void Initialize() => VerifyDiffPlex.Initialize(OutputType.Compact);
```
thomaslevesque, line 10: Fancy! I don't think I ever used a module initializer before.
But maybe this could be done in the test class static constructor? It would feel a bit less "magic"...

thomaslevesque (changes requested):
Thanks @blairconrad.
Looks like the line endings have changed in the .verified.txt files...

tests/FakeItEasy.Tests.Approval/ApprovedApi/FakeItEasy.Extensions.ValueTask/netstandard2.0.verified.txt
blairconrad: Slightly annoying, and I'm happy to update the files in 2 steps, so it's more obvious what substantive changes we have, but probably best to keep the new line endings? I did use `build force-approve` to copy these over, and I think long term we're better off not fighting with the tool's format.

blairconrad:
Thanks for the comments, @thomaslevesque. In the interest of confusing you, I just went ahead and rebased after applying all my fixups. I don't think there was much to gain by "showing my work". I think I've addressed everything, assuming you're content with the end-of-line situation.

tests/FakeItEasy.Tests.Approval/ApprovedApi/FakeItEasy.Extensions.ValueTask/netstandard2.0.verified.txt
blairconrad: Oh. I found the docs I couldn't earlier. VerifyTests.Verify indicates that the files will have only LF as end-of-line:

https://github.com/VerifyTests/Verify/blob/51953ba588c6db798548cde8ef7c4cc579802579/readme.md#verified

We don't munge any of the line endings when we commit or checkout, so we don't need to worry about the .gitattributes settings, but I think best to comply with Verify.

thomaslevesque: > happy to update the files in 2 steps, so it's more obvious what substantive changes we have

It's OK, I was able to see the actual changes by ignoring whitespace changes.

> probably best to keep the new line endings?

Sounds good to me.

thomaslevesque (approved):

thomaslevesque:
Oh. It fails on Ubuntu now :(

blairconrad:
> Oh. It fails on Ubuntu now :(

In retrospect, that makes a lot of sense!

Ugh. We gotta think. Options?

1. somehow filter away the offending attribute
2. duplicate the received/verified files, one set per OS
3. not run the API check on Ubuntu (or Windows, I'm not fussy)
4. Something I didn't think of

Me, I'm kind of inclined to go with 3, but it's not a _strong_ inclination.

thomaslevesque:
> In retrospect, that makes a lot of sense!

You're be ahead of me. I only just figured it out. The Ubuntu AppVeyor image still has only .NET 6.0.x, which doesn't have the fix to the `FrameworkDisplayName`.

thomaslevesque:
> * somehow filter away the offending attribute
> * duplicate the received/verified files, one set per OS

Seems messy, especially since it's a temporary issue (the Ubuntu image will eventually be updated to have the .NET 7 SDK, I assume)

> * not run the API check on Ubuntu (or Windows, I'm not fussy)

I guess we could, but I don't love it

> * Something I didn't think of

Force the SDK version to 6.0.x with a global.json, until .NET 7 is installed on the Ubuntu VM image?

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you!

PR #1892, opened 2022-11-26
Use GitHub Actions for CI

thomaslevesque:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+      - master
+      - support/*
+  workflow_dispatch:
+  workflow_call:
+
+env:
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+  DOTNET_CLI_TELEMETRY_OPTOUT: true
+  DOTNET_NOLOGO: true
+  DOTNET_GENERATE_ASPNET_CERTIFICATE: false
```
blairconrad, line 16: super nit: I'd've sorted. Don't bother if you aren't coming back anyhow. Or maybe just don't bother!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+            3.1
+            2.1
+      - name: Run build script
+        run: ./build.sh
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: ubuntu-artifacts
+          path: artifacts/**/*
```
blairconrad, line 59: Nothing wrong with this, but it's a shame we can't use a matrix and avoid all the duplication. But we'd need to deal with the different build commands as well as the need for extra runtimes on Linux but not Windows.
I wouldn't make any changes. Just making conversation.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+
+  build-docs:
+    name: Build documentation
+    runs-on: ubuntu-22.04
```
blairconrad, line 63: I see you prefer a named LTS version to "latest", eh?

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: windows-artifacts
+          path: /tmp/artifacts
+      - name: Set DRY_RUN_OPTION env variable
+        if: github.event.inputs.dry-run
+        run: echo "DRY_RUN_OPTION=--dry-run" >> $GITHUB_ENV
+      - name: Deploy
+        run: dotnet run --project tools-shared/FakeItEasy.Deploy -- $DRY_RUN_OPTION -r "FakeItEasy/FakeItEasy" -t "${{ github.ref_name }}" -a "/tmp/artifacts/output"
```
blairconrad, line 39: Looking at this makes me nervous, but you've probably tried not setting this environment variable and running locally to see what happens. It's fine? Doesn't think that there's an empty argument?

blairconrad (commented):
I'm content to merge once we remove AppVeyor requirement and rename.
Also, we probably want to merge the Tools changes first? I think the deployment script change from #1895 can wait until after this, no?

blairconrad:
* [ ] update required build events

.github/workflows/ci.yml
thomaslevesque: Yes, to avoid future issues when `ubuntu-latest` is updated to a newer version. Better to remain in control and update it when we want to, rather than having to fix things when they break because the image was updated behind our back.

.github/workflows/ci.yml
blairconrad: As pleases you. If you think it's fun, go ahead. Otherwise, just carry on. I have no problem with what you have.

(although, does `./build.ps1` work on both OSes?)

.github/workflows/release.yml
thomaslevesque: Yes, it's fine. No empty argument, just some whitespace.
I also tried setting a dummy env variable and adding it to the command line, just to make sure it was correctly set (rolled it back now).

blairconrad (changes requested):
Looks great, except for [the docs](https://github.com/FakeItEasy/FakeItEasy/search?q=appveyor) (and how to build, where we should probably mention the `build.ps1`.

FakeItEasy.sln
```diff
@@ -8,8 +8,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		.gitattributes = .gitattributes
 		.gitignore = .gitignore
 		.mailmap = .mailmap
-		appveyor.yml = appveyor.yml
 		build.cmd = build.cmd
+		build.sh = build.sh
```
blairconrad, line 12: nice

blairconrad (commented):
Thanks, @thomaslevesque. Great work!

blairconrad (approved):

PR #1907, opened 2023-01-04
Improve in-article tables of contents

blairconrad:
While looking at the [Argument constraints](https://fakeiteasy.github.io/docs/7.3.1/argument-constraints/) page yesterday, I noticed that unlike many other pages, the subheadings don't appear in the page's table of contents:

image

I found it's because the page uses level 1 headings instead of level 2. Typically in mkdocs, the level 1 heading is used for the page title only. 

This PR attempts to remedy that. And also suppresses the table of contents on the main landing page, as it is not helpful:

image

### DIFF
--- a/docs/argument-constraints.md
+++ b/docs/argument-constraints.md
@@ -5,7 +5,7 @@ the arguments of the call can be constrained so that only calls to the
 configured method where the arguments matches the constraint are
 selected.
 
-# Matching values exactly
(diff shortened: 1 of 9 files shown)


blairconrad:
New pages:

image

image

image

image

image


image

thomaslevesque (approved):
Nice! Thanks @blairconrad!

PR #1908, opened 2023-01-04
Lock down python requirements

blairconrad:
Should save us from surprise changes as new packages are released. Let us choose when to take advantage of new features.


thomaslevesque (approved):
LGTM, but I have just one question: was it necessary to move the requirements file to the root, or did you just do it because it seemed better this way?

blairconrad:
> LGTM, but I have just one question: was it necessary to move the requirements file to the root, or did you just do it because it seemed better this way?

Thanks. This is a slight criticism I have of the PR interface. It's sometimes hard to see the full message that accompanies a commit. Right now, the requirements are [being bundled in the documentation](https://fakeiteasy.github.io/docs/7.3.1/requirements.txt)!

It's harmless, but a little annoying, and I thought "now we'll have **2** such files"

If you prefer, I could drop that commit. After all, it's not such a big deal.
Or I can look for another mechanism to exclude the files, while keeping them in the docs dir.

blairconrad:
> I can look for another mechanism to exclude the files, while keeping them in the docs dir.

[mkdocs-exclude](https://pypi.org/project/mkdocs-exclude/) seems to do the trick!
I can use it instead, if you prefer. It's a very minor change.

thomaslevesque:
OK, thanks for the explanation. I had not seen the commit message.
Moving it to the root is fine by me, I was just wondering what the reason was.

blairconrad:
Okay, I'm not fussy. If you're content with the files at the root, merge at your leisure.

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you!

PR #1911, opened 2023-02-04
Use deploy action in release workflow

thomaslevesque:


blairconrad (approved):
So tidy. Interesting to see what the change is like without all the side quests!

Thanks for your work, @thomaslevesque.

PR #1914, opened 2023-03-02
Fix build - remove .NET 5.0 target for tests

thomaslevesque:
The first commit is actually my first fix attempt, which was to install 5.0 on Windows, hence the changes in ci.yml. I can roll them back if you prefer.


blairconrad (approved):
Thanks, @thomaslevesque. Generally looks good. As for the meandering "install 5.0, no don't!" changes, I actually like the new `dotnet-versions-to-install` matrix property, but don't think we need to see the path that got us here. Me, I'd just squash the three commits into one. But if you prefer not to, I'm happy to merge as-is.

tests/FakeItEasy.Specs/FakeItEasy.Specs.csproj
```diff
@@ -11,10 +11,6 @@
       <DefineConstants>$(DefineConstants);MISMATCH_COMPILE_AND_RUNTIME_FRAMEWORKS</DefineConstants>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(TargetFramework)' == 'net5.0'">
-      <DefineConstants>$(DefineConstants);LACKS_FAKEABLE_GENERIC_IN_PARAMETERS</DefineConstants>
-  </PropertyGroup>
```
blairconrad, line 16: Now we just need to drop 3.1 and we can get rid of a constant!

thomaslevesque:
squashed

PR #1915, opened 2023-03-15
Add Faking HttpClient recipe

blairconrad:
Adds runnable recipes (in form of tests) and includes those in a Recipe section of the documentation, by reading "snippets" right out of the executable recipe.

----
**Old description:**
Not baked yet, but a good starting point for discussion. In particular, we're going to at least want to decide whether we want these things called "case studies" or "recipes" or whatever. I picked one so I could have something to commit and push.
@thomaslevesque, everything's negotiable. For example, you might hate tone, approach of showing one way to work and then introducing the fakeable context, showing a "quicky" version of the test where we exercise the fake directly, anything. Don't hold back.

I also have ideas for hosting the code in this article (or others?) separately, similarly to the samples directory, but probably runnable. But let's look at that later…

### DIFF
--- a/docs/Recipes/faking-http-client.md
+++ b/docs/Recipes/faking-http-client.md
@@ -1,0 +1,78 @@
+
+Let's assume that you want to create a fake `HttpClient` so you can dictate the
+behavior of the
+[GetAsync(String)](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.getasync?view=net-7.0#system-net-http-httpclient-getasync(system-string))
+method. This seems like it would be a
+straightforward task, but it's complicated by the design of `HttpClient`, which
+is not faking-friendly.
+
+## A working Fake
+
+First off, let's look at the declaration of `GetAsync`:
+
+```csharp
+public Task<HttpResponseMessage>GetAsync(string? requestUri)
+```
+
+This method is neither virtual nor abstract, and so [can't be overridden by
+FakeItEasy](../what-can-be-faked.md#what-members-can-be-overridden).
+
+As a workaround, we can look at the [definition of
+GetAsync](https://github.com/dotnet/runtime/blob/ab5e28c1cab305450897749daa7393bef30d7505/src/libraries/System.Net.Http/src/System/Net/Http/HttpClient.cs#L363-L364)
+and see that the method eventually ends up calling
+[HttpMessageHandler.SendAsync(HttpRequestMessage,
+CancellationToken)](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpmessagehandler.sendasync?view=net-7.0#system-net-http-httpmessagehandler-sendasync(system-net-http-httprequestmessage-system-threading-cancellationtoken))
+on an `HttpMessageHandler` that can be supplied via the `HttpClient`
+constructor.
+
+`HttpMessageHandler.SendAsync` is protected, which makes it
+less convenient to override than a public method. We need to specify the call by
+name, and to give FakeItEasy a hint about the return type, as described in
+[Specifying a call to any method or
+property](../specifying-a-call-to-configure.md#specifying-a-call-to-any-method-or-property).
+
+With this knowledge, we can write a passing test:
+
+??? note "This is a simplified example"
+    In the interest of brevity, we create a Fake, exercise it directly, and check
+    its behavior. A more realistic example would create the Fake as a collaborator
+    of some production class (the "system under test") and the Fake would not be
+    called directly from the test code.
+
+```csharp
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/FakingHttpClient.cs:FakeAnyMethodWay
+--8<--
+```
+
+## Easier and safer call configuration
+
+The above code works, but specifying the method name and return type is a little
+awkward. A `FakeableHttpMessageHandler` class can be used to clean things up and
+to also supply a little compile-time safety by ensuring we're configuring the
+expected method.
+
+```csharp
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/FakingHttpClient.cs:FakeableHttpMessageHandler
+
+recipes/FakeItEasy.Recipes.CSharp/FakingHttpClient.cs:FakeByMakingMessageHandlerFakeable
+--8<--
+```
+
+## Faking PostAsync
+
+The techniques above can be used to intercept calls to  methods other than `GetAsync` as well.
+
+Consider this test, which ensures that the correct content is passed to `HttpClient.PostAsync`.
+
+??? bug "Does not work for .NET Framework"
+    When run under .NET Framework, the request content is disposed
+    as soon as the request is made, as explained in
+    [HttpClient source code comments](https://github.com/dotnet/runtime/blob/b0b7aaefb88aa8d01b3d64fb40ac2f73a9d98c3e/src/libraries/System.Net.Http/src/System/Net/Http/HttpClient.cs#L665-L680).
+
+```csharp
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/FakingHttpClient.cs:FakePostAsync
+--8<--
+```
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -41,8 +41,11 @@ plugins:
 markdown_extensions:
   - pymdownx.highlight:
       anchor_linenums: true
+  - pymdownx.details
   - pymdownx.inlinehilite
-  - pymdownx.snippets
+  - pymdownx.snippets:
+      check_paths: true
+      dedent_subsections: true
   - pymdownx.superfences
   - pymdownx.tilde
 
@@ -80,6 +83,8 @@ nav:
 - Bootstrapper: bootstrapper.md
 - Advanced usage: advanced-usage.md
 - Analyzer Packages: https://fakeiteasy.github.io/docs/analyzers/
+- Recipes:
+  - Faking HttpClient: Recipes/faking-http-client.md
 - Platform support: platform-support.md
 - Source Stepping: source-stepping.md
 - Why was FakeItEasy created?: why-was-fakeiteasy-created.md
--- a/recipes/FakeItEasy.Recipes.CSharp/FakeItEasy.Recipes.CSharp.csproj
+++ b/recipes/FakeItEasy.Recipes.CSharp/FakeItEasy.Recipes.CSharp.csproj
@@ -1,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>$(RecipesTargetFrameworks)</TargetFrameworks>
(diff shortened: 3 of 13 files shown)


docs/Recipes/faking-http-client.md
```diff
@@ -0,0 +1,98 @@
+
+Let's assume that you want to create a fake `HttpClient` so you can dictate the
+behavior of the
+[GetAsync(String)](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.getasync?view=net-7.0#system-net-http-httpclient-getasync(system-string))
```
thomaslevesque, line 4: I realize the goal isn't to cover every scenario, but it might be helpful to also include an example for `PostAsync` with a request body.

docs/Recipes/faking-http-client.md
```diff
@@ -0,0 +1,98 @@
...
+straightforward task, but it's complicated by the design of `HttpClient`, which
+is not faking-friendly.
+
+## A working Fake
+
+First off, let's look at the declaration of `GetAsync`:
+
+```csharp
+public Task<HttpResponseMessage>
+    GetAsync([StringSyntax(StringSyntaxAttribute.Uri)] string? requestUri)
```
thomaslevesque, line 15: I think we can omit the attribute, for the sake of readability.

docs/Recipes/faking-http-client.md
```diff
@@ -0,0 +1,98 @@
...
+
+`HttpMessageHandler.SendAsync` is protected, which makes it
+less convenient to override than a public method. We need to specify the call by
+name, and to give FakeItEasy a hint about the return type, as described in
+[Specifying a call to any method or
+property](../specifying-a-call-to-configure.md#specifying-a-call-to-any-method-or-property).
+
+With this knowledge, we can write a passing test:
+
+??? note "This is a simplified example"
```
thomaslevesque, line 37: Does `??? node` do something special that I'm not aware of? Or is it just a placeholder for something that will actually render as a note?

docs/Recipes/faking-http-client.md
```diff
@@ -0,0 +1,98 @@
...
+    In the interest of brevity, we create a Fake, exercise it directly, and check
+    its behavior. A more realistic example would create the Fake as a collaborator
+    of some production class (the "system under test") and the Fake would not be
+    called directly from the test code.
+
+```csharp
+public async Task Test()
+{
+    var response = new HttpResponseMessage();
+    response.Content = new StringContent("FakeItEasy is fun");
```
thomaslevesque, line 47: nit: maybe use object initializer syntax here?
```suggestion
    var response = new HttpResponseMessage
    {
        Content = new StringContent("FakeItEasy is fun")
    };

```

thomaslevesque (approved):
Thanks! This looks great. I have a few suggestions, nothing major.

docs/Recipes/faking-http-client.md
blairconrad: It does something you're not aware of! I thought you'd like it.

image

image

docs/Recipes/faking-http-client.md
blairconrad: I originally had it, but one one line, and it didn't render well. Will try your version.

docs/Recipes/faking-http-client.md
blairconrad: I'll work something up.

docs/Recipes/faking-http-client.md
thomaslevesque: I do like it!
It's specific to mkdocs, right?

docs/Recipes/faking-http-client.md
```diff
@@ -12,7 +12,7 @@ First off, let's look at the declaration of `GetAsync`:
 
 ```csharp
 public Task<HttpResponseMessage>
-    GetAsync([StringSyntax(StringSyntaxAttribute.Uri)] string? requestUri)
+    GetAsync(string? requestUri)
```
thomaslevesque, line 15: Without the attribute, you could put it all on one line

docs/Recipes/faking-http-client.md
blairconrad: Specifically it's a [PyMdown Extention](https://facelessuser.github.io/pymdown-extensions/extensions/details/).

docs/Recipes/faking-http-client.md
blairconrad: Is this close to what you were hoping for?

```csharp
public async Task PostAsync()
{
    using var response = new HttpResponseMessage
    {
        Content = new StringContent("FakeItEasy is fun")
    };

    var handler = A.Fake<FakeableHttpMessageHandler>();

    using var client = new HttpClient(handler);

    using var postContent = new StringContent("I am posting this.");
    var result = await client.PostAsync("https://fakeiteasy.github.io/docs/", postContent);

    A.CallTo(() => handler.FakeSendAsync(
            A<HttpRequestMessage>.That.Matches(
                m => m.Content!.ReadAsStringAsync().Result == "I am posting this."),
            A<CancellationToken>.Ignored))
        .MustHaveHappened();
}
```

thomaslevesque: Looks good to me! The "sync over async" bit is a little annoying, but I don't think there's much we can do about it... I don't think it can actually cause a deadlock in this case.

thomaslevesque: yes

blairconrad:
Hey, @thomaslevesque. Still a WIP, and builds on the yet-unmerged mkdocs upgrade.
But I thought you might enjoy seeing the slightly different route I took. In particular, note that the recipes are working tests, and I include snippets in the docs. I think this is cool. Also maybe useful.

Oh, the recipe sometimes fail on net4.5. Possibly due to sync over async?
I haven't gotten far into debugging. Part of me wonders whether we should even worry about having the recipes work on all TFMs. Another part of me says "well, yeah, if we can". Another part worries that the net45 failure may rear its ugly head on other TFMs even though I haven't seen it yet locally.

recipes/FakeItEasy.Recipes.CSharp/FakingHttpClient.cs
```diff
@@ -48,24 +51,29 @@ public async Task FakeByMakingMessageHandlerFakeable()
...
     [Fact]
     public async Task FakePostAsync()
     {
         var handler = A.Fake<FakeableHttpMessageHandler>();
 
         using var client = new HttpClient(handler);
 
-        using var postContent = new StringContent("I am posting this.");
-        var result = await client.PostAsync("https://fakeiteasy.github.io/docs/", postContent);
+        using var postContent = new StringContent("my post");
```
blairconrad, line 64: the original was a little wide when rendered

blairconrad:
As usual, I made a few assumptions that I'm happy to have overturned. Maybe you don't like the recipes being tests. Maybe you want a different folder organization. I dunno. Whatever.
Also, I'd probably arrange the commits differently, for a final merge, just going right to the snippetted version (if you like the snippets), but I thought that at least this way I can drop commits if you hate something.

recipes/FakeItEasy.Recipes.CSharp/FakeItEasy.Recipes.CSharp.csproj
thomaslevesque, line 1: Do we actually need a new project for this? Maybe we could add the recipes to the examples project? (I'm not really opposed to a new project, just thinking out loud)

thomaslevesque (commented):
> In particular, note that the recipes are working tests, and I include snippets in the docs. I think this is cool.

It _is_ pretty cool! Nice work

> Oh, the recipe sometimes fail on net4.5. Possibly due to sync over async?

This is annoying... I'll try to see what's going on.

recipes/FakeItEasy.Recipes.CSharp/FakingHttpClient.cs
```diff
@@ -0,0 +1,88 @@
...
+    //// --8<-- [start:FakePostAsync]
+    [Fact]
+    public async Task FakePostAsync()
+    {
+        var handler = A.Fake<FakeableHttpMessageHandler>();
+
+        using var client = new HttpClient(handler);
+
+        using var postContent = new StringContent("my post");
+        var result = await client.PostAsync(
```
thomaslevesque, line 65: nit: the `result` variable is never used

blairconrad:
> > Oh, the recipe sometimes fail on net4.5. Possibly due to sync over async?
> 
> This is annoying... I'll try to see what's going on.

Copying in your comments from sidechat. 

----

Looks like in .NET 4.5, the content is disposed as soon as the request is sent :/
https://github.com/dotnet/runtime/blob/b0b7aaefb88aa8d01b3d64fb40ac2f73a9d98c3e/src/libraries/System.Net.Http/src/System/Net/Http/HttpClient.cs#L665-L680
I think the best path forward is to warn that it doesn't work in .NET 4.5 (and ensure it actually work for all the other TFMs), and skip this test when the TFM is net45
Another option would be to introduce a wrapper that prevent disposal of the content:


```csharp

using var postContent = new StringContent("my post");
var result = await client.PostAsync(
    "https://fakeiteasy.github.io/docs/",
    postContent.PreventDisposal());
```

where `PreventDisposal` is an extension method returning a decorator that overrides `Dispose(bool)` to do nothing

It's not great, though...

More on this: https://github.com/dotnet/runtime/issues/14612


And here's the PR that changes the behavior: https://github.com/dotnet/corefx/pull/19082


I can't figure out in which version it was introduced, though...
Maybe .NET Core 2 (the first preview was released shortly after this PR was merged)

recipes/FakeItEasy.Recipes.CSharp/FakeItEasy.Recipes.CSharp.csproj
thomaslevesque: > "Hey! The examples were useful back when we had no docs, and are pretty thin now. What if we just made sure everything was documented or a recipe, if deserving, and remove that project"

I kinda like this idea. I doubt many people (if any) actually look at the Examples project anyway...

blairconrad: Thanks. I never do! Created #1917.

blairconrad:
Thanks for the comments and research, @thomaslevesque. I think I've incorporated everything and have squashed and removed the WIP tag.
Of course, still happy to modify.

mkdocs.yml
```diff
@@ -41,8 +41,11 @@ plugins:
 markdown_extensions:
   - pymdownx.highlight:
       anchor_linenums: true
+  - pymdownx.details
   - pymdownx.inlinehilite
-  - pymdownx.snippets
+  - pymdownx.snippets:
+      check_paths: true
+      dedent_subsections: true
```
blairconrad, line 48: This is the option that the new material for mkdocs provided.

thomaslevesque (approved):

thomaslevesque:
Great job! Thanks @blairconrad

PR #1919, opened 2023-03-22
Remove the Examples projects

### LINKED ISSUE #1917: Replace examples projects with documentation or recipes
Inspired by [a comment I made in PR 1915](https://github.com/FakeItEasy/FakeItEasy/pull/1915#discussion_r1144579014).

>  We could convert examples to be a test-like project instead of adding a new one. Alternatively, could say "Hey! The examples were useful back when we had no docs, and are pretty thin now. What if we just made sure everything was documented or a recipe, if deserving, and remove that project". (Then again, if we added recipes to the "Examples*" project(s) (under "sample", which I never noticed before, we could still thin them out later if we wanted.)
>
> Based on my actions, I slightly preferred a new \[recipes\] project (and removing the old ones)

So, once we have runnable recipes that are included in the docs via snippets, we should examine the examples projects and see if they still provide value. For each example, we should be able to

1. find it in the docs, or
2. add it to the docs (as a recipe or not)

and then remove it!

blairconrad:
Closes #1917.

I think they're all covered in the docs.
A future step might be to make sure all the code in our docs comes from the recipes project, but that can be a future step. Or never.


thomaslevesque (approved):
Thanks @blairconrad!
So much red in this diff, I love it!

image

PR #1923, opened 2023-04-01
Make ArgumentCollection.GetArgumentNames return type IEnumerable<string?>

### LINKED ISSUE #1922: ArgumentCollection.ArgumentNames has type IEnumerable<string> but may contain nulls
In F# (at least), it is possible to declare a method with anonymous parameters.
In this case, `ArgumentCollection.ArgumentNames` will contain `null`s, but we do not indicate that the elements are nullable strings.

blairconrad:
Fixes #1922

### DIFF
--- a/src/FakeItEasy/Configuration/ArgumentCollection.cs
+++ b/src/FakeItEasy/Configuration/ArgumentCollection.cs
@@ -19,7 +19,7 @@ namespace FakeItEasy.Configuration
         ///   The arguments this collection contains.
         /// </summary>
         private readonly object?[] arguments;
-        private readonly Lazy<string[]> argumentNames;
+        private readonly Lazy<string?[]> argumentNames;
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


blairconrad:
I added a bonus commit to clean up some "hints". Happy to drop if it offends.

src/FakeItEasy/Configuration/ArgumentCollection.cs
```diff
@@ -54,7 +53,7 @@ public int Count
         /// <summary>
         ///   Gets the names of the arguments in the list.
```
blairconrad, line 54: Ugh. There's an opportunity to add docs here around when the names might be null. I dropped the ball. Stay tuned.

blairconrad:
Docs added!

thomaslevesque (approved):
Perfect, thanks @blairconrad!

blairconrad:
Thanks, @thomaslevesque.

Issue #1920, opened 2023-03-29
NullReferenceException thrown when attempting to build unmet expectation message that includes an anonymous parameter


kaeedo:
Hello. I'm using FakeItEasy to test F# projects, and I much prefer it over Moq :)

However there is one problem, and that is when making assertion that a call must have happened, the assertion message it prints is misleading at best: `System.ArgumentNullException: Value cannot be null. (Parameter 'value')`. I wonder if this can be fixed somehow, either in my own code, or something you could fix on your side. I've included a small repository to to reproduce this issue, and also shows that it works as expected when testing a C# class

https://github.com/kaeedo/FakeItFSharp

I also noticed an old issue about adding an F# extension #232 . Any updates on that?

Thanks.

blairconrad:
Hi, @kaeedo. Thanks for the kind words for FakeItEasy, and sorry you're having a less than ideal time.

Regarding #232, no updates. The original idea was proposed by a community member, and no contribution was forthcoming. I'm not a F# user _at all_, and so have little insight (or use for the project). @thomaslevesque is more aware of the F# technology and may have additional insights, but the issue has never been a high priority. I'd've consider closing it, except we're not against the concept, just resource-poor.

----

I've run your repro and see the problem (and at some level, the cause). For those playing along at home, the F# output is

```
[xUnit.net 00:00:02.55]     Tests.Fake it easy with mocking [FAIL]
  Failed Tests.Fake it easy with mocking [87 ms]
  Error Message:
   System.ArgumentNullException : Value cannot be null. (Parameter 'value')
  Stack Trace:
     at FakeItEasy.Guard.AgainstNull(Object argument, String argumentName) in C:\projects\fakeiteasy\src\FakeItEasy\Guard.cs:line 24
   at FakeItEasy.DefaultOutputWriter.Write(String value) in C:\projects\fakeiteasy\src\FakeItEasy\DefaultOutputWriter.cs:line 26
   at FakeItEasy.Expressions.CallConstraintDescriber.WriteArgumentsListString(IOutputWriter writer, MethodInfo method, IEnumerable`1 argumentConstraints) in C:\projects\fakeiteasy\src\FakeItEasy\Expressions\CallConstraintDescriber.cs:line 70
[...]

PR #1924, opened 2023-04-14
Provide placeholder names for anonymous parameters in received call descriptions

blairconrad:
Follows from fix for #1920.

If a method has anonymous parameters, the description of a received call should use placeholder names just as we would for a call that is expected but not received.

### DIFF
--- a/src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
+++ b/src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
@@ -132,15 +132,14 @@ namespace FakeItEasy.Core
 
         private static ArgumentValueInfo[] GetArgumentValueInfos(IFakeObjectCall call)
         {
-            return
-                (from argument in
-                    call.Method.GetParameters()
-                        .Zip(call.Arguments, (parameter, value) => new { parameter.Name, Value = value })
-                    select new ArgumentValueInfo
-                    {
-                        ArgumentName = argument.Name,
-                        ArgumentValue = argument.Value
-                    }).ToArray();
+            return call.Method.GetParameters()
+                .Zip(call.Arguments, (parameter, value) => new { parameter.Name, Value = value })
+                .Select((argument, index) => new ArgumentValueInfo
+                {
+                    ArgumentIndex = index,
+                    ArgumentName = argument.Name,
+                    ArgumentValue = argument.Value
+                }).ToArray();
         }
 
         private void AppendArgumentsList(StringBuilder builder, IFakeObjectCall call)
@@ -166,13 +165,17 @@ namespace FakeItEasy.Core
 
         private void AppendArgumentValue(StringBuilder builder, ArgumentValueInfo argument)
         {
+            // Usually parameters will be named, but in F# (at least) it's possible
+            // to declare a method with anonymous parameters. In that case, we try to
+            // help the user by outputting names like "param1", "param2", …
+            string argumentName = argument.ArgumentName ?? $"param{argument.ArgumentIndex + 1}";
             builder
-                .Append(argument.ArgumentName)
+                .Append(argumentName)
                 .Append(": ")
                 .Append(this.GetArgumentValueAsString(argument.ArgumentValue));
         }
 
-        private string GetArgumentValueAsString(object argumentValue)
+        private string GetArgumentValueAsString(object? argumentValue)
         {
             return this.argumentValueFormatter.GetArgumentValueAsString(argumentValue);
         }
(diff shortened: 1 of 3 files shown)


src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
```diff
@@ -164,36 +164,40 @@ private void AppendArgumentsList(StringBuilder builder, IFakeObjectCall call)
...
-            var callIndex = 0;
+            var argumentIndex = 0;
             foreach (var argument in arguments)
             {
-                AppendArgumentSeparator(builder, callIndex, totalNumberOfArguments);
-                this.AppendArgumentValue(builder, argument);
-                callIndex++;
+                AppendArgumentSeparator(builder, argumentIndex, totalNumberOfArguments);
+                this.AppendArgumentValue(builder, argumentIndex, argument);
+                argumentIndex++;
```
thomaslevesque, line 192: nit: the argument index could be in `ArgumentValueInfo`, making this code slightly simpler

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -184,6 +185,34 @@ public static void CallToAField(ClassWithAField fake, Exception exception)
     1: FakeItEasy.Specs.CallMatchingSpecs+IHaveTwoGenericParameters.Bar`2[System.Int32,System.Double](baz1: 1, baz2: 2)
     2: FakeItEasy.Specs.CallMatchingSpecs+IHaveTwoGenericParameters.Bar`2[FakeItEasy.Specs.CallMatchingSpecs+GenericClass`2[System.Boolean,System.Int64],System.Int32](baz1: FakeItEasy.Specs.CallMatchingSpecs+GenericClass`2[System.Boolean,System.Int64], baz2: 3)
 
+"));
+        }
+
+        [Scenario]
+        public static void FailingMatchOCallWithAnonymousParameter(
```
thomaslevesque, line 192: What is this `O` doing here all alone in the method name? Typo?

thomaslevesque (approved):
LGTM, except for a possible typo and a minor suggestion

src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
blairconrad: Okay! I tried a thing. Tell me what you think.

src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
```diff
@@ -132,15 +132,14 @@ private static void AppendArgumentSeparator(StringBuilder builder, int argumentI
...
-                        ArgumentValue = argument.Value
-                    }).ToArray();
+            return call.Method.GetParameters()
+                .Zip(call.Arguments, (parameter, value) => new { parameter.Name, Value = value })
+                .Select((argument, index) => new ArgumentValueInfo
+                {
+                    ArgumentIndex = index,
+                    ArgumentName = argument.Name,
+                    ArgumentValue = argument.Value
+                }).ToArray();
```
thomaslevesque, line 142: I think I like it better like this anyway... the version using query syntax seemed weird with the parentheses around it.

src/FakeItEasy/Core/DefaultFakeObjectCallFormatter.cs
```diff
@@ -184,20 +183,20 @@ private string GetArgumentValueAsString(object? argumentValue)
...
-                argumentIndex++;
+                AppendArgumentSeparator(builder, argument.ArgumentIndex, totalNumberOfArguments);
+                this.AppendArgumentValue(builder, argument);
             }
         }
 
         private struct ArgumentValueInfo
         {
-            public object? ArgumentValue { get; set; }
+            public int ArgumentIndex { get; init; }
```
thomaslevesque, line 195: Looks like we're missing something to be able to use `init`. I think we can just declare `IsExternalInit` ourselves for TFMs that don't have it.

Or maybe we could use a record? (but again, we'll probably need to declare some missing types ourselves)

(and of course, the simplest fix is probably to go back to using a normal setter instead of `init`; it doesn't really matter, the type is private anyway.)

thomaslevesque (approved):
Thanks @blairconrad. Looks good to me, except the compilation errors...

blairconrad:
> except the compilation errors...

😆 

I'm lazy and just building for 6.0 at home. It's worked well enough for so long!

thomaslevesque:
> I'm lazy and just building for 6.0 at home. It's worked well enough for so long!

Yeah, I do the same...

thomaslevesque (approved):
LGTM, thanks @blairconrad!

blairconrad:
Thanks for the quality reviews, @thomaslevesque!

PR #1928, opened 2023-04-18
Ensure anonymous parameters are handled and displayed correctly in remaining cases

blairconrad:
Provides test coverage of the last 2 instances where we access parameter names, ensuring we behave somewhat sanely when the parameters have no names. Includes one minor change in how anonymous parameters are displayed, but we can only figure out how to exercise that via unit test. An "organic" use may not be possible.

Finishes work started in #1920, #1922, and #1924.

### DIFF
--- a/src/FakeItEasy/MethodInfoExtensions.cs
+++ b/src/FakeItEasy/MethodInfoExtensions.cs
@@ -170,10 +170,15 @@ namespace FakeItEasy
             for (int i = 0; i < parameters.Length; i++)
             {
                 AppendParameterSeparator(builder, i);
-                builder
-                    .Append(parameters[i].ParameterType)
-                    .Append(' ')
-                    .Append(parameters[i].Name);
+                builder.Append(parameters[i].ParameterType);
+
+                var parameterName = parameters[i].Name;
+                if (parameterName is not null)
+                {
+                    builder
+                       .Append(' ')
+                       .Append(parameterName);
+                }
             }
         }
 
--- a/tests/FakeItEasy.Specs/CreationSpecs.cs
+++ b/tests/FakeItEasy.Specs/CreationSpecs.cs
@@ -8,6 +8,7 @@ namespace FakeItEasy.Specs
     using FakeItEasy.Core;
     using FakeItEasy.Creation;
     using FakeItEasy.Tests.TestHelpers;
+    using FakeItEasy.Tests.TestHelpers.FSharp;
     using FluentAssertions;
     using Xbehave;
     using Xunit;
@@ -385,6 +386,19 @@ namespace FakeItEasy.Specs
   Failed to create fake of type System.Func`1[FakeItEasy.Specs.CreationSpecsBase+PrivateClass]:
     Can not create proxy for type System.Func`1[[FakeItEasy.Specs.CreationSpecsBase+PrivateClass, FakeItEasy.Specs, Version=1.0.0.0, Culture=neutral, PublicKeyToken=eff28e2146d5fd2c]] because type FakeItEasy.Specs.CreationSpecsBase+PrivateClass is not accessible. Make it public, or internal and mark your assembly with [assembly: InternalsVisibleTo(*DynamicProxyGenAssembly2*)] attribute*
 "));
+        }
+
+        [Scenario]
+        public void PublicDelegateWithAnonymousParameterCanBeFaked(IAmADelegateWithAnAnonymousParameter fake)
+        {
+            "Given a public delegate with an anonymous parameter "
+                .See<IAmADelegateWithAnAnonymousParameter>();
+
+            "When I create a fake of the delegate"
+                .x(() => fake = this.CreateFake<IAmADelegateWithAnAnonymousParameter>());
+
+            "Then the fake is created"
+                .x(() => fake.Should().BeAFake());
         }
 
         [Scenario]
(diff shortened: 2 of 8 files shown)


src/FakeItEasy/MethodInfoExtensions.cs
```diff
@@ -171,9 +171,14 @@ private static void AppendParameters(StringBuilder builder, ParameterInfo[] para
             {
                 AppendParameterSeparator(builder, i);
                 builder
-                    .Append(parameters[i].ParameterType)
-                    .Append(' ')
-                    .Append(parameters[i].Name);
+                    .Append(parameters[i].ParameterType);
```
thomaslevesque, line 174: nit: Maybe drop the line break here? It made sense when there were multiple calls to Append, but now it looks a bit strange.

src/FakeItEasy/MethodInfoExtensions.cs
```diff
@@ -171,9 +171,14 @@ private static void AppendParameters(StringBuilder builder, ParameterInfo[] para
...
-                    .Append(' ')
-                    .Append(parameters[i].Name);
+                    .Append(parameters[i].ParameterType);
+                var parameterName = parameters[i].Name;
+                if (parameterName is not null)
+                {
+                    builder
+                       .Append(' ')
+                       .Append(parameterName);
+                }
```
thomaslevesque, line 181: Based on the title and description of the PR, I didn't expect any change to the library. Maybe rename the PR to match what it actually does?

tests/FakeItEasy.Tests/MethodInfoExtensionsTests.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        // Arrange
+        var expected = "FakeItEasy.Tests.TestHelpers.FSharp.IHaveAMethodWithAnAnonymousParameter.Save(System.Int32)";
+
+        // Act
+        var actual = typeof(IHaveAMethodWithAnAnonymousParameter).GetMethod("Save")!.GetDescription();
+
+        // Assert
+        actual.Should().Be(expected);
+    }
```
thomaslevesque, line 27: A unit test! I thought we didn't do that anymore.
Is this because it was easier to test it like this than in a spec?

tests/FakeItEasy.Specs/CreationSpecs.cs
```diff
@@ -387,6 +388,19 @@ public void PublicDelegateWithPrivateTypeArgumentCannotBeFaked(Exception excepti
...
+        {
+            "Given a public delegate with an anonymous parameter "
+                .See<IAmADelegateWithAnAnonymousParameter>();
+
+            "When I create a fake of the delegate"
+                .x(() => fake = this.CreateFake<IAmADelegateWithAnAnonymousParameter>());
+
+            "Then the fake is created"
+                .x(() => fake.Should().BeAFake());
+        }
```
thomaslevesque, line 402: Good catch! I wouldn't have thought of testing delegates

thomaslevesque (commented):
Thanks, @blairconrad! I have a few comments and questions

tests/FakeItEasy.Tests/MethodInfoExtensionsTests.cs
blairconrad: (Read the giant comment at the top of the method.)

tests/FakeItEasy.Specs/CreationSpecs.cs
blairconrad: Wasn't all me. Just asked VS "where do we get the name of a parameter" and followed the calls.

src/FakeItEasy/MethodInfoExtensions.cs
blairconrad: done!

src/FakeItEasy/MethodInfoExtensions.cs
blairconrad: Fair point! Updated.

blairconrad (commented):
I think I addressed your concerns, @thomaslevesque. Thanks for voicing them.

I don't know what's going on with the unit tests on Windows.

tests/FakeItEasy.Tests/MethodInfoExtensionsTests.cs
thomaslevesque: That will teach me to actually read the comments!

blairconrad:
@thomaslevesque has figured out that the test failure on windows is due to dotnet/fsharp#14863. FakeItEasy.Tests.TestHelper.FSharp is delay signed. We don't understand why it's working with .Specs, except that maybe it's because .Tests sets some InternalsVisibleTo attributes (why that would make a difference, I don't know)

Options:

1. fully sign it after build. Revert after tooling catches up
2. rip the commit that requires it out. Split into another issue to implement once tooling catches up
3. drop .NET Framework support altogether
4. ??

blairconrad:
(a95d1958ed267e6aa1cbd0448054b0b75fc72208 tries fully signing after build)

thomaslevesque (approved):

blairconrad:
Created #1930 to remove the postbuild step sometime.

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you, @thomaslevesque.

Issue #1935, opened 2023-04-25
Publish target framework Support Policy


blairconrad:
Following #1925, we've been discussing which TFMs we want to support, and what policy we should have ongoing.

We've decided on

* when a new LTS .NET version is released, add a TFM for it
* when a LTS .NET version arrives at end of life, drop it
* don't add a TFM for non-LTS .NET versions

In addition, we will

* keep .NET Framework 4.6.2, at least as long as it's not causing any problems
* keep netstandard 2.0 and netstandard 2.1, as they've not been deprecated

We should add a support policy page to the documentation.

blairconrad:
We haven't talked about what we'll do when/if Castle.Core release a new version, whether they keep their supported target frameworks the same or not…

thomaslevesque:
Good point. In the past, we used the lowest possible version. Not sure if this is still a viable strategy...

blairconrad:
(We recently got some community pressure to upgrade!)

blairconrad:
Me, I'd prefer to use the lowest possible version as you note we've done. Clients are free to move to newer ones for whatever reason they might have (for example, production code that relies on a newer version). All this assumes no changes in supported platforms.

It's always possible that bugs or security vulnerabilities will be found in older versions, but those could be addressed on a case-by-case basis, I think.

PR #1936, opened 2023-04-25
Support net6.0 directly, dropping net5.0

blairconrad:
As mentioned in #1935, we intend to support certain legacy TFMs as well as the LTS .net releases.


blairconrad:
Rationalized names of build profiles while I was here. Can split out if desired.

tests/FakeItEasy.Specs/IllegalFluentApiSpecs.cs
```diff
@@ -147,15 +147,8 @@ public static void IllegalSyntax(string illegalOperation, string expectedError,
                 .x(() => diagnostics.Where(d => d.Severity == DiagnosticSeverity.Error).Should().ContainSingle()
                     .Which.GetMessage().Should().Contain(expectedError));
 
-#if MISMATCH_COMPILE_AND_RUNTIME_FRAMEWORKS
-            "And there are warnings due to a mistmach between compile-time and runtime frameworks"
-                .x(() => diagnostics.Where(d => d.Severity != DiagnosticSeverity.Error).Should().OnlyContain(d =>
-                    d.Severity == DiagnosticSeverity.Warning &&
-                    d.Id == "CS1701"));
-#else
```
thomaslevesque, line 155: I don't remember this. Was this an error that only occurred in .NET 6.0?

thomaslevesque (commented):
Thanks @blairconrad!
Looks good to me, but I think we should also update how_to_build.md and README.md (which were also missing updates for the upgrade to net462)

tests/FakeItEasy.Specs/IllegalFluentApiSpecs.cs
blairconrad: Yeah. Because we built with 5.0 and tested with 6.0.

blairconrad:
> Thanks @blairconrad! Looks good to me, but I think we should also update how_to_build.md and README.md (which were also missing updates for the upgrade to net462)

Oh, of course. My search was weak!

blairconrad:
> > Thanks @blairconrad! Looks good to me, but I think we should also update how_to_build.md and README.md (which were also missing updates for the upgrade to net462)
> 
> Oh, of course. My search was weak!

(Updating these is a pain!)

blairconrad:
(I have a typo in the first 2 commit messages, so made their amendments squashes instead of fixups.)

how_to_build.md
```diff
@@ -100,10 +100,10 @@ available:
 
 * `full`: the default profile, builds all supported target frameworks supported
   on the current platform
-* `net462`: builds only the .NET Framework 4.6.2 target framework
+* `net462`: builds and tests only the .NET Framework 4.6.2 target framework
```
thomaslevesque, line 103: Just thought of something. Maybe we should rename this profile to `netfx` or `netframework` or something similar. This way, if we change the .NET Framework version again, we won't have to rename it.

thomaslevesque (approved):
Thanks for the updates! Looks good to me now.
I just have one small suggestion; not sure if it's a good idea, let me know what you think!

how_to_build.md
thomaslevesque: True. Let's leave it as is, then.

thomaslevesque (approved):
Thanks, @blairconrad!

blairconrad:
Thank you, sir.

PR #1938, opened 2023-04-27
Add platform support policy to documentation

### LINKED ISSUE #1935: Publish target framework Support Policy

blairconrad:
Fixes #1935

### DIFF
--- a/docs/platform-support.md
+++ b/docs/platform-support.md
@@ -1,3 +1,32 @@
 # Platform support
 
+## Supported platforms for released versions
+
 See the "Dependencies" section of the [FakeItEasy NuGet Gallery page](https://www.nuget.org/packages/FakeItEasy) to see each released package's supported platforms and package dependencies.
+
+## Platform support policy
+
+Beginning with FakeItEasy 8.0.0, we intend to publish new packages that target
+
+* .NET versions designated as Long Term Support (LTS) and currently supported by Microsoft
+* .NET Standard 2.0 and 2.1[^1]
+* .NET Framework 4.6.2[^2]
+
+See the [.NET Support Policy](https://dotnet.microsoft.com/en-us/platform/support/policy) to learn which target frameworks
+are categorized as "Long Term Support", and when support begins and ends.
+
+We will add builds for new LTS .NET versions as they are released, and remove them from
+new FakeItEasy builds once support is ended.
+
+It's not clear that there's a global consensus on whether the removal of a supported target framework constitutes
+a breaking change and therefore requires a new major version. We choose to limit surprises to FakeItEasy's clients
+and issue a new major release when dropping a supported target framework.
+
+[^1]:
+    There are no explicit plans to remove support for .NET Standard target frameworks, but we may do so if Microsoft
+    explicitly ends support for them, or if supporting them within FakeItEasy becomes too onerous.
+
+[^2]:
+    There are no explicit plans to remove support for .NET Framework, but we may increase the supported
+    version or remove support if Microsoft ends support for .NET Framework, or if supporting it within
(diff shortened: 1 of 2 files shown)


docs/platform-support.md
```diff
@@ -1,3 +1,31 @@
...
+* .NET Framework 4.6.2[^2]
+
+See the [.NET Support Policy](https://dotnet.microsoft.com/en-us/platform/support/policy) to learn which target frameworks
+are categorized as "Long Term Support", and when support begins and ends.
+
+We will add builds for new LTS .NET versions as they are released, and remove them from
+new FakeItEasy builds once support is ended.
+
+The removal of a supported target framework will not be considered a breaking change, and will
+trigger a new minor release, not major.
```
thomaslevesque, line 22: Oh? I thought we were still going to publish those as major.
IIRC, there's nothing in Nuget or `dotnet add package` or the VS UI that prevents you from upgrading to an incompatible version. (I'll need to do some tests to be sure).
So, it might not _technically_ be a breaking, but from an end-user perspective, I think it will be seen as breaking.

mkdocs.yml
```diff
@@ -39,6 +39,7 @@ plugins:
   - search
 
 markdown_extensions:
+  - footnotes
```
thomaslevesque, line 42: Nice!

thomaslevesque (commented):
Thanks @blairconrad!
I have a doubt regarding the minor/major release strategy

docs/platform-support.md
blairconrad: Oh, I thought we fought about it and I got a :+1: on a comment. I may be wrong. Or maybe it wasn't you. 

Regardless, good thing we do code reviews. I can switch to say major. Or wait until we test, if you like (sounds like you're going to do). But even if nuget or `dotnet add` don't keep one from upgrading to an incompatible version, we'd get a surprise, no?

Oh. If people target net6.0 and upgrade to FIE 10.x, which drops, are you worried that they'll then silently start referencing the netstandard2.1 option?

thomaslevesque: > Oh, I thought we fought about it and I got a 👍 on a comment. I may be wrong. Or maybe it wasn't you.

Ah, yes, I saw the comment. It wasn't me!

> Or wait until we test, if you like (sounds like you're going to do)

I'll try to do it, as soon as I find the time.

> But even if nuget or dotnet add don't keep one from upgrading to an incompatible version, we'd get a surprise, no?

I'm not sure if the upgrade will fail, but at least the build will.
Anyway, the way I see it, we should always be able to upgrade to a newer minor version. I would be surprised if I couldn't. But maybe my interpretation of SemVer is wrong...

> Oh. If people target net6.0 and upgrade to FIE 10.x, which drops, are you worried that they'll then silently start referencing the netstandard2.1 option?

I had not thought of this. It's definitely a problem, as they will suddenly start pulling a much larger dependency graph (although I thought that .NET Standard 2.x solved this?). And some features might become unavailable, which will cause the build to fail if they were using these features. But even if none of those things happen, silently switching to an older TFM is certainly not ideal....

thomaslevesque (approved):
LGTM, thanks!

thomaslevesque (approved):

PR #1939, opened 2023-04-28
Publish docs preview

thomaslevesque:
Add workflow to publish docs preview when the `when the 👀 publish-docs-preview label` is added to a PR

### DIFF
--- a/.github/workflows/publish-docs-preview.yml
+++ b/.github/workflows/publish-docs-preview.yml
@@ -1,0 +1,58 @@
+name: Publish docs preview
+on:
+  pull_request_target:
+    types: labeled
+
+jobs:
+
+  publish-docs-preview:
+    name: Publish docs preview
+    runs-on: ubuntu-22.04
+    if: "${{ github.event.label.name == ':eyes: publish-docs-preview' }}"
+    steps:
+
+      - name: Download artifact
+        id: download-artifact
+        uses: dawidd6/action-download-artifact@v2
+        continue-on-error: true
+        with:
+          workflow: ci.yml
+          pr: ${{github.event.pull_request.number}}
+          name: documentation
+          path: documentation/
+
+      - name: Publish files
+        id: publish-files
+        if: ${{ steps.download-artifact.outcome == 'success' }}
+        continue-on-error: true
+        env:
+          AZURE_STORAGE_CONNECTION_STRING: ${{ secrets.DOCS_PREVIEW_STORAGE_CONNECTION_STRING }}
+        run: |
+          set -euo pipefail
+          destinationPath=docspreview/pr${{ github.event.pull_request.number }}
+          az storage blob upload-batch --destination '$web' --destination-path $destinationPath --source artifacts/docs --overwrite true
+          echo Successfully uploaded docs preview to $DOCS_PREVIEW_BASE_URL/$destinationPath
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docs-preview.yml
```diff
@@ -0,0 +1,58 @@
+name: Publish docs preview
+on:
+  pull_request_target:
```
thomaslevesque, line 3: Using `pull_request_target` instead of `pull_request` so that I can access the secrets.
The code isn't checked out anyway.

.github/workflows/publish-docs-preview.yml
```diff
@@ -0,0 +1,58 @@
...
+on:
+  pull_request_target:
+    types: labeled
+
+jobs:
+
+  publish-docs-preview:
+    name: Publish docs preview
+    runs-on: ubuntu-22.04
+    if: ${{ github.event.label.name ==  ':eyes: publish-docs-preview' }}
```
thomaslevesque: Not sure if I must use the shortcode or the real emoji character here. I entered the name with the shortcode, but it might be converted... we'll see.

.github/workflows/publish-docs-preview.yml
```diff
@@ -0,0 +1,58 @@
...
+    steps:
+
+      - name: Download artifact
+        id: download-artifact
+        uses: dawidd6/action-download-artifact@v2
+        continue-on-error: true
```
thomaslevesque, line 17: Using `continue-on-error` so that I can still publish a comment to indicate the result.

.github/workflows/publish-docs-preview.yml
```diff
@@ -0,0 +1,58 @@
...
+        if: ${{ steps.download-artifact.outcome == 'success' }}
+        continue-on-error: true
+        env:
+          AZURE_STORAGE_CONNECTION_STRING: ${{ secrets.DOCS_PREVIEW_STORAGE_CONNECTION_STRING }}
+        run: |
+          set -euo pipefail
+          destinationPath=docspreview/pr${{ github.event.pull_request.number }}
+          az storage blob upload-batch --destination '$web' --destination-path $destinationPath --source artifacts/docs --overwrite true
+          echo Successfully uploaded docs preview to $DOCS_PREVIEW_BASE_URL/$destinationPath
+          echo "docs-preview-url=$DOCS_PREVIEW_BASE_URL/$destinationPath" >> $GITHUB_OUTPUT
```
thomaslevesque, line 35: `DOCS_PREVIEW_BASE_URL` is a new repository variable (basically the same as secrets, except the value isn't a secret)

.github/workflows/publish-docs-preview.yml
```diff
@@ -0,0 +1,58 @@
...
+on:
+  pull_request_target:
+    types: labeled
+
+jobs:
+
+  publish-docs-preview:
+    name: Publish docs preview
+    runs-on: ubuntu-22.04
+    if: "${{ github.event.label.name == ':eyes: publish-docs-preview' }}"
```
thomaslevesque, line 11: Had to quote the expression here, otherwise the `:` in the string literal is a syntax error... I hate YAML 😠

thomaslevesque (commented):
Unfortunately I can't test it before it's merged...

blairconrad (approved):

blairconrad:
Worth a shot! Thanks, @thomaslevesque

PR #1949, opened 2023-06-27
Enable fake creation via constructors with in parameters

### LINKED ISSUE #1948: Failure to create fake via constructor with 'in' parameter
FakeItEasy will not attempt to create a fake via any constructor with an `in` parameter. Instead, it will give an error indicating that the `in` parameter Type "could not be resolved".

Here is an example test that does not work.

```csharp
public class ClassWithConstructorInParameter
{
	public ClassWithConstructorInParameter(in DateTimeOffset dateTimeOffset) { }
}

[TestMethod]
public void CreateFakeViaConstructorWithInParameter()
{
	A.Fake<ClassWithConstructorInParameter>();
}
```

I will submit a pull request shortly with a suggested fix.

mriehm:
Fixes #1948

### DIFF
--- a/tests/FakeItEasy.Specs/CreationSpecs.cs
+++ b/tests/FakeItEasy.Specs/CreationSpecs.cs
@@ -439,6 +439,32 @@ namespace FakeItEasy.Specs
         }
 
         [Scenario]
+        public void ClassWithConstructorRefParameterCanBeFaked(ClassWithConstructorRefParameter fake)
+        {
+            "Given a class with a constructor with a ref parameter"
+                .See<ClassWithConstructorRefParameter>();
+
+            "When I create a fake of the class"
+                .x(() => fake = this.CreateFake<ClassWithConstructorRefParameter>());
+
+            "Then the fake is created"
+                .x(() => fake.Should().BeAFake());
+        }
+
+        [Scenario]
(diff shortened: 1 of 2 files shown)


tests/FakeItEasy.Specs/CreationSpecs.cs
```diff
@@ -425,6 +425,19 @@ public void ClassWithPrivateConstructorCannotBeFaked(Exception exception)
...
+        {
+            "Given a class with a constructor with an in parameter"
+                .See<ClassWithConstructorInParameter>();
+
+            "When I create a fake of the class"
+                .x(() => fake = this.CreateFake<ClassWithConstructorInParameter>());
+
+            "Then the fake is created"
+                .x(() => fake.Should().BeAFake());
+        }
```
thomaslevesque, line 439: Could you please add tests with `ref` and `out` parameters as well? Just for completeness.

thomaslevesque (commented):
Thanks for the PR @mriehm.
I just have one request to add coverage for `ref` and `out` parameters as well. I think your fix works for those too, but we should make sure.

thomaslevesque (approved):
Thanks for the update, @mriehm! Looks good to me.
@blairconrad, any objection to merging this?

blairconrad (approved):

blairconrad:
Thanks, @mriehm. Nice addition.

Issue #1952, opened 2023-07-10
Add registry of argument comparers


blairconrad:
Inspired by #1951 (thanks, @drauch)

## The problem

The client routinely wants to match arguments using a comparator other than `object.Equals`, and finds it inconvenient to do so using either `A<T>.That.Matches(arg => arg.Value == expected.Value)` or `A<T>.That.IsEqualTo(expected, comparer)`.

## Suggested solution

We already have a mechanism for [registering argument formatters](https://fakeiteasy.github.io/docs/7.4.0/formatting-argument-values/), so we could implement a similar system for _comparing_ arguments. The default action could use `EqualityComparer<T>.Default` and clients could register their own.

We could use the same "distance between types, and priority indicator for tiebreaks" system to find an appropriate comparer in each case, or investigate another paradigm.

PR #1961, opened 2023-08-08
Choose equality comparer based on example type instead of parameter type

blairconrad:
Continues #1952.

### DIFF
--- a/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
+++ b/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
@@ -7,6 +7,7 @@ namespace FakeItEasy
     using System.Linq;
     using System.Linq.Expressions;
     using System.Threading;
+    using FakeItEasy.Expressions.ArgumentConstraints;
 
     /// <summary>
     /// Provides validation extensions for <see cref="IArgumentConstraintManager{T}"/>.
@@ -23,7 +24,7 @@ namespace FakeItEasy
         {
             Guard.AgainstNull(manager);
 
-            return manager.Matches(x => x is null, x => x.Write("NULL"));
+            return manager.Matches(NullArgumentConstraint.Instance.IsValid, NullArgumentConstraint.Instance.ConstraintDescription);
         }
 
         /// <summary>
@@ -37,7 +38,7 @@ namespace FakeItEasy
         {
             Guard.AgainstNull(manager);
 
-            return manager.Matches(x => x is null, x => x.Write("NULL"));
+            return manager.Matches(x => NullArgumentConstraint.Instance.IsValid(x), NullArgumentConstraint.Instance.ConstraintDescription);
         }
 
         /// <summary>
--- a/src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
+++ b/src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
@@ -8,12 +8,10 @@ namespace FakeItEasy.Expressions.ArgumentConstraints
         : IArgumentConstraint
     {
         private readonly object expectedValue;
-        private readonly Type parameterType;
 
-        public EqualityArgumentConstraint(object expectedValue, Type parameterType)
+        public EqualityArgumentConstraint(object expectedValue)
         {
             this.expectedValue = expectedValue;
-            this.parameterType = parameterType;
         }
 
         public string ConstraintDescription => this.ToString();
@@ -26,7 +24,7 @@ namespace FakeItEasy.Expressions.ArgumentConstraints
             }
 
             var argumentEqualityComparer = ServiceLocator.Resolve<ArgumentEqualityComparer>();
-            return argumentEqualityComparer.AreEqual(this.expectedValue, argument, this.parameterType);
+            return argumentEqualityComparer.AreEqual(this.expectedValue, argument);
         }
 
         [SuppressMessage("Microsoft.Design", "CA1031:DoNotCatchGeneralExceptionTypes", Justification = "Any type of exception may be encountered.")]
--- a/src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
+++ b/src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
@@ -7,10 +7,10 @@ namespace FakeItEasy.Expressions.ArgumentConstraints
     internal class EqualityArgumentConstraint
         : IArgumentConstraint
     {
-        private readonly object? expectedValue;
+        private readonly object expectedValue;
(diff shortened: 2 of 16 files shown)


src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
```diff
@@ -33,12 +31,11 @@ public override string ToString()
                 writer.WriteArgumentValue(this.expectedValue);
                 return writer.Builder.ToString();
             }
-            catch (Exception ex) when (!(ex is UserCallbackException))
+            catch (Exception ex) when (ex is not UserCallbackException)
```
blairconrad, line 34: Technically a bonus change, but I was right here. Will revert if you like.

blairconrad (commented):
Ah, I should make the expected value in the `ArgumentEqualityComparer` non-nullable.

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
```diff
@@ -37,7 +38,7 @@ public static class ArgumentConstraintManagerExtensions
         {
             Guard.AgainstNull(manager);
 
-            return manager.Matches(x => x is null, x => x.Write("NULL"));
+            return manager.Matches(x => NullArgumentConstraint.Instance.IsValid(x), NullArgumentConstraint.Instance.ConstraintDescription);
```
thomaslevesque, line 41: Why use a lambda instead of the method group here? (like you did for `IsNull`)

src/FakeItEasy/Expressions/ArgumentConstraints/EqualityArgumentConstraint.cs
thomaslevesque: No, it's fine!

src/FakeItEasy/IArgumentEqualityComparer.cs
```diff
@@ -19,7 +19,7 @@ public interface IArgumentEqualityComparer
         /// <param name="type">The type of the objects to compare.</param>
         /// <returns>
         /// <c>true</c> if this object can compare instances of <paramref name="type"/>. Otherwise <c>false</c>.
-        /// </returns>
+        /// </returns>g
```
thomaslevesque, line 22: I assume this wasn't intentional 😉

tests/FakeItEasy.Specs/ArgumentEqualityComparerSpecs.cs
```diff
@@ -93,6 +147,16 @@ public class CustomComparer : ArgumentEqualityComparer<ClassWithCustomArgumentEq
             }
         }
 
+        public class CustomComparerThatIsNotNullSafe : ArgumentEqualityComparer<ClassWithCustomArgumentEqualityComparerThatIsNotNullSafe>
+        {
+            protected override bool AreEqual(
+                ClassWithCustomArgumentEqualityComparerThatIsNotNullSafe? expectedValue,
```
thomaslevesque, line 153: Doesn't this produce a nullability warning? The argument is not nullable in the base class.

thomaslevesque (commented):
I like the approach you used with `NullArgumentConstraint`.
I only have a few minor comments/questions

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
blairconrad: Compiler hated the lambda.

src/FakeItEasy/IArgumentEqualityComparer.cs
blairconrad: it never it

tests/FakeItEasy.Specs/ArgumentEqualityComparerSpecs.cs
blairconrad: I didn't notice, but I do have some nullability changes coming.

blairconrad:
Thanks for the comments, @thomaslevesque. I appreciate them as always.
As it turns out, we may need to make some larger changes.

In `ArgumentEqualityComparer<T>`, we cast to `<T>` regardless of whether the received argument is a `T`. This blew up one of my future tests (thank goodness). `argumentValue` might be `null` or a different type (since the parameter type might be `object` (or some other type, but `object` is easy to talk about) and we're now basing the comparison off the type of the example object and casting accordingly. 

Ways forward with `AreEqual(object expectedValue, object? argumentValue)`, which I'm exploring:

1. double-typeparameterizing ArgumentEqualityComparer: `ArgumentEqualityComparer<TExpected, TArgument>`. More flexible. Could keep `ArgumentEqualityComparer<T>`, inheriting `ArgumentEqualityComparer<T, T>`. I think this is nice and flexible and overkill. It also opens up the can of worms about the comparison with the types _flipped_ and whether we should implement something to make that more convenient (I don't think so).
2. not converting `argumentValue` to a `T?`, which I think is a little gross since this class is all about making the simple comparison really easy. Why make the client do the work?
3. returning false when `argumentValue` is non-`null` and not a `T`. Let people implement the interface if they want to compare against non-`T`s. Now the client only has to worry about `null` argument values.
4. returning false when `argumentValue` not a `T`. Let people implement the interface if they want to compare against non-`T`s. Now the client 
5. returning false when `argumentValue` not a `T`. Also, refraining from calling a custom equality comparer _ever_ when at least one of `expectedValue` and `argumentValue` is `null`. Don't even let the interface people see this case. It does restrict clients _a bit_, but we already cut them off if `expectedValue` is `null`. They can go the `That` route, building up their own comparison, if they like `null`s so much.

I'm learning toward the last one, in case you didn't get that. But happy to hear what you think!

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
thomaslevesque: Sorry, I'm slow today. Could you elaborate?

thomaslevesque:
> 5\. returning false when `argumentValue` not a `T`. Also, refraining from calling a custom equality comparer _ever_ when at least one of `expectedValue` and `argumentValue` is `null`. Don't even let the interface people see this case. It does restrict clients _a bit_, but we already cut them off if `expectedValue` is `null`. They can go the `That` route, building up their own comparison, if they like `null`s so much.

I like this option too!

blairconrad:
> I like this option too!

Great! I may just go and amend the commits. Sorry.

thomaslevesque (approved):
Very nice. I have no comments!
I see it's still WIP. Is it ready to merge, or do you intend to push more changes?

blairconrad:
Whoops. Forgot to un-WIP it. Thanks!

thomaslevesque:
Thanks @blairconrad!

PR #1966, opened 2023-09-20
Nesting or compounding An-built constraints throws wrong exception

blairconrad:
```c#
A.CallTo(() => fake.Bar(An<int>.That.Matches(i => i % 2 == 0) +
                        An<int>.That.Matches(i => i % 2 == 1))).Returns(1));
```

yields

> FakeConfigurationException : Too many argument constraints specified. First superfluous constraint is <i => ((i % 2) == 1)>

but

```c#
A.CallTo(() => fake.Bar(A<int>.That.Matches(i => i % 2 == 0) +
                        A<int>.That.Matches(i => i % 2 == 1))).Returns(1));
```

yields

> InvalidOperationException : An argument constraint, such as That, Ignored, or _, cannot be nested in an argument.

The latter is technically more correct, since we're nesting the two constraints inside the `+` expression.

But we already have a test to detect multiple constraints, so we'd be better off with one that explicitly tests single nested constraints.

### DIFF
--- a/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -818,16 +818,37 @@ namespace FakeItEasy.Specs
         }
 
         [Scenario]
-        public static void TwoArgumentConstraints(
+        public static void NestedAArgumentConstraint(
             IHaveAnObjectParameter fake,
             Exception exception)
         {
             "Given a fake"
                 .x(() => fake = A.Fake<IHaveAnObjectParameter>());
 
-            "When I try to configure a method of the fake using two constraints on an argument"
+            "When I try to configure a method of the fake using a nested A-built constraint on an argument"
                 .x(() => exception = Record.Exception(() => A.CallTo(() =>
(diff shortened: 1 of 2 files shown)


tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -818,16 +818,37 @@ public static void UnusedArgumentMatcherDescriptionNotUsed(IHaveNoGenericParamet
...
+            "Then the call configuration throws an InvalidOperationException"
+                .x(() => exception.Should().BeAnExceptionOfType<InvalidOperationException>());
+
+            "And the exception indicates that a nested argument constraint was used"
+                .x(() => exception.Message.Should()
+                    .Be("An argument constraint, such as That, Ignored, or _, cannot be nested in an argument."));
+        }
+
+        [Scenario]
+        public static void NestedAnArgumentConstraint(
```
blairconrad, line 841: Did not throw before detector was updated.

thomaslevesque (approved):
Thanks @blairconrad, nice catch!

blairconrad:
Oh, thanks, @thomaslevesque!

PR #1967, opened 2023-09-22
Provide a mechanism for capturing arguments passed to Fakes

### LINKED ISSUE #1950: Provide a mechanism for capturing arguments passed to Fakes
## The Problem

It's possible to verify that arguments passed to Fake calls meet certain [argument constraints](https://fakeiteasy.github.io/docs/7.4.0/argument-constraints/), but these are not always rich enough for clients, who may want to capture arguments that are provided and then perform complex examination of them.

In the past, we've recommended [invoking a custom method](https://fakeiteasy.github.io/docs/7.4.0/invoking-custom-code/) to place the captured value in a variable. Something like this:

```c#
string capturedArgument;

var fake = A.Fake<IService>();

A.CallTo(() => fake.SomeMethod(A<string>._)).Invokes(call => capturedArgument = call.Get(0));

// exercise SUT

// verify capturedArgument
```

Unfortunately, this solution is not obvious to many clients.

## The Goal

We'd like to have a first-class mechanism for capturing arguments.
Ideally it would

1. be discoverable
2. not disrupt existing syntax too much
3. have a pleasing syntax
4. support capturing values from multiple arguments
5. support capturing values from multiple calls to a Fake's method
6. respect any [argument constraints](https://fakeiteasy.github.io/docs/7.4.0/argument-constraints/) applied to the argument
7. not capture methods when a call doesn't match preconfigured rules, including rules that have been activated their maximum number of times, or that don't apply due to constraints on the current argument, other arguments, or any other constraints on the method invoked

## Very rough proposal for API design (subject to change as we investigate)

```c#
var capturedArg = new CapturedArgument<int>();

A.CallTo(() => fake.Method(A<int>.That.IsCapturedTo(capturedArg).Ignored).DoesNothing();

// or
A.CallTo(() => fake.Method(A<int>.That.IsCapturedTo(capturedArg).And.Matches(() => int % 2 == 0).DoesNothing();

// …

verify(capturedArg.Values);

// or 

verify(capturedArg.LastValue); // throws if no value

```

blairconrad:
Fixes #1950.

### DIFF
--- a/docs/capturing-arguments.md
+++ b/docs/capturing-arguments.md
@@ -1,0 +1,109 @@
+# Capturing arguments
+
+Often tests will make [assertions against the calls made to a fake object](assertion.md)
+to verify the behavior of the system under test, but you can also capture arguments
+supplied in a call to a fake object and examine the values later.
+
+## Simple capture
+
+Take this contrived example:
+
+```csharp title='"production code"'
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:IListLogger
+
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:Calculator
+--8<--
+```
+
+Suppose we want to make sure that we send correct messages to the logger.
+We can record the first argument to each `Log` call in a variable of type
+`Captured` and verify the values in the "assert" phase of the test.
+
+```csharp title="simple capture" linenums="1" hl_lines="2 6 17 22"
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:SimpleCapture
+--8<--
+```
+
+This is a fairly standard test with fakes, except we:
+
+* create a `Captured` object to store received argument values
+* use the `Captured` object's `_` member to configure the call to capture any values for that argument
+  (this is analogous to the `A<T>._` member, and just like it, there is a matching `Ignored` member in
+  case you prefer that name)
+* use `Values` to access all the captured values so they can be asserted
+* (alternative flow) use `GetLastValue` to access the most recent captured value so it can be asserted.
+  This will throw a `FakeItEasy.ExpectationException` if no values have been captured.
+
+???+ note "Values are only captured if the call matches the configuration"
+    When a call configuration intends to capture one or more arguments, the argument
+    values are only captured if this specific call configuration is triggered. If an incoming call
+    does not match that configured for the method or property, no arguments are captured.
+
+## Capturing and constraining at the same time
+
+Even though you can interrogate captured values after the fact, you may want to configure
+fake behavior to take effect only when incoming arguments meet a constraint. As you might have guessed
+based on the references to `_` and `Ignored` above, you can do this using a `That` method just like the
+one supplied by the [`A<T>` argument constraints](argument-constraints.md):
+
+```csharp title="constrained capture" linenums="1" hl_lines="6"
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:ConstrainedCapture
+--8<--
+```
+
+## Capturing mutable arguments
+
+### The challenge
+
+Argument capture is shallow; there's no copying of object state.
+If a reference-based argument (e.g. a class instance, not a struct) is captured and
+subsequently modified by the test or production code, the "assert" phase of the test
+will see the updated state.
+
+```csharp title="capturing mutating values" linenums="1" hl_lines="2 7 13 14 15 16 22 23"
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:NaivelyCaptureMutatedList
+--8<--
+```
+
+Here a single list instance is passed to the production code twice, but an element is removed
+between the calls. The `Captured` object captures a reference to the list each
+time, but does not preserve the internals of the list. So by the time the `Values` are examined,
+the list has had its first element removed, and this is reflected in the failing assertion.
+
+It's the same effect as running
+
+```c#
+List<int> list = [1, 2, 3, 4];
+var a = list;
+list.RemoveAt(0);
+var b = list;
+
+// both a and b point to a list with elements { 2, 3, 4 }
+```
+
+### Capturing frozen state
+
+`Captured` objects can be created with a transforming function (or "freezer") that runs on the
+argument value before saving it away, thus insulating the captured values
+from subsequent mutations.
+
+```csharp title="freezing state of captured values" linenums="1" hl_lines="2 3 20"
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:CaptureCopiedMutatedList
+--8<--
+```
+
+You can even transform values into a different type:
+
+```csharp title="freezing state of captured values as new type" linenums="1" hl_lines="2 3 20 21"
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:CaptureCopiedMutatedListToNewType
+--8<--
+```
+
+In this case we supply an updated freezer function as well as the `Captured` typeparams
+for both the argument value and the captured value types.
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -73,6 +73,7 @@ nav:
 - Assertion:
   - Assertion: assertion.md
   - Ordered assertions: ordered-assertions.md
+- Capturing arguments: capturing-arguments.md
 - Faking async methods: faking-async-methods.md
 - Raising events: raising-events.md
 - Strict fakes: strict-fakes.md
--- a/recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
+++ b/recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
@@ -1,0 +1,184 @@
+namespace FakeItEasy.Recipes.CSharp;
+
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Xunit;
+
+public class CapturingArguments
+{
+    //// --8<-- [start:IListLogger]
+    public interface IListLogger
+    {
+        void Log(string message, IEnumerable<int> list);
+    }
+    //// --8<-- [end:IListLogger]
+
+    [Fact]
+    public void OperationsShouldLog()
+    {
+        //// --8<-- [start:SimpleCapture]
+        // Arrange
+        var capturedMessage = A.Captured<string>();
+
+        var logger = A.Fake<IListLogger>();
+        A.CallTo(() => logger.Log(
+                capturedMessage._,
+                An<IEnumerable<int>>._))
+            .DoesNothing();
+
+        var calculator = new Calculator(logger);
+
+        // Act
+        calculator.Add([1, 2, 3, 4]);
+        calculator.Square(7);
+
+        // Assert
+        capturedMessage.Values.Should().Equal(
+            "about to add",
+            "about to square");
+
+        // Or if you only care about the last value:
+        capturedMessage.GetLastValue().Should().Be(
+            "about to square");
+        //// --8<-- [end:SimpleCapture]
+    }
+
+    [Fact]
+    public void CaptureInputThatContainsThree()
+    {
+        //// --8<-- [start:ConstrainedCapture]
+        // Arrange
+        var capturedMessage = A.Captured<string>();
+
+        var logger = A.Fake<IListLogger>();
+        A.CallTo(() => logger.Log(
+                capturedMessage.That.Contains("add"),
+                An<IEnumerable<int>>._))
+            .DoesNothing();
+
+        var calculator = new Calculator(logger);
+
+        // Act
+        calculator.Add([1, 2, 3, 4]);
+        calculator.Square(7);
+        calculator.Add([8, 9]);
+
+        // Assert
+        capturedMessage.Values.Should().Equal(
+            "about to add",
+            "about to add");
+        //// --8<-- [end:ConstrainedCapture]
+    }
+
+    [Fact]
+    public void NaivelyCaptureMutatedList()
+    {
+        var test = () =>
+        {
+            //// --8<-- [start:NaivelyCaptureMutatedList]
+            // Arrange
+            var capturedOperands = A.Captured<IEnumerable<int>>();
+
+            var logger = A.Fake<IListLogger>();
+            A.CallTo(() => logger.Log(
+                    "about to add",
+                    capturedOperands._))
+                .DoesNothing();
+
+            var calculator = new Calculator(logger);
+
+            // Act
+            List<int> operands = [1, 2, 3, 4];
+            calculator.Add(operands); // capturedOperands captures operands
+            operands.RemoveAt(0);
+            calculator.Add(operands); // captures operands again - same instance
+
+            // Assert
+            // passes:
+            capturedOperands.Values[1].Should().Equal(2, 3, 4);
+
+            // fails - operands contains only 2, 3, and 4:
+            capturedOperands.Values[0].Should().Equal(1, 2, 3, 4);
+            //// --8<-- [end:NaivelyCaptureMutatedList]
+        };
+
+        test.Should().ThrowExactly<AssertionFailedException>()
+            .WithMessage("Expected * to be equal to {1, 2, 3, 4}, but {2, 3, 4} contains 1 item(s) less.");
+    }
+
+    [Fact]
+    public void CaptureCopiedMutatedList()
+    {
+        //// --8<-- [start:CaptureCopiedMutatedList]
+        // Arrange
+        var capturedOperands =
+            A.Captured<IEnumerable<int>>().FrozenBy(l => l.ToList());
+
+        var logger = A.Fake<IListLogger>();
+        A.CallTo(() => logger.Log(
+                "about to add",
+                capturedOperands._))
+            .DoesNothing();
+
+        var calculator = new Calculator(logger);
+
+        // Act
+        List<int> operands = [1, 2, 3, 4];
+        calculator.Add(operands); // capturedOperands captures copy of operands
+        operands.RemoveAt(0);
+        calculator.Add(operands); // capturedOperands captures copy of operands
+
+        // Assert
+        // both pass:
+        capturedOperands.Values[1].Should().Equal(2, 3, 4);
+        capturedOperands.Values[0].Should().Equal(1, 2, 3, 4);
+        //// --8<-- [end:CaptureCopiedMutatedList]
+    }
+
+    [Fact]
+    public void CaptureCopiedMutatedListToNewType()
+    {
+        //// --8<-- [start:CaptureCopiedMutatedListToNewType]
+        // Arrange
+        var capturedOperands =
+            A.Captured<IEnumerable<int>>().FrozenBy(l => string.Join(" + ", l));
+
+        var logger = A.Fake<IListLogger>();
+        A.CallTo(() => logger.Log(
+                "about to add",
+                capturedOperands._))
+            .DoesNothing();
+
+        var calculator = new Calculator(logger);
+
+        // Act
+        List<int> operands = [1, 2, 3, 4];
+        calculator.Add(operands); // capturedOperands captures transformed operands
+        operands.RemoveAt(0);
+        calculator.Add(operands); // capturedOperands captures transformed operands
+
+        // Assert
+        capturedOperands.Values[1].Should().Be("2 + 3 + 4");
+        capturedOperands.Values[0].Should().Be("1 + 2 + 3 + 4");
+        //// --8<-- [end:CaptureCopiedMutatedListToNewType]
+    }
+
+    //// --8<-- [start:Calculator]
+    internal class Calculator(IListLogger logger)
+    {
+        public int Add(IList<int> operands)
+        {
+            logger.Log("about to add", operands);
+            return operands.Sum();
+        }
+
+        public int Square(int input)
+        {
+            logger.Log("about to square", [input]);
+            return input * input;
+        }
+    }
+    //// --8<-- [end:Calculator]
+}
--- a/src/FakeItEasy/Captured.of.TArgument.TCapture.cs
+++ b/src/FakeItEasy/Captured.of.TArgument.TCapture.cs
@@ -1,0 +1,87 @@
+namespace FakeItEasy
+{
+    using System;
+    using System.Collections.Generic;
+    using System.ComponentModel;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
+
+    /// <summary>
+    /// Holds argument values captured from calls to a faked member.
+    /// </summary>
+    /// <typeparam name="TArgument">The type of the argument.</typeparam>
+    /// <typeparam name="TCapture">The type of the value to capture from the argument.</typeparam>
+    public class Captured<TArgument, TCapture>
+    {
+        private readonly List<TCapture> values = [];
+        private readonly Func<TArgument, TCapture> freezer;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="Captured{TArgument, TCapture}"/> class.
+        /// </summary>
+        /// <param name="freezer">
+        ///   Transforms incoming argument values before storing the result.
+        ///   Useful when argument values may be mutated between calls and you want
+        ///   to store a copy that will not be modified.
+        /// </param>
+        internal Captured(Func<TArgument, TCapture> freezer) => this.freezer = Guard.AgainstNull(freezer);
+
+        /// <summary>
+        /// Gets an argument constraint object that will be used to constrain this captured method call argument.
+        /// </summary>
+        public INegatableArgumentConstraintManager<TArgument> That =>
+            ServiceLocator.Resolve<IArgumentConstraintManagerFactory>().Create<TArgument>().IsCapturedTo(this);
+
+        /// <summary>
+        /// Gets a constraint that considers any value of this captured argument as valid.
+        /// </summary>
+        /// <remarks>This is a shortcut for the "Ignored"-property.</remarks>
+        [SuppressMessage(
+            "Microsoft.Naming",
+            "CA1707:IdentifiersShouldNotContainUnderscores",
+            Justification = "But it's kinda cool right?")]
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [CLSCompliant(false)]
+#pragma warning disable SA1300 // Element must begin with upper-case letter
+        public TArgument _ => this.Ignored;
+#pragma warning restore SA1300 // Element must begin with upper-case letter
+
+        /// <summary>
+        /// Gets a constraint that considers any value of this captured argument as valid.
+        /// </summary>
+        public TArgument Ignored => this.That.Matches(x => true, x => x.Write(nameof(this.Ignored)));
+
+        /// <summary>
+        /// Gets the captured values.
+        /// </summary>
+        /// <remarks>
+        /// Will be empty if no arguments were captured.
+        /// </remarks>
+        public IReadOnlyList<TCapture> Values => this.values;
+
+        /// <summary>
+        /// Gets the last captured value.
+        /// </summary>
+        /// <returns>
+        /// The last captured value.
+        /// </returns>
+        /// <exception cref="ExpectationException">
+        /// If no arguments were captured.
+        /// </exception>
+        public TCapture GetLastValue()
+        {
+            if (this.values.Any())
+            {
+                return this.values.Last();
+            }
+
+            throw new ExpectationException(ExceptionMessages.NoCapturedValues);
+        }
+
+        /// <summary>
+        /// Add a captured value.
+        /// </summary>
+        /// <param name="value">The new value to capture.</param>
+        internal void Add(TArgument value) => this.values.Add(this.freezer(value));
+    }
+}
--- a/src/FakeItEasy/Configuration/BuildableCallRule.cs
+++ b/src/FakeItEasy/Configuration/BuildableCallRule.cs
@@ -19,24 +19,19 @@ namespace FakeItEasy.Configuration
             call.SetReturnValue(call.GetDefaultReturnValue());
 
         private readonly List<WherePredicate> wherePredicates;
+        private readonly ICollection<Action<IFakeObjectCall>> actions;
         private Action<IInterceptedFakeObjectCall> applicator;
         private bool wasApplicatorSet;
         private bool canSetOutAndRefParametersValueProducer;
 
         protected BuildableCallRule()
         {
-            this.Actions = new LinkedList<Action<IFakeObjectCall>>();
+            this.actions = new LinkedList<Action<IFakeObjectCall>>();
             this.applicator = DefaultApplicator;
             this.OutAndRefParametersValueProducer = DefaultOutAndRefParametersValueProducer;
             this.canSetOutAndRefParametersValueProducer = true;
             this.wherePredicates = new List<WherePredicate>();
         }
-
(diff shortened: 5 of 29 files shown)


tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
```diff
@@ -0,0 +1,339 @@
...
+
+            "And a capturedArgument instance"
+                .x(() => capturedArgument = new CapturedArgument<int>());
+
+            "And a fake method is configured to capture arguments except those bigger than 10"
+                .x(() => A.CallTo(() =>
+                        fake.Invoke(A<int>.That.IsCapturedTo(capturedArgument).And.Not.IsGreaterThan(10)))
+                    .DoesNothing());
+
+            "When the method is called with a view values"
```
thomaslevesque, line 200:
```suggestion
            "When the method is called with a few values"

```

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
```diff
@@ -0,0 +1,339 @@
...
+
+        [Scenario]
+        public static void CaptureMutatingArgument(
+            Action<IList<int>> fake, IList<int> listOfInts, CapturedArgument<IList<int>> capturedArgument)
+        {
+            "Given a fake"
+                .x(() => fake = A.Fake<Action<IList<int>>>());
+
+            "And a capturedArgument instance"
+                .x(() => capturedArgument = new CapturedArgument<IList<int>>(l => new List<int>(l!)));
```
thomaslevesque, line 286: Not that it changes anything, but why `new List<int>(l)` instead of `l.ToList()`?

thomaslevesque (commented):
I have a few questions and remarks

src/FakeItEasy/Configuration/BuildableCallRule.cs
```diff
@@ -19,25 +19,20 @@ internal abstract class BuildableCallRule
...
             this.OutAndRefParametersValueProducer = DefaultOutAndRefParametersValueProducer;
             this.canSetOutAndRefParametersValueProducer = true;
             this.wherePredicates = new List<WherePredicate>();
         }
 
-        /// <summary>
-        /// Gets a collection of actions that should be invoked when the configured
-        /// call is made.
-        /// </summary>
-        public virtual ICollection<Action<IFakeObjectCall>> Actions { get; }
```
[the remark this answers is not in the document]
blairconrad: It's true. It's all us. Today. Happy to revert if you like.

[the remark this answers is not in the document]
thomaslevesque: Well, I guess it doesn't hurt to improve encapsulation, even if it's internal...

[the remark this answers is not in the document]
blairconrad: Oh, you wrote back! I was reverting but will stop now!

src/FakeItEasy/Guard.cs
```diff
@@ -13,16 +13,22 @@ internal static class Guard
...
+        public static T AgainstNull<T>(
+            [ValidatedNotNull, NotNull] T argument,
+            [CallerArgumentExpression(nameof(argument))] string argumentName = null!)
         {
             if (argument is null)
             {
                 throw new ArgumentNullException(argumentName);
             }
+
+            return argument;
```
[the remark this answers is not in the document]
thomaslevesque: Keep it! It's a good improvement.

src/FakeItEasy/A.of.T.cs
```diff
@@ -9,13 +9,14 @@ namespace FakeItEasy
     /// </summary>
     /// <typeparam name="T">The type of argument to validate.</typeparam>
     /// <remarks>If desired, <see cref="An{T}"/> may be used to specify constraints on types whose names begin with a vowel sound.</remarks>
+    [CLSCompliant(false)]
```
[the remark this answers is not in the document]
thomaslevesque: Ah. So annoying... Has anyone ever really cared about CLS compliance?

src/FakeItEasy/Guard.cs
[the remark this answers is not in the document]
blairconrad: Cool. I was too lazy to go through and update all the usages. Maybe a separate PR if I'm feeling frisky.

src/FakeItEasy/A.of.T.cs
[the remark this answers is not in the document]
blairconrad: VB people? I mean, if I wrote a `That` and a `that` method, it could mess them up?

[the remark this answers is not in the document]
thomaslevesque: > VB people?

Who? Never heard of those 😆
(but yeah, it's the main reason, I guess)

docs/capturing-arguments.md
```diff
@@ -0,0 +1,107 @@
...
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:IListLogger
+
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:Popper
+--8<--
+```
+
+While there are other ways to ensure that a particular list is passed
+to a method, we'll pretend there aren't and capture the second argument
+to `log` in a `CapturedArgument`, then verify it.
```
thomaslevesque, line 23: This is a bit confusing, I had to read it twice to understand what it was about.
I think it's because it's written with the idea of "working around limitations of argument constraints" in mind, which isn't the use case _I_ have in mind when I think about argument capture...

docs/capturing-arguments.md
```diff
@@ -0,0 +1,107 @@
...
+This is a fairly standard test with fakes, except we:
+
+* create a `CapturedArgument` variable to store a captured argument value
+* use `An<>.That.IsCapturedTo` to create a "capturing constraint" that matches any value
+* use `Values` to access all the captured values so they can be asserted
+* (alternative flow) use `GetLastValue` to access the most recent captured value so it can be asserted
+
+???+ note "Values are only captured if the call matches the configuration"
+    When a call configuration includes one or more argument-capturing constraints, the argument
+    values are only captured if the call is triggered. If an incoming call does not match that
```
thomaslevesque, line 40:
```suggestion
    values are only captured if this specific call configuration is triggered. If an incoming call does not match that

```

(more precise)

docs/capturing-arguments.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+## Capturing mutable arguments
+
+### The challenge
+
+Argument capture is shallow; there's no copying of object state.
+If a reference-based argument (e.g. a class instance, not a struct) is captured and
+subsequently modified by the test or production code, the updated state will be seen
+during the "assert" phase of the test.
```
thomaslevesque, line 62: nit: I think the sentence would be clearer in the active voice:
```suggestion
subsequently modified by the test or production code, the "assert" phase of the test
will see the updated state.
```

docs/capturing-arguments.md
```diff
@@ -0,0 +1,107 @@
...
+You can even transform values into a different type:
+
+```csharp title="freezing state of captured values as new type" linenums="1" hl_lines="2 3 19 20"
+--8<--
+recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs:CaptureCopiedMutatedListToNewType
+--8<--
+```
+
+Aside from changing the freezer function, the `CapturedArgument` needs to have the typeparams
+supplied for the argument value and the captured value.
```
thomaslevesque, line 107: I wonder if we could change the API a bit to take advantage of type inference to avoid having to specify all type params. Maybe something like this?
```csharp
var capturedLists = CapturedArgument
    .OfType<IEnumerable<string>>()
    .Freeze(l => string.Join(" - ", l!)
```

mkdocs.yml
```diff
@@ -69,7 +69,9 @@ nav:
   - Calling base methods: calling-base-methods.md
   - Calling wrapped methods: calling-wrapped-methods.md
   - Changing behavior between calls: changing-behavior-between-calls.md
-- Argument constraints: argument-constraints.md
+- Argument constraints:
+  - Argument constraints: argument-constraints.md
+  - Capturing arguments: capturing-arguments.md
```
thomaslevesque, line 74: I'm not sure I would put this under "argument constraint", because as I mentioned before, it's not necessarily related.

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        // Assert
+        capturedLists.Values.Should().BeEquivalentTo(
+            new[] { "two", "three", "four" },
+            new[] { "ate", "nine" });
+
+        // Or if you only care about the last value:
+        capturedLists.GetLastValue().Should().BeEquivalentTo(
+            new[] { "ate", "nine" });
+        //// --8<-- [end:SimpleCapture]
```
thomaslevesque, line 45: I think this example is probably more complex than it needs to be to introduce the concept of argument capture. The fact that the argument is mutated just adds confusion, IMO. I understand it's for continuity with the mutable argument capture, but for a first example, I think it would be better to use a simpler example, e.g. with a `string` or `int` argument.

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
```diff
@@ -0,0 +1,179 @@
...
+            .WithMessage(@"*but {""three"", ""four""}*contains 1 item(s) less than*{""two"", ""three"", ""four""}*");
+    }
+
+    [Fact]
+    public void CaptureCopiedMutatedList()
+    {
+        //// --8<-- [start:CaptureCopiedMutatedList]
+        // Arrange
+        var capturedLists =
+            new CapturedArgument<IEnumerable<string>>(Enumerable.ToList);
```
thomaslevesque, line 117: nit: while I also like to use method groups instead of lambdas when possible, I think a lambda will be easier to understand for the average reader.

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
```diff
@@ -0,0 +1,179 @@
...
+        //// --8<-- [end:CaptureCopiedMutatedList]
+    }
+
+    [Fact]
+    public void CaptureCopiedMutatedListToNewType()
+    {
+        //// --8<-- [start:CaptureCopiedMutatedListToNewType]
+        // Arrange
+        var capturedLists = new CapturedArgument<IEnumerable<string>, string>(
+            l => string.Join(" - ", l!));
```
thomaslevesque, line 144: Is the `!` still necessary after the nullability changes you made?

src/FakeItEasy/CapturedArgument.of.TArg.TCap.cs
```diff
@@ -0,0 +1,60 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    /// <summary>
+    /// Holds argument values captured from calls to a faked member.
+    /// </summary>
+    /// <typeparam name="TArg">The type of the argument.</typeparam>
+    /// <typeparam name="TCap">The type of the value to capture from the argument.</typeparam>
+    public class CapturedArgument<TArg, TCap>
```
thomaslevesque, line 12: nit: maybe `TArgument` and `TCapture` instead of `TArg` and `TCap`?
Characters are cheap!

thomaslevesque (changes requested):
Thanks for the updates, looks great!
I finally managed to review the docs, and I have a few remarks

blairconrad:
@thomaslevesque, I rudely force-pushed my new branch. I think I addressed all your comments as well as a few we talked about offline.

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
```diff
@@ -0,0 +1,183 @@
...
+        //// --8<-- [end:CaptureCopiedMutatedList]
+    }
+
+    [Fact]
+    public void CaptureCopiedMutatedListToNewType()
+    {
+        //// --8<-- [start:CaptureCopiedMutatedListToNewType]
+        // Arrange
+        var capturedSummands = A.Captured<IEnumerable<int>, string>(
+            l => string.Join(" + ", l));
```
thomaslevesque, line 145: This is the only part of the API that I'm not entirely happy with. Having to explicitly specify the type of the frozen argument is annoying. I think we could take advantage of type inference like this:

```csharp
        var capturedSummands = A.Captured<IEnumerable<int>>()
            .Freeze(l => string.Join(" + ", l));
```

`Freeze` (or whatever name we pick) would create a new instance of `Captured<TArgument, TCapture>`, based on the first `Captured<T>` (which would just be discarded).

What do you think?

src/FakeItEasy/Captured.of.T.cs
```diff
@@ -0,0 +1,24 @@
+namespace FakeItEasy
+{
+    using System;
+
+    /// <summary>
+    /// Holds argument values captured from calls to a faked member.
+    /// </summary>
+    /// <typeparam name="T">The type of the argument.</typeparam>
+    public class Captured<T> : Captured<T, T>
```
thomaslevesque, line 9: nit: `Captured<TArgument>`, for consistency with `Captured<TArgument, TCapture>`

src/FakeItEasy/Core/DefaultArgumentConstraintManager.cs
```diff
@@ -3,79 +3,66 @@ namespace FakeItEasy.Core
...
+        private readonly Action<MatchesConstraint> onConstraintCreated;
 
         public DefaultArgumentConstraintManager(Action<IArgumentConstraint> onConstraintCreated)
+            : this((MatchesConstraint constraint) => onConstraintCreated(constraint))
         {
-            this.onConstraintCreated = onConstraintCreated;
         }
 
-        public IArgumentConstraintManager<T> Not => new NotArgumentConstraintManager(this);
+        private DefaultArgumentConstraintManager(Action<MatchesConstraint> onConstraintCreated) =>
```
thomaslevesque, line 15: I was a bit confused by this...
Do we need to keep the old constructor? It's only used in `RootModule.ArgumentConstraintManagerFactory` (and in tests). We could just change the signature of `ArgumentConstraintTrap.ReportTrappedConstraint`.

src/FakeItEasy/ICapturableArgumentConstraintManager.cs
```diff
@@ -0,0 +1,26 @@
...
+        ///   transformation from <typeparamref name="T"/> to <typeparamref name="TCapture"/>.
+        /// </typeparam>
+        /// <param name="capturedArgument">Where to capture the argument values to.</param>
+        /// <returns>A configuration object that may be used to further constrain the supplied argument.</returns>
+        /// <remarks>
+        /// Values are only captured if the call matches the configuration.
+        /// When a call configuration includes one or more argument-capturing constraints, the argument
+        /// values are only captured if the call is triggered. If an incoming call does not match what's
+        /// configured for the method or property, no arguments are captured.
+        /// </remarks>
```
thomaslevesque, line 23: You put a lot of effort in the documentation for an internal interface 😉
(I assume you did this because it was public in a previous iteration... No harm in keeping it, I guess)

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
```diff
@@ -0,0 +1,318 @@
...
+            "And a captured argument instance"
+                .x(() => capturedArgument = A.Captured<int>());
+
+            "And a method on the fake is called"
+                .x(() => fake.Invoke(354897, -3));
+
+            "When we assert that the call must've happened with a capturing argument and another non-matching argument"
+                .x(() => exception = Record.Exception(
+                    () => A.CallTo(() => fake.Invoke(capturedArgument._, 3))
+                            .MustHaveHappened()));
```
thomaslevesque, line 49: This is actually a use case that I didn't consider before. Does it even make sense to use a `Captured` in a call assertion? It won't capture the value, since the call has already happened...
Not sure if we could/should do something to prevent it.

tests/FakeItEasy.Specs/FluentApiSpecs.cs
```diff
@@ -0,0 +1,41 @@
...
+    using Xbehave;
+    using Xunit;
+
+    /// <summary>
+    /// Check certain fluent API syntax invariants.
+    /// </summary>
+    public static class FluentApiSpecs
+    {
+        [Scenario]
+        public static void AAndAnShouldHaveTheSameMembers(IEnumerable<string> aMembers, IEnumerable<string> anMembers)
```
thomaslevesque, line 18: Good idea!

tests/FakeItEasy.Tests/CapturedArgumentTests.cs
```diff
@@ -0,0 +1,18 @@
+namespace FakeItEasy.Tests
+{
+    using System;
+    using System.Linq.Expressions;
+    using Xunit;
+
+    public class CapturedArgumentTests
+    {
+        [Fact]
+        public void Constructor_should_be_null_guarded()
```
thomaslevesque, line 10: No one can say you're not thorough!

thomaslevesque (changes requested):
Nice work, @blairconrad! I enjoyed reviewing it.
I do have a few minor remarks, and a suggestion for a small API change.

src/FakeItEasy/Core/DefaultArgumentConstraintManager.cs
thomaslevesque: Never mind that. Not all `IArgumentConstraint`s are `MatchesConstraint`s...

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    [Fact]
+    public void NaivelyCaptureMutatedList()
+    {
+        var test = () =>
+        {
+            //// --8<-- [start:NaivelyCaptureMutatedList]
+            // Arrange
+            var capturedSummands = A.Captured<IEnumerable<int>>();
```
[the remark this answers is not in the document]
thomaslevesque: Apparently it _is_ [a real word](https://en.wiktionary.org/wiki/summand). Still, using a more widely known term would probably be better.

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
blairconrad: I sympathize with the annoying bonus typeparam. I'm not sold on the `.Freeze`, though. 
Partly because I think it's a _little_ less discoverable (I figure the second constructor will show up in intellisense more strongly, but maybe people will notice the `.Freeze` as well). But I also worry a little about a type that has `_`/`Ignored`/`That` **and** `Values`/`GetLastValue` hanging off it. We're mixing responsibilities. Maybe I worry a little.
Lemme sleep on it.

src/FakeItEasy/ICapturableArgumentConstraintManager.cs
blairconrad: You assume correct!

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
blairconrad: I don't think _I'd_ ever do it. I guess the more likely scenario is that someone saves away the result of `A.CallTo`. I do do that. I can rewrite the test to use that flow.

Is it likely to capture and use `MustHaveHappened`? I dunno. But if people do it (and I don't think we should prevent it), it'd be nice if it works. 😄

tests/FakeItEasy.Specs/FluentApiSpecs.cs
blairconrad: Oh. I half forgot about this! I ended up not changing `A<T>` and `An<T>`, so this doesn't provide a tonne of value in this PR. No harm though. I"ll drop if you want.

tests/FakeItEasy.Tests/CapturedArgumentTests.cs
blairconrad: I've an image to maintain.

src/FakeItEasy/Captured.of.TArgument.TCapture.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        public TArgument Ignored => this.That.Matches(x => true, x => x.Write(nameof(this.Ignored)));
+
+        /// <summary>
+        /// Gets the captured values.
+        /// </summary>
+        /// <remarks>
+        /// Will be empty if no arguments were captured.
+        /// </remarks>
+        public IReadOnlyList<TCapture> Values => this.values;
```
blairconrad, line 60: It occurs to me that we could improve UX in the case where the `Captured` is created and never used to configure a call and then someone tries to access `Values`. They might get no results and be confused. We _could_ flip a flag once `_`, `Ignored`, or `That` is triggered. If it hasn't been flipped, `Values` could throw.

Maybe it's effort for nothing?
(Also, the analyzers will complain if the property throws what's likely to be a `FakeConfigurationException`.)

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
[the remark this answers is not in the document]
thomaslevesque: (I posted my last comment before seeing your reply, in case it wasn't clear...)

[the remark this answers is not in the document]
thomaslevesque: Anyway, feel free to ignore!

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
thomaslevesque: Yeah, I don't love having an extra member that could be used in call specifications... But we have `Values`/`GetLastValue` anyway, which have the same problem. Maybe an analyzer for this would be useful (unless we can detect that we're in a call specification, and throw an exception in that case, but I suspect it's not that simple)

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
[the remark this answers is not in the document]
blairconrad: I don't mind "operand". In my defense, I avoided "augend" as being "too out there, even for me".

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
thomaslevesque: > I can rewrite the test to use that flow.

No, don't bother.

> Is it likely to capture and use `MustHaveHappened`? I dunno. But if people do it (and I don't think we should prevent it), it'd be nice if it works. 😄

What would "works" entail? It can't capture the value, since the call already happened.

tests/FakeItEasy.Specs/FluentApiSpecs.cs
thomaslevesque: No, keep it!

src/FakeItEasy/Captured.of.TArgument.TCapture.cs
thomaslevesque: I think it's fine to just return an empty list.

> (Also, the analyzers will complain if the property throws what's likely to be a `FakeConfigurationException`.)

I don't understand. Can you elaborate?

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
blairconrad: Oh, capture, but save away the result of `A.CallTo` to use later with a `MustHavHappened`. I'd expect any constraints applied to the `Captured` to be used correctly in the `MustHaveHappened`.

```c#
var capturedArg = A.Captured<int>();

var aCallToMethod = A.CallTo(() => fake.Method(capturedArg.That.IsDivisibleBy7());

aCallToMethod.Returns(6);

…

// passes if there was one call that supplied an argument divisible by 7
aCallToMethod.MustHaveHappened(); 
```

src/FakeItEasy/Captured.of.TArgument.TCapture.cs
blairconrad: I started going down this road. Made `Values` sometimes throw. When I did, some analyzer. Probably whatever FxCop is now. Complained about throwing the unexpected exception from the property. I gave up, threw it away, later felt bad, and then rose the spectre of throwing again now.

Okay. I just checked:

> warning CA1065: get_Values creates an exception of type FakeConfigurationException, an exception type that should not be raised in a property. If this exception instance might be raised, use a
different exception type, convert this property into a method, or change this property's logic so that it no longer raises an exception. (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1065) [C:\Dev\FakeItEasy\src\FakeItEasy\FakeItEasy.c
sproj::TargetFramework=net6.0]

I quote:

> The following exceptions can be thrown from a property get method:

> 
> * [System.InvalidOperationException](https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception) and all derivatives (including [System.ObjectDisposedException](https://learn.microsoft.com/en-us/dotnet/api/system.objectdisposedexception))

> * [System.NotSupportedException](https://learn.microsoft.com/en-us/dotnet/api/system.notsupportedexception) and all derivatives

> * [System.ArgumentException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception) (only from indexed get)

> * [System.Collections.Generic.KeyNotFoundException](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.keynotfoundexception) (only from indexed get)

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
blairconrad: Trying out the scenarios in the IDE.

Your `FrozenBy` scenario:

https://github.com/FakeItEasy/FakeItEasy/assets/3275797/077b68e6-a4bb-4056-918a-ca375950c946

The current overloads scenario:

https://github.com/FakeItEasy/FakeItEasy/assets/3275797/de82a532-377b-48f5-bed6-db12536cca30

Still to me the earlier IntelliSense feedback feels better, but I'll roll it around a little more.

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
thomaslevesque: Ah, right. Sounds good to me.

src/FakeItEasy/Captured.of.TArgument.TCapture.cs
blairconrad: Which it seems we don't. I'm content to leave as empty list.

thomaslevesque: Oh, right. I was confused, I thought you were referring to the FakeItEasy analyzer

blairconrad:
@thomaslevesque, I think I addressed everything but https://github.com/FakeItEasy/FakeItEasy/pull/1967#discussion_r1439119580 (how to specify the freezer, and whether it requires the second typeparam). I don't have an improved solution yet, but figured I could at least get the other changes off my machine…

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
```diff
@@ -283,7 +283,12 @@ public static void CaptureNegatedConstrainedValues(Action<int> fake, Captured<in
 
             "Then the captured arguments instance's values match the input at the time of the calls"
                 .x(() => capturedArgument.Values.Should().BeEquivalentTo(
```
thomaslevesque, line 285: You replaced the other `BeEquivalentTo`s with `Equal`, but not this one. Is it intentional?
(I suspect it doesn't work with nested arrays, just checking)

thomaslevesque (commented):
LGTM. Let's discuss how to specify the freezer in the chat

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
blairconrad: It didn't work with nested arrays.

recipes/FakeItEasy.Recipes.CSharp/CapturingArguments.cs
thomaslevesque: Somehow I missed your reply here.
To be honest, I'm not convinced. In my experience Intellisense isn't very helpful when picking between generic methods with different arity. Or maybe it's just in Rider.
Even if it's slightly worse in terms of discoverability, I'd pick the comfort of not having to specify the type explicitly every time.

thomaslevesque: Also, I'd argue `.FrozenBy` is no less discoverable than, say, `.Returns` or `.MustHaveHappened`...

blairconrad: >  I'd pick the comfort of not having to specify the type explicitly every time.

Every time someone captures. And wants to provide a "freezer" that changes the type of the captured argument. (and probably the new type is something simple like `string`)

> Also, I'd argue `.FrozenBy` is no less discoverable than, say, `.Returns` or `.MustHaveHappened`...

And finding those is enough of a problem that we added Diagnostic <html>
FakeItEasy0001: Unused call specification. 😄

But, I do like that `.FrozenBy` is a named operation rather than a random `Func` hanging out in the constructor. I'll whip a commit. Complain soon (or later) if you don't like `FrozenBy`.

thomaslevesque: > Complain soon (or later) if you don't like `FrozenBy`.

I don't mind `FrozenBy`

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
```diff
@@ -257,13 +257,13 @@ public static void CaptureNegatedConstrainedValues(Action<int> fake, Captured<in
 
         [Scenario]
         public static void CaptureMutatingArgument(
-            Action<IList<int>> fake, IList<int> listOfInts, Captured<IList<int>> capturedArgument)
+            Action<IList<int>> fake, IList<int> listOfInts, Captured<IList<int>, List<int>> capturedArgument)
```
thomaslevesque, line 260: Doesn't this cause [CA1002](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1002)?

thomaslevesque (approved):
Looks great, thanks for humoring me!
The code is more readable this way, I think.
I just have one minor comment

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
thomaslevesque: Oh. The `List<T>` isn't directly a parameter, it's a type argument of a parameter, which is probably fine...

blairconrad: I see no evidence of the warning in VS nor the CI builds&hellip;

blairconrad:
Thanks for the review, @thomaslevesque.

tests/FakeItEasy.Specs/ArgumentCapturingSpecs.cs
thomaslevesque: Yeah, I checked the logs and didn't see it. That's what prompted me to look closer...

thomaslevesque:
Thanks @blairconrad, great work! It was a bumpy ride, but we got there eventually

PR #1975, opened 2023-12-22
Evaluate argument matchers only if needed

Miista:
When evaluating whether to apply a user rule the checks have the following order:

1. IsApplicableTo
2. HasNotBeenCalledSpecifiedNumberOfTimes

This commit changes that order to:

1. HasNotBeenCalledSpecifiedNumberOfTimes
2. IsApplicableTo

This means that we only evaluate argument matchers if the rule has not been called the specified number of times.

### DIFF
--- a/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -59,6 +59,61 @@ namespace FakeItEasy.Specs
             int Bar(object arg);
         }
 
+        public interface IHaveAnIntParameter
+        {
+            int Bar(int arg);
+        }
+
+        [Scenario]
+        public static void Test(IHaveAnIntParameter fake, List<int> results, List<int> arguments)
+        {
+            Func<int, bool> recordsArguments = i =>
+            {
+                arguments.Add(i);
+                return true;
+            };
+
+            "Given a faked class with a method taking an int parameter"
+                .x(() => fake = A.Fake<IHaveAnIntParameter>());
+
+            "And a call to a method with an int parameter configured on this fake"
+                .x(() =>
+                    A.CallTo(() => fake.Bar(A<int>.That.Matches(recordsArguments, writer => writer.Write("Hello"))))
+                        .Returns(10)
+                        .NumberOfTimes(1)
+                        .Then
+                        .Returns(20)
+                        .NumberOfTimes(1)
+                );
+
+            "And a list to record the arguments"
+                .x(() => arguments = new List<int>());
+
+            "And a list to record the results"
+                .x(() => results = new List<int>());
+
+            "When two calls with int parameters are made to the fake"
+                .x(() =>
+                    {
+                        results.Add(fake.Bar(0));
+                        results.Add(fake.Bar(1));
+                    }
+                );
+
+            "Then the fake returns the results as configured"
+                .x(() =>
+                    {
+                        results[0].Should().Be(10, because: "that is the result of the first invocation");
+                        results[1].Should().Be(20, because: "that is the result of the second invocation");
+                    }
+                );
+
+            "And only the arguments used in the calls are recorded"
+                .x(() => arguments.Should().
+                    HaveCount(2, because: "only 2 calls were made to the fake")
+                    .And.Subject.Should().Contain(new[] { 0, 1 }, because: "those are the arguments used when invoking the fake"));
+        }
+
         [Scenario]
         public static void CallToAField(ClassWithAField fake, Exception exception)
         {
--- a/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
+++ b/tests/FakeItEasy.Specs/CallMatchingSpecs.cs
@@ -65,50 +65,38 @@ namespace FakeItEasy.Specs
         }
 
         [Scenario]
-        public static void Test(IHaveAnIntParameter fake, IList<int> results, IList<int> arguments)
+        public static void CallWithSpecificNumberOfTimes(IHaveAnIntParameter fake, IList<int> arguments, Func<int, bool> recordsArguments)
         {
-            Func<int, bool> recordsArguments = i =>
-            {
-                arguments.Add(i);
-                return true;
-            };
-
             "Given a faked class with a method taking an int parameter"
                 .x(() => fake = A.Fake<IHaveAnIntParameter>());
 
-            "And a call to a method with an int parameter configured on this fake"
+            "And an argument matching predicate that records the argument and always returns true"
+                .x(() => recordsArguments = i =>
+                {
+                    arguments.Add(i);
+                    return true;
+                });
+
+            "And a call to a method with an int parameter configured on this fake using that predicate, with a specific number of times"
                 .x(() =>
                     A.CallTo(() => fake.Bar(A<int>.That.Matches(recordsArguments, writer => writer.Write("Hello"))))
-                        .Returns(10)
-                        .NumberOfTimes(1)
+                        .Returns(10).Once()
                         .Then
-                        .Returns(20)
-                        .NumberOfTimes(1));
+                        .Returns(20).Once());
 
             "And a list to record the arguments"
                 .x(() => arguments = new List<int>());
 
-            "And a list to record the results"
-                .x(() => results = new List<int>());
-
             "When two calls with int parameters are made to the fake"
                 .x(() =>
-                    {
-                        results.Add(fake.Bar(0));
-                        results.Add(fake.Bar(1));
-                    });
+                {
+                    fake.Bar(0);
+                    fake.Bar(1);
+                });
 
-            "Then the fake returns the results as configured"
-                .x(() =>
-                    {
-                        results[0].Should().Be(10, because: "that is the result of the first invocation");
-                        results[1].Should().Be(20, because: "that is the result of the second invocation");
-                    });
-
-            "And only the arguments used in the calls are recorded"
+            "Then the arguments used in the calls are recorded exactly once"
                 .x(() => arguments.Should()
-                    .HaveCount(2, because: "only 2 calls were made to the fake")
-                    .And.Subject.Should().Contain(new[] { 0, 1 }, because: "those are the arguments used when invoking the fake"));
+                    .Equal(new[] { 0, 1 }, because: "those are the arguments used when invoking the fake"));
         }
 
         [Scenario]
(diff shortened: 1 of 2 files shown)


blairconrad:
Hi, @Miista. Thanks for your interest.

Can you explain why you think this change is appropriate? Is it a feeling that this will be more efficient, or have you noticed some more dramatic ill effects with the current code?
Technically this is a breaking change, and wouldn't be eligible until a major release (which we can scheduled at any time, but we like to have a pretty strong reason).
In any event, we wouldn't merge the change without test changes that verified the new behaviour.

thomaslevesque:
> Technically this is a breaking change

Just to clarify what @blairconrad meant: this is a breaking change because if the argument matchers have side effects, those side effects will no longer happen if `HasNotBeenCalledSpecifiedNumberOfTimes` returns false.

(Obviously, having side effects in matchers is probably not a good idea, but we can't prevent people from doing it.)

Strictly speaking, I think this change would make sense, but we need to carefully consider the impacts. In fact, I think we might have discussed this in the past already and decided against it... I'll see if I can find something about it.

jrosiek:
Some additional context of why this PR was proposed: https://github.com/Miista/FakeItEasy.Capture/issues/1

thomaslevesque:
Hi @jrosiek,
Thanks for your comment, this is useful information. I was not aware of this project.

> I'm curious what is your stance on the argument capture functionality in general (I guess you must have had discussions about it).

As a matter of fact, [we have](https://github.com/FakeItEasy/FakeItEasy/issues/1950)! We've been struggling to find the right way to do it, and FakeItEasy.Capture has an interesting approach, although it has some limitations (e.g. no way to constrain the captured argument).

Given the way FakeItEasy.Capture works, I understand how evaluating the matchers every time could be a problem. Personally, I wouldn't mind swapping the condition. It would technically be a breaking change, but unlikely to affect many people. If we do it, and treat it as a breaking change, it means it won't be released until the next major version, which we haven't really started working on yet. And anyway, a built-in mechanism to capture argument values is something we're considering for the next major version, if we find an approach we like...

Miista:
@thomaslevesque I can find some time to look into support for constraining the captured argument.

thomaslevesque:
> I can find some time to look into support for constraining the captured argument.

Actually, I already gave it some thought. I think it would be nice to leverage the existing FakeItEasy argument constraints, so maybe it could be done like this (based on the example in your README):

```csharp
var singleArgument = new Capture<string>(that => that.StartsWith("foo"));
```

Miista:
@thomaslevesque Sure. That could be implemented pretty easily. We would pass on the `Func<T, bool>` to `A<T>`. 

You mentioned other limitations.

blairconrad:
Thanks for the additional context, @jrosiek. Indeed we have recently discussed making capturing of arguments a "first class citizen". See #1950. We'd been going down a path not completely dissimilar to FakeItEasy.Capture (which we'd never heard of until #1975 was raised) and then started veering into other territory.

I put the work on hold because I needed time to resolve my thoughts on which path was "best" (neither will be perfect) and to figure out how supporting capture from the already-recorded calls side would look. And work and family commitments kept me from having big enough blocks of time to really formulate something. I'd expected to come back by now, since my Christmas holidays started a week ago, but… life.

blairconrad:
@Miista 

> Edit: I understand there's interest for merging this PR. Is that correct?

Yes. In time. I like your change. I think it's superior to the order the checks were originally implemented. I'd merge it right now except for

1. the lack of test; if we're changing user-visible behaviour, I'd like to see something that tests that (and ensures we don't backslide), and
2. the timing. I think this should be in a major release. We have the power to schedule one whenever we want, but I don't _love_ the idea of putting 9.0.0 out right after 8.0.0. If nothing else, I'd like to see if we're going to end up with additional breaking changes due to the overall "Capture" conversation.

Miista:
@blairconrad What kind of tests did you have in mind?

I ran the tests locally both before and after making the change. 
I also added (and since removed) a test verifying that this change did indeed fix the problem mentioned in Miista/FakeItEasy.Capture#1

blairconrad:
> I also added (and since removed) a test verifying that this change did indeed fix the problem mentioned in https://github.com/Miista/FakeItEasy.Capture/issues/1

This would be the one. I'd add a test in Specs to verify that the argument constraints for an expiring call constraint aren't exercised once the maximum call count had been reached.

Miista:
@blairconrad The test was using code from Miista/FakeItEasy.Capture. Are you suggesting to add my code to FakeItEasy?

blairconrad:
I'm suggesting writing a test to exercise FakeItEasy (from the outside, as the Specs do) that will verify that the # of times a call has been triggered will short-circuit the "is applicable" check.
As @thomaslevesque recently said to me in sidechat, 

> […] the hardest part when making a change to FakeItEasy is often figuring out where to put the new specs 🤔

I don't think there's a perfect match here, but I'd consider adding the new test to either `CallMatchingSpecs` or `ThenSpecs`. The first is a little broad, and the second a little too narrow, but making a 3rd place at this time isn't something I'd ask of you. Nor am I sure we want a 3rd place…

Anyhow, if this is too onerous, I'd be willing, with your permission, to add the test to your branch.

Miista:
@blairconrad I will try to get it done within the next few days.

blairconrad:
Cool, cool @Miista. No rush.

thomaslevesque:
> You mentioned other limitations.

To be honest, I don't remember what else I had in mind 😅
Anyway, it's a bit off topic for this issue. We should continue the discussion about capturing approaches in #1950.

Miista:
@blairconrad I've added tests in CallMatchingSpecs.cs

blairconrad:
Thanks, @Miista! I eyeballed it and it will definitely ensure that the fix will stay fixed.
It did introduce a number of new style warnings, but those are hard to see right now in the middle of the security violations we have (and which I'm hoping #1982 will fix), so I am not surprised you missed them. These aren't a huge deal and there's probably no huge rush to address, since we'll eventually need to rebase the branch anyhow, and refrain from merging until we're ready for a major release…

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -59,6 +59,61 @@ public interface IHaveAnObjectParameter
             int Bar(object arg);
         }
 
+        public interface IHaveAnIntParameter
+        {
+            int Bar(int arg);
+        }
+
+        [Scenario]
+        public static void Test(IHaveAnIntParameter fake, List<int> results, List<int> arguments)
```
thomaslevesque, line 68: The name should be more descriptive.
Something like `ArgumentMatchersAreOnlyEvaluatedIfTheSpecifiedNumberOfTimesHaveBeenReached`
(yes, it's a mouthful... but at least we know what the test does)

thomaslevesque (approved):
Thanks, @Miista!
As mentioned by @blairconrad, there are a few StyleCop violations, but we can push a commit to your branch to fix them ourselves.

thomaslevesque:
Actually, I didn't mean to approve yet, since at least my remark about the method name should be addressed 🤦

blairconrad:
@thomaslevesque and I have been talking. Seeing the test actually drove home to me how surprising the current code is. This fix is still a change in behaviour, but I think the current behaviour is an error, not just a different-than-expected option. I've labelled the PR a bug fix. Bug fixes often change behaviour, but don't require a major version bump.

thomaslevesque:
I pushed fixes to the branch to address my comments, and fix the StyleCop/CA warnings.

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -59,6 +59,46 @@ public interface IHaveAnObjectParameter
...
+            int Bar(int arg);
+        }
+
+        [Scenario]
+        public static void CallWithSpecificNumberOfTimes(IHaveAnIntParameter fake, IList<int> arguments, Func<int, bool> recordsArguments)
+        {
+            "Given a faked class with a method taking an int parameter"
+                .x(() => fake = A.Fake<IHaveAnIntParameter>());
+
+            "And a argument matching predicate that records the argument and always returns true"
```
blairconrad, line 73:
```suggestion
            "And an argument matching predicate that records the argument and always returns true"

```

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -59,6 +59,46 @@ public interface IHaveAnObjectParameter
...
+            "When two calls with int parameters are made to the fake"
+                .x(() =>
+                {
+                    fake.Bar(0);
+                    fake.Bar(1);
+                });
+
+            "Then the arguments used in the calls are recorded only once"
+                .x(() => arguments.Should()
+                    .Equal(new[] { 0, 1 }, because: "those are the arguments used when invoking the fake"));
```
blairconrad, line 99: I never use `Equal`! Instead I use `BeEquivalentTo`, which until now I didn't realize was agnostic to the order of the collections. I have some tests to change. Thanks!

tests/FakeItEasy.Specs/CallMatchingSpecs.cs
```diff
@@ -59,6 +59,46 @@ public interface IHaveAnObjectParameter
...
+                .x(() => arguments = new List<int>());
+
+            "When two calls with int parameters are made to the fake"
+                .x(() =>
+                {
+                    fake.Bar(0);
+                    fake.Bar(1);
+                });
+
+            "Then the arguments used in the calls are recorded only once"
```
blairconrad, line 97:
```suggestion
            "Then the arguments used in the calls are recorded exactly once"
```

Or leave it. I'm not _that_ fussy.

blairconrad (changes requested):

blairconrad:
Thanks. Minor comments (the "an" is the only one I really care about). Feel free to squash/rebase/whatever.

thomaslevesque:
I amended the last commit to address your comments

blairconrad (approved):
Thanks, @thomaslevesque. Sorry, I should've been more clear. Rebasing on top of FakeItEasy/master would've also been fine. Maybe you realized this and chose not to. Whatevs!

thomaslevesque:
I didn't rebase because it's not my branch... But if you're ready to merge, I guess there's no harm. I'll do it now.

blairconrad:
Thanks @Miista (and @thomaslevesque for the tidying)!

Miista:
@blairconrad That's awesome. What are the next steps from here?

blairconrad:
Eventually we'll release! Since my capturing PR seems to be settling down, we're likely to wait until that's done (days not weeks). If it drags on longer, we may reconsider the release schedule.

blairconrad:
Thanks again, @Miista. Look for you name in the release notes! 

🥇

PR #1979, opened 2023-12-30
Upgrade StyleCop

blairconrad:
Updated rules will work better with collection expressions when we start using them.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.261" PrivateAssets="all" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.556" PrivateAssets="all" />
     <CodeAnalysisDictionary Include="$(MSBuildThisFileDirectory)src/FakeItEasy.Dictionary.xml">
       <Link>Properties\FakeItEasy.Dictionary.xml</Link>
     </CodeAnalysisDictionary>
--- a/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
+++ b/src/FakeItEasy/Core/CompletedFakeObjectCall.cs
@@ -26,4 +26,4 @@ namespace FakeItEasy.Core
 
         public int SequenceNumber { get; }
     }
-}
+}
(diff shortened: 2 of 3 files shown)


src/FakeItEasy/Core/CompletedFakeObjectCall.cs
```diff
@@ -26,4 +26,4 @@ public CompletedFakeObjectCall(IInterceptedFakeObjectCall interceptedCall, objec
 
         public int SequenceNumber { get; }
     }
-}
\ No newline at end of file
+}
```
blairconrad, line 29: New StyleCop was snippy about the lack of newline.

blairconrad:
Actually I think I see a [new warning from the build project](https://github.com/FakeItEasy/FakeItEasy/actions/runs/7365846031/job/20047530566?pr=1979#step:4:11). Which we probably can't address in here. I can try to address before we merge this if you like.

tools/FakeItEasy.Build/Program.cs
```diff
@@ -115,6 +116,10 @@
 
 RunTargetsAndExit(args, messageOnly: ex => ex is NonZeroExitCodeException);
 
+[SuppressMessage(
+    "StyleCop.CSharp.MaintainabilityRules",
+    "SA1400:AccessModifierMustBeDeclared",
+    Justification = "Cannot supply access modifier outside namespace.")]
```
thomaslevesque, line 122: What do you mean? We can absolutely specify an access modifier.
Declaring it `public` causes `CA1050`, but `internal` or `file` works fine.
I think it would be better than suppressing the warning.

thomaslevesque (changes requested):
Thanks @blairconrad. Just a small request!

tools/FakeItEasy.Build/Program.cs
blairconrad: I'm a moron. I tried `private`. Will amend.

Directory.Build.props
```diff
@@ -12,7 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.261" PrivateAssets="all" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.556" PrivateAssets="all" />
```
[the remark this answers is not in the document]
blairconrad: If only they'd stop jiggling things.

thomaslevesque (approved):
LGTM, thanks!

blairconrad:
Ugh. I left the `using`.

thomaslevesque (approved):
I think it's good this time ^^

blairconrad:
Thanks for the pushback!

PR #1992, opened 2024-03-30
Drop 'No constructor matches the passed arguments for constructor.'

### LINKED ISSUE #1929: Creation failure message may indicate that the to-be-faked type has no applicable constructor when it really does
Noticed when reading the failure message in [How to fake DbContextOptions](https://stackoverflow.com/questions/75951931/how-to-fake-dbcontextoptions):

The error message says "No constructor matches the passed arguments for constructor"
But there _is_ a matching constructor, and it's the one that gets called. It's just that calling it throws an exception

blairconrad:
Fixes #1929.


thomaslevesque (approved):
LGTM, thanks @blairconrad!

blairconrad:
Thank you.

PR #1993, opened 2024-04-01
Make EventRule shared

thomaslevesque:
While investigating what needed to be saved for the `ResetToInitialConfiguration` feature, I realized we treated `EventRule` as a "special" stateful rule, but it doesn't actually need to be stateful, and can be shared among all fakes. What actually holds the state is the `EventCallHandler`, which is specific to each fake.

This PR is a small refactoring that makes `EventRule` just like the other shared rules.


blairconrad (approved):
Fantastic find and fix. Thanks, @thomaslevesque

PR #1994, opened 2024-04-02
[WIP] Add Fake.Reset

### LINKED ISSUE #1839: Discuss alternatives to removing ClearConfiguration
In #1764, @PiotrKlecha raised a concern about how `ClearConfiguration` removes the strictness from a [strict Fake](https://fakeiteasy.readthedocs.io/en/stable/strict-fakes/), and this was unexpected, since the strictness is (usually) applied at creation time, and so it feels like it's an intrinsic aspect of the Fake, rather than some additional configuration that's added on after.

In that case, it turned out that there was a way to reformulate @PiotrKlecha's tests so ClearConfiguration wasn't needed, but the bell couldn't be unrung. @thomaslevesque and I discussed how ClearConfiguration worked and how it's very difficult to predict what it would do from the outside. Complicating factors that we identified

* configuration to make the Fake strict, to have it wrap another object, or to have it call base methods are typically applied at creation time, and may not be what the client has in mind when clearing configuration
* configuration to make a Fake act strict or to have it call a base method can be applied after creation time, although it's less convenient. Therefore retaining particular kinds of rules may not be easy.
* [implicit creation options](https://fakeiteasy.readthedocs.io/en/stable/implicit-creation-options/) further confound the picture. Since they're not in front of the user, they may be relied on without them knowing about it, so removing them from the configuration might be a surprise (conversely, _not_ removing them might be a surprise).

thomaslevesque:
Closes #1839

WIP because I haven't updated the docs yet.


docs/advanced-usage.md
```diff
@@ -2,13 +2,31 @@
...
+
+However, in some situations this might not be practical. To actually reset a fake to its initial
+state, use the `Fake.Reset` method:
+
+```csharp
+Fake.Reset(fake);
+```
+
+This method resets all changes made to the fake after it was created. This includes:
+* call configurations
```
blairconrad, line 18: It's hard to add notes to unmodified code, so I'm noting here.

Down in "Manipulating a fake's call rules" we should probably explain about `IStatefulFakeObjectCallRule`.

src/FakeItEasy/Core/EventCallHandler.cs
```diff
@@ -35,6 +41,12 @@ public void HandleEventCall(EventCall eventCall)
             }
         }
 
+        public EventCallHandler GetSnapshot()
+        {
+            var registeredHandlers = this.registeredEventHandlers.ToDictionary(kvp => kvp.Key, kvp => kvp.Value);
```
blairconrad, line 46: I may be silly, but is this a fancy way to write `var registeredHandlers = new Dictionary(this.registeredEventHandlers);`?

(By "fancy", I mean "harder to read and likely less efficient".)

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -309,6 +322,35 @@ private void UpdateLastSequenceNumber(int sequenceNumber)
...
+        {
+            // ReSharper disable once InconsistentlySynchronizedField
+            this.allUserRules = rules is null
+                ? new LinkedList<CallRuleMetadata>()
+                : new LinkedList<CallRuleMetadata>(rules);
+        }
+
+        private void ReplaceRecordedCalls(CompletedFakeObjectCall[]? calls)
+        {
+            var newRecordedCalls = calls is null
```
blairconrad, line 335: not a big deal at all, but why bother with local variable?

src/FakeItEasy/Core/FakeManager.cs
```diff
@@ -309,6 +322,35 @@ private void UpdateLastSequenceNumber(int sequenceNumber)
...
+                ? new ConcurrentQueue<CompletedFakeObjectCall>()
+                : new ConcurrentQueue<CompletedFakeObjectCall>(calls);
+            this.recordedCalls = newRecordedCalls;
+
+            var lastCall = calls?.LastOrDefault();
+            if (lastCall is not null && lastCall.SequenceNumber > this.lastSequenceNumber)
+            {
+                this.UpdateLastSequenceNumber(lastCall.SequenceNumber);
+            }
```
blairconrad, line 344: This would not have occurred to me to do. As it is, I'm not sure it's needed. How would ever a sequence number from a recorded call be bigger than our current sequence number?

src/FakeItEasy/Fake.cs
```diff
@@ -47,7 +48,7 @@ public static IEnumerable<ICompletedFakeObjectCall> GetCalls(object fakedObject)
         /// </summary>
         /// <param name="fakedObject">The faked object to clear the configuration of.</param>
         [SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "object", Justification = "The term fake object does not refer to the type System.Object.")]
-        [Obsolete("ClearConfiguration will be removed in version 8.0.0. Prefer to discard the fake and create a new one.")]
+        [Obsolete("ClearConfiguration will be removed in version 8.0.0. Use Fake.Reset, or discard the fake and create a new one.")]
```
blairconrad, line 51:
```suggestion
        [Obsolete("ClearConfiguration will be removed in version 9.0.0. Use Fake.Reset, or discard the fake and create a new one.")]
```

? I know we warned people about 8.0.0, but then we blew past it (my bad) and now the 8.1 people have been using `ClearConfiguration` all the livelong day and I'd feel bad if 8.2 bit them.

blairconrad (commented):
Very nice, as always. I have only read the docs and production code so far. Will tackle tests later.  (Probably not today. 😞)

(This should not be construed as a promise not to later complain about production code again.)

thomaslevesque:
I think I addressed your comments

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
```diff
@@ -0,0 +1,398 @@
...
+            .x(() => fake.AReferenceTypeProperty = "hello");
+
+        "When the fake is reset"
+            .x(() => Fake.Reset(fake));
+
+        "And the property is called"
+            .x(() => result = fake.AReferenceTypeProperty);
+
+        "Then it returns a default value"
+            .x(() => result.Should().NotBeNull().And.NotBe("hello"));
```
blairconrad, line 171: Technically all we know is that it's some random non-hello value, not that it's the default value. But what else _could_ it be? I'll allow it!

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
```diff
@@ -0,0 +1,398 @@
...
+
+        "And the property is called"
+            .x(() => result = fake.AReferenceTypeProperty);
+
+        "Then it returns the value set at creation time"
+            .x(() => result.Should().Be("hello"));
+    }
+
+    [Scenario]
+    public static void ReferenceTypePropertyReadAtCreationTime(IFoo fake, object initialValue, object result)
```
blairconrad, line 197: not to disparage the other tests, but this is a nice one

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
```diff
@@ -0,0 +1,398 @@
...
+        "And a call is made on the fake"
+            .x(() => fake.AMethod());
+
+        "Then the interception listener is called"
+            .x(() => A.CallTo(() => listener.OnBeforeCallIntercepted(A<IFakeObjectCall>._)).MustHaveHappenedOnceExactly()
+                .Then(A.CallTo(() => listener.OnAfterCallIntercepted(A<ICompletedFakeObjectCall>._)).MustHaveHappenedOnceExactly()));
+    }
+
+    [Scenario]
+    public static void StrictFake(IFoo fake, Exception exception)
```
blairconrad, line 384: would pass if `Reset` did nothing at all. Perhaps configure `AMethod` before resetting?

blairconrad (changes requested):
Again, very good. Tests are comprehensive. Aside from my line items, the only thing I can think of is perhaps a test to verify that a user-supplied stateful rule is cloned.

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
thomaslevesque: Yeah, not sure what I was thinking. I could just test that it's an empty string

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
```diff
@@ -0,0 +1,398 @@
...
+        "And the property is set to a given value after creation"
+            .x(() => fake.AReferenceTypeProperty = "hello");
+
+        "When the fake is reset"
+            .x(() => Fake.Reset(fake));
+
+        "And the property is called"
+            .x(() => result = fake.AReferenceTypeProperty);
+
+        "Then it returns a default value"
```
thomaslevesque, line 214: Duh, copy pasta. I should reword this.

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
thomaslevesque: Good point

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
blairconrad: but the type is `object`. It's not likely to be a `""`.

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
blairconrad: whoops, I didn't even notice!

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
thomaslevesque: Actually, no. It's an `object`, not a string. Which means I can't know what the value will be. I think the way I had done it was fine. I just need to change the step description.

docs/advanced-usage.md
```diff
@@ -21,7 +21,8 @@ This method resets all changes made to the fake after it was created. This inclu
 * [recorded calls](#getting-the-list-of-calls-made-on-a-fake)
 * [interception listeners](#intercepting-calls)
 
-However, any changes made _during_ fake creation (via fake creation options) will be preserved, including:
+However, any changes made _during_ fake creation (via fake creation options) will be preserved.
+This includes all the elements mentioned above, but also:
```
blairconrad, line 25: quality change!

src/FakeItEasy/Fake.cs
```diff
@@ -48,7 +48,7 @@ public static IEnumerable<ICompletedFakeObjectCall> GetCalls(object fakedObject)
         /// </summary>
         /// <param name="fakedObject">The faked object to clear the configuration of.</param>
         [SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "object", Justification = "The term fake object does not refer to the type System.Object.")]
-        [Obsolete("ClearConfiguration will be removed in version 8.0.0. Use Fake.Reset, or discard the fake and create a new one.")]
+        [Obsolete("ClearConfiguration will be removed in version 9.0.0. Use Fake.Reset, or discard the fake and create a new one.")]
```
blairconrad, line 51: Not for this PR, but do we want an analyzer fixup rule to help migrate? It might be a teeny bit helpful, but the "obsolete" message is pretty clear and easy to act on. Let's say no!

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
thomaslevesque: I changed the assertion to `result.Should().BeAFake()`

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
```diff
@@ -3,6 +3,7 @@
 using System;
 using FakeItEasy.Core;
 using FluentAssertions;
+using Tests.TestHelpers;
```
blairconrad, line 6: Man, the build is mad about this one.

blairconrad (changes requested):

tests/FakeItEasy.Specs/ResetFakeSpecs.cs
thomaslevesque: Oops! We might be able to configure this in .editorconfig so that the IDE and StyleCop agree with each other...

src/FakeItEasy/Fake.cs
thomaslevesque: Interesting idea. What kind of analyzer do you have in mind? What would it do exactly? Check if the current version is >= than the version mentioned in the message?

blairconrad: It would have to see a usage of `ClearConfiguration` and suggest (if the current FIE is one that has `Reset`) the rewrite to `Reset`. It wouldn't necessarily be based on the Obsolete message itself. It could work solely based on the source code (and FIE version, assuming we can get that, which I bet we can).

But I think it's of limited value.

blairconrad (changes requested):
Thanks for the updates, @thomaslevesque. I think the only comment you've neither addressed or rebutted is my suggestion for a test to see that a user-supplied stateful rule is cloned.

src/FakeItEasy/Fake.cs
thomaslevesque: Ah, OK. I thought you meant an analyzer so that we don't forget to remove things we said we would remove 😆 
Yeah, probably not worth the trouble.

blairconrad:
> Oh, right. I only addressed the inline comments 🤦‍♂️

but you did it with panache! 🧐

blairconrad (approved):
Great! Thanks for your hard work, @thomaslevesque!

blairconrad:
Thanks, @thomaslevesque!

PR #2001, opened 2024-04-11
Fix docs preview workflow

thomaslevesque:
[dawidd6/action-download-artifact](https://github.com/dawidd6/action-download-artifact) no longer works with actions/upload-artifact@v4.

Instead, find the latest successful workflow run for the PR's head commit, and download the artifact from that run, using actions/download-artifact@v4.

Also, add a final step so that the workflow fails if we didn't successfully publish the docs (currently it always passes due to the `continue-on-error` attribute)

### DIFF
--- a/.github/workflows/publish-docs-preview.yml
+++ b/.github/workflows/publish-docs-preview.yml
@@ -11,15 +11,40 @@ jobs:
     if: "${{ github.event.label.name == ':eyes: publish-docs-preview' }}"
     steps:
 
-      - name: Download artifact
-        id: download-artifact
-        uses: dawidd6/action-download-artifact@v3
+      - name: Find CI workflow run for PR
+        id: find-workflow-run
+        uses: actions/github-script@v7
         continue-on-error: true
         with:
-          workflow: ci.yml
-          pr: ${{github.event.pull_request.number}}
+          script: |
+            // Find the last successful workflow run for the current PR's head
+            const {owner, repo} = context.repo;
+            const runsResponse = await github.rest.actions.listWorkflowRuns({
+              owner,
+              repo,
+              workflow_id: 'ci.yml',
+              event: 'pull_request',
+              head_sha: context.sha,
+              status: 'success'
+            });
+            const runs = runsResponse.data.workflow_runs;
+            runs.sort((a, b) => b.created_at.getTime() - a.created_at.getTime());
+            const run = runs[0];
+            if (!run) {
+              core.setFailed('Could not find a successful workflow run for the PR');
+              return;
+            }
+            core.setOutput('run-id', run.id);
+
+      - name: Download artifact
+        if: ${{ steps.find-workflow-run.outcome == 'success' }}
+        id: download-artifact
+        uses: actions/download-artifact@v4
+        continue-on-error: true
+        with:
           name: documentation
           path: documentation/
+          run-id: ${{ steps.find-workflow-run.outputs.run-id }}
 
       - name: Publish files
         id: publish-files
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docs-preview.yml
```diff
@@ -11,15 +11,40 @@ jobs:
...
+              owner,
+              repo,
+              workflow_id: 'ci.yml',
+              event: 'pull_request',
+              head_sha: context.sha,
+              status: 'success'
+            });
+            const runs = runsResponse.data.workflow_runs;
+            runs.sort((a, b) => b.created_at.getTime() - a.created_at.getTime());
+            const run = runs[0];
```
blairconrad, line 32: I fear we'll get a nasty error if `runs` is empty. Which I guess could happen if someone (me) attached the label too quickly? Especially dangerous if the PR comes from an outside contributor for whom we haven't allow-listed the actions.

.github/workflows/publish-docs-preview.yml
```diff
@@ -11,15 +11,40 @@ jobs:
...
+              return;
+            }
+            core.setOutput('run-id', run.id);
+
       - name: Download artifact
+        if: ${{ steps.find-workflow-run.outcome == 'success' }}
         id: download-artifact
-        uses: dawidd6/action-download-artifact@v3
+        uses: actions/download-artifact@v4
         continue-on-error: true
```
blairconrad, line 43: I'm dense. Why continue-on-error? If we fail to download artifact, why not let the error fly?

blairconrad:
Generally looks good, @thomaslevesque. Thanks. I'm a little confused by one thing and a little concerned by the possibility of not finding any runs at all. If you think it's fine, I'm happy to approve.

.github/workflows/publish-docs-preview.yml
```diff
@@ -11,15 +11,40 @@ jobs:
...
         with:
-          workflow: ci.yml
-          pr: ${{github.event.pull_request.number}}
           name: documentation
           path: documentation/
+          run-id: ${{ steps.find-workflow-run.outputs.run-id }}
```
blairconrad, line 47: I was going to ask about logging the run ID that we find, but this version of download-artifact probably has decent debug logging in case of failure.

.github/workflows/publish-docs-preview.yml
```diff
@@ -11,15 +11,40 @@ jobs:
...
+            const runsResponse = await github.rest.actions.listWorkflowRuns({
+              owner,
+              repo,
+              workflow_id: 'ci.yml',
+              event: 'pull_request',
+              head_sha: context.sha,
+              status: 'success'
+            });
+            const runs = runsResponse.data.workflow_runs;
+            runs.sort((a, b) => b.created_at.getTime() - a.created_at.getTime());
```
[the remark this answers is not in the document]
thomaslevesque: Well, as far as I can tell, there isn't really a way to compare dates directly in JS. I guess I could compare the numbers and return -1/0/1 depending on the comparison, but it would be much more verbose...

.github/workflows/publish-docs-preview.yml
thomaslevesque: > I fear we'll get a nasty error if `runs` is empty

No, it just returns `undefined`.

.github/workflows/publish-docs-preview.yml
thomaslevesque: Because we want to post a comment to the PR, with information about the failure. We check if the step was successful in subsequent steps.

blairconrad: AH!

.github/workflows/publish-docs-preview.yml
blairconrad: blow me down

.github/workflows/publish-docs-preview.yml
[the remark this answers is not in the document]
blairconrad: And that's the main reason I didn't even make my original comment!

blairconrad (approved):
Great! Thanks!

thomaslevesque:
Thanks for the merge!

PR #2009, opened 2024-04-12
Acknowledge that we'll never be able to fake methods on generic types that have in parameters before .NET 6

### LINKED ISSUE #1382: Allow Faking of generic types with methods that have in parameters
Split from #1338.

As mentioned in #1337 and castleproject/Core#339, generic types containing a method with an `in` parameter cannot be faked by FakeItEasy. The ultimate cause is https://github.com/dotnet/corefx/issues/29254.

We should support this as soon as we can.

blairconrad:
Closes #1382.

### DIFF
--- a/docs/what-can-be-faked.md
+++ b/docs/what-can-be-faked.md
@@ -20,8 +20,9 @@ Note that special steps will need to be taken to
 
 ### Types whose methods have `in` parameters
 
-Generic types that contain methods having a parameter modified by the `in` keyword cannot be faked by FakeItEasy.
-This limitation is tracked as [issue 1382](https://github.com/FakeItEasy/FakeItEasy/issues/1382).
+Due to deficiencies in earlier .NET framework releases, generic types that contain methods having
+a parameter modified by the `in` keyword cannot be faked by FakeItEasy running on target frameworks
+earlier than .NET 6.
 
 ### Where do the constructor arguments come from?
   
--- a/tests/FakeItEasy.Specs/InParametersSpecs.cs
+++ b/tests/FakeItEasy.Specs/InParametersSpecs.cs
@@ -79,9 +79,7 @@ namespace FakeItEasy.Specs
 
 #if LACKS_FAKEABLE_GENERIC_IN_PARAMETERS
         // A characterization test, representing the current capabilities of the code, not the desired state.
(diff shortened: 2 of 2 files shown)


tests/FakeItEasy.Specs/InParametersSpecs.cs
```diff
@@ -79,9 +79,7 @@ public void SettingInParamDelegate(DelegateWithInParam fake, Exception exception
 
 #if LACKS_FAKEABLE_GENERIC_IN_PARAMETERS
```
blairconrad, line 80: I thought about removing this section and just not testing, but I suppose there's a very small chance something will change. We might as well find out. But if you prefer, we could just to the `#else` part.

docs/what-can-be-faked.md
```diff
@@ -43,4 +44,4 @@ Note that this means that static members, including extension methods,
 
 ### Methods that return values by reference
 
-Methods that return values by reference (officially called "[reference return values](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/ref-returns#what-is-a-reference-return-value)") cannot be invoked on a Fake. Any attempt to do so will result in a `NullReferenceException` being thrown.
\ No newline at end of file
+Methods that return values by reference (officially called "[reference return values](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/ref-returns#what-is-a-reference-return-value)") cannot be invoked on a Fake. Any attempt to do so will result in a `NullReferenceException` being thrown.
```
thomaslevesque, line 47: I don't understand why this line is shown as a difference. It's the same on both sides 🤔

thomaslevesque (approved):
> Acknowledge that we'll never be able to fake methods on generic types that have in parameters

You mean "before .NET 6", right?

docs/what-can-be-faked.md
blairconrad: Turn on whitespace differences. My editor added a new line. I can expunge if you like.

thomaslevesque: No, don't bother! Thanks

thomaslevesque:
Thanks @blairconrad!

blairconrad:
Thank you, @thomaslevesque.

Issue #1437, opened 2018-09-06
Investigate ProxyGenerator.CreateDelegateProxy


blairconrad:
@stakx is adding delegate-generating code to DynamicProxy let's create a spike that uses it to see how well it fits our model.

I'll make notes here so my in-progress investigation (and inevitable backtracking) doesn't pollute the discussion in https://github.com/castleproject/Core/pull/403.

blairconrad:
## Proxied Delegates need to be visible to the generated dynamic proxy assembly

This matches all the other proxies that DynamicProxy creates, but is a regression. As of 4.8.1, FakeItEasy can fake even private delegates.

blairconrad:
## We lose some connections between the proxy and the FakeManager

When we create the proxy, we add it to a `ConditionalWeakTable`. So, if we proxied an `Action`, we have an `Action` as the key.
If there's a failing assertion (and probably other situations), we will attempt to look up, in `DefaultFakeObjectCallFormatter.GetDescription`, the `call.FakedObject`, but the FakedObject is a `Castle.Proxies.ActionProxy`, so we can't find the manager.

thomaslevesque:
Thanks for investigating @blairconrad. So, the only issue with this change is that we will only support public delegates? I think it's acceptable; it was a bit weird that we supported non-public delegates but not non-public interfaces. But it's a breaking change, so it'll have to be done in a major release.

thomaslevesque:
Did you try to profile each approach? Not that performance is a major concern, but I'm curious...

blairconrad:
> it's a breaking change, so it'll have to be done in a major release.

Yeah, but we're due. :wink: Seriously, it'd be nice to get rid of a few "on hold" issues.

> Did you try to profile each approach? Not that performance is a major concern, but I'm curious...

I did not. I was just assessing for correctness. I can do. I have a colleague who had some cool benchmarking lying around a while ago…

thomaslevesque:
> Yeah, but we're due. 😉 Seriously, it'd be nice to get rid of a few "on hold" issues.

My thoughts exactly.

> On the other hand, if performance is not a major concern, would we rather keep whatever performance we have and not introduce a breaking change?

No, I think we should do it. I don't think the breaking change is a major issue, and it will make the behavior more consistent.

If some people really like to be able to fake private delegate, maybe we could still support the old approach as an option? But I doubt it's worth the trouble.

blairconrad:
[...]

PR #2012, opened 2024-04-17
[WIP] make delegates the new way

blairconrad:
Relates to #1437

### DIFF
--- a/src/FakeItEasy/Core/FakeManagerProvider.cs
+++ b/src/FakeItEasy/Core/FakeManagerProvider.cs
@@ -5,7 +5,7 @@ namespace FakeItEasy.Core
 
     /// <summary>
     /// Implementation of <see cref="IFakeCallProcessorProvider"/>, which returns a <see cref="FakeManager"/> as "call processor" lazily (on
-    /// the first call of <see cref="Fetch"/> or <see cref="EnsureInitialized"/>).
+    /// the first call of <see cref="Fetch"/> or <see cref="EnsureInitialized(object, object?)"/>).
     /// </summary>
     internal class FakeManagerProvider : IFakeCallProcessorProvider
     {
@@ -51,7 +51,9 @@ namespace FakeItEasy.Core
             }
         }
 
-        public void EnsureInitialized(object proxy)
+        public void EnsureInitialized(object proxy) => this.EnsureInitialized(proxy, null);
+
+        public void EnsureInitialized(object proxy, object? alias)
         {
             Guard.AgainstNull(proxy);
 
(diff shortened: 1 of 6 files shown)


blairconrad:
For discussion only. Gives me a place to reference the code if I want to look at it away from home.
Also we can yammer, but don't feel any pressure, @thomaslevesque.

src/FakeItEasy/Core/DefaultFakeManagerAccessor.cs
```diff
@@ -22,13 +23,22 @@ public FakeManager GetFakeManager(object proxy)
             Guard.AgainstNull(proxy);
 
             FakeManager? result = this.TryGetFakeManager(proxy);
+            if (result is not null)
+            {
+                return result;
+            }
 
-            if (result is null)
+            if (proxy is Delegate { Target: not null } d)
```
blairconrad: Sometimes we're passed the Delegate, and sometimes the DynamicProxy proxy. If the delegate...

Actually, I think I've double-fixed this issue. Here and in the generator. I'll see if I can simplify.

src/FakeItEasy/Creation/DelegateProxies/DelegateProxyGenerator.cs
```diff
@@ -23,212 +20,24 @@ internal static class DelegateProxyGenerator
...
+            fakeCallProcessorProvider.EnsureInitialized(proxy);
 
-            public override CompletedFakeObjectCall ToCompletedCall()
-            {
-                return new CompletedFakeObjectCall(
-                    this,
-                    this.originalArguments);
-            }
+            var a = ServiceLocator.Resolve<IFakeManagerAccessor>();
+            a.SetFakeManager(delegateProxy, a.GetFakeManager(proxy));
```
blairconrad, line 51: hack to establish a link between the delegate and the FakeManager.

blairconrad:
image

src/FakeItEasy/Creation/ProxyInterceptor.cs
blairconrad, line 1: I hoisted it to use from delegate. This is probably not the best ending place. At least in this form. Not that anything in the PR is best anything

src/FakeItEasy/Core/FakeManagerProvider.cs
```diff
@@ -51,7 +51,9 @@ public IFakeCallProcessor Fetch(object proxy)
             }
         }
 
-        public void EnsureInitialized(object proxy)
+        public void EnsureInitialized(object proxy) => this.EnsureInitialized(proxy, null);
+
+        public void EnsureInitialized(object proxy, object? alias)
```
thomaslevesque, line 56: `alias` will always be a delegate, right? So why not type it as `Delegate` and name it `delegateProxy`?

PR #2015, opened 2024-04-21
Use .NET source generators to generate strongly-typed overloads

### LINKED ISSUE #1804: Consider using .NET 5.0 source generators to generate strongly-typed overloads
Currently we use T4 templates for generating strongly-typed overloads of methods in these classes:

- ExceptionThrower
- CallbackConfigurationExtensions
- OutAndRefParameters
- ReturnValueConfigurationExtensions
- AsyncReturnValueConfigurationExtensions
- ArgumentValidationConfigurationExtensions
- ValueTaskReturnValueConfigurationExtensions
- ValueTaskAsyncReturnValueConfigurationExtensions

We could use .NET 5.0 source generators instead.

- source generators run automatically during build, without the need for additional tools
- T4 templates, on the other hand, need Visual Studio and have to be run explicitly (although VS does it automatically when you save the template). We could integrate the transformation in the build, but there aren't any simple and clean ways to do it, as far as I can tell
- We currently have the generated sources in source control, mostly because of the tooling necessary to run the templates; it means the template and generated code could become desynchronized if we commit changes to the template without running it. That wouldn't be necessary with source generators. (although there's an argument to be made in favor of having the generated code in source control, since it's easier to see what the code does)

blairconrad:
Closes #1804.


blairconrad:
Hey, @thomaslevesque. For your consideration. Of course there are more classes to convert, but I thought I'd give you a peek before I invested more hours. I am content with the changes relative using T4, but am also open to leaving T4 in the mix and/or exploring a proper templating solution.
Tell me what you think!

tools/FakeItEasy.SourceGenerators/StronglyTypedExtensionsGenerator.cs
```diff
@@ -0,0 +1,103 @@
...
+
+                              /// <content>
+                              /// Contains auto-generated functionality for the {{mainClassSimpleName}} class.
+                              /// </content>
+                              public static partial class {{mainClassSimpleName}}
+                              {
+                          {{WhenArgumentsMatchMethods(8)}}
+                              }
+                          }
+                          """;
```
thomaslevesque, line 59: nit: reduce indentation for better readability
```suggestion
        var source = $$"""
            namespace {{mainClassNamespace}}
            {
                using System;
                using System.Reflection;
                using FakeItEasy.Configuration;

                /// <content>
                /// Contains auto-generated functionality for the {{mainClassSimpleName}} class.
                /// </content>
                public static partial class {{mainClassSimpleName}}
                {
            {{WhenArgumentsMatchMethods(8)}}
                }
            }
            """;
```

thomaslevesque (commented):
Nice, thanks @blairconrad!

Two remarks:
- I see you used the original source generator model, rather than the newer "incremental" generators. In this case, I _think_ it doesn't matter, because you don't have any syntax receiver that would trigger a regeneration on every keystroke.
- I don't see the generated code in your PR. By default it's generated in the `obj` folder, which is excluded from source control. If we want to include the generated code in source control, we can specify the path where it's emitted with `CompilerGeneratedFilesOutputPath`.

tools/FakeItEasy.SourceGenerators/StronglyTypedExtensionsGenerator.cs
blairconrad: Ah, but if we did that, the output would look funny. Too outdented. If you'd still rather it look nice, and I'm too worked up about ugly generated code, I could post-process the lines to indent.

blairconrad:
> * I see you used the original source generator model, rather than the newer "incremental" generators. In this case, I _think_ it doesn't matter, because you don't have any syntax receiver that would trigger a regeneration on every keystroke.

Ah! I can look at the incremental one. I just used the first guidance I found.

> * I don't see the generated code in your PR. By default it's generated in the `obj` folder, which is excluded from source control. If we want to include the generated code in source control, we can specify the path where it's emitted with `CompilerGeneratedFilesOutputPath`.

I will explore this!

tools/FakeItEasy.SourceGenerators/StronglyTypedExtensionsGenerator.cs
thomaslevesque: Are you sure? I think the closing `"""` determines the output's indentation level.

blairconrad:
Sorry @thomaslevesque. I rebased. I even put some commits in _before_ the originals.

.globalconfig
blairconrad, line 1: rules in `.globalconfig` will be applied to generated source

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -20,9 +20,9 @@
 
   <ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'netstandard2.1'))">
     <Compile Include="../FakeItEasy/ValueTaskReturnValueConfigurationExtensions.cs" />
-    <Compile Include="../FakeItEasy/ValueTaskReturnValueConfigurationExtensions.StronglyTyped.cs" />
+    <Compile Include="../FakeItEasy/Generated/CodeGen/CodeGen.ValueTaskReturnValueConfigurationExtensionsGenerator/*.cs" />
```
blairconrad, line 23: Relying on the generated source is maybe a little gross. Can revisit if you like.

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -28,103 +44,16 @@
     <DefineConstants>$(DefineConstants);LACKS_CALLERARGUMENTEXPRESSION</DefineConstants>
   </PropertyGroup>
 
-  <!-- .NET Standard 2.0 -->
+  <!-- Pre-.NET Standard 2.1 -->
```
blairconrad, line 47: Kind of a bonus change. To be honest, I thought it was going to be more useful than it turned out to be. I don't mind the consolidation below though. Still, if you prefer, I can revert!

src/FakeItEasy/Generated/CodeGen/CodeGen.ArgumentValidationConfigurationExtensionsGenerator/ArgumentValidationConfigurationExtensions.StronglyTyped.cs
```diff
@@ -249,6 +248,5 @@ public static partial class ArgumentValidationConfigurationExtensions
                 return argumentsPredicate(args.Get<T1>(0), args.Get<T2>(1), args.Get<T3>(2), args.Get<T4>(3), args.Get<T5>(4), args.Get<T6>(5), args.Get<T7>(6), args.Get<T8>(7));
             });
         }
-
     }
-}
+}
```
blairconrad, line 252: I'm surprised there's no newline here. Will investigate!

src/FakeItEasy/Generated/CodeGen/CodeGen.AsyncReturnValueConfigurationExtensionsGenerator/AsyncReturnValueConfigurationExtensions.StronglyTyped.cs
```diff
@@ -22,7 +20,6 @@ public static partial class AsyncReturnValueConfigurationExtensions
         /// <param name="exceptionFactory">A function that returns the exception to set on the returned task when a call that matches is invoked.</param>
         /// <typeparam name="T1">The type of the first argument of the faked method call.</typeparam>
         /// <returns>The configuration object.</returns>
-        [SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "This is by design to support the fluent API.")]
```
blairconrad, line 25: Turns out we weren't triggering a warning anymore, so a bonus change. Applies elsewhere.

src/FakeItEasy/Generated/CodeGen/CodeGen.CallbackConfigurationExtensionsGenerator/CallbackConfigurationExtensions.StronglyTyped.cs
```diff
@@ -32,7 +31,7 @@ public static partial class CallbackConfigurationExtensions
                 {
                     ValueProducerSignatureHelper.AssertThatValueProducerSignatureSatisfiesCallSignature(call.Method, actionToInvoke.GetMethodInfo(), NameOfInvokesFeature);
 
-                    actionToInvoke(call.GetArgument<T1>(0));
+                    actionToInvoke(call.GetArgument<T1>(0)!);
```
blairconrad, line 34: For some reason the generated source triggered a warning about the nullability. I think it's a good change regardless.

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -28,103 +44,16 @@
...
-      <DependentUpon>ValueTaskAsyncReturnValueConfigurationExtensions.StronglyTyped.tt</DependentUpon>
-    </Compile>
-    <Compile Update="ValueTaskReturnValueConfigurationExtensions.StronglyTyped.cs">
-      <DesignTime>True</DesignTime>
-      <AutoGen>True</AutoGen>
-      <DependentUpon>ValueTaskReturnValueConfigurationExtensions.StronglyTyped.tt</DependentUpon>
-    </Compile>
-  </ItemGroup>
-
-  <ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'netstandard2.1'))">
```
blairconrad, line 123: it became more convenient to deal with these files via ifdefs, since the generated code was incorporated whether we "Remove"d the files or not. An alternative would've been to generate different code for various TFMs, but that seemed annoying and confusing!

tools/CodeGen/StronglyTypedOverloadsGenerator.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// <summary>
+    /// Generates a string by invoking the <paramref name="generator" /> function for each number
+    /// from 1 to <paramref name="maximum" /> and joining the results
+    /// with <paramref name="separator"/>.
+    /// </summary>
+    /// <param name="maximum">The maximum number to generate strings for.</param>
+    /// <param name="separator">The separator to use between generated strings.</param>
+    /// <param name="generator">The function to generate strings.</param>
+    /// <returns>The generated string.</returns>
+    protected static string Escalate(int maximum, string separator, Func<int, string> generator) =>
```
blairconrad, line 67: I don't _love_ the name of this, but struggled.

src/FakeItEasy/Generated/CodeGen/CodeGen.ArgumentValidationConfigurationExtensionsGenerator/ArgumentValidationConfigurationExtensions.StronglyTyped.cs
blairconrad: Turns out my git diff displayer (delta) has output that _I_ think looks like there used to be no newline at the end of file, and I added one. But it was opposite. Restoring newline in a fixup.

tools/CodeGen/ExceptionThrowerGenerator.cs
```diff
@@ -0,0 +1,109 @@
...
+                    }
+
+                    private partial class PropertySetterAfterCallbackConfiguredAdapter
+                    {
+            {{Indent(3, GenerateMethodImplementations("IPropertySetterConfiguration"))}}
+                    }
+                }
+            }
+            """;
+        context.AddSource("ExceptionThrower.StronglyTyped.cs", source);
```
thomaslevesque: You were worried about having long file names. We could shave a few chars by changing ".StronglyTyped.cs" to ".g.cs" (which seems to be the convention used by .NET for generated files)

tools/CodeGen/StronglyTypedOverloadsGenerator.cs
thomaslevesque: Yeah, I didn't immediately understand what it did when I saw the call. But I don't have a better idea. It's fine, I think.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
thomaslevesque: Yeah, it's not ideal. I assume it's because you can't reference the generator in this project, because it will also generate the rest of the code, unrelated to `ValueTask`?
Maybe we could do something to _conditionally_ generate the code? e.g. have an assembly-level attribute to specify which code generators we want? Or maybe something in the project's properties, if it's easily accessible from the generator?

thomaslevesque (commented):
Thanks @blairconrad! Looks good in general. I do have a few comments.
To be honest, after the first 2 "Use source generator" commits, I only gave the next ones a very cursory review, but I like the approach you took.

src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
blairconrad: Okay. Here's an option. Generate overloads based on presence of a type in the being-compiled assembly. Feels like a reasonable thing to make a decision based on. And as a bonus, we can get rid of LACKS_VALUETYPE.

src/FakeItEasy/ArgumentValidationConfigurationExtensions.StronglyTyped.cs
```diff
@@ -1,3 +1,4 @@
+
```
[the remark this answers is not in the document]
blairconrad: it was not!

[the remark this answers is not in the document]
blairconrad: But I guess it went away in the middle anyhow. You want me to track down?

tools/CodeGen/ExceptionThrowerGenerator.cs
blairconrad: 191 is the max length now, even with the FakeItEasy.Extensions.ValueTask generated files.

tools/CodeGen/StronglyTypedOverloadsGenerator.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    protected static bool IsTypeInTargetAssembly(GeneratorExecutionContext context, string fullyQualifiedMetadataName)
+    {
+        var type = context.Compilation.GetTypeByMetadataName(fullyQualifiedMetadataName);
+        return SymbolEqualityComparer.Default.Equals(context.Compilation.Assembly, type?.ContainingAssembly);
+    }
+
+    protected abstract string GenerateSource();
+
+    protected abstract bool IsApplicable(GeneratorExecutionContext context);
```
blairconrad, line 98: An alternative could be to split the ValueTask generators into their own source generator and use the original generator only from FakeItEasy. The ValueTask generator would be used from both FIE and FIE.Extensions.ValueTask and would still need to detect whether it should run (since it shouldn't when building FIE for < netstandard2.1).

thomaslevesque: But the generators would need to share some code... I think it's not worth it. I like the approach you took!

src/FakeItEasy/ArgumentValidationConfigurationExtensions.StronglyTyped.cs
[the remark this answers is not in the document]
thomaslevesque: No, don't bother (sorry, I thought I had answered that)

thomaslevesque (dismissed):
Thanks @blairconrad, looks good to me! Just squash the fixups, and I'll merge

blairconrad:
@thomaslevesque, I squashed. While I was there I dropped the commit that added LACKS_VALUETASK, as we discussed. With that change, the minor differences to the ".NET Standard 2.0"/"Pre-.NET Standard 2.1" PropertyGroups in FakeItEasy.csproj didn't do anything for us, so I dropped that commit as well.

While I was at it, I distributed the commit that added source generation to FakeItEasy.Extensions.ValueType amongst 3 other commits:

* 7ac2841ec68cfa3cf5712f7e59967c6af5b28c20 Generate strongly-typed overloads only when base type exists (which is effectively a new (or shifted back in time) commit)
* e1291db2ad8bca3cb11dc5b36b38718a0bd88ca3 Use source generator to create ValueTaskAsyncReturnValueConfigurationExtensions.StronglyTyped.cs (pre-existing)
* 4f93dd8d50d469d2a5967c7362dac415ba3dcc72 Use source generator to create ValueTaskReturnValueConfigurationExtensions.StronglyTyped.cs (pre-existing)

I think this clears stuff up. Thanks for the great suggestions. I had a ball working on this PR!

thomaslevesque (approved):
This is great, thanks!

blairconrad:
Thank you, @thomaslevesque!

PR #2019, opened 2024-05-18
Stop attempting to create ByRef-Like Dummies

### LINKED ISSUE #1362: Stop attempting to create IsByRefLike types
In #1348, we started catching exceptions when attempts to create Dummies for constructor arguments failed. This fixed #1345, but is inelegant.
We should detect when we're trying to create an `IsByRefLike` type and refuse to do so, rather than having an exception thrown.

blairconrad:
Closes #1362.

### DIFF
--- a/src/FakeItEasy/Creation/CreationResult.cs
+++ b/src/FakeItEasy/Creation/CreationResult.cs
@@ -9,9 +9,14 @@ namespace FakeItEasy.Creation
     internal abstract class CreationResult(int status)
     {
         private const int SuccessfulCreation = 0;
-        private const int FailedCreation = 1;
+        private const int InterimFailedCreation = 1;
+        private const int PermanentlyFailedCreation = 2;
 
-        public bool WasSuccessful => status == SuccessfulCreation;
+        private readonly int status = status;
+
+        public bool WasSuccessful => this.status == SuccessfulCreation;
+
+        public bool IsFinal => this.status != InterimFailedCreation;
 
         public abstract object? Result { get; }
 
@@ -19,16 +24,19 @@ namespace FakeItEasy.Creation
              new SuccessfulCreationResult(result);
 
         public static CreationResult FailedToCreateDummy(Type type, string reasonForFailure) =>
-            new FailedCreationResult(type, CreationMode.Dummy, reasonsForFailure: new List<string> { reasonForFailure });
+            new FailedCreationResult(InterimFailedCreation, type, CreationMode.Dummy, reasonsForFailure: new List<string> { reasonForFailure });
 
         public static CreationResult FailedToCreateDummy(Type type, List<ResolvedConstructor> consideredConstructors) =>
-            new FailedCreationResult(type, CreationMode.Dummy, consideredConstructors: consideredConstructors);
+            new FailedCreationResult(InterimFailedCreation, type, CreationMode.Dummy, consideredConstructors: consideredConstructors);
 
         public static CreationResult FailedToCreateFake(Type type, string reasonForFailure) =>
-            new FailedCreationResult(type, CreationMode.Fake, reasonsForFailure: new List<string> { reasonForFailure });
+            new FailedCreationResult(InterimFailedCreation, type, CreationMode.Fake, reasonsForFailure: new List<string> { reasonForFailure });
 
         public static CreationResult FailedToCreateFake(Type type, List<ResolvedConstructor> consideredConstructors) =>
-            new FailedCreationResult(type, CreationMode.Fake, consideredConstructors: consideredConstructors);
+            new FailedCreationResult(InterimFailedCreation, type, CreationMode.Fake, consideredConstructors: consideredConstructors);
+
+        public static CreationResult PermanentlyFailedToCreateDummy(Type type, string reasonForFailure) =>
+            new FailedCreationResult(PermanentlyFailedCreation, type, CreationMode.Dummy, reasonsForFailure: new List<string> { reasonForFailure });
 
         /// <summary>
         /// Returns a creation result for a dummy by combining two results.
@@ -51,11 +59,12 @@ namespace FakeItEasy.Creation
         }
 
         private class FailedCreationResult(
+            int status,
             Type type,
             CreationMode creationMode,
             IList<string>? reasonsForFailure = null,
             IList<ResolvedConstructor>? consideredConstructors = null)
-            : CreationResult(FailedCreation)
+            : CreationResult(status)
         {
             private readonly IList<string>? reasonsForFailure = reasonsForFailure;
             private readonly IList<ResolvedConstructor>? consideredConstructors = consideredConstructors;
@@ -69,6 +78,7 @@ namespace FakeItEasy.Creation
 
                 var failedOther = (FailedCreationResult)other;
                 return new FailedCreationResult(
+                    this.status,
                     type,
                     creationMode,
                     MergeReasonsForFailure(failedOther.reasonsForFailure, this.reasonsForFailure),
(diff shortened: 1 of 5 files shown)


blairconrad:
Incomplete pull request at this time. Achieves the stated goal for netstandard2.1 and above only.
Below netstandard2.1, it's less straightforward to detect attempts to create ByRef-Like objects. I'll try to add code to do so but thought you might like an early crack at the overall attempt.

A further complication: for net462 we actually can create ByRef-Like Dummies. So I left that behaviour in place for now.

If you like, we can change FakeItEasy to prevent such creation, but I see no real reason to. If not, I'll just work on changing the behaviour for netstandard2.0, to throw a better exception like in netstandard2.1 and above.

src/FakeItEasy/Creation/CreationResult.cs
```diff
@@ -46,40 +46,29 @@ private class UntriedCreationResult : CreationResult
             public override CreationResult MergeIntoDummyResult(CreationResult other) => other;
         }
 
-        private class SuccessfulCreationResult : CreationResult
+        private class SuccessfulCreationResult(object? result) : CreationResult
```
blairconrad: You have some reservations about primary constructors that I forget and didn't understand at the time. I can drop such changes if you like (and configure Rider not to suggest them to me in the future).

src/FakeItEasy/Creation/CreationResult.cs
```diff
@@ -6,11 +6,14 @@ namespace FakeItEasy.Creation
     using System.Text;
     using FakeItEasy.Core;
 
-    internal abstract class CreationResult
+    internal abstract class CreationResult(int status)
     {
+        private const int SuccessfulCreation = 0;
+        private const int FailedCreation = 1;
```
blairconrad: A private enum might be preferable, but I was lazy. Happy to introduce one if you like.

src/FakeItEasy/Creation/CreationResult.cs
```diff
@@ -6,11 +6,14 @@ namespace FakeItEasy.Creation
...
-    internal abstract class CreationResult
+    internal abstract class CreationResult(int status)
     {
+        private const int SuccessfulCreation = 0;
+        private const int FailedCreation = 1;
+
         public static CreationResult Untried { get; } = new UntriedCreationResult();
 
-        public abstract bool WasSuccessful { get; }
+        public bool WasSuccessful => status == SuccessfulCreation;
```
blairconrad: Mostly I wanted to simplify the `WasSuccessful` call. It continues to pay off in the future.

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -69,7 +69,7 @@ public CreationResult TryResolveDummyValue(Type typeOfDummy, LoopDetectingResolu
             Type typeOfDummy,
             LoopDetectingResolutionContext resolutionContext)
         {
-            CreationResult creationResult = CreationResult.Untried;
+            CreationResult creationResult = null!;
```
blairconrad, line 72: Seems like a minor and maybe silly change, but simplifies the logic in a future commit...

src/FakeItEasy/Creation/CreationResult.cs
```diff
@@ -9,26 +9,34 @@ namespace FakeItEasy.Creation
     internal abstract class CreationResult(int status)
     {
         private const int SuccessfulCreation = 0;
-        private const int FailedCreation = 1;
+        private const int InterimFailedCreation = 1;
+        private const int PermanentlyFailedCreation = 2;
 
-        public bool WasSuccessful => status == SuccessfulCreation;
+        private readonly int status = status;
```
blairconrad, line 15: A warning I didn't understand talked about duplicating the status variable in a primary constructor for the `FailedCrerationResult` and wanted me to introduce the explicit field. Maybe this is reason enough to avoid primary constructors!

thomaslevesque:
> Achieves the stated goal for netstandard2.1 and above only. Below netstandard2.1, it's less straightforward to detect attempts to create ByRef-Like objects.

To be honest, I think it doesn't matter. Frameworks that are only compatible with netstandard2.0 are all out of support now. Projects that are still maintained usually upgrade to more recent frameworks, and those that are not won't upgrade to the latest FakeItEasy anyway.

So, unless it's reasonably easy, I think we shouldn't bother. We will probably drop netstandard2.0 (and 2.1) not too far in the future anyway...

> A further complication: for net462 we actually can create ByRef-Like Dummies. So I left that behaviour in place for now.

Sounds good to me.

I'll go read the PR now...

src/FakeItEasy/Creation/CreationResult.cs
thomaslevesque: There's always something with primary constructors that makes me not use them 🤷 

In this case I haven't been able to reproduce the warning with this minimal code:

```
internal abstract class CreationResult(int status)
{
    private const int SuccessfulCreation = 0;
    private const int InterimFailedCreation = 1;
    private const int PermanentlyFailedCreation = 2;

    public bool WasSuccessful => status == SuccessfulCreation;

    public bool IsFinal => status != InterimFailedCreation;
    
    public abstract object Result { get; }

    private class FailedCreationResult(int status, object result) : CreationResult(status)
    {
        public override object Result => result;
    }
}
```

src/FakeItEasy/Creation/CreationResult.cs
thomaslevesque: My reservation is mostly that almost every time I try to use them, I end up needing additional constructor logic, so I need to create an explicit constructor... But it's not the case here.

src/FakeItEasy/Creation/CreationResult.cs
thomaslevesque: Well, an enum would actually be _less_ code, not more (at least for the declaration). I think it would be slightly better, but I won't fight you for it.

src/FakeItEasy/Creation/DummyValueResolver.cs
```diff
@@ -95,6 +101,12 @@ private class ResolveByActivatingValueTypeStrategy : ResolveStrategy
             {
                 if (typeOfDummy.IsValueType && typeOfDummy != typeof(void))
                 {
+#if !TRY_ISBYREF_CREATION
+                    if (typeOfDummy.IsByRefLike)
+                    {
+                        return CreationResult.PermanentlyFailedToCreateDummy(typeOfDummy, "It is byref-like.");
+                    }
```
thomaslevesque, line 108: Why not perform this check earlier, before we attempt any creation strategy?

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
```diff
@@ -911,6 +911,41 @@ protected override IList<T> CreateCollectionOfDummy<T>(int count)
...
+
+#if CAN_CREATE_ISBYREFLIKE
+            "Then no exception is thrown"
+                .x(() => exception.Should().BeNull());
+
+            "And the dummy was created"
+                .x(() => dummy.Should().NotBeNull());
+#elif OBLIVIOUS_TO_ISBYREFLIKE_CREATION
+            "Then it throws an exception of type DummyCreationException"
+                .x(() => exception.Should().BeAnExceptionOfType<NotSupportedException>());
```
thomaslevesque, line 931: The step description doesn't match the actual assertion. Is it `DummyCreationException` or `NotSupportedException`?

thomaslevesque (commented):
Thanks @blairconrad!
I can't help but feel like the `CreationResult` refactoring wasn't really necessary, but maybe I'm missing something?

src/FakeItEasy/Creation/DummyValueResolver.cs
blairconrad: Oh, not addressed in my new PR either. Can do.

tests/FakeItEasy.Specs/DummyCreationSpecs.cs
blairconrad: blush

blairconrad:
Closing in favour of #2022.

PR #2020, opened 2024-05-19
Stop attempting to create ByRef-Like Dummies 

blairconrad:


blairconrad:
As with #2019, an incomplete pull request at this time. Presents an alternative approach to the refactoring, which I preferred. But, as you point out, is probably not necessary.

Achieves the stated goal for netstandard2.1 and above only.
Below netstandard2.1, it's less straightforward to detect attempts to create ByRef-Like objects. I'll try to add code to do so but thought you might like an early crack at the overall attempt.

A further complication: for net462 we actually can create ByRef-Like Dummies. So I left that behaviour in place for now.

If you like, we can change FakeItEasy to prevent such creation, but I see no real reason to. If not, I'll just work on changing the behaviour for netstandard2.0, to throw a better exception like in netstandard2.1 and above.

blairconrad:
Closing in favour of #2022.

PR #2022, opened 2024-05-19
Stop trying to create ByRef-like Dummies on TMFs lower than netstandard2.1

blairconrad:
As with #2019, an incomplete pull request at this time. Avoids touching the CreationResult code at all, as you suggested, @thomaslevesque.

### DIFF
--- a/src/FakeItEasy/TypeExtensions.cs
+++ b/src/FakeItEasy/TypeExtensions.cs
@@ -2,6 +2,7 @@ namespace FakeItEasy
 {
     using System;
     using System.Diagnostics;
+    using System.Linq;
 
     /// <summary>
     /// Provides extension methods for <see cref="Type"/>.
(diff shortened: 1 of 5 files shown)


blairconrad:
@thomaslevesque, the last commit is a candidate for squashing into the previous, I think. (Under any PR.)

src/FakeItEasy/TypeExtensions.cs
```diff
@@ -31,6 +32,15 @@ public static bool IsNullable(this Type type)
                 || (type.IsGenericType && type.GetGenericTypeDefinition() == typeof(Nullable<>));
         }
 
+        public static bool IsByRefLike(this Type type)
+        {
+#if LACKS_ISBYREFLIKE
+            return type.GetCustomAttributesData().Any(att => att.AttributeType.FullName == "System.Runtime.CompilerServices.IsByRefLikeAttribute");
```
thomaslevesque, line 38: Well, that wasn't so bad after all. As I said in the other PR, I don't think we really _need_ to handle this case for netstandard2.0, but if it's no harder than that, I guess it doesn't hurt. We can always drop this method when we drop netstandard TFMs.

thomaslevesque (commented):
I think this is my favorite version. Straightforward, no major changes, no big-bang refactoring of `CreationResult`... What's not to like?

src/FakeItEasy/TypeExtensions.cs
blairconrad: Yeah, it ended up being simpler than I thought.

thomaslevesque (approved):
Thanks, @blairconrad!

blairconrad:
Thanks, @thomaslevesque. As mentioned in sidechat, my commit message was _terrible_. Redoing in #2026.

PR #2024, opened 2024-05-20
Check docs links after building

blairconrad:


tools/FakeItEasy.Build/Program.cs
```diff
@@ -56,6 +56,19 @@
...
+// The 404.html page is likewise generated, and isn't even the one we use on the live site (mike generates a new one).
+Target(
+    "docs",
+    DependsOn("generate-docs"),
+    () => Run("linkchecker", "--ignore-url=sitemap.xml --ignore-url=404.html --check-extern -F html/./artifacts/docs-link-check.html ./artifacts/docs/index.html"));
+
+Target(
+    "generate-docs",
+    () => Run("python", "-m mkdocs build --clean --site-dir artifacts/docs --config-file mkdocs.yml --strict"));
+
```
blairconrad, line 71: I waffled over how to structure the targets.
Coulda been one big target: build and check.
Coulda been 3 targets: docs that depends on check-docs-links that depends on generate-docs
Coulda been 3 targets: docs that depends on both generate-docs and check-docs-links (with checks-docs-links depending on nothing)

I went this way but would go another way!

tools/FakeItEasy.Build/Program.cs
```diff
@@ -56,6 +56,19 @@
     forEach: projectsToPack,
     action: project => Run("dotnet", $"pack {project.Path} --configuration Release --no-build --nologo --output \"{Path.GetFullPath("artifacts/output")}\""));
 
+// If mkdocs's site_url configuration is not set, sitemap.xml is full of "None" links.
```
blairconrad, line 59: our site_url is not set. I think there's some value in setting it, and in theory mike will rewrite it, but didn't want to do here.

how_to_build.md
```diff
@@ -129,7 +129,7 @@ python -m pip install --requirement requirements.txt
 from the root of the repository. After this, you can generate the docs by running
 
 ```
-python -m mkdocs build --clean --site-dir artifacts/docs --config-file mkdocs.yml --strict
+build.cmd docs
```
thomaslevesque, line 132: I'm surprised we didn't do this sooner!

tools/FakeItEasy.Build/Program.cs
```diff
@@ -56,6 +56,19 @@
     forEach: projectsToPack,
     action: project => Run("dotnet", $"pack {project.Path} --configuration Release --no-build --nologo --output \"{Path.GetFullPath("artifacts/output")}\""));
 
+// If mkdocs's site_url configuration is not set, sitemap.xml is full of "None" links.
+// Even with a valid site_url, the link check would fail if we added a new page.
+// So trust that sitemap.xml will be generated correctly.
+// The 404.html page is likewise generated, and isn't even the one we use on the live site (mike generates a new one).
```
thomaslevesque, line 62: I'm confused by this comment. If I understand correctly, because site_url is not set, the sitemap.xml is not generated correctly. But then you say that we "trust that sitemap.xml is generated correctly"?

tools/FakeItEasy.Build/Program.cs
```diff
@@ -56,6 +56,19 @@
...
     action: project => Run("dotnet", $"pack {project.Path} --configuration Release --no-build --nologo --output \"{Path.GetFullPath("artifacts/output")}\""));
 
+// If mkdocs's site_url configuration is not set, sitemap.xml is full of "None" links.
+// Even with a valid site_url, the link check would fail if we added a new page.
+// So trust that sitemap.xml will be generated correctly.
+// The 404.html page is likewise generated, and isn't even the one we use on the live site (mike generates a new one).
+Target(
+    "docs",
+    DependsOn("generate-docs"),
+    () => Run("linkchecker", "--ignore-url=sitemap.xml --ignore-url=404.html --check-extern -F html/./artifacts/docs-link-check.html ./artifacts/docs/index.html"));
```
thomaslevesque, line 66: What does the `-F` option do? Path of the report? Why `/./`?

.github/workflows/ci.yml
```diff
@@ -74,9 +74,15 @@ jobs:
...
       - name: Upload docs
         uses: actions/upload-artifact@v4
         with:
           name: documentation
           path: artifacts/docs
+      - name: Upload link check report
+        uses: actions/upload-artifact@v4
+        with:
+          name: docs-link-check
+          path: artifacts/docs-link-check.html
```
thomaslevesque, line 88: Can we not zip it? It would be more convenient, and it's really small anyway

tools/FakeItEasy.Build/Program.cs
thomaslevesque: I think I prefer this one:

> 3 targets: docs that depends on check-docs-links that depends on generate-docs

It's a bit strange that the `docs` target actually does the check

thomaslevesque (commented):
Thanks @blairconrad, nice work as usual! I have some minor comments

.github/workflows/ci.yml
thomaslevesque: Apparently it's not an option :(
There's a `compression-level` you can set to 0, but if I understand correctly it still produces an archive...

tools/FakeItEasy.Build/Program.cs
blairconrad: Format and path of the report. The `/./` is probably there because, I dunno. I got the filename via path completion in the shell and left it. Removing.

> -F TYPE[/ENCODING[/FILENAME]], --file-output TYPE[/ENCODING[/FILENAME]]
>
>    // wrap each of the following lines to match the length of the first one
>    Output to a file linkchecker-out.TYPE, $XDG_DATA_HOME/linkchecker/failures for 'failures' output,
>    or FILENAME if specified. The ENCODING specifies the output encoding, the default is that of your
>    locale. Valid encodings are listed at https://docs.python.org/library/codecs.html#standard-encodings.
>    The FILENAME and ENCODING parts of the 'none' output type will be ignored, else if the file already
>    exists, it will be overwritten. You can specify this option more than once.
>    Valid file output types are 'csv', 'xml', 'dot', 'failures', 'gml', 'gxml', 'html', 'none',
>    'sitemap', 'sql', 'text'. You can specify this option multiple times to output to more than one file.
>    Default is no file output. Note that you can suppress all console output with the option '-o none'.

.github/workflows/ci.yml
blairconrad: I'd've preferred not to archive as well. Most of the time we won't need to look at the file anyhow (we'll have good links and/or could check the build output) and downloading and clicking right into the archive in Windows File Explorer worked for me.

blairconrad:
Late-breaking update. Upgraded the documentation dependencies to go along with our new requirement. 'sides, there's a new requests since you merged dependabot's PR yesternight.

Can drop if you like.

tools/FakeItEasy.Build/Program.cs
thomaslevesque: Where is `ENCODING`? You just did `TYPE/FILENAME`, is that valid? How does it know whether it's the encoding or the filename?

blairconrad: It seems to be valid. I dunno. It sussed it out. I'll add an encoding.

thomaslevesque (approved):
LGTM, thanks @blairconrad!

blairconrad:
Thanks, @thomaslevesque.

PR #2026, opened 2024-05-21
Stop trying to create ByRef-like Dummies on TFMs above net462

blairconrad:
Redoes work in #2022, but with proper commit message.

Unfixes, then fixes, #1362.


thomaslevesque (approved):
Thanks @blairconrad. This was the quickest code review ever!
image

PR #2029, opened 2024-05-23
Build and test on macOS

blairconrad:


blairconrad:
* [ ] require the macOS check

how_to_build.md
```diff
@@ -50,7 +49,7 @@ After the build has completed, the build artifacts will be located in `artifacts
 
 * View the full list of build targets:
 
-    `build.cmd -T`
+    `build.cmd -l`
```
thomaslevesque, line 52: I usually use `-t`, which shows the dependencies between the targets.
image

I guess showing the flat list is fine, if that's what you need, but we might want to mention `-t` as well

thomaslevesque (dismissed):
Thanks @blairconrad, LGTM, just one comment/suggestion

how_to_build.md
blairconrad: I don't mind switching to `-t` if you prefer. I'd rather not list both though. We're not here to duplicate bullseye's docs…

thomaslevesque (approved):
Perfect, thanks!

blairconrad:
Thanks, @thomaslevesque!

PR #2042, opened 2024-10-03
Manage mkdocs and friends via uv

blairconrad:
Ditching "regular" python in favour of using uv to manage mkdocs and whatnot.


.python-version
```diff
@@ -0,0 +1 @@
+3.13
```
blairconrad, line 1: I don't love that this file exists, to be honest, but when uv is run in "project mode" it emits a warning if it's not here.

how_to_build.md
```diff
@@ -116,16 +116,11 @@ reload the solution. The command line build will reflect the change immediately.
...
+The CI workflow uses [mkdocs](https://www.mkdocs.org/) to build the documentation.
+We use [uv](https://docs.astral.sh/uv/) to manage mkdocs and other Python tools used for the documentation.
+To build the documentation, **[install uv](https://docs.astral.sh/uv/getting-started/installation/)**.
 
-```
-python -m pip install --upgrade pip
-python -m pip install --requirement requirements.txt
-```
-
-from the root of the repository. After this, you can generate the docs by running
```
blairconrad, line 128: I do enjoy that the end user doesn't need to care about pip.

pyproject.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0" # value isn't important; we're just building the documentation
+requires-python = ">=3.13"
+
+[tool.uv]
+dev-dependencies = [
+    "linkchecker>=10.5.0",
+    "mike>=2.1.3",
+    "mkdocs>=1.6.1",
+    "mkdocs-material>=9.5.39",
+]
```
blairconrad, line 12: Oops. I accidentally upgraded everything while I was at it!

.python-version
thomaslevesque: Maybe we can just have the workflow generate it so that we don't have it in the repo?

pyproject.toml
thomaslevesque: toml 🤢 
I'm not sure if I hate it more or less than yaml ^^

.python-version
blairconrad: Intriguing. Could rip it right out of the similar line in `pyproject.toml`. Would reduce the need to manually update in the future as well.

blairconrad: Okay, done. Now that I'm thinking about it, the build process could run the uv installer if there wasn't already an uv, but maybe we're getting a little too fancy.

tools/FakeItEasy.Build/Program.cs
```diff
@@ -65,11 +65,29 @@
...
+// Rather than maintain a duplicate of the version in pyproject.toml, we'll generate the .python-version file from the latter.
+Target(
+    "create-python-version-file",
+    () =>
+    {
+        // expect the line to look something like like
+        // requires-python = "~=x.yz"
+        // with no surrounding whitespace. We want the x.yz part.
+        var prefix = "requires-python = \"~=";
+        var pythonVersion = File.ReadAllLines("pyproject.toml")
```
thomaslevesque, line 85: nit: you can use `File.ReadLines` instead of `File.ReadAllLines` (returns a lazy `IEnumerable<string>` instead of an array)

tools/FakeItEasy.Build/Program.cs
```diff
@@ -65,11 +65,29 @@
...
+            .Where(line => line.StartsWith(prefix, StringComparison.Ordinal))
+            .Select(line => line.Substring(prefix.Length, line.Length - prefix.Length - 1))
+            .Single();
```
thomaslevesque, line 88: If the line isn't found, this will throw an exception that won't really tell us anything useful about the cause of the problem... Which I guess is OK in this case, because we're not going to remove the `requires-python` line, and the script is small enough to easily identify the problem.

thomaslevesque (dismissed):
Nice! Thanks @blairconrad
Just two minor comments

thomaslevesque (approved):

blairconrad:
Thank you, @thomaslevesque.

PR #2043, opened 2024-10-07
Ignore HTTP 429 rate limit errors when checking links to StackOverflow

blairconrad:


.linkcheckerrc
```diff
@@ -0,0 +1,3 @@
+[filtering]
+ignorewarningsforurls=
+  ^https://stackoverflow\.com ^http-rate-limited$
```
thomaslevesque, line 3: Another file at the root just for the docs? :(
Can this be avoided? I see that you pass the path of the config file on the command line, maybe it can be moved to `docs`?

thomaslevesque (approved):
Thanks @blairconrad, LGTM except for a small thing

.linkcheckerrc
thomaslevesque: Nevermind that. Since I have another PR to move everything docs-related to `docs`, I'll take care of it in my own PR after merging yours.

blairconrad:
Thanks, @thomaslevesque. I'd've been happy to react to your changes, but this works for me too!

PR #2048, opened 2024-10-09
Test on macos-15 instead of macos-12

### LINKED ISSUE #2047: Test on macos 15 instead of 12
The macos-12 github actions runner image is being retired, according to actions/runner-images#10721.

blairconrad:
Fixes #2047

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,9 +37,11 @@ jobs:
               3.1
               2.1
           - os: macos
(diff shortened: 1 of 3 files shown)


thomaslevesque (approved):
Looks great, thanks @blairconrad!
I wonder if we should include something in the build profiles to give an error when the platform is not supported?
e.g. have `profiles/netstandard2.1` throw an error on macOS?
Might be off topic for this PR anyway

PR #2051, opened 2024-10-10
Build for net462 on Linux and Mac

thomaslevesque:
We can build (but not test) the net462 TFM on Linux and Mac by referencing the [Microsoft.NETFramework.ReferenceAssemblies](https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies) package.

I also refactored the `full` build profile to avoid repetition, because:
- the TFMs for libraries are now always the same
- the TFMs for tests for a given OS are always the same

Finally, in the last commit, I cleaned up the .editorconfig (it still had stuff related to project.json...) and explicitly set the indent size to 2 for MSBuild projects (to reflect the reality, and because Rider insisted on using 4 spaces when I edited the files manually...). I can move this to a separate PR if you prefer.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,17 +18,10 @@ max_line_length = 160
 
 insert_final_newline = true
 
-[*.{csproj,vbproj}]
+[*.{csproj,vbproj,props}]
 insert_final_newline = false
-charset = utf-8-bom
-
-[project.json]
-insert_final_newline = false
-charset = utf-8-bom
-
-[*.nuget.targets]
-insert_final_newline = false
-charset = utf-8-bom
+# 2-column space indentation
+indent_size = 2
 
 [*.json]
 # 2-column space indentation
--- a/profiles/full.props
+++ b/profiles/full.props
@@ -1,33 +1,25 @@
 <Project>
   <!-- Everything -->
+  <PropertyGroup>
+    <LibraryTargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0;net8.0</LibraryTargetFrameworks>
+  </PropertyGroup>
   <PropertyGroup Condition="$([MSBuild]::IsOSPlatform('Windows'))">
-    <FakeItEasyTargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0;net8.0</FakeItEasyTargetFrameworks>
-    <ValueTaskExtensionsTargetFrameworks>net462;netstandard2.0;netstandard2.1;net6.0;net8.0</ValueTaskExtensionsTargetFrameworks>
-    <SpecsTargetFrameworks>net462;netcoreapp2.1;netcoreapp3.1;net6.0;net8.0</SpecsTargetFrameworks>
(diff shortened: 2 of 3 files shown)


.editorconfig
```diff
@@ -18,17 +18,10 @@ max_line_length = 160
 
 insert_final_newline = true
 
-[*.{csproj,vbproj}]
+[*.{csproj,vbproj,props}]
 insert_final_newline = false
-charset = utf-8-bom
```
thomaslevesque, line 23: Already specified for `*`

.editorconfig
```diff
@@ -18,17 +18,10 @@ max_line_length = 160
...
 insert_final_newline = false
-charset = utf-8-bom
-
-[project.json]
-insert_final_newline = false
-charset = utf-8-bom
-
-[*.nuget.targets]
-insert_final_newline = false
-charset = utf-8-bom
```
thomaslevesque, line 31: These files no longer exist (well, there are .nuget.g.targets files in the obj directory, but they're not meant to be edited manually anyway)

profiles/full.props
```diff
@@ -1,33 +1,25 @@
...
+    <TestHelpersTargetFrameworks>$(LibraryTargetFrameworks)</TestHelpersTargetFrameworks>
     <ApprovalTestsTargetFrameworks>net8.0</ApprovalTestsTargetFrameworks>
-    <RecipesTargetFrameworks>net6.0;net8.0</RecipesTargetFrameworks>
+    <SpecsTargetFrameworks>$(TestTargetFrameworks)</SpecsTargetFrameworks>
+    <IntegrationTestsTargetFrameworks>$(TestTargetFrameworks)</IntegrationTestsTargetFrameworks>
+    <UnitTestsTargetFrameworks>$(TestTargetFrameworks)</UnitTestsTargetFrameworks>
+    <RecipesTargetFrameworks>$(TestTargetFrameworks)</RecipesTargetFrameworks>
   </PropertyGroup>
-</Project>
+</Project>
```
blairconrad, line 25: Very cool.
It occurs to me know that we could, in the csprojs, probably replace `FakeItEasyTargetFrameworks` with `LibraryTargetFrameworks`, etc., and further simplify the `.props` files. If we wanted. Until we needed to introduce more exceptions again.

blairconrad: It might look like 

https://github.com/FakeItEasy/FakeItEasy/commit/00a04665d9e67771b33f8016fd2864157a020925

Weirdly, I get complaints from the packer when I build locally, but nothing when [the workflow runs](https://github.com/blairconrad/FakeItEasy/actions/runs/11272613138).

thomaslevesque: > Until we needed to introduce more exceptions again.

It's the main reason I didn't do it. I think it's good to keep the flexibility, I just wanted to simplify things a bit.

blairconrad (commented):
Cool, @thomaslevesque. Thanks!

blairconrad (approved):

thomaslevesque:
Thanks for the merge!

PR #2055, opened 2025-03-28
Migrate from Xbehave to LambdaTale

thomaslevesque:
Since XBehave is no longer maintained

The 1st commit upgrades xUnit to 2.9.3 (required by LambdaTale)
The 2nd commit switches from XBehave to LambdaTale

The following commits are not strictly related to this change, I can move them to a separate PR if you prefer:
- Fix a warning in FakeItEasy.Recipes.CSharp
- Disable fail-fast in the matrix build strategy (so that a failure on one OS doesn't hide the errors on other OSes)
- Disable TFM parallelism for integration tests (new in .NET 9 SDK, was causing an error when accessing the FakeItEasy.ExtensionPoints.ExternalDependency assembly from multiple processes at the same time)
- Re-enable TFM parallelism and fix the code to use different paths for the generated assemblies (could be squashed with previous commit if we want to keep it)

~~At the moment, the tests for net462 and netcoreap2.1 fail because LambdaTale doesn't have a strong name.~~ (the latest version of LambdaTale now has a strong name)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,7 @@ jobs:
   build:
     name: Build on ${{ matrix.os-name }}
     strategy:
+      fail-fast: false
       matrix:
         os: [windows, ubuntu, macos]
         include:
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -12,7 +12,7 @@ The coordinators will usually assign a priority to each issue from 1 (highest) t
 
 ## Tests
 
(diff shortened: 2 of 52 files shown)


blairconrad:
Very nice. If it's no bother, I would squash the unparallel/parallel commits. But if it a bother, I won't lose any sleep over it.

blairconrad (approved):

thomaslevesque:
Done!

blairconrad:
Thanks, sir!

PR #2056, opened 2025-04-02
Add support for default interface methods

### LINKED ISSUE #1633: Support calling default interface methods
C# 8 introduced the "default interface methods" feature, i.e. interface methods with a default implementation (this feature works only on .NET Core 3).

Currently it's not possible with FakeItEasy to make a fake that calls the default implementation. Using `CallsBaseMethods` doesn't work, because Castle.Core doesn't support this feature yet (https://github.com/castleproject/Core/issues/447).

We should support default interface methods in FakeItEasy.

API options:
- Reuse `CallsBaseMethods()` to call the default interface methods, in addition to what it currently does.
- Introduce a new `CallsDefaultImplementation()` to cover that scenario. (Maybe use a more explicit name; "default implementation" could be misleading)

Implementation options:
- Wait for Castle.Core to support it
- Do it ourselves. Default interface methods are methods whose declaring type is an interface and have `IsAbstract == false`. We need to call them non-virtually (otherwise it would cause infinite recursion). POC implementation [here](https://gist.github.com/thomaslevesque/a4310070de3d44f81144aafd136bea7c).

thomaslevesque:
Upgrade Castle.Core to 5.2.1, which introduces this support.
Fix the case of sealed default interface methods.
Reword the "non-virtual members can not be intercepted" message to also mentioned sealed members.

Fixes #1633

### DIFF
--- a/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
+++ b/src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
@@ -8,6 +8,9 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
 
     internal class CastleDynamicProxyInterceptionValidator : IMethodInterceptionValidator
     {
+        private const string NonVirtualMethodsCantBeIntercepted =
+            "Non-virtual or sealed members can not be intercepted. Only interface members and non-sealed virtual, overriding, and abstract members can be intercepted.";
+
         private readonly MethodInfoManager methodInfoManager;
 
         public CastleDynamicProxyInterceptionValidator(MethodInfoManager methodInfoManager)
@@ -15,8 +18,14 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
             this.methodInfoManager = methodInfoManager;
         }
 
-        public bool MethodCanBeInterceptedOnInstance(MethodInfo method, object? callTarget, [NotNullWhen(false)]out string? failReason)
+        public bool MethodCanBeInterceptedOnInstance(MethodInfo method, object? callTarget, [NotNullWhen(false)] out string? failReason)
         {
+            if (method.DeclaringType!.IsInterface && !method.IsVirtual)
+            {
+                failReason = NonVirtualMethodsCantBeIntercepted;
+                return false;
+            }
+
             var invokedMethod = this.GetInvokedMethod(method, callTarget);
 
             failReason = GetReasonForWhyMethodCanNotBeIntercepted(invokedMethod);
@@ -43,7 +52,7 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
                 }
                 else
                 {
-                    return "Non-virtual members can not be intercepted. Only interface members and virtual, overriding, and abstract members can be intercepted.";
+                    return NonVirtualMethodsCantBeIntercepted;
                 }
             }
 
@@ -61,7 +70,7 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
 
             if (!method.IsVirtual)
             {
-                return "Non-virtual members can not be intercepted. Only interface members and virtual, overriding, and abstract members can be intercepted.";
+                return NonVirtualMethodsCantBeIntercepted;
             }
 
             if (!Castle.DynamicProxy.ProxyUtil.IsAccessible(method, out var message))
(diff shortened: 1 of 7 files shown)


thomaslevesque:
Found a bug already. When trying to configure a sealed default interface method, we get an `InvalidOperationException: The sequence contains no element` in `MethodInfoManager.GetMethodOnTypeThatImplementsInterfaceMethod`. The interface map doesn't contain the sealed method.
Which means we can't configure this method, which makes sense since it's sealed, but we should handle it correctly.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
```diff
@@ -8,15 +8,24 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
 
     internal class CastleDynamicProxyInterceptionValidator : IMethodInterceptionValidator
     {
+        private const string NonVirtualMethodsCantBeIntercepted =
+            "Non-virtual or sealed members can not be intercepted. Only interface members and non-sealed virtual, overriding, and abstract members can be intercepted.";
```
thomaslevesque, line 12: We were already using this message in two places, and I added a third, so it was probably time to introduce a constant.

I reworded the message to also mention sealed members, which can never be intercepted.

src/FakeItEasy/Creation/CastleDynamicProxy/CastleDynamicProxyInterceptionValidator.cs
```diff
@@ -8,15 +8,24 @@ namespace FakeItEasy.Creation.CastleDynamicProxy
...
         }
 
-        public bool MethodCanBeInterceptedOnInstance(MethodInfo method, object? callTarget, [NotNullWhen(false)]out string? failReason)
+        public bool MethodCanBeInterceptedOnInstance(MethodInfo method, object? callTarget, [NotNullWhen(false)] out string? failReason)
         {
+            if (method.DeclaringType!.IsInterface && !method.IsVirtual)
+            {
+                failReason = NonVirtualMethodsCantBeIntercepted;
+                return false;
+            }
```
thomaslevesque, line 27: I had to add this check here instead of in `GetReasonForWhyMethodCanNotBeIntercepted`, because `GetInvokedMethod` throws an exception (in `MethodInfoManager.GetMethodOnTypeThatImplementsInterfaceMethod`) because it can't find the sealed method in the interface map, and doesn't account for the case where it's not found. Fixing `GetMethodOnTypeThatImplementsInterfaceMethod` to return null in that case wouldn't help, because we wouldn't know why, and anyway `GetInvokedMethod` assumes it will receive a non-null result from `GetMethodOnTypeThatWillBeInvokedByMethodInfo`.

thomaslevesque:
@blairconrad let me know if you can think of other test cases!

tests/FakeItEasy.Specs/DefaultInterfaceMethodSpecs.cs
```diff
@@ -0,0 +1,89 @@
...
+            .x(() => result.Should().Be(42));
+    }
+
+    [Scenario]
+    public static void FakeInterfaceWithDefaultMethodOverrideBehavior(IHaveDefaultMethods fake, int result)
+    {
+        "Given a faked interface with a default method"
+            .x(() => fake = A.Fake<IHaveDefaultMethods>());
+
+        "When the sealed default method is configured to return something else"
```
blairconrad, line 52:
```suggestion
        "When the default method is configured to return something else"
```

or maybe a non-sealed? So long as you're not saying the method is sealed, I'm happy

blairconrad (changes requested):
Thanks, @thomaslevesque. I enjoyed reading this. I don't know enough about default interface methods to think of more tests.

I have suggested a test script change. I think you'll enjoy it.

tests/FakeItEasy.Specs/DefaultInterfaceMethodSpecs.cs
thomaslevesque: Oops. Bad copy/paste, I guess.

thomaslevesque: Amended

blairconrad (approved):
Thanks, @thomaslevesque!

PR #2058, opened 2025-06-08
Drop support for net6.0 and netstandard TFMs 

blairconrad:
While investigating https://github.com/bbvch/LambdaTale/issues/91, it became apparent that our support for some older TFMs was becoming onerous. In particular, there are no longer supported runtimes for the net6.0, netstandard2.1, and netstandard2.0 libraries we were producing.

Among other things, supporting these libraries will hamper our efforts to upgrade our LambdaTale testing dependency. So drop them.


src/FakeItEasy.Extensions.ValueTask/FakeItEasy.Extensions.ValueTask.csproj
```diff
@@ -42,26 +42,23 @@
...
+  <!-- Include placeholder files in netX.Y folders in package -->
   <Target Name="AddTfmPlaceholders">
-    <WriteLinesToFile File="$(IntermediateOutputPath)netstandard2.1/_._" Overwrite="true" Lines="" />
     <WriteLinesToFile File="$(IntermediateOutputPath)net6.0/_._" Overwrite="true" Lines="" />
     <WriteLinesToFile File="$(IntermediateOutputPath)net8.0/_._" Overwrite="true" Lines="" />
     <ItemGroup>
-      <None Include="$(IntermediateOutputPath)netstandard2.1/_._" Pack="true" PackagePath="lib/netstandard2.1" Visible="false" />
       <None Include="$(IntermediateOutputPath)net6.0/_._" Pack="true" PackagePath="lib/net6.0" Visible="false" />
       <None Include="$(IntermediateOutputPath)net8.0/_._" Pack="true" PackagePath="lib/net8.0" Visible="false" />
     </ItemGroup>
```
thomaslevesque: I was looking at this yesterday and wondering if we could avoid hard-coding the TFMs for which we need a placeholder; it's a bit strange that we use a MSBuild property for `TargetFrameworks`, but hard-code the TFMs here. But maybe it's not worth the trouble, since it will soon be just net8.0...

thomaslevesque (approved):

PR #2059, opened 2025-06-08
Build with .NET 9.0 SDK

blairconrad:
We typically advocate using the latest SDK, and the GitHub Actions runners are already using .NET 9.0 SDK, so we might as well formalize it and clean up the warnings that have crept into the build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,9 @@ jobs:
         if: ${{ matrix.dotnet-versions-to-install }}
         with:
           dotnet-version: ${{ matrix.dotnet-versions-to-install }}
+      - name: Print .NET SDK version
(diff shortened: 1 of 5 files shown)


blairconrad:
Regard the lack of warnings!
https://github.com/FakeItEasy/FakeItEasy/actions/runs/15522862254

thomaslevesque (commented):
Thanks @blairconrad!
We also have a mention of the SDK version in how-to-build.md, should we update it?

blairconrad:
> Thanks @blairconrad! We also have a mention of the SDK version in how-to-build.md, should we update it?

Oops. I swear I had a commit for that before, but I can't find it. Glad you noticed!

thomaslevesque (approved):
Thanks!

PR #2063, opened 2025-09-02
Remove Fake.ClearConfiguration

### LINKED ISSUE #1768: Remove ClearConfiguration in 8.0.0
As discussed in  #1764, it's somewhat confusing, and untangling it is probably not worth the effort. The preferred method of getting a Fake with no configuration is to just make a new one.

thomaslevesque:
Since we already have pending breaking changes, the next release will be a major, so we can finally address this.

Closes #1768

### DIFF
--- a/tests/FakeItEasy.IntegrationTests/ConcurrentCallTests.cs
+++ b/tests/FakeItEasy.IntegrationTests/ConcurrentCallTests.cs
@@ -20,7 +20,7 @@ namespace FakeItEasy.IntegrationTests
                 new AddRuleFirst(),
                 new AddRuleLast(),
                 new RemoveRule(),
-                new ClearConfiguration(),
+                new ResetFake(),
                 new AddConfigurationViaACallTo(),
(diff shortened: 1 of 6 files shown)


tests/FakeItEasy.IntegrationTests/ConcurrentCallTests.cs
```diff
@@ -125,11 +125,9 @@ public override void MutateRules(IFoo fake)
             }
         }
 
-        private sealed class ClearConfiguration : RuleMutatingAction
+        private sealed class ResetFake : RuleMutatingAction
```
thomaslevesque, line 128: I could just remove this completely if you prefer.

blairconrad: No strong opinion. Easy to remove later if we feel like it. Let's keep it.

blairconrad (approved):

PR #2066, opened 2025-10-21
Restore how-to-build and contributing to top level

blairconrad:


thomaslevesque (approved):
Thanks, @blairconrad!
Took me a while to figure out what had happened here, with Github not clearly showing symlinks...

PR #2068, opened 2025-10-22
Stop trying to upload link check reports in CI workflow

thomaslevesque:
Since we no longer do those checks in that workflow.


blairconrad (approved):

blairconrad:
Thanks! Honestly, I'm okay with informal rules in our head.

(I think I didn't realize until yesterday that the button to "merge" the changes from the target branch could also rebase!)

PR #2069, opened 2025-10-22
Parallelize link checks (docs and README)

thomaslevesque:
This way, one isn't dependent on the other, so a failing docs check doesn't prevent the README check from running.

image


blairconrad (approved):
Looks good. Thanks, @thomaslevesque!

blairconrad (commented):
Fantastic. Thanks, @thomaslevesque.

PR #2074, opened 2025-11-14
Remove explicit reference to SourceLink

### LINKED ISSUE #2072: Remove SourceLink package reference
It's no longer necessary. Since .NET SDK 8.0, SourceLink is enabled by default when using Github (among others) for source control.
https://github.com/dotnet/sourcelink?tab=readme-ov-file#using-source-link-in-net-projects

thomaslevesque:
Since it's now enabled automatically by the SDK.

Closes #2072


src/Directory.Build.props
```diff
@@ -20,7 +20,6 @@
 
   <ItemGroup>
     <PackageReference Include="MinVer" Version="2.3.1" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
```
blairconrad: Thanks, @thomaslevesque. Looks good to me. Our [instructions for using SourceLink](../../blob/master/docs/source-stepping.md) remain valid?

blairconrad (approved):

src/Directory.Build.props
blairconrad: > Maybe we could remove the mention that we use SourceLink; we do, but we no longer do anything explicit to use it.

Agreed!

thomaslevesque: Removed

blairconrad (approved):

blairconrad:
Thanks, @thomaslevesque!

PR #2077, opened 2025-11-29
Document how to work with default interface members

blairconrad:


docs/calling-base-methods.md
```diff
@@ -19,6 +19,17 @@ var fakeShop = A.Fake<CandyShop>(options => options.CallsBaseMethods());
 
 And then [selectively override some of them](changing-behavior-between-calls.md#overriding-the-behavior-for-a-call)
 
-```chsharp
+```csharp
```
blairconrad: bonus fix. Happy to drop or split into another commit if you want

thomaslevesque (approved):
Thanks @blairconrad. Great explanation of why it doesn't work for explicitly implemented members!

PR #2079, opened 2026-01-02
Add HasSameElementsAs argument constraint

### LINKED ISSUE #1913: Proposal: IsSameSequenceAs mathcer ignoring order
Hey, I would like to propose a matcher similar to IsSameSequenceAs, but one that ignores ordering. Similar to `CollectionAssert.AreEquivalent()`.

My initial idea was to implement a new `IsEquivalentTo` method that would have a required `OrderByKey` parameter which would be used to order both enumerables and then call `IsSameSequenceAs` internally for the matching.

Would like to know your opinions on it and if I should work on it.

Thanks in advance.

thomaslevesque:
Fixes #1913

This PR adds the `IsEquivalentTo` argument constraint, which is basically the same as `IsSameSequenceAs`, but ignoring the order of elements:

```csharp
fake.SomeMethod([1, 2, 3, 2, 1]);
A.CallTo(() => fake.SomeMethod(A<int[]>.That.IsEquivalentTo(1, 1, 2, 2, 3))).MustHaveHappened();
```

Since `IsEquivalentTo` is very similar to `IsSameSequenceAs`, it was an opportunity to improve `IsSameSequenceAs` as well, to make it more like it would be if it was designed today:
- Accept an `IEnumerable<TElement>` rather than a non-generic `IEnumerable` (which means adding an additional generic type parameter)
- Accept an optional `IEqualityComparer<TElement>` to specify how to compare the elements for equality

The first point is necessary to achieve the second (technically we could accept a non-generic `IEqualityComparer`, but in practice, virtually nobody ever uses this type, and the generic one doesn't inherit the non-generic one, so it would be very inconvenient to use). Also, it's a signature change, rather than an additional overload, otherwise the two overloads with cause ambiguous invocation errors.

This is a breaking change (definitely at the binary level, possibly but rarely at the source level). However we believe it's a small one, unlikely to affect most users, because in practice:
- People rarely specify type arguments that can be automatically inferred, so the extra type parameter shouldn't be an issue
- People almost never use non-generic collections these days. Generics in C# have been there for 20 years; at this point, if you're still using the old non-generic collections, you're probably not using the latest version of FakeItEasy.

The `params` overload of `IsSameSequenceAs` has been similarly updated.

---

Note: regarding the specs, rather than adding yet more scenarios with confusing names to `CallMatchingSpecs`, I started a new pattern by creating a specs class just for `IsEquivalentTo` in a `ArgumentConstraints` subdirectory.

### DIFF
--- a/src/FakeItEasy/Compatibility/EnumerableExtensions.TryGetNonEnumeratedCount.cs
+++ b/src/FakeItEasy/Compatibility/EnumerableExtensions.TryGetNonEnumeratedCount.cs
@@ -1,0 +1,52 @@
+﻿#if LACKS_NONENUMERATEDCOUNT
+namespace System.Linq
+{
+    using System.Collections;
+    using System.Collections.Generic;
+    using FakeItEasy;
+    using JetBrains.Annotations;
+
+    internal static class EnumerableExtensions
+    {
+        /// <summary>
+        ///   Attempts to determine the number of elements in a sequence without forcing an enumeration.
+        /// </summary>
+        /// <typeparam name="TSource">The type of the elements of <paramref name="source" />.</typeparam>
+        /// <param name="source">A sequence that contains elements to be counted.</param>
+        /// <param name="count">
+        ///     When this method returns, contains the count of <paramref name="source" /> if successful,
+        ///     or zero if the method failed to determine the count.</param>
+        /// <returns>
+        ///   <see langword="true" /> if the count of <paramref name="source"/> can be determined without enumeration;
+        ///   otherwise, <see langword="false" />.
+        /// </returns>
+        /// <remarks>
+        ///   The method performs a series of type tests, identifying common subtypes whose
+        ///   count can be determined without enumerating; this includes <see cref="ICollection{T}"/>,
+        ///   <see cref="ICollection"/> as well as internal types used in the LINQ implementation.
+        ///
+        ///   The method is typically a constant-time operation, but ultimately this depends on the complexity
+        ///   characteristics of the underlying collection implementation.
+        /// </remarks>
+        public static bool TryGetNonEnumeratedCount<TSource>([NoEnumeration] this IEnumerable<TSource> source, out int count)
+        {
+            Guard.AgainstNull(source);
+
+            if (source is ICollection<TSource> genericCollection)
+            {
+                count = genericCollection.Count;
+                return true;
+            }
+
+            if (source is ICollection nonGenericCollection)
+            {
+                count = nonGenericCollection.Count;
+                return true;
+            }
+
+            count = 0;
+            return false;
+        }
+    }
+}
+#endif
--- a/src/FakeItEasy/EnumerableExtensions.cs
+++ b/src/FakeItEasy/EnumerableExtensions.cs
@@ -5,7 +5,6 @@ namespace FakeItEasy
     using System.Collections.Generic;
     using System.Linq;
     using System.Text;
-    using JetBrains.Annotations;
 
     /// <summary>
     /// Provides extension methods for generic usage of <see cref="IEnumerable{T}"/>.
@@ -72,8 +71,8 @@ namespace FakeItEasy
             Guard.AgainstNull(other);
 
             // Fail fast if counts differ, when we can get the count without enumerating
-            if (collection.TryGetCount(out int collectionCount)
-                && other.TryGetCount(out int otherCount)
+            if (collection.TryGetNonEnumeratedCount(out int collectionCount)
+                && other.TryGetNonEnumeratedCount(out int otherCount)
                 && collectionCount != otherCount)
             {
                 return false;
@@ -105,41 +104,6 @@ namespace FakeItEasy
             // Ensure we consumed all elements from other.
             // If not, collection had fewer elements than other
             return elementCounts.Count == 0;
-        }
-
-        /// <summary>
-        /// Tries to get the count of items in the collection without enumerating it.
-        /// </summary>
-        /// <param name="collection">The collection.</param>
-        /// <param name="count">The count of items, if it could be determined without enumeration; otherwise, zero.</param>
-        /// <typeparam name="T">The type of items in the collection.</typeparam>
-        /// <returns>true if the count could be determined without enumeration; otherwise, false.</returns>
-        public static bool TryGetCount<T>([NoEnumeration] this IEnumerable<T> collection, out int count)
-        {
-#if LACKS_NONENUMERATEDCOUNT
-            if (collection is ICollection<T> genericCollection)
-            {
-                count = genericCollection.Count;
-                return true;
-            }
-
-            if (collection is ICollection nonGenericCollection)
-            {
-                count = nonGenericCollection.Count;
-                return true;
-            }
-
-            if (collection is IReadOnlyCollection<T> readOnlyCollection)
-            {
-                count = readOnlyCollection.Count;
-                return true;
-            }
-
-            count = 0;
-            return false;
-#else
-            return collection.TryGetNonEnumeratedCount(out count);
-#endif
         }
 
         /// <summary>


src/FakeItEasy/EnumerableExtensions.cs
```diff
@@ -61,5 +62,111 @@ public static IList<object> AsList(this IEnumerable source)
...
+        public static bool IsEquivalentTo<T>(
+            this IEnumerable<T> collection,
+            IEnumerable<T> other,
+            IEqualityComparer<T>? comparer)
+        {
+            Guard.AgainstNull(collection);
+            Guard.AgainstNull(other);
+
+            // Fail fast if counts differ, when we can get the count without enumerating
+            if (collection.TryGetCount(out int collectionCount)
```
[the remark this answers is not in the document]
blairconrad: Oh! Because you invented it!

src/FakeItEasy/EnumerableExtensions.cs
```diff
@@ -61,5 +62,111 @@ public static IList<object> AsList(this IEnumerable source)
...
+                && other.TryGetCount(out int otherCount)
+                && collectionCount != otherCount)
+            {
+                return false;
+            }
+
+            var boxComparer = new BoxEqualityComparer<T>(comparer);
+            var itemCounts = other
+                .GroupBy(x => new Box<T>(x), boxComparer)
+                .ToDictionary(g => g.Key, g => g.Count(), boxComparer);
```
blairconrad: surprised `g.Key` and `g.Count` aren't method references or whatever. I'm sure there's some technical reason.

src/FakeItEasy/EnumerableExtensions.cs
```diff
@@ -61,5 +62,111 @@ public static IList<object> AsList(this IEnumerable source)
...
+
+            count = 0;
+            return false;
+#else
+            return collection.TryGetNonEnumeratedCount(out count);
+#endif
+        }
+
+        /// <summary>
+        /// Wraps a possibly null value in a struct to allow usage as a key in dictionaries.
```
blairconrad: OH! I thought it was to do with boxing.
Clever solution.
The name is fine as-is.

tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/IsEquivalentToTests.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        protected override string ExpectedDescription => "\"a\", \"b\", NULL, \"y\", \"z\" (in any order)";
+
+        public static IEnumerable<object?[]> InvalidValues()
+        {
+            return TestCases.FromObject(
+                new[] { "1", "2", "x", "y" },
+                Array.Empty<string>(),
+                (object?)null,
+                new[] { "a", "b", null, "z", "x" },
```
blairconrad: Ah! Here is my "null element" test. I was not expecting a unit test when I made my spec test. Feel free to disregard my [earlier comment](https://github.com/FakeItEasy/FakeItEasy/pull/2079/changes#r2658914737), which I haven't deleted, only struck out.

src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
```diff
@@ -208,38 +208,93 @@ public static T IsGreaterThan<T>(this IArgumentConstraintManager<T> manager, T v
         }
 
         /// <summary>
-        /// The tested argument collection should contain the same elements as the
+        /// Constrains the argument collection so that it must contain the same elements as the
         /// specified collection, in the same order.
         /// </summary>
         /// <param name="manager">The constraint manager to match the constraint.</param>
         /// <param name="values">The sequence to test against.</param>
+        /// <param name="comparer">A comparer to test the collection elements for equality.</param>
```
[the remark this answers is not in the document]
thomaslevesque: Will do

src/FakeItEasy/EnumerableExtensions.cs
thomaslevesque: > surprised `g.Key` and `g.Count` aren't method references or whatever. I'm sure there's some technical reason.

What are you saying? Do you mean using method groups instead of lambda expressions?
- `Key`: because it's a property, not a method. And anyway we can't use an instance method as a method group in this context, because we need an instance first, and that is provided via the lambda parameter
- `Count()`, because it's an extension method, so I don't think it could work. And anyway, just like for `Key`, we need an instance to access it. (OK, I guess I could use `Enumerable.Count`, but that wouldn't feel very natural IMO)

src/FakeItEasy/FakeItEasy.csproj
```diff
@@ -40,12 +41,12 @@
...
+    <DefineConstants>$(DefineConstants);LACKS_NULLABLE_ATTRIBUTES;LACKS_ARRAY_EMPTY;LACKS_STRING_CONTAINS_COMPARISONTYPE;LACKS_CALLERARGUMENTEXPRESSION;LACKS_ISBYREFLIKE;TRY_ISBYREF_CREATION;LACKS_GENERIC_METHODINFO_CREATEDELEGATE;LACKS_NONENUMERATEDCOUNT</DefineConstants>
   </PropertyGroup>
 
   <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net462'))">
     <Compile Remove="ValueTaskReturnValueConfigurationExtensions.cs" />
     <Compile Remove="ValueTaskAsyncReturnValueConfigurationExtensions.cs" />
   </ItemGroup>
 
-</Project>
+</Project>
```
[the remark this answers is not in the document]
thomaslevesque: Looks like I did! Not intentional. I'll roll it back. If the IDE lets me.

tests/FakeItEasy.Specs/ArgumentConstraints/IsEquivalentToSpecs.cs
```diff
@@ -0,0 +1,136 @@
...
+        [Scenario]
+        public static void Success_SameElementsDifferentOrder(IIHaveACollectionParameter fake)
+        {
+            "Given a fake with a method that has a collection parameter"
+                .x(() => fake = A.Fake<IIHaveACollectionParameter>());
+
+            "When the method is called with a specific array"
+                .x(() => fake.Bar(new[] { 3, 2, 1, 2 }));
+
+            "Then an assertion that it was called with an array equivalent to the same items in a different order should pass"
```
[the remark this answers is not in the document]
thomaslevesque: Good catch. Also, with the new method name, I think the step description will be less of a mouthful.

tests/FakeItEasy.Tests/ArgumentConstraintManagerExtensions/IsEquivalentToTests.cs
thomaslevesque: I wasn't expecting to write unit tests either, but since they existed for `IsSameSequenceAs`, I added them for the new constraint as well.

src/FakeItEasy/EnumerableExtensions.cs
thomaslevesque: Sometimes I wish we could use instance method groups like that, e.g. `SomeType::SomeInstanceMethod`. I think Java has something like this. But it's not needed very often, and anyway in this case it would be very helpful (`IGrouping<Box<T>>::Key` isn't particularly convenient)

thomaslevesque:
I think I addressed everything.
Note: some of the new commits have "real" messages to make review easier, but should actually be fixup commits (they will actually be squashed into existing commits)

src/FakeItEasy/Compatibility/EnumerableExtensions.TryGetNonEnumeratedCount.cs
```diff
@@ -0,0 +1,29 @@
+﻿#if LACKS_NONENUMERATEDCOUNT
+namespace System.Linq
+{
+    using System.Collections;
+    using System.Collections.Generic;
+    using JetBrains.Annotations;
+
+    public static class EnumerableExtensions
```
blairconrad: `    internal static class EnumerableExtensions`

src/FakeItEasy/Compatibility/EnumerableExtensions.TryGetNonEnumeratedCount.cs
```diff
@@ -0,0 +1,29 @@
...
+                count = nonGenericCollection.Count;
+                return true;
+            }
+
+            count = 0;
+            return false;
+        }
+    }
+}
+#endif
```
blairconrad: That is cleaner. Good change.

tests/FakeItEasy.Specs/ArgumentConstraints/HasSameElementsAsSpecs.cs
```diff
@@ -0,0 +1,151 @@
...
+        [Scenario]
+        public static void Failure_DifferentElements(IIHaveACollectionParameter fake, Exception? exception)
+        {
+            "Given a fake with a method that has a collection parameter"
+                .x(() => fake = A.Fake<IIHaveACollectionParameter>());
+
+            "When the method is called with a specific array"
+                .x(() => fake.Bar(new[] { 1, 2, 3 }));
+
+            "And an assertion is made that it was called with an array that has a different set of items"
```
blairconrad:
```suggestion
            "And an assertion is made that it was called with an array that has a different set of elements"
```

docs/argument-constraints.md
```diff
@@ -70,26 +70,28 @@ A.CallTo(() => foo.Bar("hello", A<int>._)).MustHaveHappened();
...
+| Contains(string)                      | substring match with ordinal string comparison                                          |
+| Contains(string, comparisonType)      | substring match with the specified comparison type                                      |
+| StartsWith(string)                    | substring at beginning of string with ordinal string comparison                         |
+| StartsWith(string, comparisonType)    | substring at beginning of string with the specified comparison type                     |
+| EndsWith(string)                      | substring at end of string with ordinal string comparison                               |
+| EndsWith(string, comparisonType)      | substring at end of string with the specified comparison type                           |
+| IsNullOrEmpty()                       | `null` or `""`                                                                          |
+| IsEmpty()                             | empty enumerable                                                                        |
+| Contains(item)                        | item's presence in an enumerable                                                        |
```
blairconrad:
```suggestion
| Contains(element)                     | element's presence in an enumerable                                                        |
```

It caught my eye, but is not directly related to this PR. Okay to leave.

tests/FakeItEasy.Specs/ArgumentConstraints/HasSameElementsAsSpecs.cs
```diff
@@ -0,0 +1,151 @@
...
+            "Then an exception should be thrown"
+                .x(() => exception.Should().NotBeNull());
+
+            "And the exception message should tell us that the call was not matched, and include the values of the actual collection elements"
+                .x(() => exception!.Message.Should().BeModuloLineEndings("""
+
+
+                      Assertion failed for the following call:
+                        FakeItEasy.Specs.ArgumentConstraints.HasSameElementsAsSpecs+IIHaveACollectionParameter.Bar(args: <4, 5, 6 (in any order)>)
+                      Expected to find it once or more but didn't find it among the calls:
```
blairconrad: I was thinking that we could give the user more hints, such as "missing element A" or "element B appeared too many times". But it looks to me like we're not well set up for that, and the few other argument constraints I checked don't, so maybe we leave as is.

src/FakeItEasy/Compatibility/EnumerableExtensions.TryGetNonEnumeratedCount.cs
```diff
@@ -0,0 +1,29 @@
+﻿#if LACKS_NONENUMERATEDCOUNT
+namespace System.Linq
+{
+    using System.Collections;
+    using System.Collections.Generic;
+    using JetBrains.Annotations;
+
+    public static class EnumerableExtensions
+    {
+        public static bool TryGetNonEnumeratedCount<TSource>([NoEnumeration] this IEnumerable<TSource> source, out int count)
```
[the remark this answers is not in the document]
thomaslevesque: Maybe not necessary once it's no longer public. I can add it though (I'll just copy from the official docs)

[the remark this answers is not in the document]
blairconrad: Fair. I'm not put out if not added.

tests/FakeItEasy.Specs/ArgumentConstraints/HasSameElementsAsSpecs.cs
thomaslevesque: Yeah, it would be nice, but I think it should be addressed separately.

docs/argument-constraints.md
thomaslevesque: done

blairconrad (approved):
Lovely. Thanks, @thomaslevesque!

thomaslevesque:
Squashed

blairconrad:
@thomaslevesque, I renamed the PR before merging. Without consultation.

thomaslevesque:
> I renamed the PR before merging. Without consultation.

Good call! I should have changed the branch name too, but that would have meant recreating the PR...

PR #2081, opened 2026-01-07
Use fixed deploy action

thomaslevesque:
Update submodule to use the latest version of the `deploy` action


blairconrad (approved):
Thanks!

PR #2083, opened 2026-01-08
Fix release workflow

thomaslevesque:
Add the `id-token` permission to `deploy` job, necessary for Github trusted publishing


blairconrad (approved):
Thanks, @thomaslevesque!

PR #2085, opened 2026-01-23
Fix issue with optional parameter in IsSameSequenceAs and HasSameElementsAs

### LINKED ISSUE #2084: Unintended breaking change with IsSameSequenceAs
From the release notes for 9.0.0:
> Breaking change : the signature of IsSameSequenceAs has [changed](https://github.com/FakeItEasy/FakeItEasy/commit/e815b77dfba2ef9c4e91fe35bc24f7910f334f87#diff-86b0fcd7c9ae751d623bf244dba6837b3393cb0c3c78a74ebf8ca316ddf209ecR73). It now has 2 generic type parameters, and accepts an optional equality comparer. This is a binary-level breaking change, and possibly a source-level breaking change, but we don't expect most users to be affected, since the generic type arguments are typically inferred by the compiler (https://github.com/FakeItEasy/FakeItEasy/issues/1913)

However, on .NET 9, it can break with: `An expression tree may not contain a call or invocation that uses optional arguments`

Sample repro:
```
using FakeItEasy;

ITest test = null!;
IEnumerable<string> strings = [];
A.CallTo(() => test.DoThings(A<IEnumerable<string>>.That.IsSameSequenceAs(strings)));

public interface ITest
{
    void DoThings(IEnumerable<string> things);
}
```

FYI: This compiles fine with .NET 10

thomaslevesque:
Optional parameters can't be used in expression trees before C# 14, so we need to have two overloads: one with the comparer parameter, and one without.

Fixes #2084

### DIFF
--- a/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
+++ b/src/FakeItEasy/ArgumentConstraintManagerExtensions.cs
@@ -213,6 +213,23 @@ public static class ArgumentConstraintManagerExtensions
     /// </summary>
     /// <param name="manager">The constraint manager to match the constraint.</param>
     /// <param name="values">The sequence to test against.</param>
+    /// <typeparam name="T">The type of argument to constrain.</typeparam>
+    /// <typeparam name="TElement">The type of the collection's elements.</typeparam>
+    /// <returns>A dummy argument value.</returns>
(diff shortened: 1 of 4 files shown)


blairconrad (approved):
Lovely. Thanks, @thomaslevesque.

PR #2086, opened 2026-01-23
Fix mistake in 'how to build' document

thomaslevesque:
Noticed this when making the changes to use `uv` in the analyzers project, and forgot to submit the change here.


blairconrad (approved):
Ah, very nice. Thanks, @thomaslevesque.

PR #2090, opened 2026-06-17
Convert FakeItEasy.Build to a file-based app

thomaslevesque:
https://learn.microsoft.com/en-us/dotnet/core/sdk/file-based-apps

No more project file!


tools/Build.cs
```diff
@@ -1,3 +1,7 @@
+#:property NoWarn=SA1200,SA1649,CA1014,CA1812
```
blairconrad: I have traditionally enjoyed having the warnings explained next to their cryptic codes, but can live without if it's tricky or you'd rather just not.
Of course we didn't have in the .csproj and I lived. I guess I just mostly like it when it looks like it's near code. Carry on how you like!

thomaslevesque: I guess we can just add comments, but I don't think we can put them inline next to the warning numbers.
We can probably do something like this:
```
// SA1649: File name should match first type name
#:property NoWarn=SA1649
```

(I removed the others, they were unnecessary)

Also, this is wrong, it should include `$(NoWarn)` as well. It's Copilot's fault!

thomaslevesque:
> (draft because I need to update the CI workflows as well)

Actually, no. The CI just calls build.ps1, which is still there.

tools/Build.cs
```diff
@@ -1,3 +1,8 @@
+// SA1649: File name should match first type name
```
blairconrad, line 1: I did enjoy this!
But would've been okay had it not made it in

blairconrad (approved):
Cool. Thanks!

